PR #607, opened 2016-03-09
Add resonance scattering to the Python API

smharper:
This PR adds the settings.xml `<resonance_scattering>` element to the PythonAPI.

In addition to that new feature, it also
* Fixes the resonance scattering test.  The test previously did not have any low energy neutrons that fell within the method window.
* Adds DBRC and WCM to the test
* Fixes a small typo in settings.xml that didn't allow fixed source simulations


smharper:
[Relevant coverage test](http://web.mit.edu/smharper/www/src/physics.F90.gcov.html) Ctrl+F for 'wcm'

openmc/settings.py
```diff
@@ -9,6 +9,7 @@
 from openmc.clean_xml import *
 from openmc.checkvalue import (check_type, check_length, check_value,
                                check_greater_than, check_less_than)
+from openmc.nuclide import Nuclide
```
wbinventor: I think this can be simplified to `from openmc import Nuclide`

openmc/settings.py
```diff
@@ -1087,3 +1136,88 @@ def export_to_xml(self):
...
+    nuclide : openmc.nuclide.Nuclide
+        The nuclide affected by this resonance scattering treatment.
+    nuclide_0K : openmc.nuclide.Nuclide
+        This should be the same isotope as `nuclide`, but it should have an
+        `xs` attribute that identifies 0 Kelvin data.
+    method : str
+        The method used to sample outgoing scattering energies.  Valid options
+        are 'ARES', 'CXS' (constant cross section), 'DBRC' (Doppler broadening
+        rejection correction), and 'WCM' (weight correction method).
+    E_min : float
```
wbinventor: To be consistent with the rest of the API, use `Real` in place of `float` here and for `E_max` below

openmc/settings.py
```diff
@@ -1087,3 +1136,88 @@ def export_to_xml(self):
...
+    def E_min(self):
+        return self._E_min
+
+    @property
+    def E_max(self):
+        return self._E_max
+
+    @nuclide.setter
+    def nuclide(self, nuc):
+        check_type('nuclide', nuc, Nuclide)
```
wbinventor: Perhaps consider allowing the user to define this as a string (*e.g.*, `'U-238') or a `Nuclide` object?

openmc/settings.py
```diff
@@ -1087,3 +1136,88 @@ def export_to_xml(self):
...
+    @nuclide_0K.setter
+    def nuclide_0K(self, nuc):
+        check_type('nuclide_0K', nuc, Nuclide)
+        if nuc.zaid == None: raise ValueError("The `nuclide_0K` must have an "
+             "explicitly defined `zaid` attribute.")
+        self._nuclide_0K = nuc
+
+    @method.setter
+    def method(self, m):
+        check_type('method', m, basestring)
```
wbinventor: This can be simplified as follows:

```python
@method.setter
def method(self, method):
    check_value('method', method, ('ARES', 'CXS', 'DBRC', 'WCM'))
    self._method = method
```

tests/test_resonance_scattering/test_resonance_scattering.py
```diff
@@ -3,9 +3,81 @@
 import os
 import sys
 sys.path.insert(0, os.pardir)
-from testing_harness import TestHarness
+from testing_harness import TestHarness, PyAPITestHarness
```
wbinventor: Does `TestHarness` need to be imported here along with `PyAPITestHarness`?

openmc/settings.py
```diff
@@ -1087,3 +1136,88 @@ def export_to_xml(self):
...
+
+class ResonanceScattering(object):
+    """Specification of the elastic scattering model for resonant isotopes.
+
+    Attributes
+    ----------
+    nuclide : openmc.nuclide.Nuclide
+        The nuclide affected by this resonance scattering treatment.
+    nuclide_0K : openmc.nuclide.Nuclide
+        This should be the same isotope as `nuclide`, but it should have an
```
wbinventor: We haven't used the "`" operator to surround parameters in docstrings elsewhere in the API. Is there a good reason to do so? If so, we can keep them here and aim to put them in the docstrings elsewhere. Otherwise, they should be removed here (personally I find them clunky to parse in the docstring itself, but perhaps they improve the rendering in sphinx).

openmc/settings.py
```diff
@@ -1043,6 +1061,36 @@ def _create_dd_subelement(self):
...
             subelement.text = str(self._dd_count_interactions).lower()
 
+    def _create_resonance_scattering_element(self):
+        if self.resonance_scattering is None: return
+
+        element = ET.SubElement(self._settings_file, "resonance_scattering")
+
+        # Create an iterable version of resonance_scattering
+        if isinstance(self.resonance_scattering, Iterable):
+            res = self.resonance_scattering
```
wbinventor: Why not just make `resonance_scattering` an `Iterable` of `ResonanceScattering` by default? The `resonance_scattering` property could enforce this perhaps.

openmc/settings.py
```diff
@@ -1043,6 +1061,36 @@ def _create_dd_subelement(self):
...
+        else:
+            res = [self.resonance_scattering]
+
+        for r in res:
```
wbinventor: If you choose to make `resonance_scattering` an `Iterable` then this `for` loop can also be simplified by iterating over `self.resonance_scattering`.

openmc/settings.py
```diff
@@ -1043,6 +1061,36 @@ def _create_dd_subelement(self):
...
+            if r.nuclide.name != r.nuclide_0K.name:
+                raise ValueError("The `nuclide` and `nuclide_0K` attributes of "
+                     "a ResonantScattering object must have identical names.")
+            scatterer = ET.SubElement(element, "scatterer")
```
wbinventor: Why not abstract the body of this loop into a new `ResonanceScattering.create_xml_subelement()` method? This would better mirror how the XML generation is done throughout the rest of the API.

openmc/settings.py
smharper: I don't think we can make that work cleanly here.  It's not enough to know the `name` of the nuclide, we also need the `xs` and the `zaid`.

openmc/settings.py
smharper: I thought it added a little extra clarity, but in retrospect it's a little weird to have this be the only place where that convention is used.  I've removed them.

openmc/settings.py
smharper: This did make the code much cleaner.  Thanks!

smharper:
Thanks for your comments, @wbinventor.  I believe I've addressed them all.

openmc/settings.py
```diff
@@ -125,6 +126,8 @@ class SettingsFile(object):
         Coordinates of the lower-left point of the UFS mesh
     ufs_upper_right : tuple or list
         Coordinates of the upper-right point of the UFS mesh
+    resonance_scattering : ResonanceScattering or iterable thereof
```
wbinventor: "iterable thereof" might best be reworded as "Iterable of ResonanceScattering" to match the context throughout the rest of the API

openmc/settings.py
```diff
@@ -125,6 +126,8 @@ class SettingsFile(object):
...
+        The elastic scattering model to use for resonant isotopes.
```
wbinventor: Remove the period on this docstring, also for consistency (I really don't care one way or the other, as long as we are consistent)

openmc/settings.py
```diff
@@ -1043,6 +1062,17 @@ def _create_dd_subelement(self):
             subelement = ET.SubElement(element, "count_interactions")
             subelement.text = str(self._dd_count_interactions).lower()
 
+    def _create_resonance_scattering_element(self):
+        if self.resonance_scattering is None: return
+
+        element = ET.SubElement(self._settings_file, "resonance_scattering")
+
+        for r in self.resonance_scattering:
```
wbinventor: Very nice!

openmc/settings.py
```diff
@@ -1087,3 +1118,101 @@ def export_to_xml(self):
...
+
+    @E_max.setter
+    def E_max(self, E):
+        check_type('E_max', E, Real)
+        check_greater_than('E_max', E, 0, True)
+        self._E_max = E
+
+    def create_xml_subelement(self, xml_element):
+        scatterer = ET.SubElement(xml_element, "scatterer")
+        subelement = ET.SubElement(scatterer, 'nuclide')
```
wbinventor: Are any of these subelements optional? If so, there should be a conditional to check if a user assigned each one.

openmc/settings.py
```diff
@@ -1087,3 +1118,101 @@ def export_to_xml(self):
...
         tree.write("settings.xml", xml_declaration=True,
                              encoding='utf-8', method="xml")
+
+
+class ResonanceScattering(object):
+    """Specification of the elastic scattering model for resonant isotopes.
+
+    Attributes
+    ----------
+    nuclide : openmc.nuclide.Nuclide
```
wbinventor: Ditto, remove periods after each attribute's comment (unless it is a multi-sentence comment) for consistency.

wbinventor:
Once my last few comments are addressed, I'm happy with this add @smharper. But I've never used the resonance scattering models so perhaps one of the core users/developers should also take a quick peek at this.

openmc/settings.py
smharper: That will make it over 80 columns.  Do I need to do anything special to show that the first line in that comment has been continued?
```python
    """
    resonance_scattering : ResonanceScattering or iterable of
        ResonanceScattering
        The elastic scattering model to use for resonant isotopes.
    """
```

wbinventor: I'm not sure about that. Seem simplest just to enforce that it always be an `Iterable`, but then again I don't know the typical use case for this (*i.e.*, if someone wants to use resonance scattering models, are they only interested in using it for a single nuclide or rather many nuclides).

smharper: As I understand it, the big three U-238 resonances account for most of the differences between the resonance scattering models.  Because of that, I expect a single `resonance_scattering` element will be the most common use case.  That's my use case, anyway.

wbinventor: Ok then I'm cool with allowing either a single element or a collection. As for the docstring, take a look at the spinx-generated documentation. My intuition tells me that the comment will have to be on one line for sphinx to render it properly (as has been the case for a few other docstrings).

smharper: Okay.  It's only 81 characters anyway.  I'll try to stop being so pedantic :p

openmc/settings.py
smharper: Ah, yes.  Some are optional.  I've added the appropriate checks.

Issue #608, opened 2016-03-10
Major Shift in Eigenvalue in CE Mode Related to #494


wbinventor:
I've been baffled recently by a major shift in the eigenvalue  (-4000pcm) computed by OpenMC for a few of the geometries I commonly simulate. I finally took some time to track this down and to my surprise, it appears to be related to the introduction of the multi-group mode in #494. The geometries for which I've observed this effect are composed of multiple assemblies from the BEAVRS benchmark. I've illustrated one example below with 2x2 assemblies next to a water reflector, with reflective BCs on the bottom and left, above and below, and vacuum on the right and bottom:

<img src="https://cloud.githubusercontent.com/assets/209492/13674451/84b2f626-e6ab-11e5-84b3-68395c097c0f.png"  width="200">

Using the *same* XML input files I observe the following eigenvalues when run with the following two commits before and after the multi-group mode merge:

**OpenMC Eigenvalues**
0.95915 +/- 0.00099 (SHA before: 713f6bf0fb6e207d315b42a485abed8bc66c8815)
0.92086 +/- 0.00092 (SHA after: 446262f709c934c59a17792da71b66bc06d611d9)

This issue came up since the eigenvalues I computed by OpenMOC (on a coarse spatial and angular mesh, but the trend is still the same) from 2-group MGXS in each case remain about the same:

**OpenMOC Eigenvalues**
0.962293 (SHA before: 713f6bf0fb6e207d315b42a485abed8bc66c8815)
0.958277 (SHA after: 446262f709c934c59a17792da71b66bc06d611d9)

Note that I only ran 100 batches of 10,000 particles/batch for the results cited here. I've confirmed these results with more particles/batch such that the two OpenMOC eigenvalues converge to one another, but OpenMC's eigenvalues remain shifted by ~4000 pcm.

I didn't review #494 in much detail and have no idea why it caused the eigenvalue to change so much for this model (and others). @nelsonag @paulromano @smharper @samuelshaner any ideas?

The XML input files for this case are in [this](https://www.dropbox.com/sh/l4l3c8v6iheb6z1/AABrmaHvpxbp4lEvomi1PFnia?dl=0) dropbox folder. Note that the "tallies.xml" file is for a distribcell nuclide-by-nuclide MGXS library with 70-groups which will easily consume your memory, but it is not needed to reproduce this issue.

wbinventor:
By the way, I've observed this same issue for a similar geometry with only the 2x2 assemblies and without the reflector.

[...]

PR #609, opened 2016-03-11
Fix for the lack of iso-in-lab scattering for nuclides since the MG mode was merged

nelsonag:
As discussed in #608, the merging of the multi-group mode made it so the iso-in-lab scattering was not accessible for nuclides (but it was for elements).  This PR fixes that issue.


wbinventor:
Thanks for the quick fix @nelsonag!

PR #615, opened 2016-03-22
Distribcell Paths in Summary Files

wbinventor:
This PR adds the distribcell "paths" - the paths through the CSG tree to reach each distribcell instance - to the `summary.h5` output files. The paths are the same strings composed for the `tallies.out` files are dumped as a string array to the `summary.h5`. The `openmc.Filter` class has a new `distribcell_paths` attribute which is populated by the `openmc.Summary` through the `openmc.StatePoint.link_with_summary(...)` method. This path provides a user with some nice metadata which can accelerate the processing of very large distribcell tally datasets.

I still have a little more testing to do (*i.e.*, run our Jupyter Notebooks) before I can confirm that this is good to go. In addition, I will submit a corollary PR to OpenCG soon which provides a few extension to make it compatible with the distribcell paths (which is needed for distribcell Pandas DataFrames with OpenCG info).


paulromano:
Python 3 test failures

wbinventor:
Alright, the Python 3 issue has been fixed and all of the tests should pass now. In addition, I've tested all of our Jupyter Notebooks...and updated them based to reflect 1) a few changes that should have been made in past PRs and 2) the latest version of OpenMOC (namely, our new mesh tally scheme to simplify the OpenMC-OpenMOC comparison of pinwise fission rates in "mgxs-part-iii.ipynb").

wbinventor:
Also for the record, the distribcell paths do give us an opportunity to eliminate OpenCG as a dependency when constructing Pandas DataFrames for distribcell filters (in one optionally chooses to embed the CSG tree / paths onto the DataFrame). I did not make the necessary changes to use the distribcell paths in place of OpenCG's CSG traversal algorithms here since 1) it is not entirely trivial and 2) it's unclear that anyone is using this feature aside from myself atm. But in theory it is very doable. 

On the flip side, the changes I did make do use the distribcell paths in a more limited way, but greatly accelerate the DataDrame construction (~1e3x faster) for distribcell tallies in large geometries such as the BEAVRS model.

openmc/filter.py
```diff
@@ -45,6 +45,9 @@ class Filter(object):
     stride : Integral
         The number of filter, nuclide and score bins within each of this
         filter's bins.
+    distribcell_paths : list of str
+        The paths traversed through the CSG tree to reach each distribcell
+         instance (for 'distribcell' filters only)
```
smharper: There's an extra space in this indentation

src/summary.F90
```diff
@@ -593,6 +598,25 @@ subroutine write_tallies(file_id)
           call write_dataset(filter_group, "bins", t%filters(j)%int_bins)
         end if
 
+        ! Write paths to reach each distribcell instance
+        if (t%filters(j)%type == FILTER_DISTRIBCELL) then
```
smharper: Spaces around these `%`

src/summary.F90
```diff
@@ -534,6 +535,10 @@ subroutine write_tallies(file_id)
     type(RegularMesh), pointer :: m
     type(TallyObject), pointer :: t
 
+    integer                     :: offset   ! distibcell offset
+    character(100), allocatable :: paths(:) ! array of distribcell paths
+    character(100)              :: path    ! temporary distribcell path
```
smharper: We should use constant parameters for these lengths.  `MAX_LINE_LEN` is probably the most appropriate based on it's name alone.  Naturally, you should also change the `path` variable in `output % get_label`

openmc/filter.py
wbinventor: Got it

src/summary.F90
wbinventor: Oh wow, I'm picking up bad habits

src/summary.F90
wbinventor: agreed and done

smharper:
Looks good!  Merged

smharper:
Shoot, I forgot something.  Can you please make another PR that updates docs/source/userguide/output/summary.rst?

paulromano:
@wbinventor With the distribcell paths embedded in summary.h5, do we even necessarily need OpenCG now to show distribcell information in the Pandas dataframe?

wbinventor:
That's precisely what I was referring to in one of my comments on this PR.
Wr should be able to get away from OpenCG for this now, but it will take a
little non-trivial refactoring to do so. But yes, it is certainly doable.

PR #618, opened 2016-03-25
Windowed multipole

smharper:
Hey, guys.  @cjosey is quite busy with depletion so I'm opening this PR to replace #560 and take multipole to the finish line.

### Verification
@cjosey wrote a [this paper](http://dx.doi.org/10.1016/j.jcp.2015.08.013) publishing the WMP method, with some verification against pointwise for a BEAVRS assembly.  I'm currently using WMP on the Mosteller Doppler-defect benchmark, and it agrees with pointwise results from JEFF-3.2 and the MCNP cross sections.  Exhibit A:

mosteller_bench

### UI details
The user needs a windowed multipole library.  It comes in the form of a directory with a bunch of ZZAAA.h5 files where Z and A are the atomic number and mass of the relevant isotope.  This library can be specified by the environment variable `OPENMC_MULTIPOLE_LIBRARY`, or with the `openmc.SettingsFile.multipole_library` attribute, or with the `<multipole_library>` XML node.  There's a new `get_multipole_data.py` script to download an example library (from my Athena locker).  The library is admittedly a bit primative, but I hope we can merge this PR anyway and iterate on the library later.

The user can specify the temperature of a cell in the geometry.  They can actually specify many temperatures to get a "distributed temperature" capability similar to our distributed material capability.  They also need to explicitly turn on the WMP capability.  In Python land, that looks like
```python
sets_file = openmc.SettingsFile()
sets_file.use_windowed_multipole=True
c11 = openmc.Cell(cell_id=11)
c11.temperature = [500, 0, 700, 800]
```
or in XML
```xml
<settings>
    <use_windowed_multipole>True</use_windowed_multipole>
</settings>
<geometry>
    <cell id="11" material="2" region="-10000" temperature="500 0 700 800" universe="11" />
</geometry>
```
If no cell temperature is provided, OpenMC will try to figure it out by looking at the metadata for the pointwise cross sections of the material in that cell.


.travis.yml
```diff
@@ -45,6 +45,9 @@ before_script:
   - cat nndc_xs/nndc.tar.gza* | tar xzvf -
   - rm -rf nndc_xs
   - export OPENMC_CROSS_SECTIONS=$PWD/nndc/cross_sections.xml
+  - wget http://web.mit.edu/smharper/Public/multipole_lib.tar.gz
```
paulromano: I feel we should have this stored somewhere a little more permanent. We have NNDC binaries stored in @bhermanmit's nndc_xs github repo; can we just create a github repo for this too?

data/get_nndc_data.py
```diff
@@ -84,7 +84,7 @@
         raise IOError("MD5 checksum for {} does not match. If this is your first "
                       "time receiving this message, please re-run the script. "
                       "Otherwise, please contact OpenMC developers by emailing "
-                      "openmc-users@googlegroups.com.")
+                      "openmc-users@googlegroups.com.".format(f))
```
paulromano: Oops, thanks!

src/input_xml.F90
```diff
@@ -1088,6 +1105,20 @@ subroutine read_settings_xml()
       end select
     end if
 
+    ! Check to see if windowed multipole functionality is requested
+    if (check_for_node(doc, "use_windowed_multipole")) then
+      call get_node_value(doc, "use_windowed_multipole", temp_str)
+      select case (to_lower(temp_str))
+      case ('true', 't', '1', 'y')
```
paulromano: For XML, I generally follow the [XML schema definition](https://www.w3.org/TR/xmlschema-2/#dt-boolean) of datatypes, meaning for a boolean the following values are valid: {'true', 'false', 1, 0}.

docs/source/usersguide/input.rst
```diff
@@ -281,6 +281,18 @@ based on the recommended value in LA-UR-14-24530_.
...
+``<multipole_library>`` Element
+-------------------------------
+
+The ``<multipole_library>`` element indicates the directory containing a
+windowed multipole library. If a windowed multipole library is available,
+OpenMC can use it for on-the-fly Doppler-broadening of resolved resonance range
+cross sections. If this element is absent from the settings.xml file, the
+:envvar:`OPENMC_MULTIPOLE_LIBRARY` environment variable will be used.
+
+  .. note:: The <use_windowed_multipole> element must also be set to "True"
```
paulromano: lowercase "true"

src/input_xml.F90
```diff
@@ -164,12 +165,28 @@ subroutine read_settings_xml()
       end if
     end if
 
+    ! Find the windowed multipole library
+    if (run_mode /= MODE_PLOTTING) then
+      if (.not. check_for_node(doc, "multipole_library") .and. &
+           run_mode /= MODE_PLOTTING) then
```
paulromano: redundant check of run_mode here

src/input_xml.F90
```diff
@@ -164,12 +165,28 @@ subroutine read_settings_xml()
...
+        ! No library location specified in settings.xml, check
+        ! environment variable
+        call get_environment_variable("OPENMC_MULTIPOLE_LIBRARY", env_variable)
+        path_multipole = trim(env_variable)
```
paulromano: I think we ought to have a check here whether this is empty and if so write a warning. Otherwise, if a user sets `<use_windowed_multipole>true</use_windowed_multipole>` but has no library specified, it just runs as normal with no warning.

tests/test_multipole/test_multipole.py
```diff
@@ -0,0 +1,134 @@
...
+import os
+import sys
+sys.path.insert(0, os.pardir)
+from testing_harness import TestHarness, PyAPITestHarness
+import openmc
+from openmc.stats import Box
+from openmc.source import Source
+
+
+class DistribmatTestHarness(PyAPITestHarness):
```
paulromano: distribmat?

paulromano: Also, perhaps we should check that OPENMC_MULTIPOLE_LIBRARY is set and if not raise an error saying that it is not. Otherwise, you just get 'results don't match' which is not very helpful.

openmc/universe.py
```diff
@@ -446,6 +472,13 @@ def create_surface_elements(node, element):
             # Call the recursive function from the top node
             create_surface_elements(self.region, xml_element)
 
+        if self.temperature is not None:
+            if isinstance(self.temperature, Iterable):
+                element.set("temperature", ' '.join(
+                            [str(t) for t in self.temperature]))
```
paulromano: I'm sure we're not consistent about this, but it's generally better to use a generator expression, i.e.
```
' '.join(str(t) for t in self.temperature)
```

src/ace.F90
```diff
@@ -418,6 +423,68 @@ subroutine read_ace_table(i_table, i_listing)
...
+!===============================================================================
+
+  subroutine read_multipole_data(i_table)
+
+    integer, intent(in) :: i_table   ! index in nuclides/sab_tables
+
+    logical       :: file_exists   ! does multipole library exist?
+    character(7)  :: readable      ! is multipole library readable?
+    character(6)  :: zaid_string   ! String of the ZAID
+    character(MAX_FILE_LEN+9)  :: filename  ! path to multipole xs library
```
paulromano: +9?

src/multipole.F90
```diff
@@ -0,0 +1,184 @@
...
+
+      call read_dataset(group_id, "MT_list", MT)
+
+      call close_group(group_id)
+
+      accumulated_fission = .false.
+
+      ! Loop over each MT entry and load it into a reaction.
+      do i = 1, NMT
+        write(MT_string, '(I3.3)') MT(i)
```
paulromano: You can use the `zero_padded` function from string.F90 here instead of an internal write.

src/multipole.F90
```diff
@@ -0,0 +1,184 @@
...
+                end if
+
+                deallocate(nuc % reactions(j) % sigma)
+                allocate(nuc % reactions(j) % sigma(nuc % n_grid))
+
+                call read_dataset(group_id, "MT_sigma", nuc % reactions(j) % sigma)
+                call read_dataset(group_id, "Q_value", nuc % reactions(j) % Q_value)
+                call read_dataset(group_id, "threshold", nuc % reactions(j) % threshold)
+                nuc % reactions(j) % threshold = 1 ! TODO: reconsider implications.
+                nuc % reactions(j) % Q_value = nuc % reactions(j) % Q_value / 1.0D6
```
paulromano: `1.0e6_8` is the prevailing style

src/multipole.F90
```diff
@@ -0,0 +1,184 @@
...
+                  accumulated_fission = .false.
+                end if
+
+                deallocate(nuc % reactions(j) % sigma)
+                allocate(nuc % reactions(j) % sigma(nuc % n_grid))
+
+                call read_dataset(group_id, "MT_sigma", nuc % reactions(j) % sigma)
+                call read_dataset(group_id, "Q_value", nuc % reactions(j) % Q_value)
+                call read_dataset(group_id, "threshold", nuc % reactions(j) % threshold)
+                nuc % reactions(j) % threshold = 1 ! TODO: reconsider implications.
```
paulromano: Can you elaborate on the TODO?

src/multipole.F90
```diff
@@ -0,0 +1,184 @@
...
+      allocate(nuc % nu_fission(nuc % n_grid))
+      allocate(nuc % absorption(nuc % n_grid))
+
+      nuc % total = ZERO
+      nuc % absorption = ZERO
+      nuc % fission = ZERO
+
+      ! Read in new energy axis (converting eV to MeV)
+      call read_dataset(group_id, "energy_points", nuc % energy)
+      nuc % energy = nuc % energy / 1.0D6
```
paulromano: `1.0e6_8`

src/multipole.F90
```diff
@@ -0,0 +1,184 @@
...
+              if (nuc % reactions(j) % MT == MT(i)) then
+                ! Match found
+
+                ! Individual Fission components exist, so remove the combined
+                ! fission cross section.
+                if ( (MT(i) == N_F .or. MT(i) == N_NF .or. MT(i) == N_2NF &
+                      .or. MT(i) == N_3NF) .and. accumulated_fission) then
+                  nuc % total = nuc % total - nuc % fission
+                  nuc % absorption = nuc % absorption - nuc % fission
+                  nuc % fission = 0.0_8
```
paulromano: Use `ZERO` here, and generally when assigning to an array, it's smart to write `nuc % fission(:) = ...` to indicate that no reallocation occurs (strictly speaking the F2003+ standard allows you to reallocate an allocatable variable on assignment).

src/cross_section.F90
```diff
@@ -141,99 +157,133 @@ subroutine calculate_nuclide_xs(i_nuclide, i_sab, E, i_mat, i_nuc_mat, i_log_uni
...
-      i_grid = mat % nuclide_grid_index(i_nuc_mat, i_log_union)
+      ! Call multipole kernel
+      call multipole_eval(nuc % multipole, E, sqrtkT, sigT, sigA, sigF)
 
-    case (GRID_LOGARITHM)
-      ! Determine the energy grid index using a logarithmic mapping to reduce
-      ! the energy range over which a binary search needs to be performed
+      micro_xs(i_nuclide) % total = sigT
+      micro_xs(i_nuclide) % absorption = sigA
+      micro_xs(i_nuclide) % elastic = sigT - sigA
```
paulromano: What happens with inelastic scattering reactions? Can we really guarantee that there are no threshold inelastic reactions that go into the resolved resonance region (IIRC, I've seen evaluations where inelastic reactions can have thresholds within the resolved range)? Might need @cjosey to answer this one.

tests/test_filter_distribcell/case-1/test.py
```diff
@@ -0,0 +1,7 @@
+import openmc
+
+
+su = openmc.Summary('summary.h5')
+sp = openmc.StatePoint('statepoint.1.h5')
+sp.link_with_summary(su)
+print(sp.tallies[1].get_pandas_dataframe(summary=su))
```
paulromano: Oops?

src/input_xml.F90
smharper: Good point, but I decided to put this check in `ace % read_ace_xs` where we can also catch the case of a user specifying the environment variable, but specifying it incorrectly.

src/ace.F90
smharper: `path_multipole` is a `character(MAX_FILE_LEN)` so adding the `+9` (for ZZZAAA.h5) avoids the character truncation warning.

src/multipole.F90
smharper: Using `zero_padded` causes a character truncation warning.  I'd prefer the write statement to the compiler warning.  I did remove the unnecessary `MT_string` variable though.

src/multipole.F90
smharper: Hmm, I cannot elaborate on that TODO.  I'll get back to you on that.

src/cross_section.F90
cjosey: Just did analysis.  I was under the impression that inelastic scattering below the RRR/URR cutoff was not supposed to happen.  Basically, since the LRF=3 form of Reich-Moore does not support inelastic channels, that anyone using LRF=3 as a curve fitting basis would be doing things incorrectly.  Whether or not it is correct is irrelevant, though, since at least Fe-57 has MT=4 below RRR/URR cutoff.

Currently in discussion on how to fix this, but right now, nuclides in which MT=4 is below RRR/URR (and similarly, all non-(n,gamma) MT=101(?) channels) should just be treated as pointwise and should not be converted into windowed multipole, as the solution is likely to require something other than windowed multipole to fix, such as Pablo's new pointwise algorithm.

src/cross_section.F90
```diff
@@ -1,19 +1,31 @@
...
+  use random_lcg,       only: prn, future_prn, prn_set_stream
+  use sab_header,       only: SAlphaBeta
+  use search,           only: binary_search
 
   implicit none
 
+  ! Allocatable arrays for multipole that are allocated once for speed purposes
+  complex(8), allocatable :: sigT_factor(:)
+  real(8), allocatable :: twophi(:)
+  real(8), allocatable :: broadened_polynomials(:)
```
paulromano: Is there any reason these arrays can't just be local variables in `multipole_eval`?

smharper: They are always the same size so we only want to allocate them once per run.  I should probably move that to something called in initialization.

paulromano: To be clear, the size of the arrays is known at the time of the subroutine invocation, which means they can be fixed-size local arrays that are allocated on the stack. Thus, the allocation is "free" (all it entails is incrementing a stack pointer) and has the benefit that you don't need to make more global, threadprivate variables. It should also exhibit better spatial locality in memory.

src/math.F90
```diff
@@ -702,6 +718,95 @@ function watt_spectrum(a, b) result(E_out)
...
+    else
+      wv = -conjg(faddeeva_w(conjg(z), relerr))
+    end if
+
+  end function faddeeva
+
+!===============================================================================
+! BROADEN_WMP_POLYNOMIALS Doppler broadens the windowed multipole curvefit.  The
+! curvefit is a polynomial of the form
+! a/En + b/sqrt(En) + c + d sqrt(En) ...
```
paulromano: It would be great to put a reference in this comment to the JCP paper so someone looking at the source can figure out where these equations come from.

src/math.F90
```diff
@@ -702,6 +718,95 @@ function watt_spectrum(a, b) result(E_out)
...
+    ! The MIT Faddeeva function evaluates w(z) = exp(-z^2)erfc(-iz). These
+    ! two forms of the Faddeeva function are related by a transformation.
+    !
+    ! If we call the integral form w_int, and the function form w_fun:
+    ! For imag(z) > 0, w_int(z) = w_fun(z)
+    ! For imag(z) < 0, w_int(z) = -conjg(w_fun(conjg(z)))
+
+    relerr = ZERO
+    if (aimag(z) > ZERO) then
+      wv = faddeeva_w(z, relerr)
```
paulromano: Probably worth putting in a comment that the MIT faddeeva function implementation interprets zero as machine epsilon.

paulromano: Also a question for @cjosey -- I know you were looking at alternate Faddeeva function implementations for a while, but I see you've settled on just using the MIT one here. Can you talk about that decision a bit? Or is it discussed in one of your papers?

src/cross_section.F90
```diff
@@ -508,6 +559,189 @@ pure function find_energy_index(mat, E) result(i)
...
+
+    ! Locate us.
+    i_window = floor((sqrtE - sqrt(multipole % start_E)) / multipole % spacing &
+         + ONE)
+    startw = multipole % w_start(i_window)
+    endw = multipole % w_end(i_window)
+
+    ! Fill in factors.
+    if (startw <= endw) then
+      call fill_factors(multipole, sqrtE, sigT_factor, twophi, &
```
paulromano: Looks like `twophi` is never actually used here in `multipole_eval`, so in fact it could be a local variable in `fill_factors`

src/math.F90
cjosey: So, the cost of MIT W is 50% higher than the WHOPPER W when just evaluated on a long array of Z values.  MIT W costs 12% of a cross section lookup.  As such, at best, we'd see a 4% speed increase switching.  In reality, it would be far closer to 1 or 2%, and last time I actually tested it, there was no measurable difference in speed.

Choice of Fadeeva has a moderate impact on accuracy. In the absolute worst case, WHOPPER W would introduce 10^-3 relerr on XS.  QuickW was nearer to 10^-2. 

As I could not find an algorithm that was comfortably 10^-5 everywhere while being faster than MIT W, I stuck with it.

paulromano:
@cjosey Another technical question for you -- in all the papers, the multipole form shows a correction term that becomes important at low energies. How do we get around not having to calculate that term at thermal energies?

cjosey:
This is one of the few reasons windowed multipole is more accurate than raw multipole.  At low energies, my optimization algorithm adds no poles to a window and does the entire domain using the curve fits.  These curve fits are exactly broadened and no terms are missing.

This does pose some issues.  Some nuclides (U235, some Gd isotope I forgot, etc.) have extremely low E resonances.  For these, the curve fit order has to become very large in order to get an accurate representation of the left side of the resonance.  I believe U235 is usually best between 4th and 5th order, and that one Gd nuclide had to go as high as 9.

smharper:
I added some basic theory documentation.  [It can be previewed here](http://web.mit.edu/smharper/www/methods/cross_sections.html#windowed-multipole-representation).  @cjosey, could you please check it for accuracy?

src/cross_section.F90
smharper: It seems like these are perfectly happy as local arrays.  A nice side-effect of making these local is that I was also able to eliminate the `max_L`, `max_poly`, and `max_poles` global variables

src/math.F90
smharper: Do you still want a JCP reference here now that it's been added to the docs?

src/cross_section.F90
smharper: To sidestep the issue just a bit, I added a note to the theory doc that explicitly states that we are neglecting inelastic scattering in the resolved range.

src/multipole.F90
```diff
@@ -0,0 +1,188 @@
...
+                deallocate(nuc % reactions(j) % sigma)
+                allocate(nuc % reactions(j) % sigma(nuc % n_grid))
+
+                call read_dataset(group_id, "MT_sigma", &
+                                  nuc % reactions(j) % sigma)
+                call read_dataset(group_id, "Q_value", &
+                                  nuc % reactions(j) % Q_value)
+                call read_dataset(group_id, "threshold", &
+                                  nuc % reactions(j) % threshold)
+                nuc % reactions(j) % threshold = 1 ! TODO: reconsider implications.
```
smharper: Just pinging this TODO again so I don't forget about it.  My understanding is that the cross sections in the library have been set to zero where necessary to account for threshold reactions... but I still need to look into this more.

src/math.F90
paulromano: Yes, I think it would still be useful for someone looking through the source code.

docs/source/methods/cross_sections.rst
```diff
@@ -63,6 +63,53 @@ Other Methods
...
+---------------------------------
+
+In addition to the usual pointwise representation of cross sections, OpenMC
+offers support for an experimental data format called windowed multipole (WMP).
+This data format requires less memory than pointwise cross sections, and it
+allows on-the-fly Doppler broadening to arbitrary temperature.
+
+The multipole method was introduced by [Hwang]_ and the faster windowed
+multipole method by [Josey]_.  In the multipole format, cross section resonances
+are represented by poles, `p_j`, and residues, `r_j`, in the complex plane.  The
```
paulromano: Use ``:math:`p_j` `` instead of `` `p_j` ``. Same goes for r_j

docs/source/methods/cross_sections.rst
```diff
@@ -63,6 +63,53 @@ Other Methods
...
+.. math::
+   \sigma(E, T) = \frac{1}{2 E \sqrt{\xi}} \sum_j \text{Re} \left[i r_j
+   \sqrt{\pi} W(z) \right]
+.. math::
+   z = \frac{\sqrt{E} - p_j}{2 \sqrt{\xi}}
+.. math::
+   \xi = \frac{k_B T}{4 A}
+
+where `T` is the temperature of the resonant scatterer, `k_B` is the Boltzmann
+constant, `A` is the mass of the target nucleus, and `W` is the Faddeeva
```
paulromano: Use `:math:` on above two lines

docs/source/methods/cross_sections.rst
```diff
@@ -63,6 +63,53 @@ Other Methods
...
+.. math::
+   \sigma(E, T=0\text{K}) = \frac{1}{E} \sum_j \text{Re} \left[
+   \frac{i r_j}{\sqrt{E} - p_j} \right]
+
+Assuming free-gas thermal motion, cross sections in the multipole form can be
+analytically Doppler broadened to give (after some approximation) the form:
+
+.. math::
+   \sigma(E, T) = \frac{1}{2 E \sqrt{\xi}} \sum_j \text{Re} \left[i r_j
+   \sqrt{\pi} W(z) \right]
```
paulromano: I would personally like to see discussion about the second term that appears as that was one point that I was confused about looking at the source code. But, I understand you've gone above and beyond by even including this extra documentation (thank you!) so if you're reluctant to go into too much detail that's ok. If you'd rather defer, we can just open an issue for it.

smharper: Which term are you referring to?  The sigma_T factor or the polynomial?

smharper: Or are you referring to the `C` term?  Neglecting that term is what I meant by "(after some approximation)"...

paulromano: Yes, I was referring to `C`. The fact that the `C` term gets absorbed into the polynomials is, to me, an important point.

cjosey: I'm fixing up the documentation.  I'll PR it here in a few hours.

smharper:
@cjosey improved the documentation.  Right now the only open comments are adding the JCP reference to the source code, and taking care of that TODO.  I will try to address those shortly...

smharper:
I should mention the docs I have in [my athena locker](http://web.mit.edu/smharper/www/methods/cross_sections.html#windowed-multipole-representation) have been updated.

paulromano:
One more thing that we ought to include in this pull request is documentation on the format of the WMP HDF5 file. Having seen one too many instances of codes that don't have any documentation on their data formats, I'd really like for us to not commit that same sin.

cjosey:
I don't mind taking up that mantle since I wrote the format, but this leads to three questions.  The first is whether or not this format is close to final.  I did a reorganization and got essentially no comments on that front, so I will assume so.  The second is whether or not much will change when absorbed into the ACE replacement.  I do not believe so (horizontal concatenation and all that), but I thought I'd bring up the concern.  The third is if there is a documentation structure you'd like me to follow, perhaps similar to the endf102 document but with less obtuse-ness?

paulromano:
It will probably change but I'm not really concerned about that. As long as the format and its documentation change simultaneously it's fine.

As for format, probably the easiest thing is to use the same format that describes our output files, e.g. look at `docs/source/usersguide/output/statepoint.rst`. The format I will end up using for the ACE replacement is going to be a bit different, mostly because I make judicious use of attributes in HDF5. You can see a sneak peak [here](https://github.com/paulromano/openmc/blob/pyapi-data/docs/source/usersguide/data_hdf5.rst).

paulromano:
That being said, expecting that it will change means we should version the format itself.

cjosey:
I have PR'ed some rudimentary documentation to @smharper's repo.  I probably left a bit too much implicit understanding in my documentation, so let me know if something does not make sense.

smharper:
When I merged @cjosey's docs, I made a new directory for all our I/O files and moved the userguide/output/* files and /userguide/mgxs_library into it.

smharper:
That last commit makes a big change to the way non-RR data is handled.  Rather than relying on an energy grid and inelastic cross sections from the WMP library, it will use the existing values from the ACE library.  I think this is preferable to the previous implementation because it confines the effects of WMP to the `multipole % start_E` to `multipole % end_E` range.  This required some tedious remapping of the energy grid and therefore the cross sections.  I'd appreciate a careful review of the changes in multipole.F90 to make sure I haven't made an off-by-one error or missed some important physics.

The results in test_multipole had a last-digit change.  I've rerun one of my application problems with many more nuclides, and I got the same k-eff resolved to 300 pcm.

smharper:
That new energy grid scheme has an error because of #652.  It looks like `multipole % start_E` always coincides with `i_grid = 1`, so low energy neutrons will get cross sections extrapolated from `i_grid = 1` and the bogus `i_grid = 2` that I created.  Fixing this issue will either be linked with resolving #652 or leaving two energy grid points to extrapolate from.

openmc/settings.py
```diff
@@ -75,6 +75,11 @@ class Settings(object):
         continuous-energy calculations and
         :envvar:`OPENMC_MG_CROSS_SECTIONS` will be used for multi-group
         calculations to find the path to the XML cross section file.
+    multipole_library : str
+        Indicates the path to a directory containing a windowed multipole
+        cross section library. If it is not set, the
+        :envvar:`OPENMC_MULTIPOLE_LIBRARY' environment variable will be used. A
```
paulromano: Need a back-tick instead of an apostrophe at the end of LIBRARY.

docs/source/io_formats/index.rst
```diff
@@ -1,13 +1,15 @@
-.. _usersguide_output:
+.. _io_file_formats:
 
-===================
-Output File Formats
-===================
+===============
+IO File Formats
```
paulromano: How about just "File Formats" or "File Format Specifications"?

docs/source/io_formats/data_wmp.rst
```diff
@@ -0,0 +1,92 @@
+.. _io_data_wmp:
+
+==========================================
+The Windowed Multipole Library Format v0.2
```
paulromano: I'd make this "Windowed Multipole Library Format" and then have a sentence below saying this is version 0.2. More importantly, the library itself really should have the version information within it!

docs/source/io_formats/data_wmp.rst
```diff
@@ -0,0 +1,92 @@
...
+    - **curvefit** (*double[][][]*)
+        Curve fit coefficients. Indexed by (reaction type, coefficient index,
+        window index).
+    - **data** (*complex[][]*)
+        Complex poles and residues.  Each pole has a corresponding set of
+        residues.  For example, the `i`th pole and corresponding residues are
+        stored as `data[:,i] = [pole, residue_1, residue_2, ...]`.  The 
+        residues are in the order: total, competitive if present, absorption,
+        fission. Complex numbers are stored by forming a type with `"r"` and
+        `"i"` identifiers, similar to how `h5py` does it.
```
paulromano: This description renders weirdly. Inline literals should use double backticks. Inline math should use `` :math:`...` ``. The h5py reference could be a link to h5py documentation.

docs/source/io_formats/data_wmp.rst
```diff
@@ -0,0 +1,92 @@
...
+    - **start_E** (*double*)
+        Lowest energy the windowed multipole part of the library is valid for.
```
paulromano: Everything is ordered alphabetically except start_E

docs/source/io_formats/data_wmp.rst
```diff
@@ -0,0 +1,92 @@
...
+        scattering radius for a given :math:`l`.
+    - **spacing** (*double*)
+        .. math::
+            \frac{\sqrt{E_{max}}- \sqrt{E_{min}}}{n_w}
+        
+        Where :math:`E_{max}` is the maximum energy the windows go up to.  This
+        is not equivalent to the maximum energy for which the windowed multipole
+        data is valid for.  It is slightly higher to ensure an integer number of
+        windows. :math:`E_{min}` is the minimum energy and equivalent to
+        `start_E`, and :math:`n_w` is the number of windows, given by `windows`.
```
paulromano: start_E and windows should probably be double backtick-ed

docs/source/io_formats/data_wmp.rst
```diff
@@ -0,0 +1,92 @@
...
+        The pole to start from for each window.
+    - **w_end** (*int[]*)
+        The pole to end at for each window.
+    - **windows** (*int*)
+        Number of windows.
+
+**/nuclide/reactions/MT<i>**
+    - **MT_sigma** (*double[]*) -- Cross section value for this reaction.
+    - **Q_value** (*double*) -- Energy released in this reaction, in eV.
+    - **threshold** (*int*) -- The first non-zero entry in `MT_sigma`.
```
paulromano: Double backticks on MT_sigma?

docs/source/io_formats/index.rst
```diff
@@ -1,13 +1,15 @@
...
+===============
 
 .. toctree::
    :numbered:
    :maxdepth: 3
 
+   data_wmp
+   mgxs_library
```
paulromano: It's a little weird to me to have an XML format here alongside a bunch of binary formats.

data/get_multipole_data.py
```diff
@@ -0,0 +1,125 @@
...
+import subprocess
+import sys
+import tarfile
+import glob
+import hashlib
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-b', '--batch', action = 'store_true',
+                    help = 'supresses standard in')
```
paulromano: No spaces around the = sign for action and help (you can fix in get_nndc_data.py too if you wish)

data/get_multipole_data.py
```diff
@@ -0,0 +1,125 @@
...
+                      "time receiving this message, please re-run the script. "
+                      "Otherwise, please contact OpenMC developers by emailing "
+                      "openmc-users@googlegroups.com.".format(f))
+
+# ==============================================================================
+# EXTRACT FILES FROM TGZ
+
+for f in files:
+    fname = f[:-9] if f.endswith('?raw=true') else f
+    if not fname in filesComplete:
```
paulromano: `if fname not in filesComplete`

docs/source/io_formats/index.rst
nelsonag: Hint, hint, nudge, nudge? 😉

nelsonag: By that I mean I still intend to get on that h5 mgxs lib format.

src/multipole.F90
```diff
@@ -73,113 +77,138 @@ subroutine multipole_read(filename, multipole, i_table)
...
-                  nuc % absorption(:) = nuc % absorption &
-                                        + nuc % reactions(j) % sigma
-                end if
-              end if
-            end do
-        end select
-
-        call close_group(group_id)
+      !=========================================================================
+      ! Remove the uneeded/inconsitent pointwise data.  This step enforces the
```
paulromano: unneeded/inconsistent

src/multipole.F90
```diff
@@ -73,113 +77,138 @@ subroutine multipole_read(filename, multipole, i_table)
...
+
+      ! Adjust the nu-fission cross section.
+      allocate(new_xs(new_n_grid))
+      new_xs(1:cut1) = nuc % nu_fission(1:cut1)
+      new_xs(cut1+1) = (ONE - f1) * nuc % nu_fission(cut1) &
+                       + f1 * nuc % nu_fission(cut1+1)
+      new_xs(cut1+2:cut1+3) = ZERO
+      new_xs(cut1+4) = (ONE - f2) * nuc % nu_fission(cut2-1) &
+                       + f2 * nuc % nu_fission(cut2)
+      new_xs(cut1+5:new_n_grid) = nuc % nu_fission(cut2:nuc % n_grid)
```
paulromano: It doesn't look like nu_fission has actually been calculated yet at this point, so it probably suffices to just resize nu_fission.

docs/source/io_formats/index.rst
paulromano: Yeah, geez Adam, what are we paying you for!!

paulromano:
@smharper Your new changes look ok to me. I looked through the "cutting" method pretty carefully and can't find anything obviously wrong with it. I think this is good to go pending resolution of the last series of comments.

smharper:
I gave up on the attempt to rework the ACE data in-place.  I encountered a lot of issues with reactions not being sampled, and I assume this code will change when we move to all-HDF5 data anyway.  I've also addressed @paulromano's most recent comments.

@cjosey, I noticed that boron isn't in the library.  Is WMP able to handle (n, alpha) reactions, or is that precluded thanks to R-matrix's limitations on charged particles?

cjosey:
Boron isn't in the library for two reasons.  One, if we had charged particle resonance data, we couldn't convert it yet.  Pablo's actively working on this now.  Two, we don't have resonance data, so we couldn't convert it even if we wanted to.  Both of these would be solved by the new pointwise Doppler broadening algorithm, but that's a ways away.

docs/source/io_formats/data_wmp.rst
```diff
@@ -0,0 +1,102 @@
...
+        Curve fit coefficients. Indexed by (reaction type, coefficient index,
+        window index).
+    - **data** (*complex[][]*)
+        Complex poles and residues. Each pole has a corresponding set of
+        residues. For example, the :math:`i`-th pole and corresponding residues
+        are stored as
+        
+        .. math::
+        \text{data}[:,i] = [\text{pole},~\text{residue}_1,~\text{residue}_2,
+        ~\ldots]`
```
paulromano: You need to separate this by a blank line and indent the equation. Also, there is a backtick at the end that looks like it should be removed.

openmc/cell.py
```diff
@@ -58,7 +58,6 @@ class Cell(object):
     rotation : Iterable of float
         If the cell is filled with a universe, this array specifies the angles
         in degrees about the x, y, and z axes that the filled universe should be
-        rotated. The rotation applied is an intrinsic rotation with specified
```
paulromano: Oops?

src/cross_section.F90
```diff
@@ -133,107 +140,142 @@ end subroutine calculate_xs
...
     nuc => nuclides(i_nuclide)
     mat => materials(i_mat)
 
-    ! Determine index on nuclide energy grid
-    select case (grid_method)
-    case (GRID_MAT_UNION)
+    ! If MP, don't interpolate, it's all already baked in.
+    if (nuc % mp_present .and. &
+         (E >= nuc % multipole % start_E/1.0e6_8 .and.&
+          E <= nuc % multipole % end_E/1.0e6_8)) then
```
paulromano: Can we break this into two `if`s (one to check `mp_present` and another nested one to check energies)? Fortran doesn't guarantee short-circuit evaluation so this is not safe if no multipole data is present.

smharper:
Comments addressed

paulromano:
@smharper Thanks for carrying this to the finish line!

PR #623, opened 2016-04-08
Updated MGXS Notebooks and Minimalistic Python Imports

wbinventor:
This PR updates all of our MGXS Jupyter Notebooks to use `openmoc.opencg_compatible` in place of `openmoc.compatible` to conform with a recent change to the OpenMOC codebase. Furthermore, I reduced the number of Python imports used in the notebooks and Python example inputs. In particular, the current examples some variation of the following:

```python
import openmc
from openmc.source import Source
from openmc.stats import Box
from openmc.summary import Summary
from openmc.statepoint import StatePoint
```

All but the first imports are superfluous, so they have been removed.


paulromano:
I'd personally like to see a more logical organization of packages/modules/classes for OpenMC. Right now almost everything gets imported into a single namespace. That being said, I'm not sure what the best organization would be. I was thinking we could have an `openmc.geometry` package that includes everything geometry-related, but then I'm not sure about materials, settings, and tallies-related classes.

Two more specific comments/questions -- we should have `from openmc.source import *` in `openmc/__init__.py` (right now it is indirectly imported there). Also, I don't see how `Box` would end up in the `openmc` namespace, so that would require a separate import. Am I missing something there?

wbinventor:
@paulromano - unfortunately, I don't know how to answer your question about how `Box` ends up in the `openmc` namespace. After looking through our `__init__.py` I'm confused about this as well. Without digging into this more, all I know is that it works for some unknown reason!

As for the logical organization of packages/modules/classes, I would initially agree with you but think this should be done in a principled fashion. The reason I say this is that most uses cases I have encountered require don't simply require one of geometric primitives, materials, settings or tallies. Rather, most use cases (from an input standpoint) I've seen require all of these simultaneously. Requiring separate imports would seem to be 

In the case of something like `scipy` it makes sense to separate out different algorithms into different modules. For example, there is not an obviously strong correlation between use cases for `scipy.optimize` and `scipy.stats`. In our case, however, our API is much more tightly coupled. If anything, it might make the most sense to couple together code into `openmc.input` and `openmc.output` to strike a tradeoff between code organization and import simplicity for our users. But this demarcation doesn't really make sense either since the geometry and materials in `openmc.input` would also be used by `openmc.output` through the `Summary`.

Although it's not clear that we need to separate our API into modules, I'm like the idea if there were a way to do so which wouldn't *always* lead to an "import explosion" of 4+ imports atop every script which uses the API.

paulromano:
@wbinventor All good points that I tend to agree with. I would point out that import explosion is not absolutely necessary, i.e. even with just `import openmc`, you could still reference classes in subpackages via `openmc.subpackage.Class` or something like that, but at that point it does get very verbose. Anyway, you've convinced me we should leave most everything in the `openmc` namespace for now.

Re: the Box import, it actually doesn't work for me:
```
In [1]: import openmc

In [2]: openmc.Box
AttributeError                            Traceback (most recent call last)
<ipython-input-2-44053e964dbe> in <module>()
----> 1 openmc.Box

AttributeError: module 'openmc' has no attribute 'Box'
```

wbinventor:
The nested reference case is something I realized I neglected to mention,
it would be one alternative though I'm not yet convinced that it would be
preferable to one big namespace.

I was able to import `Box` the other day, I think try `openmc.stats.Box` in
cell 2. I *think* this is how I refactored all of the examples to work in
this PR as well.
On Apr 13, 2016 8:49 AM, "Paul Romano" <notifications@github.com> wrote:

@wbinventor <https://github.com/wbinventor> All good points that I tend to
agree with. I would point out that import explosion is not absolutely
necessary, i.e. even with just import openmc, you could still reference
classes in subpackages via openmc.subpackage.Class or something like that,
but at that point it does get very verbose. Anyway, you've convinced me we
should leave most everything in the openmc namespace for now.

Re: the Box import, it actually doesn't work for me:

In [1]: import openmc

In [2]: openmc.Box
AttributeError                            Traceback (most recent call last)
<ipython-input-2-44053e964dbe> in <module>()
----> 1 openmc.Box

AttributeError: module 'openmc' has no attribute 'Box'

—
You are receiving this because you were mentioned.

Reply to this email directly or view it on GitHub
<https://github.com/mit-crpg/openmc/pull/623#issuecomment-209415145>

wbinventor:
Also, as an aside I'll remark that I do see some future promise in
modularizing the API into sub-modules/packages for different use cases. For
example, if we wanted to provide pre-built reactor components /materials
for PWRs, BWRs, CANDUs, SFRs, TRISOs, etc. then we might include
`openmc.pwr`, `openmc.bwr`, `openmc.candu`, `openmc.sfr`, `openmc.triso`,
etc. modules. Just a thought, but this is where I currently see real
promise for demarcating code into modules by use case.

nelsonag:
Can confirm, last night I saw I was using Box via openmc.stats.Box

docs/source/pythonapi/examples/mgxs-part-ii.ipynb
```diff
@@ -34,8 +34,16 @@
...
      "output_type": "stream",
      "text": [
-      "/usr/local/lib/python2.7/dist-packages/IPython/kernel/__main__.py:11: QAWarning: pyne.rxname is not yet QA compliant.\n",
-      "/usr/local/lib/python2.7/dist-packages/IPython/kernel/__main__.py:11: QAWarning: pyne.ace is not yet QA compliant.\n"
+      "/usr/local/lib/python2.7/dist-packages/matplotlib-1.5.1+1178.ga40c9ec-py2.7-linux-x86_64.egg/matplotlib/__init__.py:884: UserWarning: axes.color_cycle is deprecated and replaced with axes.prop_cycle; please use the latter.\n",
+      "  warnings.warn(self.msg_depr % (key, alt_key))\n",
+      "/usr/local/lib/python2.7/dist-packages/matplotlib-1.5.1+1178.ga40c9ec-py2.7-linux-x86_64.egg/matplotlib/__init__.py:1362: UserWarning:  This call to matplotlib.use() has no effect\n",
+      "because the backend has already been chosen;\n",
+      "matplotlib.use() must be called *before* pylab, matplotlib.pyplot,\n",
+      "or matplotlib.backends is imported for the first time.\n",
```
paulromano: We should indeed heed this warning and put `%matplotlib inline` before importing pyplot. Can you change that on this notebook and all others?

paulromano:
Ah, gotcha, didn't realize you used the fully qualified `openmc.stats.Box`.

docs/source/pythonapi/examples/mgxs-part-ii.ipynb
wbinventor: Sure, I can certainly put that before importing `pyplot`. But it actually won't eliminate the warning message here since it is *not* due to the order of our imports in the notebook, but rather the way in which we setup matplotlib in OpenMOC to plot on headless environments.

wbinventor: Also interestingly enough the plots do not show up inline for mgxs-part-ii.ipynb and mgxs-part-iii.ipynb unless `%matplotlib inline` appears *after* the `openmoc` imports. It does work just fine, however, for our other notebooks, *but* the plot formatting is different than what is shown in our current documentation.

wbinventor: I just pushed updated notebooks which move `%matplotlib inline` ahead of `import matplotlib.*`. All notebooks except those two which use OpenMOC were affected.

Issue #629, opened 2016-04-24
Transposed Scatter XS from MGXS


friedmud:
Codes other than OpenMOC might expect Scatter XS's to be transposed so that they are "in on column" and "out on rows".  This is the natural choice inside a reactor physics application (so the application of the XS can be done as a vectorizable matvec with the scalar flux).

Would be nice to have the option in MGXS.

wbinventor:
Fixed with #638.

PR #638, opened 2016-04-30
Optional Transpose of ScatterMatrixXS Data

wbinventor:
This short PR is intended to address #629. As pointed out by @friedmud, the `ScatterMatrixXS` reports its data first indexed by indexed by incoming group then outgoing group. Although this is typically the way the matrix is expressed in textbooks, it is not how a deterministic code likely needs the data. Instead, it is more common to organize the data by outgoing then incoming group in order to compute the scattering source with a matrix-vector product in a cache friendly manner.

This PR adds a new `row_column` parameter to the `ScatterMatrix.get_xs(...)` method. This parameter can take on the values `"inout"` (default) or `"outin"` (as requested by @friedmud). I've chosen to leave the default as it is since I do think it matches the common notation used in the reactor physics community.

### DIFF
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -575,7 +575,8 @@ class Library(object):
         return subdomain_avg_library
 
     def build_hdf5_store(self, filename='mgxs.h5', directory='mgxs',
-                         subdomains='all', nuclides='all', xs_type='macro'):
+                         subdomains='all', nuclides='all', xs_type='macro',
+                         row_column='inout'):
         """Export the multi-group cross section library to an HDF5 binary file.
 
         This method constructs an HDF5 file which stores the library's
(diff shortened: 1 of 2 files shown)


wbinventor:
The `MGXS.build_hdf5_store(...)` method now also accepts `row_column` as a parameter so one can specify the order in which scattering matrices are store in HDF5 files.

friedmud:
Awesome Will!  Thanks!

nelsonag:
@wbinventor: should this also  'row_column' as a parameter to  `Library.build_hdf5_store(...)`?

nelsonag:
Ran the code and looked it over; it seems fine to me pending resolution of the above comment.

wbinventor:
@nelsonag - excellent point, I just added `row_column` as a parameter to `Library.build_hdf5_store(...)`.

nelsonag:
Great! Thanks @wbinventor!

PR #643, opened 2016-05-07
Integrated generation of Multi-Group Cross Section libraries with openmc.mgxs.Library

nelsonag:
This PR adds the capability to generate a multi-group cross section library for OpenMC's multi-group mode with @wbinventor's great `mgxs.Library` API.  This is done via the `mgxs.Library.write_mg_library(...)` routine.  

I've also created an example notebook showing this off.  This has been incorporated in to the docs.  

Finally, to make it a little easier for the user, I've made a list in the `mgxs.Library` module called 'OPENMC_MG_MGXS_TYPES` which can be used by the users to set the `mgxs_types` attribute of a `Library` object and to automatically get the types they'd need for a multi-group MC calculation.

Note this PR is based off of #611 and so that would need to be merged first (... or just skip to this one after reviewing that one).

PR #592, opened 2016-02-23
Add KappaFission capability to MGXS

friedmud:
Simply add the ability to compute `kappa-fission` based group XS.  I hate code duplication so I did this by generalizing the existing `nu-fission` object and deriving `NuFission` and `KappaFission` from it.  Let me know if you prefer the code duplication instead and I'll switch it around.

PR #644, opened 2016-05-09
Scattering Matrix Moments

wbinventor:
This PR introduces higher order scattering matrix moments to the `openmc.mgxs` module. The `ScatterMatrixXS` class now has a `legendre_order` attribute/property which may be set to an integer between 0 and 10 (default is 0). The data access and storage methods for `ScatterMatrixXS` now include a `moment` parameter which may take on a value less than or equal to `legendre_order`.

None of the Jupyter Notebooks highlight the scattering matrix moment feature, but it could be introduced in either one of the existing three or @nelsonag's new one. I don't have a way to test out the higher order moments (OpenMOC is isotropic), but I understand that OpenMC's new MG mode can make use of them, so perhaps the new notebook is a good place to highlight this feature.


wbinventor:
It's worthwhile to note that the `MGXS.build_hdf5_store(...)` method does support the scattering moments without any changes. It does so by storing the data to HDF5 as a 3D array indexed by energy group in, energy group out, and scattering moment. If users prefer the storage to be more transparent then individual HDF5 datasets can be explicitly designated for each moment. An HDF5 attribute with the max. legendre moment order can also be added to the scattering matrix HDF5 group if needed. Since I will not be using this feature of `openmc.mgxs`, I'll leave it up to any first adopters out there (@paulromano - run this by ANL-NE?) to communicate a standard that best works for them.

nelsonag:
I <3 this

As for a notebook showing this off, the one I have in #643 makes the most sense.  If this PR gets merged first, I'll update the notebook in #643 to use this.  If not, then there will be a separate PR with minor code changes from me anyways to get the code currently in #643 up to date with this PR and so we'll get the notebook changes in then.

I'll take a look at this when I can, though it may not be wise to wait for me if ANL-NE prefers this in short order.

openmc/mgxs/library.py
```diff
@@ -280,15 +289,21 @@ def domains(self, domains):
...
 
-    @energy_groups.setter
-    def energy_groups(self, energy_groups):
-        cv.check_type('energy groups', energy_groups, openmc.mgxs.EnergyGroups)
-        self._energy_groups = energy_groups
+    @legendre_order.setter
+    def legendre_order(self, legendre_order):
+        cv.check_type('legendre_order', legendre_order, Integral)
+        cv.check_greater_than('legendre_order', legendre_order, 0, equality=True)
+        self._legendre_order = legendre_order
```
nelsonag: Where is the right place to check if `legendre_order is > 0`, then it doesn't make sense to have a `correction=="P0"`?  In the setter of each `legendre_order` and `correction`?

Of course, if you wanted to implement the extended transport approximation and allow the `0` in `P0` to be `legendre_order + 1`, then that would work too.

openmc/mgxs/mgxs.py
```diff
@@ -1547,7 +1546,7 @@ def tallies(self):
         if self._tallies is None:
 
             # Create a list of scores for each Tally to be created
-            scores = ['flux', 'total', 'scatter-P1']
+            scores = ['flux', 'total', 'scatter-1']
```
nelsonag: Good catch, the `scatter-0` order that was brought in by `scatter-P1` was wasteful.
However, should this be `scatter-1` or `nu-scatter-1`?

nelsonag: Did I ask that before? I feel like I did

openmc/mgxs/mgxs.py
```diff
@@ -1884,13 +1886,19 @@ def tallies(self):
...
-            else:
-                scores = ['flux', 'scatter']
-                filters = [[energy], [energy, energyout]]
+            # Create lists of scores, filters for each Tally to be created
+            scores = ['flux']
+            filters = [[energy]]
+
+            # Create separate tallies for each moment
+            for moment in range(self.legendre_order+1):
+                scores.append('scatter-{}'.format(moment))
```
nelsonag: This inputs `scatter-n` for n in range(0:N), inclusive.  OpenMC's tallying routines are more efficient if you use `scatter-pN` instead.  Why did you choose this approach?

openmc/mgxs/mgxs.py
```diff
@@ -1922,9 +1939,30 @@ def rxn_rate_tally(self):
     @correction.setter
     def correction(self, correction):
         cv.check_value('correction', correction, ('P0', None))
+
+        if correction == 'P0' and self.legendre_order > 0:
```
nelsonag: Oh, so these always over-ride the setters in `Library`?

openmc/mgxs/mgxs.py
```diff
@@ -2115,7 +2183,76 @@ def get_xs(self, in_groups='all', out_groups='all',
...
+             moment will be returned.
+        xs_type: {'macro', 'micro'}
+            Return macro or micro cross section in units of cm^-1 or barns.
+            Defaults to 'macro'.
+        summary : None or openmc.Summary
+            An optional Summary object to be used to construct columns for
+            distribcell tally filters (default is None). The geometric
+            information in the Summary object is embedded into a multi-index
+            column with a geometric "path" to each distribcell intance.
+            NOTE: This option requires the OpenCG Python package.
```
paulromano: Not true anymore, right? I'm assuming this was copy-pasted so that means it should also be updated elsewhere.

openmc/mgxs/mgxs.py
```diff
@@ -1977,9 +2031,9 @@ def get_slice(self, nuclides=[], in_groups=[], out_groups=[]):
         return slice_xs
 
     def get_xs(self, in_groups='all', out_groups='all',
```
nelsonag: I've seen codes expect the moments to have been already multiplied by `(2l+1)/2` and others that do not.  OpenMC's MG mode does not expect the multiplication to have already occurred.  The docs should be updated in some way to let the user know what they're getting.  Hopefully they'd get very nonsensical results if they didn't know that, but 1) maybe not, and 2) might as well save them some time.

openmc/mgxs/mgxs.py
```diff
@@ -2242,28 +2386,35 @@ def tallies(self):
...
-                scores = ['flux', 'nu-scatter']
-                estimator = 'analog'
-                keys = ['flux', 'scatter']
-                filters = [[energy], [energy, energyout]]
+            # Create lists of scores, filters for each Tally to be created
+            scores = ['flux']
+            filters = [[energy]]
+            keys = ['flux']
+
+            # Create separate tallies for each moment
```
nelsonag: same comment as above re: tallying efficiency is better with `nu-scatter-pN` than with a set of `nu-scatter-n` scores

nelsonag:
Just gave it a quick look through, looks good.  That's probably about as detailed as I can get until I'm able to put it through its paces.  Like I said in my previous comment - not sure when that will be.

paulromano:
@nelsonag I don't think there's a great urgency. They will be happy to know that it is in the works and will be in soon-ish.

paulromano:
@wbinventor A request not really related to this pull request, per se -- can we copy the parameters/attributes docstring sections from MGXS to all of its children. The problem is that if you're in ipython or a notebook and you want to get help on one of the subclasses, the only thing you see is the few things that show up currently, e.g. for ScatterMatrixXS you'd only see the two attributes correction and legendre_order.

paulromano:
Only hackers who know `inspect.getmro()` will know to look at the parent class...

src/state_point.F90
```diff
@@ -330,11 +330,11 @@ subroutine write_state_point()
           MOMENT_LOOP: do j = 1, tally % n_user_score_bins
             select case(tally % score_bins(k))
             case (SCORE_SCATTER_N, SCORE_NU_SCATTER_N)
-              str_array(k) = 'P' // trim(to_str(tally % moment_order(k)))
+              str_array(k) = trim(to_str(tally % moment_order(k)))
```
paulromano: Why don't you like the "P"?

openmc/mgxs/library.py
wbinventor: I just added a check with a warning message to both the `correction` and `legendre_order` property setters. These mirror those in the properties for the `MGXS` class.

openmc/mgxs/mgxs.py
wbinventor: Hmm good question - I actually don't know, but think it should be "nu-scatter-1" in place of "scatter-1" correct?

openmc/mgxs/mgxs.py
wbinventor: Yes this was a dilemma for me and I'm certainly open to refactoring. The rationale behind this implementation is that it makes it simple to search for the tallies in the output. Right now if you specify a "scatter-PN" tally in your input, the statepoint/summary will expand the scores into a list of "scatter-0", "scatter-1", ... " scatter-N". Although this can be very nice for downstream data processing, it means that you cannot search for a tally in the output with the score "scatter-PN" that was used in the input. 

Of course I can refactor to get around this with a hack like the following:
* use "scatter-PN" in the input
* manually expand "scatter-PN" to "scatter-0", ... "scatter-N" when loading data from a statepoint
* slice apart the different moment tallies for ease of handling

Of course I should note that if one *merges*  the tallies in the input file (*i.e.*, `library.add_to_tallies_file(tallies_file, merge=True)`) then a single `Tally` will encapsulate all of the "scatter-0", ..., "scatter-N" scores even though they are separate in the `MGXS`. The tally merging was implemented to improve tallying efficiency. But perhaps a "scatter-PN" score is more efficient than a single tally with multiple "scatter-X" scores - perhaps you or @paulromano can clarify if this is indeed the case.

openmc/mgxs/mgxs.py
wbinventor: Yes, yes they do. Sync'ing up between `Library` and `ScatterMatrixXS` would be challenging but doable. I've simply presumed that anyone using `Library` should employ the appropriate "safety measures" and avoid directly manipulating the `MGXS` objects in the `Library`, which is the objective of the `Library` class to some extent.

openmc/mgxs/mgxs.py
wbinventor: Good point - just added a clarification to the docstring on this point.

openmc/mgxs/mgxs.py
wbinventor: Correct, I'll remove this. I'm surprised this shows up, but perhaps I need to sync with develop first.

openmc/mgxs/mgxs.py
wbinventor: I'll wait to make a move on this until I hear more from you about my rationale behind this.

src/state_point.F90
wbinventor: My issue with it is related to my rationale behind using explicit scores "scatter-0", ... "scatter-N" above. If you specify a "scatter-1" tally the output in the statepoint is "scatter-P1". It seems to me that it would be better for the output to match the input so that one can easily find a `Tally` created for the input in the output (which is not the case with these "P"'s added to the scores).

paulromano: Oh, right. Agreed.

openmc/mgxs/mgxs.py
paulromano: Both ways will result in the same number of calls to `calc_pn`, but using the "scatter-Pn" scores will cut down on some of the overhead above that. Hard to say how much of a difference it would make, although it would be easy to test empirically.

openmc/mgxs/mgxs.py
nelsonag: I *think* so... I mean starting from like the bell-Hansen-sandmeier derivation, the right hand side _should_ be a nu-scatter sigS, and therefore we would be subtracting off nu-scatter-1.

openmc/mgxs/mgxs.py
nelsonag: It is true that "scatter-PN" is more efficient than "scatter-X".  The cycling through the orders is an inner loop for "scatter-PN", while if you had "scatter-0 scatter-1 scatter-2 ..." then you have to cycle through the score loop each time and incur that overhead.

wbinventor: Alright - any issues if I get around this issue by making my tally merging
code "smarter" and merging "scatter-X" scores into a single "scatter-PN"
score?

paulromano: In principle I don't see anything wrong with that.

nelsonag: Seems like a good plan to me.

openmc/mgxs/mgxs.py
nelsonag: For the record, our CMFD solver also does this (uses scatter-1 vice nu-scatter-1), but it should be of no consequence there since in the end you force the currents to match anyways.

wbinventor: Alright, I've migrated this over to "nu-scatter-1" since I agree that it should have been using that score all along.

wbinventor: Oh wait, actually I was confused and thought you were commenting on the correction applied to `NuScatterMatrixXS`. I'm not so sure about `TransportXS`... I *think* that I agree with you in the case one is using `nu` in the scattering cross section. But in the case that one is not using `nu`, then we would *not* want the correction to account for `nu`. This requires a new setting in `TransportXS` to reflect the need for "nu-scatter-1" rather than "scatter-1." I've implemented this as a new option for the `correction` attribute as `"nuP0"` alongside the original `"P0"` setting. What do you think @nelsonag?

nelsonag: That sounds reasonable, if we allow users to use scatter instead of nu-scatter there doesn't seem to be any other choice.  Though it does make sense to have `"nuP0"' be default since that is preferred.  

While we are at it, maybe we can improve on the `"P0"` and `"nuP0"` names.  Maybe `"out-scatter"` and `"nu-out-scatter"` or `"out-nu-scatter"`?  What do you think @wbinventor?

wbinventor: I've actually opted for an alternative since my last post. How about a new `NuTransportXS` class? I really hate adding in new properties/attributes since it seems like a sub-optimal hacky solution. Plus, I can foresee an explosion of them in the future with more complex transport corrections and the like.

nelsonag: Hmmmm, thats possible.  Will this path also explode when more complex transport corrections are applied?  Is there something the Library can do for us instead of leaving it up to the Mgxs objects themselves?  That might not make much sense but its getting late for me.

nelsonag: I see a few options:
1) Don't support a TransportXS which doesn't use nu-scattering.  That doesn't seem so bad since its not technically correct to be using normal scattering matrices vice nu-scatter matrices.
2) If we don't want to go that route, your NuTransportXS class idea seems good for now, though like I said above, that might explode with future corrections too.
3) In the long-term (or now if it seems trivial enough), what about simplifying this considerably by recognizing that all transport corrections work by including a deltaSigma_tr which is applied to a total xs and the scattering matrix diagonal.  What if we just have an extra MGXS class for deltaSigma_tr? Then, Library can know how to apply it with some `apply_correction()` function, or we can leave it to the downstream code to do something with.

Thoughts @wbinventor or @paulromano or anyone else who has thought about such corrections before?

wbinventor:
This PR has been updated a great deal based on the comments. The `ScatterMatrixXS` class now uses `'scatter-PN'` scores for tallying efficiency. There is a new `NuTransportXS` class to account for neutron multiplicity in scattering reactions. The `MGXS.domain_type` attribute is now set based on the type of the `MGXS.domain` attribute to simplify things for the user.

In the process of making this work, I refactored much of the `MGXS` API in a way that I *think* is much simpler. The refactorization applies some of the concepts that were discussed in #592. In summary this refactoring led to the following:

* Removed the `tallies` property for each `MGXS` subclass
* Added new `MGXS.scores`, `MGXS.filters`, `MGXS.tally_keys` and `MGXS.estimator` properties which are used directly by the `MGXS.tallies` property
  * Generic implementations of each property are given in the `MGXS` class
  * Only a few subclasses over-ride the properties, such as `TransportXS` and `ScatterMatrixXS`
* Made greater use of the `MGXS.rxn_type` attribute to construct tallies
* Removed the `FissionXSBase` class introduced in #592 
* Added a new `MGXS.hdf5_key` property since the `MGXS.rxn_type` properties are no longer unique
* Added a new `MGXS.loaded_sp` property to facilitate loading tallies with `'scatter-0'`, `'scatter-1'`, ... scores for `ScatterMatrixXS` which expects a `'scatter-PN'` tally score

@nelsonag - I'd prefer to wait on diving into the transport corrections, that wasn't really something I aimed to tackle with this PR. But I can see merit in your suggestion of using something like a `DeltaTransportXS` to correct both `TotalXS` and `ScatterMatrixXS` in the future.

paulromano:
@wbinventor MGXS part 3 notebook is all hosed up

openmc/mgxs/mgxs.py
```diff
@@ -212,11 +229,70 @@ def num_groups(self):
         return self.energy_groups.num_groups
 
     @property
+    def scores(self):
+        return ['flux', self.rxn_type]
+
+    @property
+    def filters(self):
+        group_edges = self.energy_groups.group_edges
+        energy_filter = openmc.Filter('energy', group_edges)
```
paulromano: Not sure if it would ever really be an issue, but looking at this property multiple times will result in `Filter`s that have different `id`s

wbinventor: Except that filters don't have IDs. Point taken, but I'm not sure it
matters.

paulromano: Oh sorry, that was confusing. What I meant is that `[id(f) for f in mgxs_object.filters]` will return a different list each time you evaluate it.

wbinventor: Oh....yes good point. Do you see any reason to avoid this? Of course it
would be simple to store the filters in a new class attribute if necessary.

wbinventor:
Fixed my accidental hosing of MGXS Part III

nelsonag:
This is probably related to the reason why my Part IV notebook needs cells
to be re-generated from scratch; I seem to be getting different cell ids
every time.

wbinventor:
I don't think that my notebook getting hosed is related to your issues @nelsonag - the problem is that you can't run two of our notebook examples at the same time since they will clobber each other's statepoint/summary files! But per your issue related to IDs - the part-iv notebook should be designed to be agnostic to the IDs just like all of the other notebooks. In fact, I removed the ID setters in your notebook while hacking away at it and it worked just fine with changing IDs.

openmc/mgxs/mgxs.py
```diff
@@ -2114,7 +2058,77 @@ def get_xs(self, in_groups='all', out_groups='all',
...
+        if self.legendre_order > 0:
+            # Insert a column corresponding to the Legendre moments
+            moments = ['P{}'.format(i) for i in range(self.legendre_order+1)]
+            moments = np.tile(moments, df.shape[0] / len(moments))
+            df['moment'] = moments
+
+            # Place the moment column before the mean column
+            mean_index = df.columns.get_loc('mean')
+            columns = df.columns.tolist()
+            df = df[columns[:mean_index] + ['moment'] + columns[mean_index:-2]]
```
paulromano: Should that be -1 at the end there?

paulromano:
@wbinventor We should add higher moment scattering matrix generation to one of the tests. Other than that, I think it looks pretty good now. Thanks so much for taking a stab at this!

openmc/mgxs/mgxs.py
wbinventor: Using -1 would leave the "mean" column at the end. Instead, we want to
truncate it off the end and move it to the center of the DataFrame.

paulromano: What I'm seeing is that it ends up chopping off the 'std. dev.' column. After the second line, columns would be something like `['cell', 'group in', 'group out', 'nuclide', 'mean', 'std. dev.', 'moments']`, so inserting 'moments' in the middle and indexing to -1 would prevent 'moments' from appearing twice but not remove anything.

paulromano:
test_mgxs_library_nuclides/results_true.dat is getting pretty big (245 kB). We should either hash it or otherwise try to cut it down.

openmc/mgxs/mgxs.py
wbinventor: Oh wow, I was confused before but absolutely agree.

wbinventor:
The MGXS tests now all use `legendre_order=3`. The nuclide-based MGXS are now hashed to avoid the big data problem. This should be good to go.

paulromano:
Still getting test failures on this.

wbinventor:
Yeah the problem is that I updated the test results on my workstation with
one version of Pandas that adds whitepsace in a different way than the
conda-distributed version. I need to update using the conda Python on my
laptop, bit I'm having issues with that machine right now. Will let you
know once I sort things out.

wbinventor:
The test results are up-to-date and now pass Travis.

openmc/mgxs/mgxs.py
paulromano: I think it's ok as is. If this is just getting generated from the tallies property, *that* is saved in an attribute and wouldn't be generated again.

Issue #597, opened 2016-02-24
Update MGXS Module of Python API To Support Data Used by Multi-Group Mode


nelsonag:
The multi-group capability in OpenMC (OpenMG, if you will) can take advantage of different data than is currently provided by the MGXS and MGXSLibrary Python APIs.  This includes the following items which are very likely to be needed by users:
- [x] scatter moment matrices (i.e., moment as in Legendre moments) [Edit: closed in #644]
- [x] nu-scatter moment matrices [Edit: closed in #644]

And the following items which are lower-need since many users wont use this feature, but OpenMC still supports the functionality and therefore would be useful to have:
- [x] polar and azimuthal angle dependent cross sections generated with constant width bins unique to each dimension (polar and azimuthal) (this applies for all current data in the develop branch and this issue, save for the scattering multiplication matrix, as that's overkill).
- [x] scattering matrices with change-in-angle information provided in equal-width mu bins.
- [x] nu-scattering matrices with change-in-angle information provided in equal-width mu bins.
- [x] nu-fission matrix
- [x] scattering multiplication matrix (i.e., nu-scatter-0[g'->g]/scatter-0[g'->g] used to multiply particle weights after a scattering event.

nelsonag:
Closed by #644, #653, #758, and #778

We now have a MGXS API that is fully integrated with the MG Mode! Yay!

wbinventor:
Whew, you made all of this happen @nelsonag! Nicely done sir, nicely done.

nelsonag:
with lots of help from you and the great MGXS API!!

PR #653, opened 2016-05-24
Addition of NuFissionMatrixXS and MultiplicityMatrix MGXS classes in MGXS Python API

nelsonag:
This mild-mannered PR addresses the bottom two checkboxes in #597: giving the Python API's Library class the ability to generate scattering multiplicity matrices and nu-fission matrices (the latter replacing the combination of nu-fission and chi vectors).
In doing this, I also added in a MatrixMGXS type which becomes the base for all energyin x energyout matrix MGXS types.  This is now the parent of *ScatterMatrixXS, NuFissionMatrixXS, and MultiplicityMatrixXS.

The incorporation of these changes would give us the ability to automagically generate all the isotropically-weighted and Legendre-expanded data types one would need for OpenMC's multi-group mode [adding in angle-dependent MGXS and/or tabular-mu scattering data is a tougher nut to crack].

### DIFF
--- a/docs/source/pythonapi/examples/mgxs-part-iv.ipynb
+++ b/docs/source/pythonapi/examples/mgxs-part-iv.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This Notebook illustrates the use of the openmc.mgxs.Library class specifically for application in OpenMC's multi-group mode. This example notebook follows the same process as was done in MGXS Part III, but instead uses OpenMC as the multi-group solver. This Notebook illustrates the following features:\n",
+    "This Notebook illustrates the use of the openmc.mgxs.Library class specifically for application in OpenMC's multi-group mode. This example notebook follows the same process as was done in MGXS Part III, but instead uses OpenMC as the multi-group solver. During this process, this notebook will illustrate the following features:\n",
     "\n",
     "   - Calculation of multi-group cross sections for a fuel assembly\n",
     "   - Automated creation and storage of MGXS with openmc.mgxs.Library\n",
@@ -334,7 +334,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "With the geometry and materials finished, we now just need to define simulation parameters. In this case, we will use 10 inactive batches and 40 active batches each with 2500 particles."
+    "With the geometry and materials finished, we now just need to define simulation parameters. In this case, we will use 10 inactive batches and 40 active batches each with 5000 particles."
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


openmc/mgxs/mgxs.py
```diff
@@ -425,7 +427,11 @@ def get_mgxs(mgxs_type, domain=None, domain_type=None,
 
         Parameters
         ----------
-        mgxs_type : {'total', 'transport', 'nu-transport', 'absorption', 'capture', 'fission', 'nu-fission', 'kappa-fission', 'scatter', 'nu-scatter', 'scatter matrix', 'nu-scatter matrix', 'chi'}
+        mgxs_type : {'total', 'transport', 'nu-transport', 'absorption',
```
wbinventor: This was a single long line in order for each type to be properly tracked by Sphinx.

openmc/mgxs/mgxs.py
```diff
@@ -1521,6 +1531,415 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
...
+        Returns
+        -------
+        openmc.mgxs.MGXS
+            A new tally which encapsulates the subset of data requested for the
+            nuclide(s) and/or energy group(s) requested in the parameters.
+
+        """
+
+        # Call super class method and null out derived tallies
+        slice_xs = super(NuFissionMatrixXS, self).get_slice(nuclides,
```
wbinventor: This should use `MGXSMatrixXS` in the call to `super`

openmc/mgxs/mgxs.py
nelsonag: Fixed, here and in library.py

openmc/mgxs/mgxs.py
nelsonag: Oh shoot, good catch, thanks.

docs/source/pythonapi/examples/mgxs-part-iv.ipynb
```diff
@@ -952,12 +952,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We will now use the `Library` to produce a multi-group cross section data set for use by the OpenMC multi-group solver.  "
+    "We will now use the `Library` to produce a multi-group cross section data set for use by the OpenMC multi-group solver.  \n",
+    "Note that since we have ran so few histories, is not unreasonable to expect some divisions by zero errors.  This will show up as a runtime warning in the following step."
```
wbinventor: Note that these errors divide-by-zero warnings exist in the other MGXS notebook I as well. No mention is made of these errors, though I agree that it may be worthwhile to point it out as you do here.

docs/source/pythonapi/examples/mgxs-part-iv.ipynb
```diff
@@ -513,7 +513,7 @@
    "source": [
     "# Specify multi-group cross section types to compute\n",
     "mgxs_lib.mgxs_types = ['total', 'absorption', 'nu-fission', 'fission',\n",
-    "                       'nu-scatter matrix', 'scatter matrix', 'chi']"
+    "                       'nu-scatter matrix', 'multiplicity matrix', 'chi']"
```
wbinventor: Why don't you use the new "nu-fission matrix" here?

openmc/mgxs/mgxs.py
```diff
@@ -1521,6 +1527,414 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
...
+
+    @property
+    def estimator(self):
+        return 'analog'
+
+    def get_xs(self, in_groups='all', out_groups='all',
+               subdomains='all', nuclides='all',
+               xs_type='macro', order_groups='increasing',
+               row_column='inout', value='mean', **kwargs):
+        r"""Returns an array of multi-group cross sections.
```
wbinventor: Does this docstring need to be a raw string?

openmc/mgxs/mgxs.py
```diff
@@ -1521,6 +1527,414 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
...
+
+        This method constructs a 2D NumPy array for the requested multiplicity
```
wbinventor: This is not necessarily a "multiplicity" matrix, correct?

openmc/mgxs/mgxs.py
```diff
@@ -1521,6 +1527,414 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
...
+            Return the cross section indexed according to increasing or
+            decreasing energy groups (decreasing or increasing energies).
+            Defaults to 'increasing'.
+        row_column: {'inout', 'outin'}
+            Return the cross section indexed first by incoming group and
+            second by outgoing group ('inout'), or vice versa ('outin').
+            Defaults to 'inout'.
+        value : str
+            A string for the type of value to return - 'mean', 'std_dev', or
+            'rel_err' are accepted. Defaults to the empty string.
```
wbinventor: This may be my error, but the default is not the "empty string" but rather "mean"

openmc/mgxs/mgxs.py
```diff
@@ -1521,6 +1527,414 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
...
+            xs = xs[:, ::-1, ::-1, :]
+
+            # Eliminate trivial dimensions
+            xs = np.squeeze(xs)
+            xs = np.atleast_2d(xs)
+
+        return xs
+
+    def get_slice(self, nuclides=[], in_groups=[], out_groups=[]):
+        """Build a sliced NuFissionMatrix for the specified nuclides and
```
wbinventor: This isn't necessarily a "NuFissionMatrix"

openmc/mgxs/mgxs.py
```diff
@@ -1521,6 +1527,414 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
...
+        in_groups : list of int
+            A list of incoming energy group indices starting at 1 for the high
+            energies (e.g., [1, 2, 3]; default is [])
+        out_groups : list of int
+            A list of outgoing energy group indices starting at 1 for the high
+            energies (e.g., [1, 2, 3]; default is [])
+
+        Returns
+        -------
+        openmc.mgxs.MGXS
```
wbinventor: Clarify that this returns a `openmc.mgxs.MGXSMatrixXS`

wbinventor: Also, this is probably my fault, but the docstring is wrong - this does not return a "tally" but rather an "MGXS object"

openmc/mgxs/mgxs.py
```diff
@@ -2452,7 +2866,7 @@ def __init__(self, domain=None, domain_type=None,
         self._rxn_type = 'nu-scatter'
 
 
-class ScatterMatrixXS(MGXS):
+class ScatterMatrixXS(MatrixMGXS):
```
wbinventor: Shouldn't you be able to eliminate the `ScatterMatrixXS` class methods `get_xs(...)`, `get_slice(...)` and `print_xs(...)` since they are defined in the parent `MGXSMatrixXS` class? Also, what about the `get_pandas_dataframe(...)` method? Have you run MGXS notebooks I, II and III?

openmc/mgxs/mgxs.py
```diff
@@ -3182,6 +3592,229 @@ def __init__(self, domain=None, domain_type=None,
         self._hdf5_key = 'nu-scatter matrix'
 
 
+class MultiplicityMatrixXS(MatrixMGXS):
+    """The scattering multiplicity matrix.
+
+    This class can be used for both OpenMC input generation and tally data
+    post-processing to compute spatially-homogenized and energy-integrated
+    multi-group cross sections for deterministic neutronics calculations.
```
wbinventor: I suppose that the "deterministic" in this docstring is no longer true :-)

openmc/mgxs/mgxs.py
```diff
@@ -3182,6 +3592,229 @@ def __init__(self, domain=None, domain_type=None,
...
+        scores = ['nu-scatter', 'scatter']
+        return scores
+
+    @property
+    def filters(self):
+        # Create the non-domain specific Filters for the Tallies
+        group_edges = self.energy_groups.group_edges
+        energy = openmc.Filter('energy', group_edges)
+        energyout = openmc.Filter('energyout', group_edges)
+
```
wbinventor: this can be simplified as:

```python
return [[energy, energyout], [energy, energyout]]
```

but not necessary

openmc/mgxs/mgxs.py
```diff
@@ -3182,6 +3592,229 @@ def __init__(self, domain=None, domain_type=None,
...
+        return self._xs_tally
+
+
+class NuFissionMatrixXS(MatrixMGXS):
+    """A fission production matrix multi-group cross section.
+
+    This class can be used for both OpenMC input generation and tally data
+    post-processing to compute spatially-homogenized and energy-integrated
+    multi-group cross sections for deterministic neutronics calculations.
```
wbinventor: ditto, remove "deterministic" if you wish

openmc/mgxs/mgxs.py
```diff
@@ -3182,6 +3592,229 @@ def __init__(self, domain=None, domain_type=None,
...
+    hdf5_key : str
+        The key used to index multi-group cross sections in an HDF5 data store
+
+    """
+
+    def __init__(self, domain=None, domain_type=None,
+                 groups=None, by_nuclide=False, name=''):
+        super(NuFissionMatrixXS, self).__init__(domain, domain_type,
+                                                groups, by_nuclide, name)
+        self._rxn_type = 'nu-fission matrix'
```
wbinventor: So if you want this class to mirror the functionality I introduced to the `MGXS` subclasses in #644, you would make the following changes:

* Set the `_rxn_type` property to `'nu-fission'`
* Assign the `_hdf5_group` property to `'nu-fission matrix'`
* Remove the `scores` property getter since the `MGXS.scores` property *should* work just fine 
* Remove the `rxn_rate_tally` property getter since `MGXS.rxn_rate_tally` *should* work just fine

I recognize that some of my changes in #644 which permit the changes above may not lend themselves to fantastic code transparency, though they did eliminate a few hundred lines of redundant code. If you have a better idea to go about this then I'm all ears.

openmc/mgxs_library.py
```diff
@@ -512,8 +523,8 @@ def kappa_fission(self, kappa_fission):
 
     @chi.setter
     def chi(self, chi):
-        if self._use_chi is not None:
-            if not self._use_chi:
+        if self.use_chi is not None:
+            if not self.use_chi:
```
wbinventor: How about:

```python
msg = 'Providing "chi" when "nu-fission already provided as a matrix'
```

to mirror our MGXS type strings as closely as possible.

openmc/mgxs_library.py
```diff
@@ -744,7 +753,10 @@ def set_nu_fission_mgxs(self, nu_fission, nuclide='total',
             msg = 'Angular-Dependent MGXS have not yet been implemented'
             raise ValueError(msg)
 
-        self._use_chi = True
+        if type(nu_fission) is openmc.mgxs.NuFissionMatrixXS:
```
wbinventor: I think the more Pythonic version of this would be to use `isinstance`, but I don't know of a strict rule either way.

openmc/mgxs_library.py
```diff
@@ -923,23 +937,33 @@ def set_multiplicity_mgxs(self, nuscatter, scatter, nuclide='total',
...
-                    [self.energy_groups])
         check_value('domain_type', nuscatter.domain_type,
                     ['universe', 'cell', 'material'])
-        check_value('domain_type', scatter.domain_type,
-                    ['universe', 'cell', 'material'])
+        if scatter is not None:
+            check_type('scatter', scatter, openmc.mgxs.ScatterMatrixXS)
+            if isinstance(nuscatter, openmc.mgxs.MultiplicityMatrixXS):
+                msg = 'Either an MultiplicityMatrixXS object must be passed ' \
+                      'for "nuscatter" or the "scatter" argument must be ' \
```
wbinventor: add a hyphen to "nuscatter" to mirror our MGXS type strings

wbinventor:
@nelsonag this PR looks nice! The refactoring of the matrix `MGXS` classes into an abstract `MatrixMGXS` is precisely what I have been wanting to see for a little while now. I think this PR should be a quick one to merge in pending the resolution of my few comments. Also, have you run all of the MGXS notebooks? If not, and you wish to save the time of installing OpenMOC (I swear it's not too bad!) let me know and I can run them just to check everything out.

nelsonag:
FYI - it looks like that build failed because an Ubuntu repo mirror was down.

docs/source/pythonapi/examples/mgxs-part-iv.ipynb
nelsonag: I haven't done one of my good old convergence tests on it, but it looks like (and stands to reason) the nu-fission matrix is slower to resolve than good ol' nu-fission and chi.  When I used nu-fission matrix the CE-MG bias was 1,000pcm at these few history levels, dropping to low levels (<20pcm i think it was) as histories increased.  Since I don't want to artificially scare anyone away from using MG mode with a high bias, I stuck with the nu-fission/chi pair.

openmc/mgxs/mgxs.py
nelsonag: I left those there as that's where all the moment and correction related code lives, which doesn't necessarily need to live in MatrixMGXS.

nelsonag: But no I haven't re-run those notebooks. I suppose I should eventually get OpenMOC running.

openmc/mgxs/mgxs.py
nelsonag: Yeah, one of the classes I was comparing used the above, the other uses whats in my PR.  I just settled on this but I don't quite care either way.

openmc/mgxs/mgxs.py
nelsonag: Leaving off with this comment right here for the day. Will get back to this tonight.

openmc/mgxs_library.py
paulromano: Yes, it's actually mentioned in PEP8.

docs/source/pythonapi/examples/mgxs-part-iv.ipynb
wbinventor: Ah, I see! Seems like sound logic to me.

openmc/mgxs/mgxs.py
wbinventor: Ok, I forgot about the moments in my red eye flight stupor. Sounds good, keep those methods as they are, but do your own set of checks to ensure that each of those methods work for each of the `MarrixsMGXS` subclasses (esp. `NuFissionMatrixXS`).

wbinventor:
@nelsonag what do you think of adding a new test to the suite which runs a CE case with MGXS tallies,  pipes the MGXS back in and runs the corresponding MG mode case? Also, as we continue to expand our set of MGXS types, I'm concerned that we aren't adequately accounting for them in the test suite. Right now the MGXS tests only consider "total", "nu-scatter matrix", "nu-fission" and "chi" since this is the core of what I use for my research with OpenMOC.  But this doesn't account for what you need with MG mode. The downside of adding more MGXS types to the suite is that it may take longer to run and the output will be larger, but the code coverage may be worth it. What do you think?

nelsonag:
@wbinventor sigh I am having trouble with OpenMOC.  either not getting complete compilation in python2 or in python3, i hit a wall because i dont have OpenCG.  Would you actually mind running the other 3 mgxs notebooks and for me whenever you get a chance?

wbinventor:
I just ran each of the MGXS I-III notebooks and they work fine. What do you think about my ideas for the test suite? Perhaps that can come in its own separate PR.

nelsonag:
Working on the test suite now. I am running in to some bugs w/ TransportXS and NuTransportXS which should not be showing up and I am quite perplexed by them

nelsonag:
simply put, I modified the test_mgxs_library_hdf5 test to include every score type possible and the test quits within rxn_rate_tally for nu-transport or transport (whichever is first), with an error that the operands (total and scatter-1) could not be broadcast together b/c total is of shape (2,1,1) and scatter-1 is of shape (4,1,1).  Ok, thats confusing. So I just printed out the tally for scatter-1, and the material it fails on has two energy filters, and no energyout filter. Whaaat?  Why does it do that with all these scores but not in the test as it is now? So weird.

nelsonag:
Given these test issues, why dont we put them into a separate PR since it looks like there may be more findings which will complicate the merging of whats already in this PR.  I'll put in the typo you pointed out, and then if you agree that your comments are satisfactorily addressed, and that its OK to pull the test to another PR, you can go ahead and merge

wbinventor:
@nelsonag that sounds like a good plan to me. If you can open up an issue mentioning the problem you spotted that would be nice. Sounds like there is a bug lurking somewhere.

nelsonag:
Sure thing. I'll recreate and post the issue next time ive got the PC on.

PR #660, opened 2016-05-29
New Optional Strictly Exact Option for StatePoint.get_tally(...)

wbinventor:
This short PR introduces a new optional boolean `exact` parameter for the `StatePoint.get_tally(...)` method. This parameter *should* resolve issue #657 based on my assessment. Although the `MGXS.load_from_statepoint(...)` method now uses `exact=True` in order to mitigate #657, the test suite remains unchanged in this PR since I presume @nelsonag is considering updates to the MGXS tests for a separate PR.


nelsonag:
Quick, simple, and works.  Thanks @wbinventor!

PR #661, opened 2016-05-29
Added further tests of MGXS data

nelsonag:
Included in this PR are:

- test of MGXS created by CE being piped in to MG mode.
- test of all MGXS types in test_mgxs_*nuclides.
- Same all-inclusive list of all MGXS types as in previous bullet not necessary for other test_mgxs_* tests since the current mgxs types explicitly test functions on base classes, and therefore we might as well save Travis' time by not doing it again.


nelsonag:
Travis is failing on my new test.  Not sure why yet, as its OK locally.

nelsonag:
Should be good to go now, hope I dont have to eat my words.

The failure was caused by a division by zero in OpenMC's MG mode due to many cross sections being zero. I fixed this in OpenMC, and changed the model to not be the HM problem, but instead a simple pincell. Now everybody get's a score with so few histories.

tests/input_set.py
```diff
@@ -570,6 +570,105 @@ def build_defualt_plots(self):
...
+        self.materials = openmc.Materials()
+        self.geometry = openmc.Geometry()
+        self.tallies = None
+        self.plots = None
+
+    def export(self):
+        self.settings.export_to_xml()
+        self.materials.export_to_xml()
+        self.geometry.export_to_xml()
+        if self.tallies is not None: self.tallies.export_to_xml()
```
wbinventor: Per PEP8, split these conditionals across separate lines

wbinventor: Actually I see that the other `InputSet` subclasses do the very same thing.

nelsonag: Meh, I'll still fix it.

tests/test_mgxs_library_ce_to_mg/test_mgxs_library_ce_to_mg.py
```diff
@@ -0,0 +1,98 @@
...
+import openmc.mgxs
+
+
+class MGXSTestHarness(PyAPITestHarness):
+    def _build_inputs(self):
+        # Set the input set to use the pincell model
+        self._input_set = PinCellInputSet()
+
+        # The openmc.mgxs module needs a summary.h5 file
+        self._input_set.settings.output = {'summary': True}
```
wbinventor: This is (no longer) necessary since the "summary.h5" is reported by default.

tests/test_mgxs_library_ce_to_mg/test_mgxs_library_ce_to_mg.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Generate inputs using parent class routine
+        super(MGXSTestHarness, self)._build_inputs()
+
+        # Rewrite file
+        # self._input_set.settings.export_to_xml()
```
wbinventor: Remove commented out debugging code

tests/test_mgxs_library_ce_to_mg/test_mgxs_library_ce_to_mg.py
```diff
@@ -0,0 +1,98 @@
...
+
+        assert returncode == 0, 'CE OpenMC calculation did not exit' \
+                                'successfully.'
+
+        # Build MG Inputs
+        # Get data needed to execute Library calculations.
+        statepoint = glob.glob(os.path.join(os.getcwd(), self._sp_name))[0]
+        sp = openmc.StatePoint(statepoint)
+        self.mgxs_lib.load_from_statepoint(sp)
+        self._input_set.mgxs_file, self._input_set.materials, \
```
wbinventor: Nice!

tests/test_mgxs_library_distribcell/results_true.dat
```diff
@@ -1,8 +1,11 @@
                                     avg(distribcell)  group in nuclide      mean  std. dev.
-0  (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,...         1   total  0.718919   0.520644                                    avg(distribcell)  group in nuclide  mean  std. dev.
-0  (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,...         1   total   0.0        0.0                                    avg(distribcell)  group in  group out nuclide moment      mean  std. dev.
+0  (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,...         1   total  0.718919   0.520644
```
wbinventor: I'm a little confused what changed here, can you elaborate?

tests/test_mgxs_library_distribcell/test_mgxs_library_distribcell.py
```diff
@@ -59,7 +59,7 @@ def _get_results(self, hash_output=False):
             for mgxs_type in avg_lib.mgxs_types:
                 mgxs = avg_lib.get_mgxs(domain, mgxs_type)
                 df = mgxs.get_pandas_dataframe()
-                outstr += df.to_string()
+                outstr += df.to_string() + '\n'
```
wbinventor: I presume this is what changed the diff above, why is this necessary? Perhaps it makes the results cleaner.

tests/test_mgxs_library_nuclides/test_mgxs_library_nuclides.py
```diff
@@ -25,8 +25,12 @@ def _build_inputs(self):
         # Initialize MGXS Library for a few cross section types
         self.mgxs_lib = openmc.mgxs.Library(self._input_set.geometry)
         self.mgxs_lib.by_nuclide = True
-        self.mgxs_lib.mgxs_types = ['transport', 'nu-fission',
-                                    'nu-scatter matrix', 'chi']
+        self.mgxs_lib.mgxs_types = ['total', 'transport', 'nu-transport',
```
wbinventor: How about using `MGXS_TYPES` so that these tests will "scale" with the introduction of `MGXS` subclasses in the future:

```python
self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES
```

Though I can see the disadvantage of this being that the "mgxs_*_nuclide" tests will *always* fail each time a new `MGXS` subclass is introduced.

wbinventor:
This looks great and I'm happy to merge as soon as my few tiny comments are addressed and Travis passes.

tests/test_mgxs_library_ce_to_mg/test_mgxs_library_ce_to_mg.py
nelsonag: I thought you'd like the simplicity of that :-)

tests/test_mgxs_library_distribcell/results_true.dat
nelsonag: I simply added an endl after each df print in these test_mgxs_library* tests because when those results_true.dat were visually not useful.  The header for the next dataset was on the same line as the previous data set.

tests/test_mgxs_library_distribcell/test_mgxs_library_distribcell.py
nelsonag: exactly, see above

tests/test_mgxs_library_nuclides/test_mgxs_library_nuclides.py
nelsonag: actually thats a good idea.  Its probably good behavior to remind the developer to check to make sure the test is OK

wbinventor: That's my leaning too. Also, what do you think about using your pin cell problem for all of the MGXS tests save for the distribcell one?

wbinventor: Also, I would lean towards using all `MGXS_TYPES` in each MGXS test. The reason is that we *do* want to ensure that the functionality tested in each test (*i.e.*, energy condensation, HDF5 data stores, etc.) does indeed work for all of our supported MGXS types. If this results in output that is too large for some of the tests, then we can simply hash the "results_true.dat" as we already do for the distribcell test (though this may be less problematic if we use your new pin cell input set everywhere anyway). Thoughts @nelsonag?

nelsonag: Yeah, let me convert the MGXS problems to the pincell and then I will add in the MGXS_TYPES run as well.

wbinventor: Alright cool. Sorry to up the ante, this just seems like the right time and place for it.

wbinventor: Also, looks like none of your commits have passed Travis yet, so there may still be a lingering issue. Make sure your version of Pandas is consistent with that used on the Travis image, that is always a hang up.

nelsonag: There was. I didnt update the inputs after removing the summary=True option
:-(

nelsonag:
This travis run is looking like it will complete just fine.  All comments are incorporated.

wbinventor:
Looks golden!

PR #662, opened 2016-05-29
fixed issue with expanded elements into their naturally-occurring isotopes

samuelshaner:
This PR fixes an issue in `element.py` where elements represented by a single letter (e.g. O, C, B, U etc) will expand to include isotopes of all elements that start with that letter (e.g. O would expand to include O *and* Os isotopes).


wbinventor:
Looks good!

PR #664, opened 2016-05-30
[HOTFIX] Improved Exact Paramter for StatePoint.get_tally(...)

### LINKED ISSUE #663: Unable to find merged tally when loading tally from statepoint
The addition of the `exact` option in the `StatePoint.get_tally()` method in PR #660 has resulted in the `MGXS.load_from_statepoint()` method producing the following error in `mgxs-part-iii.ipynb`:

    # Initialize MGXS Library with OpenMC statepoint data
    mgxs_lib.load_from_statepoint(sp)

    LookupError                               Traceback (most recent call last)
    <ipython-input-28-76d7abb36a81> in <module>()
          1 # Initialize MGXS Library with OpenMC statepoint data
    ----> 2 mgxs_lib.load_from_statepoint(sp)
    
    /home/shaner/.local/lib/python2.7/site-packages/openmc-0.7.1-py2.7.egg/openmc/mgxs/library.pyc in load_from_statepoint(self, statepoint)
        446             for mgxs_type in self.mgxs_types:
        447                 mgxs = self.get_mgxs(domain, mgxs_type)
    --> 448                 mgxs.load_from_statepoint(statepoint)
        449                 mgxs.sparse = self.sparse
        450 
    
    /home/shaner/.local/lib/python2.7/site-packages/openmc-0.7.1-py2.7.egg/openmc/mgxs/mgxs.pyc in load_from_statepoint(self, statepoint)
        694             sp_tally = statepoint.get_tally(
        695                 tally.scores, tally.filters, tally.nuclides,
    --> 696                 estimator=tally.estimator, exact=True)
        697             sp_tally = sp_tally.get_slice(
        698                 tally.scores, filters, filter_bins, tally.nuclides)

    /home/shaner/.local/lib/python2.7/site-packages/openmc-0.7.1-py2.7.egg/openmc/statepoint.pyc in get_tally(self, scores, filters, nuclides, name, id, estimator, exact)
        620         # If we did not find the Tally, return an error message
        621         if tally is None:
    --> 622             raise LookupError('Unable to get Tally')
        623 
        624         return tally

    LookupError: Unable to get Tally

If the tallies are not merged, MGXS is able to load the tallies from statepoint just fine. This error seems to occur regardless of the cross sections used in:

    mgxs_lib.mgxs_types = ['transport', 'nu-fission', 'fission', 'nu-scatter matrix', 'chi']

For instance, even if `mgxs_lib.mgxs_types` is set to `['total']`, the error is still thrown. I haven't done too much digging into the cause of the error, but it should be pretty easy to reproduce by running `mgxs-part-iii.ipynb`.

wbinventor:
This PR fixes #663 by using finer granularity for the `exact` parameter introduced for `StatePoint.get_tally(...)` in #660. In the case of MGXS, the following rules must be adhered to when loading tallies from a statepoint (whether or not the tallies were merged in the input):

* number of scores need not match (*i.e.*, scores may have been merged)
* number of nuclides need not match (*i.e.*, nuclides may have been merged)
* number of filters must match
* filters may be subsets of one another (*i.e.*, "cell" or "material" filter bins may have been merged)

This PR simply introduces `exact_filters`, `exact_nuclides` and `exact_scores` parameters in place of the single `exact` parameter. In order to meet the rules above, the `MGXS.load_from_statatepoint(...)` method simply sets `exact_filters` to `True`.

### DIFF
--- a/openmc/statepoint.py
+++ b/openmc/statepoint.py
@@ -504,16 +504,6 @@ class StatePoint(object):
         This routine searches the list of Tallies and returns the first Tally
         found which satisfies all of the input parameters.
 
-        If the "exactness" parameter is 0 (default), the input parameters
-        do not need to match the complete Tally specification and may only
-        represent a subset of the Tally's properties. If the "exactness"
-        parameter is 1 then the length of the scores, filters, nuclides
-        parameters must precisely match those of any matching Tally, but the
-        filter bins may represent a subset of those in any matching Tally
-        (useful if tallies are merged in the input). If the "exactness"
-        parameter is 2 then the values of the scores, nuclides and filters
-        parameters must precisely match those of any matching Tally.
-
         NOTE: If any of the "exact" parameters are False (default), the input
         parameters do not need to match the complete Tally specification and
         may only represent a subset of the Tally's properties. If an "exact"
--- a/openmc/statepoint.py
+++ b/openmc/statepoint.py
@@ -497,8 +497,8 @@ class StatePoint(object):
                 self.tallies[tally_id].sparse = self.sparse
 
     def get_tally(self, scores=[], filters=[], nuclides=[],
-                  name=None, id=None, estimator=None, exact_nuclides=False,
-                  exact_filters=False, exact_scores=False):
+                  name=None, id=None, estimator=None, exact_filters=False,
+                  exact_nuclides=False, exact_scores=False):
         """Finds and returns a Tally object with certain properties.
 
         This routine searches the list of Tallies and returns the first Tally
--- a/openmc/statepoint.py
+++ b/openmc/statepoint.py
@@ -497,17 +497,28 @@ class StatePoint(object):
                 self.tallies[tally_id].sparse = self.sparse
 
     def get_tally(self, scores=[], filters=[], nuclides=[],
-                  name=None, id=None, estimator=None, exact=False):
+                  name=None, id=None, estimator=None, exact_nuclides=False,
+                  exact_filters=False, exact_scores=False):
         """Finds and returns a Tally object with certain properties.
 
         This routine searches the list of Tallies and returns the first Tally
         found which satisfies all of the input parameters.
 
-        NOTE: If the "exact" parameter is False (default), the input parameters
+        If the "exactness" parameter is 0 (default), the input parameters
         do not need to match the complete Tally specification and may only
-        represent a subset of the Tally's properties. If the "exact" parameter
-        is True then the scores, filters, nuclides and estimator parameters
-        must precisely match those of any matching Tally.
+        represent a subset of the Tally's properties. If the "exactness"
+        parameter is 1 then the length of the scores, filters, nuclides
+        parameters must precisely match those of any matching Tally, but the
+        filter bins may represent a subset of those in any matching Tally
+        (useful if tallies are merged in the input). If the "exactness"
+        parameter is 2 then the values of the scores, nuclides and filters
+        parameters must precisely match those of any matching Tally.
+
+        NOTE: If any of the "exact" parameters are False (default), the input
+        parameters do not need to match the complete Tally specification and
+        may only represent a subset of the Tally's properties. If an "exact"
+        parameter is True then number of scores, filters, or nuclides in the
+        parameters must precisely match those of any matching Tally.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


openmc/statepoint.py
```diff
@@ -497,17 +497,28 @@ def sparse(self, sparse):
...
-        represent a subset of the Tally's properties. If the "exact" parameter
-        is True then the scores, filters, nuclides and estimator parameters
-        must precisely match those of any matching Tally.
+        represent a subset of the Tally's properties. If the "exactness"
+        parameter is 1 then the length of the scores, filters, nuclides
+        parameters must precisely match those of any matching Tally, but the
+        filter bins may represent a subset of those in any matching Tally
+        (useful if tallies are merged in the input). If the "exactness"
+        parameter is 2 then the values of the scores, nuclides and filters
+        parameters must precisely match those of any matching Tally.
```
samuelshaner: From your implementation, it doesn't look like an `exactness` parameter of 2 yields any different behavior than having a value of 1. 

Also, what do you think about having an `exact_filter_bins` parameter instead of allowing the exactness parameter to go to 2? It seems like an exactness parameter of 1 for nuclides and scores will be guaranteed to have the same behavior as an exactness parameter of 2, per your description.

openmc/statepoint.py
```diff
@@ -497,17 +497,28 @@ def sparse(self, sparse):
...
+
+        NOTE: If any of the "exact" parameters are False (default), the input
+        parameters do not need to match the complete Tally specification and
+        may only represent a subset of the Tally's properties. If an "exact"
+        parameter is True then number of scores, filters, or nuclides in the
+        parameters must precisely match those of any matching Tally.
```
samuelshaner: It looks like this note should now be removed as it duplicates what is said above.

openmc/tallies.py
```diff
@@ -762,6 +762,8 @@ def _can_merge_filters(self, other):
 
                 # If filters are the second mergeable filters encountered
                 elif filter1.can_merge(filter2) and merge_filters:
+                    merge_filters = True
+                    mergeable_filter = True
```
samuelshaner: Is there any reason these lines were added? Since they are local variables with a return statement right after they are set, it seems setting has no impact.

samuelshaner:
Thanks for the quick fix on this @wbinventor!

openmc/statepoint.py
wbinventor: This comment was from an early stage implementation, I should have removed it before but will now.

wbinventor: Also, I'll leave it simple with `exact_filters` only for now until (or if) we have a use case that requires even more granularity (it's not obvious to me that we ever would).

openmc/statepoint.py
wbinventor: Correct.

openmc/tallies.py
wbinventor: This was added when I was toying around with a fix and accidentally was retained in my commit.

openmc/statepoint.py
```diff
@@ -523,9 +524,18 @@ def get_tally(self, scores=[], filters=[], nuclides=[],
...
+            to those in the matching Tally. If False (default), the filters in
+            the parameters may be a subset of those in the matching Tally.
+        exact_nuclides : bool
+            If True, the number of nuclides in the parameters must be identical
+            to those in the matching Tally. If False (default), the nuclides in
+            the parameters may be a subset of those in the matching Tally.
+        exact_scores : bool
+            If True, the number of scores in the parameters must be identical
+            to those in the matching Tally. If False (default), the scores
+            in the parameters may be a subset of those in the matching Tally.
```
samuelshaner: The parameters in the method declaration are ordered `exact_nuclides`, `exact_filters`, `exact_scores` while the ordering here is `exact_filters`, `exact_nuclides`, `exact_scores`. Is consistency in the ordering of the parameters in the method declaration and docstring comment a standard that should be maintained? Looking at other files it appears so.

samuelshaner:
Tested this out with `mgxs-part-iii.ipynb` and my c5g7 ipynb and everything seems to be working as expected. I think this is ready to merge as soon as my last comment is addressed.

openmc/statepoint.py
wbinventor: I reordered the parameters to match the docstring - they are in the order of the dimensions of the tally data.

Issue #655, opened 2016-05-25
Tally Merging not performing correctly


nelsonag:
When using the Python API to generate some problems, I've found that the tally merge process is not fully merging all the tallies that it can.  It seems like every merge call merges one tally with another but no more.  
This behavior can be readily seen by running the MGXS Example notebook iv locally and viewing the continuous-energy tallies.xml file it produces on the back-end.  Even though a merge was requested, there are massive amounts of tallies present in the file with substantial room for further combining.

FYI: As a local temporary workaround I have been calling the merge routine for every domain that I am tallying over.  This seems to work and others should adopt this in the meantime.  

When this is fixed it would also be wise to update the Python API Example notebooks so the OpenMC active neutron calculation rate is not portrayed as significantly lower than what it would be.

wbinventor:
Closed with #665.

PR #665, opened 2016-05-30
Fix for #655: tally merging not thoroughly merging all tallies which can be merged.

nelsonag:
I first noticed this problem when creating the `mgxs-part-iv` notebook.  In that case, the tallies were not being merged thoroughly, resulting in many more tallies being present than actually needed.  

The problem was that `tallies._can_merge_scores()` only thought a score was merge-able if either no scores conflicted or if all scores were the same.  However, there is nothing wrong with having only one of many scores overlap.  I fixed this.

I also updated the `mgxs-part-iv` notebook and found that the runtimes didn't change drastically, indicating to me that the mesh tally is dominating the total tally penalty.  Therefore the other mgxs example notebooks don't require similar updating.

### DIFF
--- a/docs/source/pythonapi/examples/mgxs-part-iv.ipynb
+++ b/docs/source/pythonapi/examples/mgxs-part-iv.ipynb
@@ -433,7 +433,7 @@
    "outputs": [
     {
      "data": {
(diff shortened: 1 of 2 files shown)


wbinventor:
Thanks for the quick fix @nelsonag!

PR #681, opened 2016-07-05
Fixed issue in tally.get_slice() method for mesh filters

samuelshaner:
In the tally.get_slice(…) method, the filter_bins parameter is supposed to be a list of tuples. The tuple containing the bins for a mesh filter contains 3-tuples with the mesh indices. This looks correct as an input for the nested tally.get_values(…) method. However, lower down the tuple of 3-tuples are converted to filter bins and on line 2993 of tallies.py these filter bins are assigned to the filter. The problem is that a mesh filter only has one filter bin, the mesh id, although the num_bins for the filter equal the number of mesh cells. The bin_indices and num_bins should really be set in the same way that they are set for distribcell, as shown below. This PR makes that fix in the tally.get_slice(...) method and in my testing seems to fix cases where a slice of a tally with a mesh filter is requested.


wbinventor:
@samuelshaner I'll make a quick review of this and your other PR sometime today. One thought that comes to mind - can you add a test of this functionality to the "test_tally_slice_merge" test?

tests/test_tally_slice_merge/test_tally_slice_merge.py
```diff
@@ -69,8 +77,18 @@ def _build_inputs(self):
         for nuclide in self.nuclides:
             distribcell_tally.add_nuclide(nuclide)
 
+        mesh_tally = openmc.Tally(name='mesh tally')
+        mesh_tally.estimator = 'tracklength'
+        mesh_tally.add_filter(mesh_filter)
+        mesh_tally.add_filter(merged_energies)
+        for score in self.scores:
```
wbinventor: This can simply be:

```python
mesh_tally.scores = self.scores
mesh_tally.nuclides = self.nuclides
```

tests/test_tally_slice_merge/results_true.dat
```diff
@@ -1,36 +1,37 @@
...
-8           6.25e-07           2.00e+01    21   U-235     fission 3.12e-02   2.56e-03
-9           6.25e-07           2.00e+01    21   U-235  nu-fission 7.65e-02   6.24e-03
-10          6.25e-07           2.00e+01    21   U-238     fission 2.00e-02   1.30e-03
-11          6.25e-07           2.00e+01    21   U-238  nu-fission 5.56e-02   3.78e-03
-12          6.25e-07           2.00e+01    27   U-235     fission 1.39e-02   1.06e-03
-13          6.25e-07           2.00e+01    27   U-235  nu-fission 3.40e-02   2.61e-03
-14          6.25e-07           2.00e+01    27   U-238     fission 9.72e-03   1.21e-03
-15          6.25e-07           2.00e+01    27   U-238  nu-fission 2.71e-02   3.80e-03         sum(distribcell)  energy low [MeV]  energy high [MeV] nuclide       score     mean  std. dev.
+   cell  energy low [MeV]  energy high [MeV] nuclide    score     mean  std. dev.
+0    21          0.00e+00           6.25e-07   U-235  fission 1.08e-01   7.94e-03   cell  energy low [MeV]  energy high [MeV] nuclide       score     mean  std. dev.
```
wbinventor: Can you just clarify what changed here? It looks like the new results include the cell ID for some reason?

tests/test_tally_slice_merge/test_tally_slice_merge.py
```diff
@@ -138,7 +156,24 @@ def _get_results(self, hash_output=False):
 
         # Append merged Tally Pandas DataFrame to output string
         df = merge_tally.get_pandas_dataframe()
-        outstr += df.to_string()
+        outstr += df.to_string() + '\n'
+
+        # Extract the mesh tally
+        mesh_tally = sp.get_tally(name='mesh tally')
+
+        # Sum up a few subdomains from the mesh tally
```
wbinventor: Looks like a thorough test to me

wbinventor:
I am fine with merging this in once my few comments are addressed.

tests/test_tally_slice_merge/results_true.dat
samuelshaner: The old results have the cell ID, but it was after the energy columns. Fixing the typo on lines 41-42 of `test_tally_slice_merge.py` results in the energy and cell filters being added in the opposite order so now the cell ID column shows up before the energy columns.

samuelshaner:
I addressed the comments for this PR and think it is ready to be merged or further reviewed.

PR #683, opened 2016-07-12
Tally for non-fission heating

smharper:
This replaces #682 

This PR allows us to compute the energy production from capture and inelastic scattering reactions.  The new score name is simply "Q".  I'm open to changing the name if yall have a better idea.  Maybe "non-fission-heat"?

As for verification, I'm checking against "CASMO-5 energy release per fission model" by Rhodes, Smith, and Xu ([link](https://www.researchgate.net/profile/Joel_Rhodes/publication/267788591_CASMO-5_energy_release_per_fission_model/links/5575949e08ae7521586ac607.pdf)).  They give a few "mean energy released per capture" values.  Essentially, it's a reaction-rate-averaged estimate of reaction Q-values.  To get the same value from OpenMC, I commented out all the reactions with outgoing neutrons (the CASMO paper only mentions capture) in the new ace.F90 subroutine, run a simulation, and divide the tallied Q by absorption - fission.

The CASMO paper gives results for various LWR lattices.  In particular, I'm trying to recreate a result they give for a 17x17 PWR lattice with 800 ppm boron and 4.5% enrichment.  For geometry and material compositions, I used a 2D BEAVRS assembly, no BP, modified for 800 ppm and 4.5%.  The CASMO paper gives a Q of 5.40 MeV/capture and I get 4.78 +/- 0.03.  These numbers aren't the same, but I think they're close enough.  Not having the exact geometry and materials from the C5 paper makes the comparison somewhat uncertain, and it looks like C5 also increases the Q-values they use to account for subsequent fast decays.


paulromano:
To account for heating properly, we need to be using heating numbers instead of just the Q value. To understand why, it's instructive to look at the example of elastic scattering. As discussed in the [NJOY/HEATR manual](http://t2.lanl.gov/nis/codes/NJOY12/NJOY2012.50.pdf), in an ideal world, you'd have the explicit distribution of all charged reaction products (including recoil nucleus) and could just calculate the heating using the sum of the average outgoing energy of charged products. However, in most cases we don't have enough information to do that so instead NJOY defaults to using an energy-balance method where the heating number is calculated as:

image

where E is the incoming energy of the neutron and En and Eg are the average outgoing energies of neutrons and photons. For elastic scattering, Q is zero and there are no photons, so the heating is

image

For elastic scattering, one can show the average outgoing energy is:

image

Thus, the heating from elastic scattering is

image

So, you see that even though the Q value is zero, there is most definitely still heating that results from the recoil nucleus. From my understanding, heating from elastic scattering is not negligible, especially in water.

smharper:
Ah, now I remember why I chose to call this tally `Q` rather than something involving the word `heat`, and this comment is the reason.

For the purposes of depletion we don't care where the heat is deposited only that it is.  So it is sufficient to know that every neutron will deposit it's kinetic energy _somewhere_ in the reactor.  That kinetic energy is actually captured by our `kappa-fission` tally (since our neutrons got their kinetic energy (most of it anyway) from fission).

paulromano:
Gotcha -- that makes sense now given the context. How do you propose that we handle leakage of particles (and hence energy)? Neutrons probably don't carry away much energy but it's harder to make the same argument about gammas.

smharper:
I'm totally ignoring leakage for now.  I'm not convinced that even the gamma leakage is significant for most reactors.  In a PWR or BWR, a gamma needs to make it all the way through a baffle, reflector, barrel, and downcomer, before you can argue that it's heat won't make its way into the coolant.

paulromano:
Ok, I think for starters this should be acceptable. We can talk to our friends in ANL/NE about how much leakage matters for non-LWRs.

paulromano:
Also, I seem to remember that capturing the energy from fission product decays can actually become important as burnup increases (i.e., there's a good reason CASMO increases Q values for a few nuclides).

nelsonag:
Good point @paulromano. Would it be good then to give the user the ability
to override the Q value you would get from ACE/this PR so they can do
something like this? Obviously it doesn't really matter until depletion is
in, but this PR seems geared towards that

paulromano:
@nelsonag It will be easy to do that with the new `openmc.data` module (coming soon to a Python API near you).

smharper:
After further thought, I've decided to close this PR.  For depletion purposes, we already need to compute all the per-nuclide reaction rates of all the capture and neutron multiplying reactions and you can simply multiply those reaction rates by Q.  And you need post-processing anyway to account for the subsequent decays.

PR #689, opened 2016-08-01
Fix unassigned energy in SCORE_PROMPT_NU_FISSION

smharper:
It was previously possible for prompt-nu-fission to use an energy variable that was never assigned.  This caused NaN's in one of our current tests and lots MGXS of regressions in the #678 tests.

Relatedly, maybe we should modify the test suite so that it fails if any of the tallies output a NaN or an Inf


samuelshaner:
This looks good to me. Thanks for cleaning up my mistake. I'm fine with this being merged, but will let this simmer overnight before merging in case there are any other comments.

PR #691, opened 2016-08-04
Extension of MGXS to support multi-delayed-group XS (MDGXS)

samuelshaner:
This PR introduces the abstract MDGXS class which extends the abstract MGXS class to support multi-delayed-group cross section generation. Concrete classes have been implemented to support multi-energy-group and multi-delayed-group `chi-delayed`, `delayed-nu-fission`, and `beta` (delayed neutron fraction) cross sections.

**MDGXS**

The MDGXS abstract class has been implemented to generate cross sections that can depend on energy *and* delayed groups. However, unlike the MGXS class where the energy groups filter is mandatory, including the `delayedgroup` filter is optional. Therefore, cross sections like `chi-delayed` that are not expected to have a large dependence on delayed group can be tallied without the delayedgroup filter if the user prefers. 

**IPython Notebooks**

Two IPython notebooks, `mdgxs-part-i.ipynb` and `mdgxs-part-ii.ipynb`, have been included to demonstrate how to use the MDGXS class to produce useful data, such as the emission spectra for prompt and delayed neutrons (shown below). Additionally, the `mdgxs-part-ii.ipynb` notebook demonstrates how to generate a MGXS library with a mesh domain and tally surface currents. These notebooks were adapted from `mgxs-part-i.ipynb` and `mgxs-part-iii.ipynb`.

mdgxs

**Mesh Surface Current Tallies**

Currently, the mesh surface current tallies in the Python API are not being computed properly. The root of the issue is that the Python API assumes that each mesh cell contains the six surface currents for its surfaces (in 3D) while the fortran code assigns each cell the outgoing and incoming surface currents for the x-max, y-max, and z-max surfaces, using ghost cells to get the x-min, y-min, and z-min currents for cells on the x-min, y-min, and z-min geometry boundaries. I saw three options to fix this issue:
- Have a function in the Python API to translate the surface currents in the statepoint file from the ordering currently in the fortran code to the ordering in the Python API.
- Change the Python API to arrange the surface currents using the format in the fortran code (i.e. store the x-max, y-max, and z-max outgoing and incoming surface currents with each cell). However, this might seem like an odd arrangement from the users perspective.
- Change the fortran code to store the six outgoing surface currents with each cell.

I chose to implement option 3 since this seemed like the most intuitive way to store the surface currents. Now, the fortran code stores the outgoing currents for each surface with the cell, as printed in the `tallies.out` and Python API output currents below:

    ==================>     TALLY 10036: CURRENT TALLY     <===================

    Mesh Index (1, 1, 1)
        Outgoing Current to Left     0.0            +/- 0.0
        Outgoing Current to Right    2.98600E-02    +/- 6.78241E-04
        Outgoing Current to Back     0.0            +/- 0.0
        Outgoing Current to Front    3.09100E-02    +/- 6.36011E-04
        Outgoing Current to Bottom   0.0            +/- 0.0
        Outgoing Current to Top      0.0            +/- 0.0

        mesh 1	    surface	nuclide	score	mean	std. dev.
        x	y	z					
    0	1	1	1	x-min	total	current	0.00000	0.000000
    1	1	1	1	x-max	total	current	0.02986	0.000678
    2	1	1	1	y-min	total	current	0.00000	0.000000
    3	1	1	1	y-max	total	current	0.03091	0.000636
    4	1	1	1	z-min	total	current	0.00000	0.000000
    5	1	1	1	z-max	total	current	0.00000	0.000000


wbinventor:
@samuelshaner I'll take a look at this over the next day or two. One question comes to mind from my cursory overview - is there a true need to add new MDGXS tests? Or can you simply extend the existing MGXS tests by combining the `MGXS_TYPES` and `MDGXS_TYPES` into a single list of items to tally for a `Library`?

samuelshaner:
The `MDGXS_TYPES` could be tested in combination with the `MGXS_TYPES` in the MGXS tests. I'll remove the new tests and modify the MGXS tests.

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,129 @@ def merge(self, other):
...
+
+    @groups.setter
+    def groups(self, groups):
+        cv.check_type('groups', groups, Iterable, Integral)
+        cv.check_greater_than('number of delayed groups', len(groups), 0)
+
+        # Check that the groups are within [1, MAX_DELAYED_GROUPS]
+        for group in groups:
+            cv.check_greater_than('delayed group', group, 0)
+            cv.check_less_than('delayed group', group, MAX_DELAYED_GROUPS + 1)
```
paulromano: There's an optional `equality` argument on `check_less_than` that you can use here.

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,129 @@ def merge(self, other):
...
+
+        self._groups = np.array(groups, dtype=int)
```
paulromano: Use `np.asarray`

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,129 @@ def merge(self, other):
...
+        -------
+        bool
+            Whether the delayed groups can be merged
+
+        """
+
+        if not isinstance(other, DelayedGroups):
+            return False
+        else:
+            return True
```
paulromano: All four lines can be replaced by
`return isinstance(other, DelayedGroups)`

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,129 @@ def merge(self, other):
...
+        if not self.can_merge(other):
+            raise ValueError('Unable to merge delayed groups')
+
+        # Create deep copy to return as merged delayed groups
+        merged_groups = copy.deepcopy(self)
+
+        # Merge unique filter bins
+        groups = np.concatenate((self.groups, other.groups))
+        groups = np.unique(groups)
+        groups = sorted(groups)
```
paulromano: Use the `sort()` method of `ndarray` to avoid another copy.

openmc/mgxs/library.py
```diff
@@ -239,22 +255,34 @@ def name(self, name):
 
     @mgxs_types.setter
     def mgxs_types(self, mgxs_types):
+        all_mgxs_types = np.append(openmc.mgxs.MGXS_TYPES,
+                                   openmc.mgxs.MDGXS_TYPES)
```
paulromano: Is there a reason you want an `ndarray` here rather than a list, i.e. `openmc.mgxs.MGXS_TYPES + openmc.mgxs.MDGXS_TYPES`?

openmc/mgxs/library.py
```diff
@@ -239,22 +255,34 @@ def name(self, name):
...
-                cv.check_value('mgxs_type', mgxs_type, openmc.mgxs.MGXS_TYPES)
+                cv.check_value('mgxs_type', mgxs_type, all_mgxs_types)
             self._mgxs_types = mgxs_types
 
     @by_nuclide.setter
     def by_nuclide(self, by_nuclide):
         cv.check_type('by_nuclide', by_nuclide, bool)
+
+        if by_nuclide == True and self.domain_type == 'mesh':
+            raise ValueError('Unable to create MGXS library by nuclide with ' +
```
paulromano: The + is superfluous. String concatenation is implicit when separated only by whitespace.

openmc/mgxs/library.py
```diff
@@ -239,22 +255,34 @@ def name(self, name):
...
+                             'mesh domain')
+
         self._by_nuclide = by_nuclide
 
     @domain_type.setter
     def domain_type(self, domain_type):
         cv.check_value('domain type', domain_type, openmc.mgxs.DOMAIN_TYPES)
+
+        if self.by_nuclide == True and domain_type == 'mesh':
+            raise ValueError('Unable to create MGXS library by nuclide with ' +
```
paulromano: + is superfluous

src/cmfd_data.F90
```diff
@@ -236,71 +236,101 @@ subroutine compute_xs()
...
+                  matching_bins(i_filter_surf) = OUT_RIGHT
+                  score_index = sum((matching_bins(1:size(t % filters)) - 1) &
+                       * t % stride) + 1 ! incoming
+                  cmfd % current(2,h,i,j,k) = t % results(1,score_index) % sum
+                end if
 
                 ! Right surface
+                if (i < nx) then
+                  matching_bins(i_filter_mesh) = mesh_indices_to_bin(m, &
+                     (/ i+1, j, k /) )
```
paulromano: Continuation line should be indented five spaces per style guide

openmc/mgxs/library.py
samuelshaner: Nope, I'll change that to a list.

samuelshaner:
It looks like travis has failed for my last two commits due to an error downloading/initializing miniconda. Has anyone had travis fail during this step? Any ideas on what the issue could be?

paulromano:
@samuelshaner I would just wait a little while and then restart the build.

samuelshaner:
Can you restart the build with pushing another commit? If so, how is that done?

paulromano:
Look for the button I've highlighted in yellow

screenshot from 2016-08-05 09-25-49

samuelshaner:
Hmmm....I don't see that button.

travis_restart_build

paulromano:
Hmm, looks like you weren't on the "openmc-read-write" github team. I just added you so hopefully that will fix it?

samuelshaner:
That fixed it! I also hadn't synced my github account to travis so I did that too.

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,127 @@ def merge(self, other):
...
         return merged_groups
+
+
+class DelayedGroups(object):
+    """A delayed groups structure used for multi-delayed-group parameters.
+
+    Parameters
+    ----------
+    groups : Iterable of Int
+        The delayed groups
```
wbinventor: Perhaps it is just me, but I'm not clear on what this would actually be. If my cross section library has 6 delayed groups, then this allows me to specify some list - e.g., `[1,2,4,6]` - of the groups that I want to tally?

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,127 @@ def merge(self, other):
...
+        # If this object has been copied before, return the first copy made
+        else:
+            return existing
+
+    def __eq__(self, other):
+        if not isinstance(other, DelayedGroups):
+            return False
+        elif self.num_groups != other.num_groups:
+            return False
+        elif np.allclose(self.groups, other.groups):
```
wbinventor: These last four lines can be collapsed into two:

```python
else:
    return np.allclose(self.groups, other.groups)
```

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1564 @@
...
+        mdgxs.by_nuclide = by_nuclide
+        mdgxs.name = name
+        return mdgxs
+
+    def get_xs(self, groups='all', subdomains='all', nuclides='all',
+               xs_type='macro', order_groups='increasing',
+               value='mean', delayed_groups='all', **kwargs):
+        """Returns an array of multi-delayed-group cross sections.
+
+        This method constructs a 2D NumPy array for the requested
```
wbinventor: Would it be worth clarifying what the indices in the array correspond to?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1564 @@
...
+        if self.domain_type == 'mesh':
+            mesh_str = 'mesh {0}'.format(self.domain.id)
+            df.sort_values(by=[(mesh_str, 'x'), (mesh_str, 'y'), \
+                               (mesh_str, 'z')] + columns, inplace=True)
+        else:
+            df.sort_values(by=[self.domain_type] + columns, inplace=True)
+        return df
+
+        # Select out those delayed groups the user requested
+        if not isinstance(delayed_groups, basestring):
```
wbinventor: Is any of the code above this needed? Or can you simply call the parent `MGXS`

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1564 @@
...
+
+        # Make directory if it does not exist
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+
+        filename = os.path.join(directory, filename)
+        filename = filename.replace(' ', '-')
+
+        # Get a Pandas DataFrame for the data
+        df = self.get_pandas_dataframe(groups=groups, xs_type=xs_type,
```
wbinventor: It looks like this is the only line of code that differs from the parent class method? Sure would be nice to gut a lot of this redundant code throughout this class, but I know that isn't entirely straightforward.

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1564 @@
...
+        if self.nuclides != other.nuclides:
+
+            # The nuclides must be mutually exclusive
+            for nuclide in self.nuclides:
+                if nuclide in other.nuclides:
+                    msg = 'Unable to merge Chi Delayed with shared nuclides'
+                    raise ValueError(msg)
+
+            # Concatenate lists of nuclides for the merged MGXS
+            merged_mdgxs.nuclides = self.nuclides + other.nuclides
```
wbinventor: Hmmm, is this really how we have been merging nuclides? Seems like we should eliminate any common nuclides.

wbinventor: Oh I see that there is an error check above for this shared nuclides.

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1564 @@
...
+    num_nuclides : int
+        The number of nuclides for which the multi-group cross section is
+        being tracked. This is unity if the by_nuclide attribute is False.
+    nuclides : Iterable of str or 'sum'
+        The optional user-specified nuclides for which to compute cross
+        sections (e.g., 'U-238', 'O-16'). If by_nuclide is True but nuclides
+        are not specified by the user, all nuclides in the spatial domain
+        are included. This attribute is 'sum' if by_nuclide is false.
+    sparse : bool
+        Whether or not the MGXS' tallies use SciPy's LIL sparse matrix format
```
wbinventor: You use "MDGXS" in some of the docstrings but "MGXS" in this one.

openmc/mgxs/groups.py
```diff
@@ -299,3 +303,127 @@ def merge(self, other):
         # Assign merged edges to merged groups
         merged_groups.group_edges = list(merged_edges)
         return merged_groups
+
+
+class DelayedGroups(object):
```
wbinventor: So a high level question - why is a `DelayedGroups` class needed? It seems like it is effectively just a list of integers without any unique functionality. What is your reasoning for creating a class rather than simply storing the list of integers directly to the `MDGXS.delayed_groups` attribute?

openmc/mgxs/groups.py
samuelshaner: Correct, it is just a list of delayed groups and allows the user to specify some list of the delayed groups. Since the number of delayed groups differs by cross section library and so users can tally to some subset of the delayed groups, I figured having this object would be best.

tests/test_cmfd_feed/results_true.dat
```diff
@@ -124,6 +124,10 @@ tally 3:
 1.020705E+00
 5.413570E-02
 tally 4:
+3.049469E+00
```
wbinventor: Why is there additional info in the CMFD test results given your changes to the surface tallies? Sorry for my ignorance, I didn't review any of the Fortran diff yet since I'm not familiar with that part of the codebase.

tests/test_mgxs_library_distribcell/test_mgxs_library_distribcell.py
```diff
@@ -6,29 +6,39 @@
 import hashlib
 sys.path.insert(0, os.pardir)
 from testing_harness import PyAPITestHarness
+from input_set import PinCellInputSet
```
wbinventor: Why are you now using the `PinCellInputSet` for this test? It seems to me that this defeats the purpose of this test of distribcell MGXS.

openmc/mgxs/mdgxs.py
samuelshaner: Much of the code is redundant, but some of the operations require the `num_delayed_groups` for tiling the column values for the energy groups. I'll look into seeing if I can call the `MGXS` class `get_pandas_dataframe(...)` method and then just re-tile the column values.

openmc/mgxs/mdgxs.py
samuelshaner: I'll look into splitting things up to reduce redundant code since I know there is a fair amount.

wbinventor:
I went through my first review of this and overall things look good. My main comments are summarized below:

* **FORTRAN** - I have not reviewed any of the changes to the Fortran code. It would be great if another developer could help me out in this area.
* **Redundant Code** - It seems like a lot of common code between the `MDGXS` and `MGXS` classes. This perhaps reflects the poor design of the `MGXS` class methods which are not easy to inherit and extend. I don't see this as a dealbreaker for this PR, but something to consider if more `MGXS` subclasses are added in the future.
* **Distribcell MGXS Test** - The distribcell MGXS test should utilize a geometry with more than one replicated cell. The new `PinCellInputSet` may be needed for some reason I'm not aware of, but is probably not a good one to stress the distribcell algorithm and data processing infrastructure.

Overall, this should be a pretty straightforward merge.

tests/test_cmfd_feed/results_true.dat
samuelshaner: The way that the fortran code stores currents is different. Before the fortran code stored the inward and outward current for the x-max, y-max, and z-max surfaces with each cell, using ghost cells to account for the currents on the x-min, y-min, and z-min surfaces of the x-min, y-min, and z-min geometry boundaries. Kinda confusing. Now each cell just stores the outward current for all six of its surfaces. This will result in the currents being stored in a different order and hence why the cmfd tests needed to be updated. You will notice the only difference in the test is with regards to tally 4, the surface current tally, as expected.

tests/test_mgxs_library_distribcell/test_mgxs_library_distribcell.py
samuelshaner: I chose to change the input because the distribcell chosen on the previous input did not have fissile isotopes and hence would result in zeros for all the MDGXS tallies. Choosing a fissile material cell would then drastically increase the time to run this test because the only fissile material cells are replicated many, many times across the geometry. It would be nice to have an assembly test or something where there are fissile material cells present and replicated, but not too many times to make the test too onerous to run. Thoughts? Change back to the old test and accept the zeros? Make a new test with an assembly?

wbinventor: Ok, I see your point. I think the optimal solution would be to have an assembly geometry with replicated pin cells which produces non-zero M(D)GXS data. Is it reasonable to ask for a `AssemblyInputSet` for this PR? If not, revert back to the original `InputSet` for now and we can aim for an assembly in a future PR.

samuelshaner: I'll go ahead and add an `AssemblyInputSet`.

samuelshaner:
Does anyone have any thoughts about why the Python 3 tests could be failing? The tests pass just fine locally with Python 2 and 3 but fail with Python 3 on travis. I haven't ventured much into Python 3 so any help would be much appreciated!

wbinventor:
Which version of Python 3 do you have? I believe that Travis is using 3.5. I'll take a quick look through your code again to see if I can spot anything obvious.

samuelshaner:
I was using Python 3.4. I'll give Python 3.5 a try.

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1448 @@
...
+    def num_delayed_groups(self):
+        if self.delayed_groups == None:
+            return 0
+        else:
+            return len(self.delayed_groups)
+
+    @delayed_groups.setter
+    def delayed_groups(self, delayed_groups):
+
+        cv.check_type('delayed groups', delayed_groups, list, int)
```
wbinventor: Does `delayed_groups` have to be a `list` or can it be any `Iterable`?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1448 @@
...
+        openmc.mgxs.MDGXS
+            A new MDGXS object which encapsulates the subset of data requested
+            for the nuclide(s) and/or energy group(s) and/or delayed group(s)
+            requested in the parameters.
+
+        """
+
+        cv.check_iterable_type('nuclides', nuclides, basestring)
+        cv.check_iterable_type('energy_groups', groups, Integral)
+        cv.check_type('delayed groups', delayed_groups, list, int)
```
wbinventor: ditto - perhaps `delayed_groups` can be more generally be any `Iterable` but it's nbd

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1448 @@
...
+
+        if len(delayed_groups) != 0:
+            filter_bins.append(tuple(delayed_groups))
+            filters.append('delayedgroup')
+
+        if filters != []:
+
+            # Slice nu-fission-out tally along energyout filter
+            delayed_nu_fission_out = slice_xs.tallies['delayed-nu-fission-out']
+            tally_slice = delayed_nu_fission_out.get_slice\
```
wbinventor: space before the `\` line continuation character

wbinventor:
Nothing stands out to me as being Python 3 incompatible. Since the MGXS tests are the only ones that fail, you might try reverting those tests to only use `MGXS_TYPES`. If it passes, then that isolates the issue to your new "mdgxs.py" file. If it fails, then the issue is elsewhere. What do you think?

openmc/mgxs/mdgxs.py
samuelshaner: It could technically be an Iterable, but would require some changes in other places. For instance, the behavior is different for lists and np.array (which both qualify as Iterables) when trying to do things like concatenations, which worries me. Some of the Iterables in mgxs.py (e.g. `nuclides`) use concatenations which would fail if they were np.arrays. What are your thoughts on the general classification of things as Iterables and concerns over things like concatenations.

samuelshaner:
I'll give that a try.

openmc/mgxs/mdgxs.py
wbinventor: Ok, your concerns make sense to me. Perhaps you can allow the user to pass in an `Iterable` of any type, then convert it to a `list` for storage to the `_delayed_groups` attribute? This would give the user flexibility while allowing you to flexibly do list concatenation under-the-hood.

samuelshaner:
I've gone through this PR and addressed the comments. A few notes on what I've changed:

- **Redundant Code** - I've tried to minimize the amount of redundant code as much as possible. Further reducing redundant code will probably require a bigger refactor effort.
- **Distribcell MGXS Test** - I added an `AssemblyInputSet` to allow the distribcell test to be run in a reasonable amount of time. This will probably be useful for future tests as well.
- **mgxs.get_xs(...)** - Added `squeeze` option to this method. I've found that having the squeeze done without an option is kind of confusing since it outputs an array with a different structure than is described in the comment. The squeeze option now makes this a little more transparent.
- **Fortran Code** - I'm not too familiar with the cmfd code, so it would be nice to get someone to take a look at that.

openmc/mgxs/mdgxs.py
samuelshaner: I went through the comments and tried to fix all uses of `MDGXS` and `MGXS`.

docs/source/pythonapi/examples/mgxs-part-iii.ipynb
```diff
@@ -32,7 +32,7 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "/home/romano/miniconda3/envs/default/lib/python3.5/site-packages/matplotlib/__init__.py:1350: UserWarning:  This call to matplotlib.use() has no effect\n",
+      "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/matplotlib/__init__.py:1357: UserWarning:  This call to matplotlib.use() has no effect\n",
```
wbinventor: Do you want to add your new `MDGXS` types to this Notebook? Specifically, the text under Cell 18 has a list of all of the types supported by a `Library`, perhaps add your new ones there?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1587 @@
...
+        # energy groups such that data is from fast to thermal
+        if self.domain_type == 'mesh':
+            mesh_str = 'mesh {0}'.format(self.domain.id)
+            df.sort_values(by=[(mesh_str, 'x'), (mesh_str, 'y'), \
+                               (mesh_str, 'z')] + columns, inplace=True)
+        else:
+            df.sort_values(by=[self.domain_type] + columns, inplace=True)
+
+        return df
+
```
wbinventor: Looks like something happened here - perhaps a mistaken copy-paste. Remove lines 845-857.

docs/source/pythonapi/examples/mgxs-part-iii.ipynb
samuelshaner: Sure.

openmc/mgxs/mdgxs.py
samuelshaner: Yeah, I thought I could shorten it to just the lines below, but then realized it didn't work so I had to revert to the mostly copied code above. I'll remove the lines 845-857.

openmc/mgxs/mgxs.py
```diff
@@ -1806,8 +1819,6 @@ def get_xs(self, in_groups='all', out_groups='all',
                                           filter_bins=filter_bins,
                                           nuclides=query_nuclides, value=value)
 
-        xs = np.nan_to_num(xs)
```
wbinventor: Why remove this?

openmc/mgxs/mgxs.py
```diff
@@ -1351,8 +1360,6 @@ def build_hdf5_store(self, filename='mgxs.h5', directory='mgxs',
                 std_dev = self.get_xs(subdomains=[subdomain], nuclides=[nuclide],
                                       xs_type=xs_type, value='std_dev',
                                       row_column=row_column)
-                average = average.squeeze()
```
wbinventor: Why remove this?

openmc/mgxs/mgxs.py
```diff
@@ -3638,8 +3658,6 @@ def get_xs(self, in_groups='all', out_groups='all',
                                           filter_bins=filter_bins,
                                           nuclides=query_nuclides, value=value)
 
-        xs = np.nan_to_num(xs)
```
wbinventor: Ditto

openmc/mgxs/mgxs.py
samuelshaner: I just moved in down to after where the `micro` could be computed. I went through all the `get_xs(...)` methods and tried to make them all consistent with where they do nan to num and squeeze the array.

openmc/mgxs/mgxs.py
```diff
@@ -4541,6 +4564,10 @@ def get_xs(self, groups='all', subdomains='all', nuclides='all',
             Defaults to 'increasing'.
         value : {'mean', 'std_dev', 'rel_err'}
             A string for the type of value to return. Defaults to 'mean'.
+        squeeze : bool
+            A boolean representing whether to eliminate the extra dimensions
+            of the multi-dimensional array this is to be retured. Defaults to
```
wbinventor: typo: "retured" -> "returned"

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,1587 @@
...
+            Return the cross section indexed according to increasing or
+            decreasing energy groups (decreasing or increasing energies).
+            Defaults to 'increasing'.
+        value : {'mean', 'std_dev', 'rel_err'}
+            A string for the type of value to return. Defaults to 'mean'.
+        delayed_groups : list of int or 'all'
+            Delayed groups of interest. Defaults to 'all'.
+        squeeze : bool
+            A boolean representing whether to eliminate the extra dimensions
+            of the multi-dimensional array this is to be retured. Defaults to
```
wbinventor: typo: "retured" -> "returned"

wbinventor: Also, perhaps re-phrase as "this is to be returned" -> "to be returned"

openmc/mgxs/mgxs.py
samuelshaner: The `get_xs(...)` method now has a squeeze parameter that defaults to true, so this would be redundant.

tests/test_mgxs_library_condense/test_mgxs_library_condense.py
```diff
@@ -9,6 +9,7 @@
 from input_set import PinCellInputSet
 import openmc
 import openmc.mgxs
+import numpy as np
```
wbinventor: Doesn't look like this is needed any more.

tests/test_mgxs_library_distribcell/test_mgxs_library_distribcell.py
```diff
@@ -6,29 +6,40 @@
 import hashlib
 sys.path.insert(0, os.pardir)
 from testing_harness import PyAPITestHarness
+from input_set import AssemblyInputSet
 import openmc
 import openmc.mgxs
+import numpy as np
```
wbinventor: Ditto

tests/test_mgxs_library_mesh/test_mgxs_library_mesh.py
```diff
@@ -8,6 +8,7 @@
 from testing_harness import PyAPITestHarness
 import openmc
 import openmc.mgxs
+import numpy as np
```
wbinventor: Ditto

tests/test_mgxs_library_no_nuclides/test_mgxs_library_no_nuclides.py
```diff
@@ -9,6 +9,7 @@
 from input_set import PinCellInputSet
 import openmc
 import openmc.mgxs
+import numpy as np
```
wbinventor: ditto

tests/test_mgxs_library_nuclides/test_mgxs_library_nuclides.py
```diff
@@ -9,6 +9,7 @@
 from input_set import PinCellInputSet
 import openmc
 import openmc.mgxs
+import numpy as np
```
wbinventor: and here

openmc/mgxs/mgxs.py
wbinventor: Alright I see

openmc/mgxs/mgxs.py
wbinventor: Ok that's fine. I was just worried since I thought I saw one or two place where it was removed but not added back in. Must have been mistaken. This `nan_to_num` is pretty important if you generate "noisy" MGXS with few particle histories, and is hence important to my work!

wbinventor:
I made a final pass through the Python code in this PR and am happy. One final thought - @samuelshaner, did you want to add your new MDGXS Jupyter Notebooks to the documentation? I didn't see the links show up in the Sphinx source anywhere.

Otherwise, this PR looks good to go to me. I haven't thoroughly reviewed the changes to the Fortran source, so if @paulromano or @smharper can help that would be great.

samuelshaner:
Thanks for the quick review @wbinventor! I just added the MDGXS Jupyter Notebooks to the documentation.

wbinventor:
@samuelshaner can you update this? Looks like merging in #704 resulted in a merge conflict. I don't want to sit on this much longer, and will merge it in once it is up-to-date.

samuelshaner:
There was just one minor merge conflict that I fixed. I also fixed a minor issue in `tally.F90` that caused a seg fault when the an energy filter and mesh filter were both used and the energy filter was not the first filter read in. 

All of the tests pass locally, but almost all of them fail on travis. @wbinventor @paulromano - any ideas on what could recent modifications might be causing this behavior?

paulromano:
Try clearing the cache on Travis and restarting the build

samuelshaner:
Once again @paulromano to the rescue! Clearing the caches did the trick and all tests are passing.

PR #696, opened 2016-08-06
Support MGXS Tallied over a mesh in Library.create_mg_library and related

nelsonag:
This PR builds off the great work in #687 to support the mesh domain for MGXS.  This PR adds  the library back-end components needed to take these mesh MGXS and output a library for MG mode.


openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
 
         return element
+
+    def build_cells(self, bc=['reflective'] * 6):
+        """Generates a list of cells which mimic the mesh geometry.
+
+        Parameters
+        ----------
+        bc : iterable of {'reflective', 'periodic', or 'vacuum'}
+            Boundary conditions for each of the six faces of a parallelopiped:
```
samuelshaner: I think it is `parallelepiped`

openmc/mesh.py
```diff
@@ -181,6 +183,21 @@ def __repr__(self):
         string += '{0: <16}{1}{2}\n'.format('\tPixels', '=\t', self._width)
         return string
 
+    def cell_generator(self):
+        """Generator function to traverse through every [i,j,k] index
+        of the mesh.
+
```
wbinventor: This is great! I've thought that we needed something like this for a long time. Could you add an "Examples" block to this docstring to illustrate how one can use this method?

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
             subelement.text = ' '.join(map(str, self._width))
 
         return element
+
+    def build_cells(self, bc=['reflective'] * 6):
+        """Generates a list of cells which mimic the mesh geometry.
```
wbinventor: "mimicking the mesh geometry" -> perhaps clarify this to mean that you build separate parallelepipeds for each individual Cartesian mesh cell.

wbinventor: As opposed to building a single parallelepiped for the outermost boundaries of the mesh.

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
+
+        """
+
+        cv.check_length('bc', bc, length_min=6, length_max=6)
+        for entry in bc:
+            cv.check_value('bc', entry, ['transmission', 'vacuum',
+                                         'reflective', 'periodic'])
+
+        if len(self.dimension) == 2:
+            twod = True
```
wbinventor: This `if-else` block can be compressed to a single line:

```python
twod = len(self.dimension) == 2
```

openmc/mesh.py
samuelshaner: Also, it looks like `transmission` surfaces are allowed as well.

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
+        xplanes = [openmc.XPlane(x0=self.lower_left[0],
+                                 boundary_type=bc[3]),
+                   openmc.XPlane(x0=self.upper_right[0],
+                                 boundary_type=bc[1])]
+        yplanes = [openmc.YPlane(y0=self.lower_left[1],
+                                 boundary_type=bc[2]),
+                   openmc.YPlane(y0=self.upper_right[1],
+                                 boundary_type=bc[0])]
+        if twod:
+            zplanes = [openmc.ZPlane(z0=-1.E50, boundary_type=bc[5]),
```
wbinventor: How about using `np.inf` in place of -1.E50 here?

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
         return element
+
+    def build_cells(self, bc=['reflective'] * 6):
+        """Generates a list of cells which mimic the mesh geometry.
+
+        Parameters
+        ----------
+        bc : iterable of {'reflective', 'periodic', or 'vacuum'}
+            Boundary conditions for each of the six faces of a parallelopiped:
+            North, East, South, West, Up, and Down.
```
samuelshaner: There seems to be several ways to describe the six surfaces of a rectangle in OpenMC:

    x-min, x-max, y-min, y-max, z-min , z-max
    Left , Right, Back , Front, Bottom, Top
    East , West , South, North, Down  , Up        
    
I personally prefer using min and max because it leaves no ambiguity as to what dimension and direction the surface applies to while the others leave some ambiguity. For instance, the picture below indicates what I would think of the surfaces as. The front surface would be the y-min surface in cartesian coordinates, but OpenMC treats it as a y-max surface. This is more of personal preference, but my vote would be to refer to the surfaces with dimension + min/max.

cube-4

openmc/mgxs/library.py
```diff
@@ -859,65 +866,80 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
...
+                                           subdomain=subdomain_val)
         # If multiplicity matrix is available, prefer that
         if 'multiplicity matrix' in self.mgxs_types:
             mymgxs = self.get_mgxs(domain, 'multiplicity matrix')
             xsdata.set_multiplicity_mgxs(mymgxs, xs_type=xs_type,
-                                         nuclide=[nuclide])
+                                         nuclide=[nuclide],
+                                         subdomain=subdomain_val)
             using_multiplicity = True
         # multiplicity wil fall back to using scatter and nu-scatter
```
samuelshaner: wil -> will

openmc/mesh.py
wbinventor: Excuse me, `np.inf` would be a bad choice - instead, use `np.finfo(np.float).max`

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
+        for entry in bc:
+            cv.check_value('bc', entry, ['transmission', 'vacuum',
+                                         'reflective', 'periodic'])
+
+        if len(self.dimension) == 2:
+            twod = True
+        else:
+            twod = False
+
+        # Build enclosing cell
```
wbinventor: Can you elaborate on your comment about what the cell encloses?

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
+            zplanes = [openmc.ZPlane(z0=self.lower_left[2],
+                                     boundary_type=bc[5]),
+                       openmc.ZPlane(z0=self.upper_right[2],
+                                     boundary_type=bc[5])]
+        root_cell = openmc.Cell()
+        root_cell.region = ((+xplanes[0] & -xplanes[1]) &
+                            (+yplanes[0] & -yplanes[1]) &
+                            (+zplanes[0] & -zplanes[1]))
+
+        # Build our universes
```
wbinventor: Ditto - elaborate on this comment a bit to describe what each universe represents

openmc/mesh.py
```diff
@@ -210,3 +227,90 @@ def get_mesh_xml(self):
...
+        universes = np.ndarray(self.dimension[::-1], dtype=np.object)
+        cells = []
+        for [i, j, k] in self.cell_generator():
```
wbinventor: Beautiful!

openmc/mesh.py
wbinventor: On a related note, what is the purpose of this method anyway? Seems like a `Lattice` with a `distribcell` tally would accomplish the same thing with much less code, but perhaps I'm not following your logic for implementing this method.

openmc/mgxs/library.py
```diff
@@ -792,6 +792,10 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
             parameter is set to `True`.  In this case, this parameter sets the
             number of equally-spaced points in the domain of [-1,1] to be used
             in building the tabular distribution. Default is `33`.
+        subdomain : iterable of int
```
wbinventor: Add a `NOTE` to the end of this docstring mentioning that `distribcell` subdomains are not yet supported. In the future, this should be relatively easy to implement using @smharper's distributed materials feature. In any case, the `subdomain` notation now corresponds to both `mesh` and `distribcell` spatial tally domains, so it would be good to indicate what is and is not supported.

openmc/mesh.py
```diff
@@ -181,6 +183,21 @@ def __repr__(self):
...
+        """
+        if len(self.dimension) == 2:
+            for x in range(self.dimension[0]):
+                for y in range(self.dimension[1]):
+                    yield [x + 1, y + 1, 1]
+        else:
+            for x in range(self.dimension[0]):
+                for y in range(self.dimension[1]):
+                    for z in range(self.dimension[2]):
+                        yield [x + 1, y + 1, z + 1]
```
samuelshaner: Is there are reason to use the `+ 1`'s here? One place where the output of this method is used, lines 285-293 of this file, `- 1`'s are used to negate the `+ 1`'s.

openmc/mgxs/library.py
```diff
@@ -859,65 +866,80 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
             xsdata.zaid = self._nuclides[nuclide][0]
             xsdata.awr = self._nuclides[nuclide][1]
 
+        if subdomain is None:
```
wbinventor: Is there a reason to create a new `subdomain_val` variable here? Or can you simply reuse `subdomain` as so:

```python
if subdomain is None:
    subdomain = 'all'
else:
    subdomain = [subdomain]
```

openmc/mgxs/library.py
```diff
@@ -1005,32 +1018,79 @@ def create_mg_library(self, xs_type='macro', xsdata_names=None,
...
-                nuclides = list(domain.get_all_nuclides().keys())
+        if xs_ids is not None:
+            if isinstance(xs_ids, basestring):
+                # If we only have a string lets convert it now to a list
+                # of strings.
+                xs_id_input = xs_ids
+                xs_ids = []
+                for domain in self.domains:
+                    if self.domain_type is 'mesh':
+                        for subdomain in domain.cell_generator():
```
wbinventor: This seems like a very slow way to populate the `xs_ids` list - can you simply determine the number of mesh cells in `domain` and then do:

```python
xs_ids = [xs_id_input] * num_mesh_cells
```

wbinventor: This might require adding a new `num_mesh_cells` property to the `Mesh` class, which would be useful for other applications anyway.

openmc/mgxs/library.py
```diff
@@ -1005,32 +1018,79 @@ def create_mg_library(self, xs_type='macro', xsdata_names=None,
...
-            for nuclide in nuclides:
-                # Build & add metadata to XSdata object
-                if xsdata_names is None:
-                    xsdata_name = 'set' + str(i + 1)
+                cv.check_iterable_type('xs_ids', xs_ids, basestring)
+        else:
+            xs_ids = []
+            for domain in self.domains:
+                if self.domain_type is 'mesh':
+                    for subdomain in domain.cell_generator():
```
wbinventor: ditto, seems like the `xs_ids` list can be populated much faster as discussed above

openmc/mgxs/library.py
```diff
@@ -1005,32 +1018,79 @@ def create_mg_library(self, xs_type='macro', xsdata_names=None,
...
 
-                xsdata = self.get_xsdata(domain, xsdata_name, nuclide=nuclide,
-                                         xs_type=xs_type, xs_id=xs_ids[i],
-                                         tabular_legendre=tabular_legendre,
-                                         tabular_points=tabular_points)
+        if self.domain_type == 'mesh':
+            # Create the xsdata objects and add to the mgxs_file
+            i = 0
+            for domain in self.domains:
+                for subdomain in domain.cell_generator():
```
wbinventor: Ditto, I would recommend avoiding the mesh cell iteration unless explicitly needed.

wbinventor: In this case, you might actually need it, but probably can come up with a way to do the string construction with `i` much more quickly in a single line.

openmc/mgxs/library.py
```diff
@@ -1060,6 +1120,9 @@ def create_mg_mode(self, xsdata_names=None, xs_ids=None,
             parameter is set to `True`.  In this case, this parameter sets the
             number of equally-spaced points in the domain of [-1,1] to be used
             in building the tabular distribution. Default is `33`.
+        bc : iterable of {'reflective', 'periodic', or 'vacuum'}
+            Boundary conditions for each of the six faces of a parallelopiped:
```
wbinventor: "parallelopiped" -> "parallelepiped"

openmc/mgxs/library.py
```diff
@@ -1092,60 +1155,121 @@ def create_mg_mode(self, xsdata_names=None, xs_ids=None,
...
         if xs_ids is not None:
             if isinstance(xs_ids, basestring):
                 # If we only have a string lets convert it now to a list
                 # of strings.
-                xs_ids = [xs_ids for i in range(len(self.domains))]
+                xs_id_input = xs_ids
+                xs_ids = []
+                for domain in self.domains:
+                    if self.domain_type is 'mesh':
+                        for subdomain in domain.cell_generator():
```
wbinventor: ditto eliminate mesh cell iteration

wbinventor: And actually, I think you can simplify further within something like:

```python
all_xs_ids = []
for domain in self.domains:
    if self.domain_type is 'mesh':
        all_xs_ids.extend([xs_ids] * domain.num_mesh_cells)
    else:
        all_xs_ids.append(xs_ids)
```

I may be picky, but I typically don't like the redefinition of variables as you do with `xs_id_input` and `xs_ids` since it often is not the most succinct way to express an algorithm.

openmc/mgxs/library.py
```diff
@@ -1092,60 +1155,121 @@ def create_mg_mode(self, xsdata_names=None, xs_ids=None,
...
+            xs_ids = []
+            for domain in self.domains:
+                if self.domain_type is 'mesh':
+                    for subdomain in domain.cell_generator():
+                        xs_ids.append('1m')
+                else:
+                    xs_ids.append('1m')
+
+        if self.domain_type == 'mesh':
+            # Before continuing, we cant build a model automatically if
```
wbinventor: I'm not sure I understand what this comment actually means. Can you replace "So check and complain as needed" with something more informative - i.e., what goes in and what comes out?

openmc/mgxs/library.py
```diff
@@ -1092,60 +1155,121 @@ def create_mg_mode(self, xsdata_names=None, xs_ids=None,
...
+            # the user provided multiple mesh domains for library generation.
+            # So check and complain as needed.
+            cv.check_length("domains", self.domains, 1, 1)
+
+            # Create the xsdata objects and add to the mgxs_file
+            # and assign to materials and geometry as we go
+            materials = openmc.Materials()
+            root = openmc.Universe(name='root', universe_id=0)
+            # Add cells representative of the mesh with reflective BC
```
wbinventor: White space before this comment

openmc/mgxs_library.py
```diff
@@ -603,7 +603,8 @@ def nu_fission(self, nu_fission):
         if np.sum(self._nu_fission) > 0.0:
             self._fissionable = True
 
-    def set_total_mgxs(self, total, nuclide='total', xs_type='macro'):
+    def set_total_mgxs(self, total, nuclide='total', xs_type='macro',
+                       subdomain='all'):
```
wbinventor: Elsewhere this parameter is typically plural as `subdomains` since a user can specify a list of one, some or all subdomains (i.e., distribcell IDs or mesh cell indices), or the `'all'` string. In this case, it looks like you can either specify a single subdomain (mesh cell indices only) or `'all'`. Can this be expanded to be more flexible and accept one, some or all subdomains as is done elsewhere in `openmc.mgxs`?

wbinventor: And this comment applies to the other `set_*_mgxs(...)` methods below.

openmc/mesh.py
nelsonag: This will produce a lattice, with an infinite cell for every [i,j,k].  The user then has to fill in the materials for each.  Perhaps a distribcell would be better but I dont fully understand how to input/use them.

wbinventor:
Wow, another sweet PR @nelsonag! I've done an initial review, though I think it would be great if any others who have used MG mode - such as @samuelshaner - could review this as well. A few high level comments that come to mind:

* **Jupyter Notebook** - would it be worthwhile to document this in a new Jupyter Notebook example? I realize that our database of Notebooks is quickly growing, and the MGXS notebook are coming to overwhelm our other examples. But I don't see how users will be able to figure out how to use complicated features like this one without an example. What do you think?

* **Distribmats** - I may be wrong, but my intuition tells me that a coupling of distribcell tallies with distribmats may be the best long-term approach to running simulations with pin-wise or assembly-wise spatially-homogenized MGXS in MG mode. If I understand correctly, this new code creates a very large system of cells/universes to mimic a rectilinear mesh, which be very slow for ray tracing since there is not a hierarchical system of nested universes/lattices to optimize cell finding. An `MGXS` library defined for `distribcell` tally domains, transformed into a distributed materials input file, would give us the best of both worlds. Again, this is just me postulating since I'm not sure I entirely understand what your code does without seeing an example tying it all together.

openmc/mesh.py
nelsonag: I will be going with your way in an upcoming commit

openmc/mesh.py
nelsonag: Done in the code, but for sake of this comment: its the cell which will hold/contain the lattice universe itself.

openmc/mesh.py
nelsonag: Yeah, so the mesh indices technically begin at 1.  So, this is true to that.  The instance where you see me take off the 1 is because I am using the produced [i,j,k] to index a numpy array which is 0-based.

openmc/mgxs/library.py
nelsonag: Ok, so I wasnt sure about this.  But isn't there something where if you change the value of a parameter from default it will be that way the next time the code is executed, even if the default isnt left behind?  I feel like that used to be a thing

wbinventor: Nope!

openmc/mgxs/library.py
nelsonag: Agree, but the loop exists mainly for the xsdata building anyways, which needs the mesh iteration.  So I am just piggy-backing off that loop.

nelsonag:
Regarding your last points @wbinventor:

- Notebook example: That can be done, however, the create_mg_mode functionality is so specialized (i.e., running the exact same problem in CE and MG) that last time around we decided not to feature it in a notebook.  I think thats still probably the case, and therefore writing a new notebook for the mesh wouldn't be any more instructive than what we already have (the only difference would be switching the domain type and then manually making the lattice to mimic the mesh.

- Distribcells: what you are describing isn't substantially different than whats done now. I am using the lattice to mimic a mesh where I have a cell (with no bounding surfaces) for every [i,j,k] of the mesh, which are then assigned a unique universe, and those universes are assigned to the correct [i,j,k] locations of the lattice.  Can you use distribcells inside a lattice where every lattice location has a unique material?

nelsonag:
Finally, @wbinventor , had a comment about the set_*_mgxs using subdomains instead of subdomain (i.e., lists of subdomains) to be similar to other places.  The xsdata object though cant do anything with mgxs from multiple domains/subdomains, it only wants one set of data per object.  Therefore subdomains is misleading, since only one is possible.  Does that make sense or am I misunderstanding?

openmc/mesh.py
nelsonag: - [x]

openmc/mesh.py
nelsonag: - [x]

openmc/mesh.py
nelsonag: - [x] for all, though distribcell discussion will continue elsewhere (@ bottom of comments chain right now)

openmc/mesh.py
nelsonag: - [x] good catch

openmc/mesh.py
nelsonag: Ah yes, i didnt know about finfo, thanks!

- [x]

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mesh.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: Great then, thanks!
- [x]

openmc/mgxs/library.py
nelsonag: Good idea, and, to my surprise, there already was a num_mesh_cells property! Yay!
- [x]

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: sigh
- [x]

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: - [x]

nelsonag:
@wbinventor to help with understanding what this does, here is the geometry.xml file created for a 1-D (z-direction) problem as generated using Library.create_mg_mode():
```
<geometry>
    <cell fill="10100" id="10000" region="(((10000 -10001) (10002 -10003)) (10004 -10005))" universe="0" />
    <cell id="10001" material="10000" universe="10000" />
    <cell id="10002" material="10001" universe="10001" />
    <cell id="10003" material="10002" universe="10002" />
    ...
    ...
    ...
    <cell id="10100" material="10099" universe="10099" />
    <lattice id="10100">
        <pitch>10.0 10.0 0.05</pitch>
        <dimension>1 1 100</dimension>
        <lower_left>0.0 0.0 0.0</lower_left>
        <universes>
10000 
10001 
10002 
10003 
...
...
...
10099 </universes>
    </lattice>
    <surface boundary="reflective" coeffs="0.0" id="10000" type="x-plane" />
    <surface boundary="reflective" coeffs="10.0" id="10001" type="x-plane" />
    <surface boundary="reflective" coeffs="0.0" id="10002" type="y-plane" />
    <surface boundary="reflective" coeffs="10.0" id="10003" type="y-plane" />
    <surface boundary="reflective" coeffs="0.0" id="10004" type="z-plane" />
    <surface boundary="reflective" coeffs="5.0" id="10005" type="z-plane" />
</geometry>
```

nelsonag:
Anything else needed w/ this PR?

openmc/mesh.py
```diff
@@ -6,6 +6,8 @@
 import numpy as np
 
 import openmc.checkvalue as cv
+import openmc.surface
```
wbinventor: Do these need to be imported individually? You can access `Cell` and `Surface` directly from the `openmc` namespace.

openmc/mesh.py
```diff
@@ -181,6 +183,32 @@ def __repr__(self):
...
+
+        will produce the following output for a 3-D 2x2x2 mesh in mymesh:
+        [1, 1, 1]
+        [1, 1, 2]
+        [1, 2, 1]
+        [1, 2, 2]
+        ...
+
+        """
+        if len(self.dimension) == 2:
```
wbinventor: Add a blank space between the docstring and the method implementation.

openmc/mesh.py
```diff
@@ -210,3 +238,99 @@ def get_mesh_xml(self):
             subelement.text = ' '.join(map(str, self._width))
 
         return element
+
+    def build_cells(self, bc=['reflective'] * 6):
+        """Generates a lattice of universes with the same dimensionality
+        as the mesh object in self.  The individual cells/universes produced
```
wbinventor: I think this first sentence would be confusing to a Python novice - perhaps re-word as "...with the same dimensionality as the mesh object."

openmc/mgxs/library.py
```diff
@@ -1072,32 +1087,80 @@ def create_mg_library(self, xs_type='macro', xsdata_names=None,
...
+                                              "support nuclidic tallies")
+                for subdomain in domain.cell_generator():
+                    # Build & add metadata to XSdata object
+                    if xsdata_names is None:
+                        xsdata_name = 'set' + str(i + 1)
+                    else:
+                        xsdata_name = xsdata_names[i]
+
+                    # Create XSdata and Macroscopic for this domain
+                    xsdata = self.get_xsdata(domain, xsdata_name,
```
wbinventor: Are you using all of the keyword arguments supported by `get_xsdata(...)`? If so, can you remove the keywords here to compress this code  a bit?

openmc/mgxs/library.py
```diff
@@ -1072,32 +1087,80 @@ def create_mg_library(self, xs_type='macro', xsdata_names=None,
...
+                for nuclide in nuclides:
+                    # Build & add metadata to XSdata object
+                    if xsdata_names is None:
+                        xsdata_name = 'set' + str(i + 1)
+                    else:
+                        xsdata_name = xsdata_names[i]
+                    if nuclide is not 'total':
+                        xsdata_name += '_' + nuclide
+
+                    xsdata = self.get_xsdata(domain, xsdata_name,
```
wbinventor: ditto

openmc/mgxs/library.py
```diff
@@ -1157,62 +1226,77 @@ def create_mg_mode(self, xsdata_names=None, xs_ids=None,
...
-        mgxs_file = openmc.MGXSLibrary(self.energy_groups)
-
-        # Create a copy of the Geometry to differentiate for these Macroscopics
-        geometry = copy.deepcopy(self.openmc_geometry)
-        materials = openmc.Materials()
-
-        # Get all Cells from the Geometry for differentiation
-        all_cells = geometry.get_all_material_cells()
+        # If the domain type is a mesh, then there can only be one domain for
+        # this method. Thi is because we can build a model automatically if
```
wbinventor: typo: "Thi" -> "This"

openmc/mgxs_library.py
wbinventor: What is your take on this earlier question @nelsonag?

openmc/mgxs_library.py
```diff
@@ -619,6 +620,9 @@ def set_total_mgxs(self, total, nuclide='total', xs_type='macro'):
         xs_type: {'macro', 'micro'}
             Provide the macro or micro cross section in units of cm^-1 or
             barns. Defaults to 'macro'.
+        subdomain : iterable of int
+            If the MGXS contains a mesh domain type, the subdomain parameter
+            specifies which mesh cell ([i, j, k]) to use.
```
wbinventor: Should this really be a tuple of a list?

openmc/mgxs_library.py
```diff
@@ -691,6 +701,9 @@ def set_fission_mgxs(self, fission, nuclide='total', xs_type='macro'):
...
```
wbinventor: ditto

openmc/mgxs_library.py
wbinventor: Actually nvm, I see your comment on the main page for this PR. Sounds good to me.

wbinventor:
@nelsonag I'm fine with this PR save for a few final comments. I'll merge once they are addressed.

openmc/mgxs/library.py
```diff
@@ -891,6 +897,9 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
                       (type(None), bool))
         if tabular_points is not None:
             cv.check_greater_than('tabular_points', tabular_points, 1)
+        if subdomain is not None:
+            cv.check_iterable_type('subdomain', subdomain, Integral,
+                                   max_depth=2)
```
samuelshaner: Can you change this to `max_depth=3`. I have found the subdomains for a mesh require a max depth of 3.

openmc/mgxs/library.py
```diff
@@ -1127,6 +1190,12 @@ def create_mg_mode(self, xsdata_names=None, xs_ids=None,
             parameter is set to `True`.  In this case, this parameter sets the
             number of equally-spaced points in the domain of [-1,1] to be used
             in building the tabular distribution. Default is `33`.
+        bc : iterable of {'reflective', 'periodic', 'transmission', or 'vacuum'}
+            Boundary conditions for each of the four faces of a rectangle
+            (if aplying to a 2D mesh) or six faces of a parallelepiped
```
samuelshaner: aplying -> applying

openmc/mesh.py
nelsonag: - [x]

openmc/mesh.py
nelsonag: - [x]

openmc/mesh.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: I was able to remove a few for this one, but not for the next since those all have the possibility of not being default

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs_library.py
nelsonag: Oh, that is a parenthetical not a tuple, I clarified it.

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: - [x]

nelsonag:
Alright, I factored everything in so long as this test passes (which they should as my local ones did - oops did I just jinx it??) we should be good to go.

PR #703, opened 2016-08-12
Fix mesh filter max iterator check

smharper:
Found a bug from #678.  There was a check for an infinite loop that never actually incremented it's loop count.  I've fixed that and I've changed the call from `fatal_error` to `warning`.  @wbinventor found some cases on really big runs where this loop would cycle infinitely.  There is likely some finite precision error related to the particle's position behind this error.


wbinventor:
Thanks @smharper! Just tried this out and it solves my lost particle problems!

PR #712, opened 2016-08-24
Removal of redundant temperature-independent data from HDF5 Datasets

nelsonag:
This PR extends the great work by @paulromano to create an HDF5 CE Data format by reorganizing the data such that there is one HDF5 file per CE nuclide, including all temperatures available for that file (See #697).  This cuts down on disk space significantly.  This PR does not propagate the changes in data structure to the Fortran side aside from being able to read in the new format, and therefore at this stage there are no memory savings to be had.  This PR also does not do the same for MGXS mode, that PR will come after this is merged.

This re-organization was done by removing redundant angle/energy and product data from every temperature.  The only thing replicated as a function of temperature are the cross sections, consistent with the processing that NJOY does.  Finally, I'll note that some high-energy x/s (and thus the entire x/s set for inelastic levels, etc) are not touched by NJOY so there could be some more savings to be found there, at the expense of additional complexity and lack of flexibility should something upstream change).

To see memory savings, here is a list of directory sizes for each data set.  *_new indicates the data for this PR.  The savings depends on the number of temperatures present (NNDC only has one, so we have an overhead cost without any benefit) and the resolution of the cross section energy grid ( JEFF is coarser than the MCNP endf70 set).
```
822M ./nndc
835M ./nndc_new
12G	  ./jeff
1.2G  ./jeff_new
2.9G  ./endf70
1.6G  ./endf70_new
```

To accomodate the change in format, the good ol' xs identifiers (e.g., `.71c` and `.10t`) are gone.  Therefore the `materials.xml` file can have a `<default_temperature>` instead of `<default_xs>`, and each `<material>` can have a `<temperature>` instead of each nuclide/element/macroscopic having a `xs` attribute.

After this PR I will either work on moving to doing the same with MG mode or converting nuclide_header.F90 and associated code to utilize the new format to actually save memory during simulations w/ multiple temperatures (i.e., when you need it most).  So I will be answering comments with an eye for what makes sense in this PR vs in the next one or two.

P.s. right now I have Travis pulling the processed data from my dropbox account.  I'd actually prefer to get rid of that account in general, so if its possible it'd be good if another PR moved the data elsewhere after this is merged.


openmc/data/neutron.py
```diff
@@ -40,8 +98,9 @@ class IncidentNeutron(EqualityMixin):
         Metastable state of the nucleus. A value of zero indicates ground state.
     atomic_weight_ratio : float
         Atomic mass ratio of the target nuclide.
-    temperature : float
-        Temperature of the target nuclide in MeV.
+    kTs : Iterable float
```
paulromano: Iterable of float

openmc/data/neutron.py
```diff
@@ -21,6 +21,64 @@
...
+        a metastable nuclide. Because the normal ZAID (=1000*Z + A) does not
+        encode metastable information, different conventions are used among
+        different libraries. In MCNP libraries, the convention is to add 400
+        for a metastable nuclide except for Am242m, for which 95242 is
+        metastable and 95642 (or 1095242 in newer libraries) is the ground
+        state. For NNDC libraries, ZAID is given as 1000*Z + A + 100*m.
+
+    Returns
+    -------
+
```
paulromano: Need to describe return values here

docs/source/io_formats/nuclear_data.rst
```diff
@@ -16,23 +16,40 @@ Incident Neutron Data
              - **metastable** (*int*) -- Metastable state (0=ground, 1=first
                excited, etc.)
              - **atomic_weight_ratio** (*double*) -- Mass in units of neutron masses
-             - **temperature** (*double*) -- Temperature in MeV
+             - **kTs** (*double[]*) -- Temperatures (in MeV) contained in the library
```
paulromano: Doesn't look like this attribute exists?

docs/source/io_formats/nuclear_data.rst
```diff
@@ -16,23 +16,40 @@ Incident Neutron Data
...
 
 :Datasets: - **energy** (*double[]*) -- Energy points at which cross sections are tabulated
 
+**/<nuclide name>/kTs/**
+
+<TTTK> is the temperature in Kelvin, rounded to the nearest integer, of the
+temperature-dependent data set.  For example, the data set corresponding to
+300 Kelvin would be located at `300K`.
+
+:Datasets: - **<TTTK>** (*double[]*) -- kT values (in MeV) for each Temperature
```
paulromano: I would write this as `<TTT>K` to be clear that only TTT is the parameter. Also, it should be `double`, not `double[]`.

docs/source/io_formats/nuclear_data.rst
```diff
@@ -16,23 +16,40 @@ Incident Neutron Data
...
              - **label** (*char[]*) -- Name of the reaction
              - **Q_value** (*double*) -- Q value in MeV
-             - **threshold_idx** (*int*) -- Index on the energy grid that the
-               reaction threshold corresponds to
              - **center_of_mass** (*int*) -- Whether the reference frame for
                scattering is center-of-mass (1) or laboratory (0)
              - **n_product** (*int*) -- Number of reaction products
 
-:Datasets: - **xs** (*double[]*) -- Cross section values tabulated against the nuclide energy grid
+**/<nuclide name>/reactions/reaction_<mt>/<TTTK>/**
```
paulromano: Same thing here, `<TTT>K`

openmc/data/data.py
```diff
@@ -218,3 +218,7 @@ def atomic_mass(isotope):
         isotope = isotope[:isotope.find('_')]
 
     return _ATOMIC_MASS.get(isotope.lower())
+
+def kT_to_K(kT):
+    K = kT / 8.6173324e-11
```
paulromano: I'd say just make this value a module variable, `K_BOLTZMANN`, and just get rid of the function.

Also, I just noticed that the value changed in CODATA 2014... hmmmm

openmc/data/neutron.py
```diff
@@ -69,23 +130,28 @@ class IncidentNeutron(EqualityMixin):
     summed_reactions : collections.OrderedDict
         Contains summed cross sections, e.g., the total cross section. The keys
         are the MT values and the values are Reaction objects.
-    temperature : float
-        Temperature of the target nuclide in MeV.
+    temperatures : Iterable of str
+        List of string representations the temperatures of the target nuclide
+        in the data set.  The temperatures are strings with 1 decimal place,
```
paulromano: It looks like its an integer rather than 1 decimal, no?

openmc/data/neutron.py
```diff
@@ -51,8 +110,10 @@ class IncidentNeutron(EqualityMixin):
         Atomic symbol of the nuclide, e.g., 'Zr'
     atomic_weight_ratio : float
         Atomic weight ratio of the target nuclide.
-    energy : numpy.ndarray
+    energy : dict of numpy.ndarray
         The energy values (MeV) at which reaction cross-sections are tabulated.
+        They keys of the dict are the temperature string ('296.3K') for each
```
paulromano: Same comment here, isn't it just an integer?

openmc/data/neutron.py
nelsonag: - [x]

openmc/data/neutron.py
nelsonag: - [x]

openmc/data/data.py
nelsonag: Ahh shoot. Well I think we should keep the same value until maybe the new ENDF release; might as well be consistent with what the evaluators used.

openmc/data/neutron.py
nelsonag: - [x]

openmc/data/neutron.py
nelsonag: - [x]

docs/source/io_formats/nuclear_data.rst
nelsonag: - [x]

docs/source/io_formats/nuclear_data.rst
nelsonag: - [x]

docs/source/io_formats/nuclear_data.rst
nelsonag: - [x]

openmc/data/neutron.py
```diff
@@ -218,6 +276,87 @@ def urr(self, urr):
...
+                        xsdata = Sum([self.reactions[mt_i].T_data[strT].xs
+                                      for mt_i in mts])
+
+                        self.summed_reactions[mt].add_temperature(strT, 0,
+                                                                  xsdata)
+            else:
+                raise ValueError('Data provided for an incorrect nuclide')
+
+        else:
+            raise Warning('Temperature data set already within '
```
paulromano: warnings.warn?

openmc/data/neutron.py
```diff
@@ -327,19 +470,25 @@ def from_hdf5(cls, group_or_filename):
         mass_number = group.attrs['A']
         metastable = group.attrs['metastable']
         atomic_weight_ratio = group.attrs['atomic_weight_ratio']
-        temperature = group.attrs['temperature']
+        kTg = group['kTs']
+        kTs = []
+        for temp in kTg:
+            kTs.append(kTg[temp].value)
+        temperatures = [str(int(round(kT / K_BOLTZMANN))) + "K" for kT in kTs]
```
paulromano: `"{}K".format(int(round(kT / K_BOLTZMANN)))` is clearer IMO

paulromano: I would actually make `temperatures` a property of `IncidentNeutron` that is computed automatically based on `self.kTs`. That way, you don't need to define it multiple times.

openmc/data/reaction.py
```diff
@@ -281,9 +353,10 @@ class Reaction(EqualityMixin):
     threshold_idx : int
         The index on the energy grid corresponding to the threshold of this
         reaction.
-    xs : callable
+    T_data : dict of openmc.data.XS
```
paulromano: I would personally leave this named `xs`.

openmc/data/reaction.py
```diff
@@ -342,13 +406,9 @@ def products(self, products):
...
         self._products = products
 
-    @xs.setter
-    def xs(self, xs):
-        cv.check_type('reaction cross section', xs, Callable)
-        if isinstance(xs, Tabulated1D):
-            for y in xs.y:
-                cv.check_greater_than('reaction cross section', y, 0.0, True)
-        self._xs = xs
+    def add_temperature(self, temperature, threshold_idx, xs):
```
paulromano: Need a docstring on this guy

openmc/material.py
```diff
@@ -41,11 +41,19 @@ class Material(object):
...
+    temperature : str, optional
+        The temperature identifier applied to this material. The units are
+        in Kelvin and the temperature rounded to the nearest integer.
+        For example, a tempreature of 293.6K would be provided as '294K'
 
     Attributes
     ----------
     id : int
         Unique identifier for the material
+    temperature : str
```
paulromano: We actually already have temperature as a property of `Cell`, so it seems redundant to me to have temperature specified on the material. How easy would it be to revert to using the cell temperature?

scripts/openmc-ace-to-hdf5
```diff
@@ -124,47 +124,81 @@ for filename in ace_libraries:
...
+                # Add nuclide to list
+                nuclides[name] = outfile
+            else:
+                # Then we only need to append the data
+                print('Converting {} (ACE) to {} (HDF5)'.format(table.name,
+                                                                neutron.name))
+                neutron = \
+                    openmc.data.IncidentNeutron.from_hdf5(nuclides[name])
+                neutron.add_temperature_from_ace(table, args.metastable)
+                neutron.export_to_hdf5(outfile + '_1', 'w')
```
paulromano: Why is it necessary to use a separate file and then move it? I would have thought using 'w' overwrites the original file in the first place.

paulromano:
A general design comment -- the new `XS` class doesn't seem like it adds much value. It's basically just a 1D function with an extra attribute indicating the threshold index. In my mind, since `IncidentNeutron` and `Reaction` are meant to be general abstractions independent of the source data, they shouldn't publicly have metadata specific to a single format (ACE in this case). I would suggest making the following changes:

- As I've already suggested, rename `Reaction.T_data` to `Reaction.xs`
- Have the `xs` attribute be a dict of `Tabulated1D`, i.e., `Reaction.xs[temperature]` gives you a `Tabulated1D` object. This makes it extremely readable: `Reaction.xs[temperature](energy)` gives you the reaction cross section at the specified temperature and energy.
- In `Reaction.from_ace()`, store the threshold index as a private attribute, so something like:

  ```
  tabulated_xs = Tabulated1D(energy, xs)
  tabulated_xs._threshold_idx = threshold_idx
  rx.xs[strT] = tabulated_xs
  ```
- Revert to having `Reaction.to_hdf5` write the `xs` dataset and attaching `threshold_idx` as an attribute. `Reaction.from_hdf5` would also read in the attribute.

paulromano:
By the way, those reductions in the JEFF 3.2 library size are amazing!

src/input_xml.F90
```diff
@@ -5856,7 +5813,8 @@ subroutine read_ce_cross_sections(libraries, library_dict)
           ! Read nuclide data from HDF5
           file_id = file_open(libraries(i_library) % path, 'r')
           group_id = open_group(file_id, name)
-          call nuclides(i_nuclide) % from_hdf5(group_id)
+          call nuclides(i_nuclide) % from_hdf5(group_id, &
+                                               materials(i) % temperature)
```
paulromano: Because `nuclide_dict` and `already_read` now use a name without a cross section (temperature) identifier, it seems like this would only get hit for the first temperature. Moreover, when nuclides are assigned to materials (up around line 2480), it seems like a second temperature would never get assigned in the first place. Am I misunderstanding something here?

src/nuclide_header.F90
```diff
@@ -215,17 +221,53 @@ subroutine nuclide_from_hdf5(this, group_id)
...
+    kT_group = open_group(group_id, 'kTs')
+    ! Before accessing the temperature data, see if the user-provied temperature
+    ! exists.  We can find this out by looking at the datasets within kT_group
+    temp = adjustr(trim(temperature))
+    temperature_integer = str_to_int(temp(1:len(temp) - 1))
+    call get_datasets(kT_group, temperatures)
+    allocate(temperatures_integer(size(temperatures)))
+    do i = 1, size(temperatures)
+      temp = adjustr(trim(temperatures(i)))
+      temperatures_integer(i) = str_to_int(temp(1:len(temp) - 1))
```
paulromano: Why not use `adjustl` and `len_trim`?

scripts/openmc-ace-to-hdf5
nelsonag: You're thought process is correct, however I got a surprise from this:
- IncidentNeutron.from_hdf5 opens the hdf5 file and populates an IncidentNeutron object with the data.
- h5py doesnt do a copy of the data as you load it, it just gives you the reference in some cases.  So I can't close the file, or I lose the data (or I add lots of code to do a deep copy)
- So when you do neutron.export_to_hdf5, you cant open the file again (this time to write), so it will fail.  I could have just passed around the file object and originally opened it under mode 'a' but that seemed an odd thing to do from an API standpoint

src/nuclide_header.F90
nelsonag: - [x]

src/input_xml.F90
nelsonag: No you're right.  This was an oversight.  I'm trying to think of the best way to fix this.  If I can't fix it without a substantial change that wont persist after we switch to the `Nuclide` class containing all relevant temperature information, then I will just plow forward to get the T dependence within the `Nuclide` class now in this PR.

openmc/data/neutron.py
nelsonag: - [x]

openmc/data/neutron.py
nelsonag: - [x]

openmc/data/reaction.py
nelsonag: - [x]

openmc/data/reaction.py
nelsonag: - [x]

openmc/material.py
nelsonag: The more I think abou this the more OK with it I become.  On the one hand putting `temperature` in `Cell` is both more correct and more useful when using temperature feedback, but on the other, when you are not using temperature feedback or depletion the inputs become way more verbose as there are way more cells than materials in your standard BOC, zero power case.

Given we have the pyAPI now (and probably should keep a default_temperature somewhere), verbosity isnt such a big deal anymore, so I guess I agree, lets go with correctness and put temperature only in the `Cell`.

Further, thinking about what to do on the code side, you'd need to do have the following steps included in the input processing:
 1. Read geometry.xml and materials.xml
 2. For every nuclide/sab identify the temperatures you need by parsing the geometry
 3. If interpolating on temperature, for each nuclide:
 
   a. Identify the upper and lower bounds of all temperatures needed, if we are really advanced you could also identify multiple sets of boundaries (as a simple example, O-16 is in the water and the fuel. In the fuel the range of temperatures is in >1000K whereas in the water the range of temperatures is much closer to 600K.  You'd like to say 'i dont need 700K-1000K' if possible).
    b. load all data at all temperature sets within your temperature ranges needed

 4. If using WMP or other future Doppler broadening algorithm, for each nuclide:

    a. Same as 3.a, identify your ranges
    b. load only the data you need at all temperature sets within your temperature ranges needed (i.e., reaction/product/distribution data, URR data vs T, and alll sab data)
    c. load the WMP data

 5. Otherwise we are just loading the nearest data sets to the temperatures from the geometry for each nuclide

nelsonag:
Quick note: 
`openmc_ace_to_hdf5.py` doesnt work for JEFF data right now (i.e., as called from `get_jeff_data.py`.  The problem, I think, is that some of the tables themselves claim to be one nuclide, but then you read the zaid and its another nuclide and the script then gets confused.  At least that is what seems to be going on.  Diving in today.

nelsonag:
New update: 5GB for JEFF-3.2 vs 12GB before this PR. Not nearly as good, but a little better 'compression' than endf70, still likely due to endf70 being generated with a finer thinning tolerance (thats my guess at least, I havent compared NJOY inputs).

openmc/material.py
paulromano: You're right -- with the Python API, assigning temperatures to all cells that contain the same material is easy. The other thing we could do, as a user convenience, is to have a default material temperature. So, while the ultimate temperature is still by cell and is stored as such, it would inherit from the material default temperature (which itself would inherit from a global default temperature). So you'd have:
```
default temperature
  |
  |-> material default temprature
        |
        |-> cell temperature
```

I'm plowing ahead with adding temperature dependence on the `Nuclide` and `SabData` derived types and pretty much following your outline here. I already have 1, 2, and 5 pretty much taken care of. We need to decide on what the extra input for the user would look like to decide on their temperature treatment. Right now I'm thinking they could either select to have exact temperatures (within some tolerance) or to apply temperature interpolation.

paulromano: How's this for user input:
```Python
settings = openmc.Settings()
settings.temperature = {
  'default': 293.6,
  'method': 'nearest',
  'tolerance': 5.0
}
```
which would indicate that in the absence of material/cell temperatures, a temperature of 293.6 K is to be applied and cross sections within (293.6 +/- 5) K would be acceptable (the nearest one chosen). The 'method' could also be set to 'interpolation' which would indicate that interpolation is to be applied. On the XML side, this would translate to:
```XML
<temperature_default>293.6</temperature_default>
<temperature_method>nearest</temperature_method>
<temperature_tolerance>5.0</temperature_tolerance>
```
Note that this is a marked departure from what we have now where the default is effectively specified in materials.xml. I just think it makes more sense to group it with the settings for method/tolerance.

nelsonag: I think that's a great strategy.  How does this interplay with the WMP?
WMP simply over-rides this as needed?

paulromano:
Note that I submitted a PR to Adam's branch. Although I reviewed Adam's code and he reviewed mine, it would be good to get an independent third reviewer (@smharper? @wbinventor?). Below is the description of that PR:

-----------------------------------------------------

This PR is a major overhaul of `Nuclide` and `SAlphaBeta`. The `Nuclide` derived type now stores cross sections for multiple temperatures, if present, and directly maps to the HDF5 data. There are three temperature-dependent arrays: `grid`, `sum_xs`, and `urr_data`. The `Reaction` type itself also now has temperature dependent cross sections, `xs`.

The basic means for assigning temperatures is now as follows. The canonical place that temperature is stored is by cell. However, you can assign a default material temperature that is inherited by any cell/distribcell that uses that material (unless overridden). Similar, you can set a default temperature (now from `Settings` / settings.xml that is applied when no other specification exists. The global default temperature effectively replaces default_xs.

The `Settings` class now has a temperature attribute that looks something like:
```Python
settings.temperature = {
  'default': 293.6,
  'method': 'nearest',
  'tolerance': 5.0
}
```
There will be three methods (nearest, interpolation, and multipole). For nearest, the tolerance gives a range that it will accept temperatures within if the nuclear data doesn't exactly match the cell temperature. Temperature interpolation hasn't been implemented yet -- that will come in a future PR.

Multipole deserves its own discussion because a lot has changed here. The multipole library supplies its own set of pointwise cross sections, so currently in the develop branch OpenMC will strip out any existing data and use these instead. This has been changed now so that any existing data that is read from an HDF5 library just stays as is, and only the WMP parameters are read in (these alone are sufficient to calculate cross sections in the RRR). This means that in theory the URR and fast cross sections might be different. However, fast cross sections are not Doppler broadened (by NJOY) so there really isn't any meaningful difference in fast cross sections between the multipole library and our regular NNDC HDF5 data. Things will get tricky if and when we want to use multipole in the RRR and, say, temperature interpolation on URR / thermal. I'm going to leave that discussion for another day though.

A bug was fixed relating to multipole. Before when a target velocity was sampled for elastic scattering kinematics, the temperature used was whatever was assigned to the nuclide rather than the actual cell temperature. This has been fixed. I separated this out in a commit after the major refactor. You can see that in the original refactor, the results for multipole changed only in the last digit. After the bug was fixed, they changed noticeably.

A few notes on other things that changed:

- The use of zaid is gone. `Nuclide` now stores an atomic number, mass number, and metastable state, just like the `IncidentNeutron` class on the Python side. (Note that this changed the library format for S(a,b) since zaids was replaced by nuclides).
- Note that before we needed to do some trickery with mapping zaids for probability table sampling (to treat temperature correlation correctly). Now we don't need an extra dictionary for that and can use i_nuclide directly.
- A new `algorithm` module has binary search, sort, and find.
- The cross_sections.out file is gone. The `openmc.data` Python package is a much better way to find information on cross sections.
- The material union energy grid has been removed (sorry @walshjon!). Given that a material can be applied to multiple cells with different temperatures now, the abstraction just doesn't fit. The log grid is now the only method.

Overall, there is actually a reduction in LOCs on the Fortran side :tada:. The Python API grew by a few hundred LOCs, but not a huge change.

paulromano:
@nelsonag Looks like there's a merge conflict. Cam you pull develop and resolve?

nelsonag:
I can. Its only on an ipython notebook which I just need to re-run.  Re-generating the hdf5 data now and then I will be able to push the resolution.

Note that it looks like I fixed the S(a,b) table error and the He-4 (and H-3) issue about MT=27.  Again, after JEFF finished I can tell you for sure.

nelsonag:
Alright, all is set.

I've fixed the JEFF data issues, and merged in develop! This should be set for some fresh eyes to take a look!

openmc/data/neutron.py
```diff
@@ -21,6 +22,74 @@
     basestring = str
 
 
+def _get_metadata(zaid, metastable_scheme='nndc'):
+    """Method to obtain the complete element name, element, Z, mass_number,
+    and metastable state
```
smharper: The first line of a docstring should only be one line long (PEP 257).  Ideally, it should also be formatted like a sentence.  How about "Return basic identifying data for a nuclide with given ZAID."?

smharper:
I'm a little sad to see us lose the memory savings promised by multipole.  That was really pushed as one of the major advantages of multipole in the various papers.  I will agree though that simply storing a new set of pointwise cross sections in the multipole files was the wrong way to go.

I think we should remove the RR pointwise data in multipole runs.  We could try to do that as a pre-processing step in Fortran, but maybe there is a way we can do that on the HDF5 data side.  Thoughts?

smharper:
Oh, and thanks @nelsonag and @paulromano for this PR.  I'm all for getting rid of redundant data, and it'll be nice not having to fool around with those .##c suffixes.

openmc/data/neutron.py
```diff
@@ -214,10 +276,62 @@ def summed_reactions(self, summed_reactions):
...
-                      (ProbabilityTables, type(None)))
+        cv.check_type('probability table dictionary', urr, MutableMapping)
+        for key, value in urr:
+            cv.check_type('probability table temperature', key, basestring)
+            cv.check_type('probability tables', value, ProbabilityTables)
         self._urr = urr
 
+    def add_temperature_from_ace(self, ace_or_filename, metastable_scheme='nndc'):
+        """Add data to the IncidentNeutron object from an ACE file at a
+        different temperature.
```
smharper: PEP 257.  How about "Append data from an ACE file at a different temperature."?

src/nuclide_header.F90
```diff
@@ -175,24 +182,28 @@ module nuclide_header
...
     integer :: storage_type
     integer :: max_corder
     integer :: n_links
     integer :: hdf5_err
+    integer :: i_closest
+    integer :: n_temperature
     integer(HID_T) :: urr_group, nu_group
-    integer(HID_T) :: energy_dset
+    integer(HID_T) :: energy_group, energy_dset
+    integer(HID_T) :: kT_group, kT_dset
```
smharper: `kT_dset` and `n` are unused.

src/sab_header.F90
```diff
@@ -66,248 +63,253 @@ module sab_header
...
     integer :: n_energy, n_energy_out, n_mu
+    integer :: i_closest
+    integer :: n_temperature
     integer :: hdf5_err
+    integer(SIZE_T) :: name_len, name_file_len
+    integer(HID_T) :: T_group
     integer(HID_T) :: elastic_group
     integer(HID_T) :: inelastic_group
     integer(HID_T) :: dset_id
+    integer(HID_T) :: kT_group, kT_dset
```
smharper: `kT_dset` and `n` are unused

src/input_xml.F90
```diff
@@ -5764,10 +5696,10 @@ subroutine assign_sab_tables()
         ASSIGN_SAB: do k = 1, size(mat % i_sab_tables)
           ! In order to know which nuclide the S(a,b) table applies to, we need
           ! to search through the list of nuclides for one which has a matching
-          ! zaid
+          ! name
           associate (sab => sab_tables(mat % i_sab_tables(k)))
             FIND_NUCLIDE: do j = 1, size(mat % nuclide)
-              if (any(sab % zaid == nuclides(mat % nuclide(j)) % zaid)) then
+              if (any(sab % nuclides == nuclides(mat % nuclide(j)) % name)) then
```
smharper: My compiler really doesn't like this line of code.  For some reason, it goes into an infinite loop here.  Running `write(*, *) sab % nuclides` also gives me an infinite loop

smharper: Hmm, not actually an infinite loop, just very long.  For some reason, `size(sab % nuclides) = 251129904`

paulromano: Sounds like a bug, shouldn't be that bug

src/hdf5_interface.F90
```diff
@@ -2347,6 +2426,67 @@ subroutine read_attribute_string(buffer, obj_id, name)
...
+    integer(HSIZE_T) :: dims(1)
+    integer(HSIZE_T) :: maxdims(1)
+
+    call h5aopen_f(obj_id, trim(name), attr_id, hdf5_err)
+
+    if (allocated(buffer)) then
+      dims(:) = shape(buffer)
+    else
+      call h5aget_space_f(attr_id, space_id, hdf5_err)
+      call h5sget_simple_extent_dims_f(space_id, dims, maxdims, hdf5_err)
```
smharper: This is closer to the source of the error.  I'm seeing `dims = 140732188196672` coming out of this guy.  Which leads to a cool error I've never seen before:
```
Operating system error: Cannot allocate memory
Allocation would exceed memory limit
```

paulromano: What compiler / HDF5 are you using?

src/nuclide_header.F90
nelsonag: Good catch, thanks! (same for sab_header)

src/hdf5_interface.F90
nelsonag: Yeah thats odd.... any more information you can give would be appreciated.  Also what model and do you have any indication which location in the h5 file this comes from?

openmc/data/neutron.py
nelsonag: Yeah I like that. I didn't know about PEP 257, thanks for pointing that out.

nelsonag:
@smharper First off, don't feel too sad about losing some memory savings offered by WMP; this PR does save alot of memory - but it will still use too much, especially when one gets to storing temperatures every 5K, 10K, or 50K!

As for modifications to remove pointwise RR data when doing multipole, do you mean removing /nuclide/reactions/MT<i>/MT_sigma from the WMP library or removing the temperature-dependent data obtained from the ACE files but now in the h5 files?  If the latter, I say we do that on the Fortran side if running WMP as opposed to on the library side - I can imagine myself wanting to run both with and without WMP.

src/hdf5_interface.F90
```diff
@@ -2347,6 +2426,67 @@ subroutine read_attribute_string(buffer, obj_id, name)
...
+    integer(HID_T) :: memtype
+    integer(SIZE_T) :: size
+    integer(SIZE_T) :: n
+    type(c_ptr) :: f_ptr
+
+    ! Make sure buffer is large enough
+    call h5aget_type_f(attr_id, filetype, hdf5_err)
+    call h5tget_size_f(filetype, size, hdf5_err)
+    if (size > len(buffer(1)) + 1) then
+      print *, size, len(buffer(1))
```
paulromano: Oops, can you get rid of this @nelsonag?

src/hdf5_interface.F90
smharper: GCC 5.3.0 and parallel HDF5-1.8.16

The model I've been testing with is test_iso_in_lab.py.  The trouble comes from this guy:
```
HDF5 "c_H_in_H2O.h5" {
GROUP "/" {
   GROUP "c_H_in_H2O" {
      ATTRIBUTE "nuclides" {
         DATATYPE  H5T_STRING {
            STRSIZE 6;
            STRPAD H5T_STR_NULLPAD;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "['H1']"
         }
      }
```

paulromano: Ah-ha. This helped me to figure it out. So, you are most definitely using Python 2 (and I am most definitely using Python 3), and apparently when you do `np.string_(['H1'])` in Python 2 you get a scalar and in Python 3 an array. The call really should be `np.array(self.nuclides, dtype='S')`. I'll go ahead and fix this.

smharper:
I mean that in a multipole run, I don't want the unneeded resolved resonance pointwise data hanging around hogging up my gigabytes.

The way WMP used to handle that is by essentially writing most of the ACE cross section data into an H5 file and omitting all the energy grid points inside the RRR.  In the middle of #618, I tried to replace that with a system that would use the given ACE data and cut out the RRR in a preprocessing step.  You can see this with the commit "Use ACE data over WMP data where possible".  But then I ran into hella issues with extrapolating bad cross section values down to low energy neutrons (all nuclear data stops at 1e-5 eV) and some problems sampling reactions at the high energy edge of the WMP data.  (The tests passed btw; these are issues you need 10s or 100s of millions of neutrons to see.)  The debugging was difficult and my thesis was looming so I gave up and went back to the old method.

Probably the correct answer is to revisit that commit and try again to debug those issues.  I was just wondering out loud if there is some clever way we can structure our HDF5 data so we never even load the unneeded RRR pointwise for WMP simulations.

smharper:
Also, I'll still be happy to merge the PR without addressing the WMP data thing.  The pointwise data is by far the priority.  Just voicing my long-term goals.

paulromano:
@smharper I agree with you that long-term it doesn't make sense to keep the ACE data around. I think the most sensible place to "fix" the data is on the Fortran side, i.e., there's no reason to hack around with the WMP or HDF5 libraries. So, it seems like resurrecting your logic from #618 is probably the easiest way forward. My personal preference would be to keep that separate from this PR, but I'd be open to fixing it here if others feel strongly.

src/hdf5_interface.F90
smharper: Ahh there's something tricky with the Python2/Python3 thing here.  I had ruled that out by running `python3 get_nndc_data.py`.  But `get_nndc_data.py` simply calls `openmc-ace-to-hdf5` which uses just `python` aka Python2 on my system.  Glad I haven't gotten around to aliasing `python` to `python3` yet!

nelsonag:
That latest push (which was a merge of @paulromano's code) should address the comments from last night.

docs/source/usersguide/input.rst
```diff
@@ -714,6 +709,47 @@ survival biasing, otherwise known as implicit capture or absorption.
...
+.. _temperature_method:
+
+``<temperature_method>`` Element
+--------------------------------
+
+The ``<temperature_method>`` element has an accepted value of "nearest" or
+"interpolation". A value of "nearest" indicates that for each cell, the nearest
+temperature at which cross sections are given is to be applied, within a given
+tolerance (see :ref:`temperature_tolerance`). A value of "interpolation"
+indicates that cross sections are to be interpolated between temperatures at
```
smharper: I'd prefer we remove "interpolation" from the docs until we have it implemented.

openmc/settings.py
```diff
@@ -103,6 +101,14 @@ class Settings(object):
         Coordinates of the lower-left point of the Shannon entropy mesh
     entropy_upper_right : tuple or list
         Coordinates of the upper-right point of the Shannon entropy mesh
+    temperature : dict
+        Defines a default temperature and method for treating intermediate
+        temperatures at which nuclear data doesn't exist. Accepted keys are
+        'default', 'method', and 'tolerance'. The value for 'default' should be
+        a float representing the default temperature in Kelvin. The value for
+        'method' should be 'nearest' or 'interpolation'. If the method is
```
smharper: Remove "interpolation".  Add "multipole".

smharper:
@paulromano's fix works as advertised.  After those last two comments are resolved, I'm :+1: for a merge.

smharper:
Excellent!  Merging

PR #715, opened 2016-08-30
[HOTFIX] MGXS Pandas DataFrames with Nuclides

wbinventor:
This short PR fixes a bug that must have been introduced recently to the `MGXS.get_pandas_dataframe(...)` method. In particular, the column of energy groups (in/out) is incorrectly inserted into the DataFrame. Leave it up to me to find these obscure bugs in `openmc.mgxs` after a bit of late night head scratching!

PR #716, opened 2016-08-31
[HOTFIX] MDGXS Pandas DataFrames with Nuclides 

samuelshaner:
This PR is a follow up to @wbinventor's PR on fixing this issue in mgxs.py (#715). Also, the now nonexistent `MGXS.get_all_nuclides()` has been replaced with `MGXS.get_nuclides()` in mdgxs.py.


wbinventor:
Looks like the test suite doesn't like these changes.

samuelshaner:
@wbinventor I fixed the issue that was causing that one test to fail so this PR should be ready for further review.

PR #719, opened 2016-09-11
Implement HDF5-Based Multi-Group XS Library Format

### LINKED ISSUE #594: Implement HDF5-Based MGXS Library Format
As discussed in #336, it would be nice to have an HDF5-based mgxs input file format.  This can be done in one of two ways:

1. We just take the current xml format as is and port it to HDF5.
2. Gain some experience with different use cases of the current capability and see if format changes exist which would be more useful.

I'm open to either an am curious to hear other's opinions.  One facet to consider is if this should be included with 0.8 or not.  If so, that would push us to #1.

nelsonag:
This PR resolves #594 while also updating to reflect the temperature-dependent data format in use through the rest of the code as a result of #712.  I have also replaced direct hdf5 calls from nuclide_header, sab_header, reaction_header with wrapper functions in hdf5_interface.  

The python API has been updatd, and a script provided (openmc-update-mgxs.py) to convert old XML libs to the h5 format.  

@friedmud, @wbinventor, and @samuelshaner could you please review with an eye on the suitability of this format to your needs?  As we discussed in #594 it would be best if we all used the same format.  I believe I've captured all your concerns in the discussions within #594, but double check for me.  The only differences in this library from the one I was sharing there are that the temperature-dependent data was slightly re-organized to match what we do with the continuous-energy HDF5 files, and, I've got in the sparse scattering matrices @friedmud requested.  Thanks guys!


nelsonag:
Oh yeah, since the library should not contain the user option to represent the Legendre scattering data as tabular data (<tabular_legendre> in the old xml lib), and I didnt want to add another attribute to material for something I dont think any one would ever actually use (i.e., material-specific changes on this), I put it as one global setting in settings.xml

openmc/mgxs/library.py
```diff
@@ -979,48 +962,52 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
...
 
         return xsdata
 
-    def create_mg_library(self, xs_type='macro', xsdata_names=None,
-                          tabular_legendre=None, tabular_points=33):
+    def create_mg_library(self, xs_type='macro', xsdata_names=None):
         """Creates an openmc.MGXSLibrary object to contain the MGXS data for the
-        Multi-Group mode of OpenMC.
+        Multi-Group mode of OpenMC. Note that this library will not make use
+        of nested temperature tables. Every dataset in te library will be
```
samuelshaner: te -> the

openmc/mgxs/library.py
```diff
@@ -1117,41 +1090,29 @@ def create_mg_library(self, xs_type='macro', xsdata_names=None,
...
+    def create_mg_mode(self, xsdata_names=None, bc=['reflective'] * 6):
         """Creates an openmc.MGXSLibrary object to contain the MGXS data for the
         Multi-Group mode of OpenMC as well as the associated openmc.Materials
         and openmc.Geometry objects. The created Geometry is the same as that
         used to generate the MGXS data, with the only differences being
         modifications to point to newly-created Materials which point to the
         multi-group data. This method only creates a macroscopic
         MGXS Library even if nuclidic tallies are specified in the Library.
+        Note that this library will not make use of nested temperature tables.
+        Every dataset in te library will be treated as if it was at the same
```
samuelshaner: ditto

openmc/mesh.py
```diff
@@ -288,9 +288,16 @@ def build_cells(self, bc=['reflective'] * 6):
...
+            # Choosing a safe and sane default.
+            # Values of +/-1000 are used here as there seems to be an
+            # inconsistency between what numpy uses as the max float and what
+            # Fortran expects for a real(8), so this avoids code complication
+            # and achieves the same goal.
+            zplanes = [openmc.ZPlane(z0=-1000.,
                                      boundary_type='reflective'),
-                       openmc.ZPlane(z0=np.finfo(np.float).max,
+                       openmc.ZPlane(z0=1000.,
                                      boundary_type='reflective')]
```
samuelshaner: Can the `boundary_type='reflective')` bits be moved on to the previous lines to clean things up? It looks like the lines should still be under 80 characters.

openmc/mgxs/library.py
```diff
@@ -822,11 +822,12 @@ def load_from_file(filename='mgxs', directory='mgxs'):
...
     def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
-                   order=None, tabular_legendre=None, tabular_points=33,
-                   subdomain=None):
+                   order=None, subdomain=None):
         """Generates an openmc.XSdata object describing a multi-group cross section
         data set for eventual combination in to an openmc.MGXSLibrary object
-        (i.e., the library).
+        (i.e., the library).  Note that this method does not build an XSdata
+        object with nested temperature tables.  The temperature of each
+        XsData object will be left at the default value of 300K.
```
samuelshaner: To be consistent `XsData` should be `XSdata`.

openmc/mgxs_library.py
```diff
@@ -160,16 +96,16 @@ class XSdata(object):
         azimuthal angles times the number of polar angles, with the
         inner-dimension being groups, intermediate-dimension being azimuthal
         angles and outer-dimension being the polar angles.
-    scatter : numpy.ndarray
+    scatter_matrix : numpy.ndarray
```
samuelshaner: The comments here are getting quite long and repetitive. How do you feel about shortening the comments to just say what it is and then having a Note at the bottom that describes how the cross section arrays are structured? I'm working on adding the MGXS for delayed parameters that are not currently in the multi-group solver (e.g. chi_prompt, chi_delayed, prompt_nu_fission, delayed_nu_fission, and beta) and this comment section seems unnecessarily large as more cross sections are added.

openmc/mgxs_library.py
```diff
@@ -85,40 +24,37 @@ class XSdata(object):
...
-        Temperature (in units of MeV).
+    temperatures : numpy.ndarray
+        Temperatures (in units of Kelvin) of the provided datasets.  Defaults
+        to a single temperature at 294K.
     energy_groups : openmc.mgxs.EnergyGroups
         Energy group structure
     fissionable : bool
         Whether or not this is a fissionable data set.
-    scatt_type : {'legendre', 'histogram', or 'tabular'}
+    scatter_type : {'legendre', 'histogram', or 'tabular'}
```
samuelshaner: Could `scatter_type` be changed to `scatter_dist` (or some other name that is a little more prescriptive)? When I think scatter type, I think (n,n)/(n,2n)/etc or elastic/inelastic.

openmc/mgxs_library.py
```diff
@@ -218,29 +154,28 @@ class XSdata(object):
 
     """
 
-    def __init__(self, name, energy_groups, representation='isotropic'):
+    def __init__(self, name, energy_groups, temperatures=[294.],
```
samuelshaner: Any significance of 294?

openmc/mgxs_library.py
```diff
@@ -457,61 +366,173 @@ def use_chi(self, use_chi):
...
+            to 294K
+
+        See also
+        --------
+        openmc.mgxs_library.set_total_mgxs()
+
+        """
         check_type('total', total, Iterable, expected_iter_type=Real)
         # Convert to a numpy array so we can easily get the shape for
         # checking
```
samuelshaner: Can you move this (and other instances of this comment) to the previous line? I think it will still be under the 80 character limit.

openmc/mgxs_library.py
```diff
@@ -457,61 +366,173 @@ def use_chi(self, use_chi):
...
+        self._absorption[i] = npabsorption
+
+    def set_fission(self, fission, temperature=294.):
+        """This method sets the cross section for this XSdata object at the
+        provided temperature.
 
-        self._absorption = npabsorption
+        Parameters
+        ----------
+        fission: nd.nparray
```
samuelshaner: np.ndarray

openmc/mgxs_library.py
```diff
@@ -457,61 +366,173 @@ def use_chi(self, use_chi):
...
 
-    @kappa_fission.setter
-    def kappa_fission(self, kappa_fission):
+    def set_kappa_fission(self, kappa_fission, temperature=294.):
+        """This method sets the cross section for this XSdata object at the
+        provided temperature.
+
+        Parameters
+        ----------
+        kappa_fission: nd.nparray
```
samuelshaner: ditto

openmc/mgxs_library.py
```diff
@@ -970,109 +1103,189 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+                matrix = self._scatter_matrix[i]
+                flat_scatt = []
+                for g_in in range(G):
+                    for g_out in range(g_out_bounds[g_in, 0],
+                                       g_out_bounds[g_in, 1] + 1):
+                        for l in range(len(matrix[:, g_in, g_out])):
+                            flat_scatt.append(matrix[l, g_in, g_out])
+                # And write it.
+                scatt_grp = xsgrp.create_group('scatter data')
+                scatt_grp.create_dataset("scatter matrix",
```
samuelshaner: I like the readability a bit more here with the underscore before `grp`. Could you add an underscore in `xsgrp` above?

openmc/mgxs_library.py
```diff
@@ -970,109 +1103,189 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+                    scatt_grp.create_dataset("multiplicity matrix",
+                                             data=np.array(flat_mult),
+                                             compression=compression)
+                # And finally, adjust g_out_bounds for 1-based group counting
+                # and write it.
+                g_out_bounds[:, :] += 1
+                scatt_grp.create_dataset("g_min", data=g_out_bounds[:, 0],
+                                         compression=compression)
+                scatt_grp.create_dataset("g_max", data=g_out_bounds[:, 1],
+                                         compression=compression)
```
samuelshaner: This section of code is pretty compact. Can you add a blank line before the comments to aid in readability?

openmc/mgxs_library.py
```diff
@@ -970,109 +1103,189 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+
+            elif self.representation == 'angle':
+                Np = self.num_polar
+                Na = self.num_azimuthal
```
samuelshaner: Can `Np` and `Na` be made all lower case?

openmc/mgxs_library.py
```diff
@@ -970,109 +1103,189 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+                scatt_grp.create_dataset("multiplicity matrix",
+                                         data=np.array(flat_mult),
+                                         compression=compression)
+                # And finally, adjust g_out_bounds for 1-based group counting
+                # and write it.
+                g_out_bounds[:, :, :, :] += 1
+                scatt_grp.create_dataset("g_min", data=g_out_bounds[:, :, :, 0],
+                                         compression=compression)
+                scatt_grp.create_dataset("g_max", data=g_out_bounds[:, :, :, 1],
+                                         compression=compression)
```
samuelshaner: Could you add some blank lines before the comments?

openmc/mgxs_library.py
```diff
@@ -970,109 +1103,189 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
 
 class MGXSLibrary(object):
     """Multi-Group Cross Sections file used for an OpenMC simulation.
     Corresponds directly to the MG version of the cross_sections.xml input
     file.
 
+    Parameters
+    ----------
+    energy_groups : openmc.mgxs.EnergyGroups
+        Energygroup structure
```
samuelshaner: Could you add a space between energy and group here and other places in this file?

openmc/mgxs_library.py
```diff
@@ -1163,52 +1379,35 @@ def remove_xsdata(self, xsdata):
...
-    def _create_xsdata_subelements(self):
-        for xsdata in self._xsdatas:
-            xml_element = xsdata._get_xsdata_xml()
-            self._cross_sections_file.append(xml_element)
-
-    def export_to_xml(self, filename='mgxs.xml'):
-        """Create an mgxs.xml file that can be used for a
-        simulation.
+    def export_to_hdf5(self, filename='mgxs.h5', compression=None):
+        """Create an mgxs.h5 file that can be used for a simulation.
```
samuelshaner: Since filename is an input variable, perhaps just have this be `Create an hdf5 file...`

openmc/mgxs_library.py
```diff
@@ -1163,52 +1379,35 @@ def remove_xsdata(self, xsdata):
...
 
         Parameters
         ----------
-        filename : str, optional
-            filename of file, default is mgxs.xml
+        filename : str
+            Filename of file, default is mgxs.xml.
```
samuelshaner: Change to `mgxs.h5`.

openmc/mgxs_library.py
```diff
@@ -1163,52 +1379,35 @@ def remove_xsdata(self, xsdata):
...
+        compression : None or int
+            The compression level to use in the datasets within the hdf5 file.
+            A value of None implies no compression, numbers between 0 and 9
+            refer to the compression level using the gzip algorithm.
+            Defaults to None.
```
samuelshaner: Is having the compression level option necessary? I feel like this will never get used.

openmc/mgxs_library.py
```diff
@@ -1084,10 +1297,9 @@ class MGXSLibrary(object):
     """
 
     def __init__(self, energy_groups):
-        self._xsdatas = []
-        self._energy_groups = energy_groups
+        self.energy_groups = energy_groups
         self._inverse_velocities = None
```
samuelshaner: I tend to think of the inverse velocities just like any other cross section. They show up in the transport equation with space dependence and the MGXS module tallies them with space dependence. Is there a reason to they couldn't just be treated like any other xs?

openmc/settings.py
```diff
@@ -101,6 +101,13 @@ class Settings(object):
         Coordinates of the lower-left point of the Shannon entropy mesh
     entropy_upper_right : tuple or list
         Coordinates of the upper-right point of the Shannon entropy mesh
+    tabular_legendre : dict
+        Determines if a multi-group scattering moment kernel expanded via
+        Legendre polynomials is to be converted to a tabular distribution or
+        not. Accepted keys are 'enable', and 'num_points'. The value for
```
samuelshaner: no comma after enable.

openmc/settings.py
```diff
@@ -101,6 +101,13 @@ class Settings(object):
         Coordinates of the lower-left point of the Shannon entropy mesh
     entropy_upper_right : tuple or list
         Coordinates of the upper-right point of the Shannon entropy mesh
+    tabular_legendre : dict
```
samuelshaner: What do you think about just having this be a bool and then having another attribute called perhaps `num_tabular_points` that takes on the value of `num_points`?

src/global.F90
```diff
@@ -128,6 +128,12 @@ module global
   ! Maximum Data Order
   integer :: max_order
 
+  ! Whether or no to convert Legendres to tabulars
```
samuelshaner: no -> not

openmc/settings.py
samuelshaner: Maybe split this into two attributes - `legendre_to_tabular` and `legendre_to_tabular_points` - to be consistent with the variables in `global.F90`.

src/hdf5_interface.F90
```diff
@@ -213,12 +218,11 @@ end subroutine file_close
 
   subroutine get_groups(object_id, names)
     integer(HID_T), intent(in)  :: object_id
-    character(len=255), allocatable, intent(out) :: names(:)
+    character(len=150), allocatable, intent(out) :: names(:)
```
samuelshaner: Any reason why this is change?

src/hdf5_interface.F90
```diff
@@ -2486,6 +2557,28 @@ subroutine read_attribute_string_1D_explicit(attr_id, dims, buffer)
...
+    integer, target :: int_buffer
+    integer         :: hdf5_err
+    integer(HID_T)  :: attr_id
+    type(c_ptr)     :: f_ptr
+
+    call h5aopen_f(obj_id, trim(name), attr_id, hdf5_err)
+    f_ptr = c_loc(int_buffer)
+    call h5aread_f(attr_id, H5T_NATIVE_INTEGER, f_ptr, hdf5_err)
+    call h5aclose_f(attr_id, hdf5_err)
+    ! Convert to Fortran logical
```
samuelshaner: Can you add a blank line before this comment for readability?

src/input_xml.F90
```diff
@@ -4575,44 +4606,43 @@ subroutine read_ce_cross_sections_xml(libraries)
...
-    type(Node), pointer :: doc => null()
-    type(Node), pointer :: node_xsdata => null()
-    type(NodeList), pointer :: node_xsdata_list => null()
+    logical :: file_exists ! does mgxs.h5 exist?
+    integer(HID_T) :: file_id
     real(8), allocatable :: rev_energy_bins(:)
+    character(len=MAX_WORD_LEN), allocatable :: names(:)
 
-    ! Check if mgxs.xml exists
+    ! Check if mgxs.h5 exists
```
samuelshaner: Could this made generic since the file doesn't have to be named `mgxs.h5`?

src/mgxs_data.F90
```diff
@@ -20,52 +21,40 @@ module mgxs_data
...
-    character(MAX_LINE_LEN) :: temp_str
     logical :: get_kfiss, get_fiss
     integer :: l
     type(DictCharInt) :: xsdata_dict
+    type(VectorReal), allocatable :: temps(:)
 
     ! Check if cross_sections.xml exists
     inquire(FILE=path_cross_sections, EXIST=file_exists)
     if (.not. file_exists) then
       ! Could not find cross_sections.xml file
```
samuelshaner: Update this comment and the one on line 41 to reflect change from xml to hdf5

openmc/mesh.py
nelsonag: - [x]

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs_library.py
nelsonag: I agree, what I just did seems to summarize it well.

- [x]

openmc/mgxs_library.py
nelsonag: Agree, replaced with `scatter-format`

- [x]

openmc/mgxs_library.py
nelsonag: Eh, just room temperature.  Its the same default as in CE mode so I kept that here.

- [x]

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: hahahaha sorry about that (i dont know why i find what I did funny, but I do)

- [x]

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: Ahh this one I have to keep as Np (or something equally short, which I might as well just keep as is) to avoid collisions with numpy (as np).

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: You're probably right.  I put this in before there ever were sparse scattering matrices.  Now that they are in, this probably isnt very useful.  I've removed it.

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: You're correct.  I don't know why I got it in my head that this was any different.

openmc/settings.py
nelsonag: - [x]

openmc/settings.py
nelsonag: I kind of like it the way it is since they are directly linked together in this format here.

src/hdf5_interface.F90
nelsonag: Oh to be consistent with some code which uses this and to stop getting the warnings about truncation.  I figured even 150 was overkill, so I just matched it.

src/hdf5_interface.F90
nelsonag: - [x]

src/input_xml.F90
nelsonag: - [x]

src/mgxs_data.F90
nelsonag: - [x]

openmc/mgxs_library.py
samuelshaner: Thats right - I'm fine with it as is then.

- [ x]

openmc/settings.py
samuelshaner: Okay.

- [x]

src/hdf5_interface.F90
samuelshaner: Okay 

- [x]

samuelshaner:
Is there are reason why the temperature is in MeV instead of K?

nelsonag:
@samuelshaner good question, where did you see that?

samuelshaner:
I see it in `mgxs_header.F90` where you use `kT` for the temperature in a lot of places. Since the continuous energy cross sections are set in Kelvin, it would just seem to make more sense to have temperatures in K.

samuelshaner:
The description for the `mgxs_from_hdf5` subroutine in `mgxs_header.F90` needs to be updated.

samuelshaner:
Do you know if chi is properly normalized when it is interpolated or should that be a concern?

nelsonag:
> Do you know if chi is properly normalized when it is interpolated or should that be a concern?

Chi is normalized already in both *_from_hdf5 and *_combine.  The one in *_combine (after the interpolation) should be combining two already normalized chi values and therefore any deviance from a normalized result would just be floating error anyway.  Either way, I still normalize it again just in case.

nelsonag:
@samuelshaner:
> I see it in mgxs_header.F90 where you use kT for the temperature in a lot of places. Since the continuous energy cross sections are set in Kelvin, it would just seem to make more sense to have temperatures in K.

Oh, got it.  Yeah, in mgxs_from_hdf5 I use the temperatures in Kelvin to find the dataset from the library (as thats how they are organized), but I dont drag around those temperatures in Kelvin in the actual MGXS object, only in kT.  So, when we get to mgxs*_combine I only have the temperatures in kT, so I just go ahead and use those instead of converting (since Im going to use those anyways for the actual interpolation).  

Your question did cause me to find 2 bugs: I wasnt converting tolerance in mgxs*_combine to kT, and I wasnt converting a kT to a K when printing an error.

nelsonag:
@friedmud and @wbinventor any perspective on the MGXS format itself and suitability for your uses?

nelsonag:
Is there any action remaining on this PR?

paulromano:
@nelsonag Let me take a crack at reviewing tomorrow. I was going to defer to others, but thus far only @samuelshaner has chimed in.

wbinventor:
I apologize for not helping out with the review (#thesiswoes) -- I'll make some time this afternoon to make a high-level pass through this PR.

friedmud:
I've been completely buried in my own stuff - sorry for not chiming in here.  Let me take a look...

nelsonag:
Thanks guys! I'm just afraid I'll be getting busy pretty quick here in the
coming weeks and don't want to let this knowledge decay!

friedmud:
It looks like you used the same idea as YakXS for the scattering matrix: storing the `min` and `max` for each group to effectively "compress" out a lot of the zeros.  That's good, and should cut down on the amount of storage for scattering matrices by quite a lot.

My only issue: it looks like the scattering matrix is the _transpose_ of what a computational code really wants.  We ( @wbinventor and I ) have talked about this before... and I believe it's much better to have the scattering matrix set up so that it is "in on columns" and "out on rows".

This is because it makes the application of the scattering matrix into a perfect matrix-vector product... which maximizes the opportunity for vectorization on CPUs.

YakXS uses that storage (in on columns, out on rows) as does my own format for my code.  My code is set up to take advantage of this by doing vectorized mat-vecs for the scattering source computation.

Of course... I can always read it in and transpose it... but that gets to be egregious when you're dealing with many temperatures and many (hundreds/thousands) of groups.  With a large enough XS library you might not want to have the whole thing in memory all the time... and just opportunistically read the pieces you need right now... in which case you would have to do the transpose on the fly as well... which would get expensive.

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+             - **group structure** (*double[]*) -- Monotonically increasing
+               list of group boundaries, in units of MeV.  The length of this
+               array should be the number of groups plus 1.
+
+**/<library name>/**
+
+The data within <library name> contains the temperature-dependent multi-group
+data for the nuclide or material that it represents.
+
+:Attributes: - **awr** (*double*) -- The atomic weight ratio (optional, i.e. it
```
nelsonag: This one is, sadly.  It is needed for converting atom-percent materials input in to the correct number densities.  Your question, however, showed that I wasn't doing that correctly and that should be fixed now.

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+             - **fissionable** (*int*) -- Whether the dataset is fissionable
+               (1) or not (0).
+             - **representation** (*char[]*) -- The method used to generate and
+               represent the multi-group cross sections.  That is, whether they
+               were generated with scalar flux weighting (or reduced to a
+               similar representation) and thus are angle-independent, or if the
+               data was generated with angular dependent fluxes and thus the
+               data is angle-dependent.  Valid values are either "isotropic" or
+               "angle".
+             - **num-azimuthal** (*int*) -- Number of equal width angular bins
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+               that the azimuthal angular domain is subdivided if the
+               `representation` attribute is "angle". This parameter is
+               ignored otherwise.
+             - **num-polar** (*int*) -- Number of equal width angular bins
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+               that the polar angular domain is subdivided if the
+               `representation` attribute is "angle". This parameter is
+               ignored otherwise.
+             - **scatter-format** (*char[]*) -- The representation of the
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+               default of "legendre" will be assumed.
+             - **order** (*int*) -- Either the Legendre order, number of bins,
+               or number of points (depending on the value of `scatter-format`)
+               used to describe the angular distribution associated with each
+               group-to-group transfer probability.
+
+**/<library name>/kTs/**
+
+:Datasets: - **<TTT>K** (*double*) -- kT values (in MeV) for each Temperature
+             TTT (in Kelvin), rounded to the nearest integer
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+             is fissionable and fission-tallies are expected to be used.
+           - **nu-fission** (*double[]* to *double[][][][]*) -- Nu-Fission
+             cross section.
+             If **chi** is provided, then `nu-fission` has the same
+             dimensionality as `fission`.  If **chi** is not provided, then
+             the `nu-fission` data must represent the fission neutron energy
+             spectra as well and thus will have one additional dimension
+             for the outgoing energy group.  In this case, `nu-fission` has the
+             same dimensionality as `multiplicity matrix`.
+           - **inverse velocities** (*double[]*) -- Average inverse velocity
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+:Datasets: - **g_min** (*int[]* or *int[][][]) --
+             Minimum (most energetic) outgoing groups with non-zero values of
+             the scattering matrix. These group numbers use the standard
+             ordering where the fastest neutron energy group is group 1 while
+             the slowest neutron energy group is group G.
+             The dimensionality of `g_out bounds` is:
+             `g_min[g_in]`, or `g_min[num-polar][num-azimuthal][g_in]`.
+             The former is used when `representation` is "isotropic", and the
+             latter when `representation` is "angle".
+           - **g_max** (*int[]* or *int[][][]) --
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+               array should be the number of groups plus 1.
+
+**/<library name>/**
+
+The data within <library name> contains the temperature-dependent multi-group
+data for the nuclide or material that it represents.
+
+:Attributes: - **awr** (*double*) -- The atomic weight ratio (optional, i.e. it
+               is not meaningful for material-wise data)
+             - **fissionable** (*int*) -- Whether the dataset is fissionable
```
wbinventor: Can't the code easily determine whether a material is fissionable by inspection of the fission MGXS?

wbinventor: I recognize that this isn't something new you introduce here (I don't think so anyway), but is just me thinking out loud

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+**/<library name>/kTs/**
+
+:Datasets: - **<TTT>K** (*double*) -- kT values (in MeV) for each Temperature
+             TTT (in Kelvin), rounded to the nearest integer
+
+**/<library name>/<TTT>K/**
+
+Temperature-dependent data, provided for temperature <TTT>K.
+
+:Datasets: - **total** (*double[]* or *double[][][]*) -- Total cross section.
```
wbinventor: Whew, temperature- AND angle-dependent MGXS...

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+           - **g_max** (*int[]* or *int[][][]) --
+             Maximum (least energetic) outgoing groups with non-zero values of
+             the scattering matrix. These group numbers use the standard
+             ordering where the fastest neutron energy group is group 1 while
+             the slowest neutron energy group is group G.
+             The dimensionality of `g_out bounds` is:
+             `g_max[g_in]`, or `g_max[num-polar][num-azimuthal][g_in]`.
+             The former is used when `representation` is "isotropic", and the
+             latter when `representation` is "angle".
+           - **scatter matrix** (*double[]*) -- Flattened representation of the
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+           - **absorption** (*double[]* or *double[][][]*) -- Absorption
+             cross section.
+             This is a 1-D vector if `representation` is "isotropic", or a 3-D
+             vector if `representation` is "angle" with dimensions of
+             [groups, azimuthal, polar].
+           - **fission** (*double[]* or *double[][][]*) -- Fission
+             cross section.
+             This is a 1-D vector if `representation` is "isotropic", or a 3-D
+             vector if `representation` is "angle" with dimensions of
+             [groups, azimuthal, polar].  This is only required if the data set
```
wbinventor: This is just me being OCD, but might not "dataset" be more appropriate than "data set"? This would better correspond to the spelling used by HDF5, as well as the broader "data science" community.

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+             latter when `representation` is "angle".  Note that if the value of
+             `scatter-format` is "legendre", the order dimension will be one
+             larger than the value of `order`, otherwise it will match `order`.
+             Finally, the g_out dimension has a dimensionality of
+             `g_out bounds`[0] to `g_out bounds`[1].
+           - **multiplicity matrix** (*double[]*) -- Flattened representation of
+             the scattering moment matrices. This dataset provides the code with
+             a scaling factor to account for neutrons being produced in (n,xn)
+             reactions. This is assumed isotropic and therefore is not repeated
+             for every Legendre moment or histogram/tabular bin. This dataset is
```
wbinventor: And in fact you use "dataset" here (correctly :wink:)

docs/source/pythonapi/examples/mgxs-part-iv.ipynb
```diff
@@ -1433,7 +1444,7 @@
     "collapsed": true
    },
    "source": [
-    "We also see very good agreement between the fission rate distributions, though these should converge closer together with an increasing number of particle histories in both the continuous-energy run to generate the multi-group cross sections, and in the multi-group calculation itself."
+    "We also see good agreement between the fission rate distributions, though these should converge closer together with an increasing number of particle histories in both the continuous-energy run to generate the multi-group cross sections, and in the multi-group calculation itself."
```
wbinventor: hehe

examples/python/pincell_multigroup/build-xml.py
```diff
@@ -21,44 +21,47 @@
...
-                       [0.0000000, 0.0000000, 0.0000000, 0.0000000, 0.0000000, 0.0085458, 0.2730800]]]
-uo2_xsdata.fission = [7.21206E-03, 8.19301E-04, 6.45320E-03,
-                      1.85648E-02, 1.78084E-02, 8.30348E-02,
-                      2.16004E-01]
-uo2_xsdata.nu_fission = [2.005998E-02, 2.027303E-03, 1.570599E-02,
-                         4.518301E-02, 4.334208E-02, 2.020901E-01,
-                         5.257105E-01]
-uo2_xsdata.chi = [5.8791E-01, 4.1176E-01, 3.3906E-04, 1.1761E-07,
-                  0.0000E+00, 0.0000E+00, 0.0000E+00]
+uo2_xsdata.set_total(
```
wbinventor: I liked the old property setters, but I suppose I'll find out why you now use `set_*` methods below

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+             follows (in row-major format):
+             `scatter matrix[order(+1)][g_in][g_out]`, or
+             `scatter matrix[num-polar][num-azimuthal][order(+1)][g_in][g_out]`
+             The former is used when `representation` is "isotropic", and the
+             latter when `representation` is "angle".  Note that if the value of
+             `scatter-format` is "legendre", the order dimension will be one
+             larger than the value of `order`, otherwise it will match `order`.
+             Finally, the g_out dimension has a dimensionality of
+             `g_out bounds`[0] to `g_out bounds`[1].
+           - **multiplicity matrix** (*double[]*) -- Flattened representation of
```
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
```diff
@@ -8,297 +8,155 @@ OpenMC can be run in continuous-energy mode or multi-group mode, provided the
...
+             for every Legendre moment or histogram/tabular bin. This dataset is
+             optional, if it is not provided no multiplication (i.e., values of
+             1.0) will be assumed.
+             The pre-flattened array is shaped as follows (in row-major format):
+             `multiplicity matrix[g_in][g_out]`, or
+             `multiplicity matrix[num-polar][num-azimuthal][g_in][g_out]`
+             The former is used when `representation` is "isotropic", and the
+             latter when `representation` is "angle". Finally, the g_out
+             dimension has a dimensionality of `g_out bounds`[0] to
+             `g_out bounds`[1].
```
nelsonag: - [x]

openmc/mesh.py
```diff
@@ -286,21 +286,24 @@ def build_cells(self, bc=['reflective'] * 6):
...
-                                     boundary_type='reflective')]
+            # Would prefer to have the z ranges be the max supported float, but
+            # these values are apparently different between python and Fortran.
+            # Choosing a safe and sane default.
+            # Values of +/-1000 are used here as there seems to be an
+            # inconsistency between what numpy uses as the max float and what
+            # Fortran expects for a real(8), so this avoids code complication
+            # and achieves the same goal.
+            zplanes = [openmc.ZPlane(z0=-1000., boundary_type='reflective'),
+                       openmc.ZPlane(z0=1000., boundary_type='reflective')]
```
nelsonag: - [x]

examples/xml/pincell_multigroup/mg_cross_sections.xml
```diff
@@ -1,383 +0,0 @@
...
-      0.00000000E+00  0.00000000E+00  0.00000000E+00  6.55562000E-05  2.07838000E-01  3.81486000E-03  3.69760000E-09
-      0.00000000E+00  0.00000000E+00  0.00000000E+00  0.00000000E+00  1.02427000E-03  2.02465000E-01  4.75290000E-03
-      0.00000000E+00  0.00000000E+00  0.00000000E+00  0.00000000E+00  0.00000000E+00  3.53043000E-03  6.58597000E-01
-    </scatter>
-
-    <total>
-      2.16767595E-01 4.80097720E-01 8.86369232E-01 9.70009150E-01 9.10481420E-01 1.13775017E+00 1.84048743E+00
-    </total>
-  </xsdata>
-</library>
```
wbinventor: Does the "mgxs.h5" file need to be shipped with the code? It looks like it is created each time the code is run, and hence does not need to be version controlled (nor should it be)

openmc/mgxs_library.py
```diff
@@ -85,40 +24,37 @@ class XSdata(object):
...
-    name : str, optional
+    name : str
         Name of the mgxs data set.
     energy_groups : openmc.mgxs.EnergyGroups
-        Energygroup structure
+        Energy group structure
     representation : {'isotropic', 'angle'}, optional
         Method used in generating the MGXS (isotropic or angle-dependent flux
         weighting). Defaults to 'isotropic'
+    temperatures : numpy.ndarray
```

openmc/mgxs/library.py
```diff
@@ -822,11 +822,12 @@ def load_from_file(filename='mgxs', directory='mgxs'):
         return pickle.load(open(full_filename, 'rb'))
 
     def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
-                   order=None, tabular_legendre=None, tabular_points=33,
-                   subdomain=None):
+                   order=None, subdomain=None):
         """Generates an openmc.XSdata object describing a multi-group cross section
```
wbinventor: Shorten the initial docstring description to no more than 2 lines so that it can be rendered inline with use of "CTRL+Tab" in IPython Notebook or other rich IDEs. A more detailed docstring can then be given in a separate paragraph below. This is how all of our other docstrings are formatted at least, for the above reason.

openmc/mgxs_library.py
```diff
@@ -142,105 +78,79 @@ class XSdata(object):
...
-        providing the matrix, it should be ordered the same as the
-        ``multiplicity`` matrix.
+        provided), or is the group-wise fission production matrix.
+    inverse_velocities : dict of numpy.ndarray
+        Inverse of velocities, in units of sec/cm.
+
+    Notes
+    -----
+    The parameters containing cross section data have dimensionalities which
+    depend upon the value of ``representation`` as well as the number of
```

nelsonag: Cool, I didnt know that (obviously)

openmc/mgxs_library.py
```diff
@@ -901,24 +1017,27 @@ def set_scatter_mgxs(self, scatter, nuclide='total', xs_type='macro',
...
 
         elif self.representation is 'angle':
             msg = 'Angular-Dependent MGXS have not yet been implemented'
             raise ValueError(msg)
 
-    def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
-                              xs_type='macro', subdomain=None):
+    def set_multiplicity_matrix_mgxs(self, nuscatter, scatter=None,
+                                     temperature=294., nuclide='total',
+                                     xs_type='macro', subdomain=None):
```
nelsonag: Yeah, I would much prefer that sort of strategy, but sadly I cant :-(

openmc/mgxs/library.py
```diff
@@ -979,48 +962,52 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
...
+                xsdata.set_scatter_matrix_mgxs(nuscatt_mgxs, xs_type=xs_type,
+                                               nuclide=[nuclide],
+                                               subdomain=subdomain)
 
                 # Since we are not using multiplicity, then
                 # scattering multiplication (nu-scatter) must be
                 # accounted for approximately by using an adjusted
                 # absorption cross section.
                 if 'total' in self.mgxs_types:
-                    xsdata.absorption = \
```
wbinventor: what if a transport-corrected total is used?

openmc/mgxs_library.py

openmc/mgxs_library.py
```diff
@@ -457,102 +340,262 @@ def use_chi(self, use_chi):
...
+        self._absorption.append(None)
+        self._scatter_matrix.append(None)
+        self._multiplicity_matrix.append(None)
+        self._fission.append(None)
+        self._nu_fission.append(None)
+        self._kappa_fission.append(None)
+        self._chi.append(None)
+        self._inverse_velocities.append(None)
+
+    def set_total(self, total, temperature=294.):
```
nelsonag: (OBE)

openmc/mgxs/library.py
```diff
@@ -979,48 +962,52 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
...
+                                xsdata._scatter_matrix[i][0, :, :], axis=1))
 
         return xsdata
 
-    def create_mg_library(self, xs_type='macro', xsdata_names=None,
-                          tabular_legendre=None, tabular_points=33):
+    def create_mg_library(self, xs_type='macro', xsdata_names=None):
         """Creates an openmc.MGXSLibrary object to contain the MGXS data for the
-        Multi-Group mode of OpenMC.
+        Multi-Group mode of OpenMC. Note that this library will not make use
```
wbinventor: ditto, split docstring

openmc/mgxs_library.py
```diff
@@ -380,34 +286,34 @@ def representation(self, representation):
...
+        self._fissionable = fissionable
+
+    @temperatures.setter
+    def temperatures(self, temperatures):
+        check_type('temperatures', temperatures, Iterable,
+                   expected_iter_type=Real)
+        # Convert to a numpy array so we can easily get the shape for checking
+        nptemperatures = np.asarray(temperatures)
+
+        check_value('temperatures dimensionality', nptemperatures.ndim, [1])
```

openmc/mgxs_library.py
```diff
@@ -380,34 +286,34 @@ def representation(self, representation):
...
         self._awr = awr
 
-    @kT.setter
-    def kT(self, kT):
-        # Check validity of type and that the kT value is >= 0
-        check_type('kT', kT, Real)
-        check_greater_than('kT', kT, 0.0, equality=True)
-        self._kT = kT
+    @fissionable.setter
+    def fissionable(self, fissionable):
```
wbinventor: ditto per my earlier comment - cannot this be determined under-the-hood from the input data?

openmc/mgxs_library.py
```diff
@@ -380,34 +286,34 @@ def representation(self, representation):
...
+        check_type('fissionable', fissionable, bool)
+        self._fissionable = fissionable
+
+    @temperatures.setter
+    def temperatures(self, temperatures):
+        check_type('temperatures', temperatures, Iterable,
+                   expected_iter_type=Real)
+        # Convert to a numpy array so we can easily get the shape for checking
+        nptemperatures = np.asarray(temperatures)
```
wbinventor: nbd, but why not simply re-define `temperatures` instead of defining the new variable `nptemperatures`?

openmc/mgxs_library.py
wbinventor: I agree. Though I do recall a day nearly two years ago where I tried to persuade @paulromano to let me keep my beloved `set_*` methods in my initial PR of the Python API. But he convinced me otherwise to use properties everywhere, and my view of the world has changed :-)

openmc/mgxs_library.py
```diff
@@ -970,144 +1095,202 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
-
-            if self._chi is not None:
-                subelement = ET.SubElement(element, 'chi')
-                subelement.text = ndarray_to_string(self._chi)
-
-        return element
+        self._multiplicity_matrix[i] = \
+            np.nan_to_num(self._multiplicity_matrix[i])
+
+    def _get_xsdata_group(self, file):
```

openmc/mgxs_library.py
```diff
@@ -457,102 +340,262 @@ def use_chi(self, use_chi):
...
+        openmc.mgxs_library.set_fission_mgxs()
 
-    @fission.setter
-    def fission(self, fission):
+        """
         check_type('fission', fission, Iterable, expected_iter_type=Real)
-        # Convert to a numpy array so we can easily get the shape for
-        # checking
+        # Convert to a numpy array so we can easily get the shape for checking
         npfission = np.asarray(fission)
```
wbinventor: ditto, I prefer to see `np` prefixes removed everywhere for simplicity

openmc/mgxs_library.py
```diff
@@ -457,102 +340,262 @@ def use_chi(self, use_chi):
...
+        """This method sets the cross section for this XSdata object at the
+        provided temperature.
+
+        Parameters
+        ----------
+        fission: np.ndarray
+            Fission Cross Section
+        temperature : float
+            Temperature (in units of Kelvin) of the provided dataset. Defaults
+            to 294K
```
wbinventor: Perhaps clarify explicitly that "Defaults to room temperature (294K)." and end with a period, here and everywhere else.

openmc/mgxs_library.py
```diff
@@ -457,102 +340,262 @@ def use_chi(self, use_chi):
...
+        # Convert to a numpy array so we can easily get the shape for checking
         npmultiplicity = np.asarray(multiplicity)
         check_iterable_type('multiplicity', npmultiplicity, Real,
                             max_depth=len(npmultiplicity.shape))
         check_value('multiplicity shape', npmultiplicity.shape,
                     [self.matrix_shape])
+        check_type('temperature', temperature, Real)
+        check_value('temperature', temperature, self.temperatures)
+
+        i = self.temperatures.tolist().index(temperature)
```
wbinventor: alternatively:

```
i = np.where(self.temperatures == temperature)
```

It certainly won't make a difference on performance (who cares about that for this anyway) since the number of temperatures will always be small, but just throwing it out there since we use that approach to compute array indices throughout the rest of the API.

openmc/mgxs_library.py
```diff
@@ -582,14 +626,43 @@ def nu_fission(self, nu_fission):
...
+            if npnu_fission.shape == self.vector_shape:
+                self.use_chi = True
+            else:
+                self.use_chi = False
 
-        self._nu_fission = npnu_fission
-        if np.sum(self._nu_fission) > 0.0:
+        i = self.temperatures.tolist().index(temperature)
+        self._nu_fission[i] = npnu_fission
+        if np.sum(npnu_fission) > 0.0:
```
wbinventor: ditto, since the `fissionable` attribute is set here, why give the user an option to set the attribute themselves?

openmc/mgxs_library.py
```diff
@@ -970,144 +1095,202 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+                # Now create the flattened scatter matrix array
+                flat_scatt = []
+                for p in range(Np):
+                    for a in range(Na):
+                        matrix = self._scatter_matrix[i][p, a, :, :, :]
+                        for g_in in range(G):
+                            for g_out in range(g_out_bounds[p, a, g_in, 0],
+                                               g_out_bounds[p, a, g_in, 1] + 1):
+                                for l in range(len(matrix[:, g_in, g_out])):
+                                    flat_scatt.append(matrix[l, g_in, g_out])
```
nelsonag: That won't allow me to get rid of the zeros between our g_min and g_max from the data, so I chose against it.

.gitignore
```diff
@@ -50,6 +50,8 @@ tests/ctestscript.run
 
 # HDF5 files
 *.h5
+!tests/1d_mgxs.h5
+!examples/xml/pincell_multigroup/mgxs.h5
```
nelsonag: - [x]

.gitignore
```diff
@@ -89,3 +91,7 @@ docs/source/pythonapi/examples/mgxs
 docs/source/pythonapi/examples/tracks
 docs/source/pythonapi/examples/fission-rates
 docs/source/pythonapi/examples/plots
+
+# VSCode workspace directory
+.vscode
+.vscode/*
```
nelsonag: Meh, I gave up on that software anyways.

- [x]

openmc/mgxs_library.py
```diff
@@ -621,16 +697,19 @@ def set_total_mgxs(self, total, nuclide='total', xs_type='macro',
...
-                                       subdomains=subdomain)
+            self._total[i] = total.get_xs(nuclides=nuclide, xs_type=xs_type,
+                                          subdomains=subdomain)
         elif self.representation is 'angle':
             msg = 'Angular-Dependent MGXS have not yet been implemented'
             raise ValueError(msg)
 
-    def set_absorption_mgxs(self, absorption, nuclide='total',
-                            xs_type='macro', subdomain=None):
+    def set_absorption_mgxs(self, absorption, temperature=294.,
```
wbinventor: Is it just me, or is 293K more commonly used to refer to room temperature?

openmc/mgxs_library.py
```diff
@@ -884,8 +998,10 @@ def set_scatter_mgxs(self, scatter, nuclide='total', xs_type='macro',
                     [self.energy_groups])
         check_value('domain_type', scatter.domain_type,
                     ['universe', 'cell', 'material', 'mesh'])
+        check_type('temperature', temperature, Real)
+        check_value('temperature', temperature, self.temperatures)
 
-        if self.scatt_type != 'legendre':
+        if (self.scatter_format != 'legendre'):
```
wbinventor: why are enclosing parentheses needed here?

scripts/openmc-update-mgxs.py
```diff
@@ -0,0 +1,232 @@
...
+                        help='input XML file')
+    parser.add_argument('-o', '--output', nargs='?', default='',
+                        help='output file, in HDF5 format')
+    parser.add_argument('-c', '--compression', type=int,
+                        help='HDF5 Compression Level')
+    args = vars(parser.parse_args())
+
+    if args['output'] == '':
+        filename = args['input'].name
+        extension = filename[filename.rfind('.'):]
```
nelsonag: - [x]

openmc/mgxs_library.py
wbinventor: Also it seems like you use a mixture of `==` vs. `is` and `!=` and vs. `is not` for string comparisons throughout the code. Any reason not to be consistent?

scripts/openmc-update-mgxs.py
```diff
@@ -0,0 +1,232 @@
...
+            order_dim = order + 1
+        else:
+            order_dim = order
+
+        if i != -1:
+            xsd[i].add_temperature(temperature)
+
+        temp = get_data(xsdata_elem, 'total')
+        if temp is not None:
+            temp = np.array(temp.split())
```
nelsonag: - [x]

openmc/mgxs_library.py
wbinventor: clever numpy tricks :+1:

scripts/openmc-update-mgxs.py
```diff
@@ -0,0 +1,232 @@
...
+            total = temp.astype(np.float)
+            total = np.reshape(total, xsd[i].vector_shape)
```
nelsonag: - [x]  (Here and later)

scripts/openmc-update-mgxs.py
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env python
```
wbinventor: Just curious, but is this new script necessary? How many people will need to use it?

src/hdf5_interface.F90
```diff
@@ -246,16 +250,62 @@ subroutine get_groups(object_id, names)
...
+
+  function check_dataset(object_id, name) result(exists)
+    integer(HID_T), intent(in) :: object_id
+    character(*),   intent(in) :: name ! name of group
+    logical :: exists
+
+    ! Wrap check_group since the method used there will work here too
+    exists = check_group(object_id, name)
+
+  end function check_dataset
```
nelsonag: - [x]

src/input_xml.F90
```diff
@@ -4625,45 +4655,28 @@ subroutine read_mg_cross_sections_xml(libraries)
...
-
       ! Get name of material
       allocate(libraries(i) % materials(1))
-      call get_node_value(node_xsdata, "name", libraries(i) % materials(1))
+      libraries(i) % materials(1) = names(i)
     end do
 
-    ! Close cross sections XML file
-    call close_xmldoc(doc)
+    ! Close MGXS HDF file
```
wbinventor: typo: "HDF" -> "HDF5"

src/tracking.F90
```diff
@@ -92,6 +92,9 @@ subroutine transport(p)
         ! Since the MGXS can be angle dependent, this needs to be done
         ! After every collision for the MGXS mode
         if (p % material /= MATERIAL_VOID) then
+          ! Update the temperature index
+          call macro_xs(p % material) % obj % find_temperature(p % sqrtkT)
+          ! Get the data
           call macro_xs(p % material) % obj % calculate_xs(p % g, &
                p % coord(p % n_coord) % uvw, material_xs)
         else
```
wbinventor: ditto, any reason to version control the "mgxs.h5" file for this test?

src/mgxs_data.F90
```diff
@@ -172,28 +167,120 @@ end subroutine read_mgxs
...
+            ! Add temperature if it hasn't already been added
+            if (find(temps(i_nuclide), temperature) == -1) then
+              call temps(i_nuclide) % push_back(temperature)
+            end if
+          end do NUC_NAMES_LOOP
+        end associate
+      end do
+    end do
+
+  end subroutine get_temperatures
```
nelsonag: Yep, I made the sab argument in input_xml.F90 optional and then moved this in to geometry_header.F90.  Based on the dependencies it was either that or material_header.F90.

wbinventor:
I generally agree with @friedmud's comment about the transpose of the scattering matrix...but for a different reason. While I agree with the need to vectorize the mat-vec for deterministic codes, I don't see the same relevance for Monte Carlo (as implemented in OpenMC anyway). There shouldn't be any reason to use in-out vs. out-in for performance here.

Nevertheless, it may be a good idea to use the transpose for consistency/continuity of the user interface iff the YakXS format is something that the general reactor physics community will adopt (will it?). Given the relatively tight coupling of the MOOSE and OpenMC developer communities, I tend to think it is a good idea to provide a uniform format and go with the YakXS standard to avoid confusion.

src/mgxs_data.F90
nelsonag: - [x]

scripts/openmc-update-mgxs.py
```diff
@@ -0,0 +1,232 @@
...
+        args['output'] = filename
+
+    # Parse and return commandline arguments.
+    return args
+
+
+def get_data(element, entry):
+    try:
+        value = element.find(entry).text
+    except:
```
nelsonag: Good point.  I revised this to avoid the try/except  completely and made the later try/except more specific with the Exceptions we were worried about.

docs/source/io_formats/mgxs_library.rst
nelsonag: Boooo-Yaaaaa!!!

docs/source/io_formats/mgxs_library.rst
nelsonag: Absolutely can, and in fact when writing the library, this is done automagically.  I chose to keep this here since to reduce the Fortran side complexity low.  This would come about because the proper way to check to see if something is fissionable would be via nu_fission.  If nu_fission is not present, then its easy, it isnt fissionable.  BUT, if it is present, then I need to read in either as a matrix or a vector, depending on if chi is present too, AND then I need to see if the data is all zeros or if it has any non-zero values.  I figured this was alot of cruft just to enter an if-branch and so I might as well just store the boolean and make it cleaner.

docs/source/io_formats/mgxs_library.rst
nelsonag: - [x]

docs/source/io_formats/mgxs_library.rst
nelsonag: haha whoops.  It'd be better if I was consistently wrong ;-)

examples/python/pincell_multigroup/build-xml.py
nelsonag: me too :-( there were just too many arguments involved.

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs_library.py
nelsonag: Yes, it actually is set by the set_*fission guys as well so the user doesnt need to use this at all.

openmc/mgxs_library.py
nelsonag: OBE, but I should have.

openmc/mgxs_library.py
nelsonag: OH i remember why i did this.  The user of this library passed absorption, which could be a list.  Its not my call to change what they passed me, so I introduced npfission (etc) to simply be a local version I can change.

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: Its actually 293.6K

wbinventor:
But it won't change what they passed I'm since the variable name is local
to the scope of the method.

openmc/mgxs_library.py
nelsonag: Thats not unreasonable - i just removed the setter.

scripts/openmc-update-mgxs.py
nelsonag: Hopefully noone, but since we have a release out the door I cant assume thats true.

src/tracking.F90
nelsonag: This comment was from tracking.F90 which isn't what you intended.  Is there an mgxs.h5 file that is in version control from the testing folder?

nelsonag:
Ok, nap time is over.  Almost done with all these comments.  Whats left is figuring out how I broke one test and further discussions re: the scatter matrix ordering.  What I have now is optimum for MGMC, but what you want is optimum for determinstic transport.  Sounds like an arm-wrasslin match is in order! ;-)

friedmud:
Well - I've contributed very little overall... so feel free to ignore me!  😄 

Worse comes to worse I can always pre-process the HDF5 files and generate new ones with a transposed scattering matrix.

That said - we could always provide an identifier to say which way the matrix is stored... and we could provide an option to write it out either way.

docs/source/io_formats/mgxs_library.rst
samuelshaner: Could this also be changed to `inverse-velocity` to have consistency with the tally score and MGXS object?

samuelshaner:
Once the conflicts and any lingering review requests are made, I think this is ready to go. I've tested the code in this PR out quite extensively and haven't found any run-time issues that were not caught by the test suite.

nelsonag:
Roger that. Sorry about the delays this week. I should be able to get to
this in the next few days.  As for the scattering matrix ordering, I'm
leaning towards implementing the ordering in the ordering preferred by the
deterministic codes.  Those have a better chance of being used for large
problems than MG MC does.  I'd rather support one format for this than two.
Us this reasonable?

friedmud:
Certainly wouldn't hear any complaints from me! :-)

wbinventor:
I agree, good plan!

nelsonag:
Ah never mind, I take that back for two reasons:  1) I'm lazy and this seems like a PITA with no guarantee anyone else will actually use this library, and 2) I need some convincing you are settled on the location of the Scattering Order index ([Gout][L || Nmu][Gin] vs [Gout][Gin][L || Nmu]) - the options of course would be dependent upon where your energy loop is nested) and so I dont want to put in to much effort on something that could be of no value anyways in the end.  I suspect this isn't fully fleshed out since I *think* you dont yet have anisotropy in either of the two codes we're talking about yet.  

So with the above, I recommend we go back to the user option for matrix ordering for now and therefore the Library will have a descriptor that states the ordering of the matrix.  In row major ordering: [Gout][Gin][L || Nmu] (deterministic transport) OR [Gin][Gout][L || Nmu] (MC).  

I will modify this PR to add this option to the documentation, set up the PyAPI to write the flag, and set up the Fortran to read the flag.  I will not yet write (PyAPI) or read (Fortran) in the deterministic mode in this PR, instead an equivalent of a NotImplementedYet error will be raised.  

Thoughts?

wbinventor:
Frankly I'm fine either way since neither ordering will impact OpenMC performance. The flag sounds like a reasonable approach to me - who cares how the data is stored under-the-hood if one can input it any way they please?

I do think that any smart deterministic code developers should adopt the convention promoted by @gastdr. That said, I agree that it is highly questionable that mainstream codes will all adopt the YakXS convention any time soon (for better or worse).

friedmud:
Sounds good to me.

nelsonag:
Well, to be fair, the deterministic ordering will impact OpenMC performance.  OpenMC doesn't do a vector dot product on the data like deterministic can, but it still does take advantage of caching (and thus contiguous access on the gout dimension is preferred).  

I have only heard about YakXS via this PR (and the preceeding issue).  Is there any actual push to make this library itself 'the' standard?  If by convention you are simply referring to the scattering matrix ordering itself as being a common, then I mostly agree with that, as the specific ordering of the scattering matrix is certainly not rare - the old LANL DTF format is ordered in the same way too.  Most of the variance I've seen on the scattering matrix ordering is where the Legendre order is, though the old MATXS format (i.e., TRANSX' output, among other codes) is similar to the MC format in this library.

samuelshaner:
Looking through `mgxs_header.F90` I notice that some of the keys for items in the mgxs hdf5 file have underscores and some have hyphens. Hyphens is definitely the dominant punctuation. Can any keys such as `scatter_data`, `scatter_matrix`, and `inverse_velocities` (and perhaps others) be changed to use hyphens for consistency? Also can inverse velocities be changed to inverse velocity to be consistent throughout OpenMC?

wbinventor:
@nelsonag good point on the energy out caching - I suppose even MC needs to access all of those points at each collision to build the sampling distribution over outgoing energies. Perhaps deterministic and MC are not so dissimilar after all...

samuelshaner:
Actually looking at `xstype` character array in the `get_xs(...)` function where only underscores are used, I would recommend that underscores be the punctuation of choice instead of hyphens.

nelsonag:
Hmmm, without the source, I'm willing to bet the only hyphenated tags in
mgxs_header are ones which are consistent with the MGXS_TYPES in
openmc.MGXS, is that right?

For those tags, we all need to be aware that this change will break every
user script which uses the MGXS tallying API (not just those piping the
data back in to MG OpenMC).  So let's all agree on the format and rules for
underscore vs space vs hyphens before going further.

wbinventor:
It's fine by me if you need to break things to make the notation consistent. I recently had to resign myself to stop updating to develop since I couldn't keep up and write my thesis at the same time. All of my are broken with the #725 anyway, but it's a minor woe in the name of progress. Even still, the sooner we settle things down the better ;-)

paulromano:
All of our existing HDF5 file formats use underscores exclusively, so while it's not written in stone anywhere, I'd highly suggest sticking to that.

@wbinventor Sorry! The changes to develop have been rather intrusive lately but it is in the interest of getting to a more stable state sooner rather than later.

samuelshaner:
Currently just `nu-fission` and `kappa-fission` use hyphens. This does make it consistent with the MGXS_TYPES and with the way scores are listed in xml files. However, when the cross sections are read in `mgxs_header.F90` this results in there being `'nu-fission'` (in `from_hdf5(...)`) and `'nu_fission'` (in `get_xs(...)`) in the same file. When adding new cross sections in my PR, I just found that a little confusing. Looking at it closer, the underscores in the cross section strings in the `get_xs(...)` functions in `mgxs_header.F90` seems to be the inconsistency. Changing the underscores to hyphens in the `get_xs(...)` function should make all strings that contain nu-fission and kappa-fission have a hyphen in the python, xml, hdf5, and fortran files (I think). Also, the `inverse_velocity` should be changed to have hyphens throughout. How about just making those changes?

That covers the nu-fission, kappa-fission, and inverse-velocity. There's also the underscores in `'scatter_matrix'`, `'scatter_data'`, and `'multiplicity_matrix'`. The convention in the MGXS module is to have a space before matrix. What do you think about changing the underscores to spaces in these three items?

nelsonag:
Starting up on the library scattering matrix ordering issue now; @samuelshaner it looks like your follow-on PR to this just dealt with the hyphen vs underscore issue and so I dont need to resolve that in this PR, is that right?

samuelshaner:
Correct, the hyphen/underscore dilemma is addressed in my PR. I was doing it more in preparation for what would be included in your PR, but since the route of just modifying the strings in the `get_xs(...)` function doesn't result in any modifications to input files, I'm fine with this item just be addressed in my PR.

nelsonag:
Ok, fixed, as described in my previous comment.

Additionally, I undid a previous change that I incorporated per @wbinventor to be consistent with my string comparisons (i.e., `a_string == 'test_string'` and `a_string is `test_string`)  I went with the `is` path before, now all are `==`.  Apparently the `'test_string'`, while immutable, does not have the same id as `a_string` which should contain it, so the `is` test would sometimes fail.

nelsonag:
This one is almost ready - figured I would improve MG test coverage.  Working on that right now, should be done very soon

nelsonag:
Ok, I submitted the modified tests.  I raised coverage from 63% to 74% by throwing in more angular data, histogram scatter distributions and combinations of the two.  The rest will come with the build of a new MGXS library for testing that will have data needed to get to the rest of the coverage.

nelsonag:
Life is hard.  Raised the mgxs_header coverage to 76% by getting nuclidic tallies in there.  The rest definitely requires me to make a new library so I can access the rest.  Thats a longer term effort and not for this PR.

Ok, so long as Travis' tests pass I consider this to be done and all comments to be addresed so its ready for the final review & A-OK from a lucky soul.

nelsonag:
... dont hold your breath

nelsonag:
NOW it's "done"

samuelshaner:
Just gave this a final read over and I think its ready to go. Since @paulromano and @wbinventor also reviewed this PR, I'll let them weight in on final approval and merge this in.

wbinventor:
The Python changes look good to me as well, though I didn't give the Fortran diff a good read.

openmc/mesh.py
```diff
@@ -286,21 +286,24 @@ def build_cells(self, bc=['reflective'] * 6):
                    openmc.XPlane(x0=self.upper_right[0],
                                  boundary_type=bc[1])]
         if len(self.dimension) == 1:
-            yplanes = [openmc.YPlane(y0=np.finfo(np.float).min,
-                                     boundary_type='reflective'),
-                       openmc.YPlane(y0=np.finfo(np.float).max,
-                                     boundary_type='reflective')]
+            yplanes = [openmc.YPlane(y0=1000., boundary_type='reflective'),
+                       openmc.YPlane(y0=1000., boundary_type='reflective')]
```
paulromano: Same deal here with 1000 vs 1e10

openmc/mgxs/library.py
```diff
@@ -1319,7 +1283,7 @@ def check_library_for_openmc_mgxs(self):
              ('nu-transport' not in self.mgxs_types))):
             error_flag = True
             warn('A "nu-transport" MGXS type is required since a "P0" '
-                 'correction is applied, but a "nu-transport" MGXS is '
+                 'correction is applied, but a "nu-transport" MGXS == '
```
paulromano: Got a little overzealous with search-and-replace?

openmc/mgxs_library.py
```diff
@@ -970,144 +1102,213 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
-        return element
+        """
+        grp = file.create_group(self.name)
+        if self.atomic_weight_ratio is not None:
+            grp.attrs['atomic_weight_ratio'] = self.atomic_weight_ratio
+        if self.fissionable is not None:
+            grp.attrs['fissionable'] = self.fissionable
+        if self.representation is not None:
+            grp.attrs['representation'] = np.array(self.representation,
+                                                   dtype='S')
```
paulromano: Replace RHS with `np.string_(self.representation)`

openmc/mgxs_library.py
```diff
@@ -970,144 +1102,213 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+            if self.representation == 'angle':
+                if self.num_azimuthal is not None:
+                    grp.attrs['num_azimuthal'] = self.num_azimuthal
+                if self.num_polar is not None:
+                    grp.attrs['num_polar'] = self.num_polar
+        if self.scatter_format is not None:
+            grp.attrs['scatter_format'] = np.array(self.scatter_format,
+                                                   dtype='S')
```
paulromano: Use `np.string_`

openmc/mgxs_library.py
```diff
@@ -970,144 +1102,213 @@ def set_multiplicity_mgxs(self, nuscatter, scatter=None, nuclide='total',
...
+        if self.scatter_shape is not None:
+            grp.attrs['scatter_shape'] = np.array(self.scatter_shape,
+                                                  dtype='S')
```
paulromano: Use `np.string_`

openmc/mgxs/library.py
nelsonag: - [x]

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mesh.py
nelsonag: - [x]

paulromano:
Thanks @nelsonag!

PR #732, opened 2016-10-06
Track Density-Weighted Average MGXS

wbinventor:
This PR fixes an issue in the `MGXS.get_subdomain_avg_xs(...)` method, among a few other minor bug fixes. Currently, this method will take the geometric average of the MGXS for a bunch of distribcells or mesh cells. However, the geometric average does not preserve global reactivity.

Instead, a _track density-weighted (TD-weighted)_ average MGXS should be computed. Fortunately, this is as simple as substituting `Tally.summation(...)` for `Tally.average(...)` in the `get_subdomain_avg_xs(...)` method. In effect, this computes the  average MGXS from tally sums which are equivalent to what one would obtain if an explicit cell or material tally were specified for the subdomains of interest. And as my thesis broadly confirms, the TD-weighted average MGXS does indeed preserve global reactivity.

I made these changes a few months ago, and ran into some issues when merging with the develop branch earlier today due to #723 and #725. As a result, this PR is likely not ready for merging - I'm not yet sure I've fixed all of the issues. I haven't been able to run my scripts which use TD-weighted MGXS since something with the new `MeshFilter` is breaking when loading from a statepoint. I'll post here again once I have a chance to sort that out.


smharper:
Sorry, #725 broke statepoint backwards compatibility with mesh filters.  Previously, the `bins` for mesh filters were mesh indices but #725 changed that to mesh ids.

If your statepoints are for big runs that are hard to rebuild, I can probably write a script for you to manually convert the indices to ids in your statepoints.

openmc/filter.py
```diff
@@ -300,7 +300,43 @@ def get_bin_index(self, filter_bin):
...
+                    nx, ny, nz = self.mesh.dimension
+                    val = (filter_bin[0] - 1) * ny * nz + \
+                          (filter_bin[1] - 1) * nz + \
+                          (filter_bin[2] - 1)
+                else:
+                    nx, ny = self.mesh.dimension
+                    val = (filter_bin[0] - 1) * ny + \
+                          (filter_bin[1] - 1)
+
+                filter_index = val
```
smharper: I would prefer you get rid of all the `isinstance` blocks and just override `get_bin_index` on the relevant subclasses.

wbinventor:
Haha, no problem, I've mastered the art of breaking backwards compatibility the last few years so I understand. But that won't stop me from likening #725 to Apple removing the headphone jack on the latest iPhone ;-)

Thanks for offering to help convert my statepoints - I don't think that's necessary at this point though. Interestingly, I ran the latest "develop" branch version to generate a new statepoint and it still broke on [this](https://github.com/mit-crpg/openmc/blob/develop/openmc/filter.py#L63) line in "filter.py". I don't have a lot of time to debug into this right now, but will certainly clear things up at some point to help merge this PR.

Separately, I am curious if your change to 'bins' for mesh filters will break the `Tally.summation(...)` and `Tally.average(...)` methods when applied across a range of mesh indices (I think it will).

openmc/mgxs/mgxs.py
```diff
@@ -50,6 +50,7 @@
 # Filter types corresponding to each domain
 _DOMAIN_TO_FILTER = {'cell': openmc.CellFilter,
                      'distribcell': openmc.DistribcellFilter,
+                     'sum(distribcell)': openmc.DistribcellFilter,
```
smharper: You can also average/sum over other domains, right?  Take a look at `MGX.num_subdomains`.  I was trying to handle these guys in #725 by stripping the `avg()` form the name before indexing `_DOMAIN_TO_FILTER`.  Not sure what's the better solution, but we should probably stick to one or the other.

wbinventor:
Closing this PR to open a new one which resolves the test failures.

PR #735, opened 2016-10-08
Upgrade Travis to Trusty and thus drop support for gfortran 4.6

nelsonag:
This PR addresses issue #562 by updating the .travis.yml to use an Ubuntu 14.04 distribution on Travis (and thus use gfortran 4.8) and also updates the docs to state that gfortran 4.8 is the minimum version of gfortran supported.  
I also removed the most obvious instance of gfortran 4.6 specific code based on comments left by developers.


nelsonag:
Interestingly the Travis job timed out, leading to a failure.  It didnt on my local branch (only took 22 minutes there).  There must have been some server outage somewhere (I blame a certain hurricane...).

nelsonag:
ugh.  Looks like Trusty doubles the total test time and thus pushes us to the limit.  I'm re-running this again to make sure there was nothing wrong with Travis' servers yesteray, but otherwise it looks like we may need to address #262 first.

paulromano:
It looks like the trusty infrastructure uses 2 "bursted" cores (i.e., the 2 cores are not guaranteed resources) whereas with the contained-based infrastructure, you're actually guaranteed 2 dedicated cores, so that may explain some of the different in run times. We are also using 3 OpenMP threads and 3 MPI processes by default for build configurations with those options turned on, so that is also probably not helping (especially with both where you'd have 9 total threads running on 1.5 cores!) Nevertheless, the more sensible thing to do is to judiciously go through our test suite and cut down on the unneeded complexity of many of the tests while maintaining the same level of coverage.

nelsonag:
Good point.  I revised the number of cores/processes used to less over-subscribe our 2 bursted cores.  I also am now building mpich, hdf5 and phdf5 with make -j2.  We'll see how this goes.  Next will be simply reducing the number of isotopes in the default input set.  That should help out a bit as well.  

Finally, I was looking in to whether or not we can use Trusty's own hdf5 packages like our docs suggest (i.e., on Debian derivatives, one can use `sudo apt-get install libhdf5-8 libhdf5-dev hdf5-helpers`), but it doesn't look like the Ubuntu/Debian libhdf5 packages are built with --enable-fortran2003 enabled.  So, are the docs wrong, or is my reading of the Debian packaging scripts wrong (I give the latter a 75% probability, that process is wayyy more confusing than an Arch PKGBUILD).  As a follow-on, can we pull MPICH/OpenMPI from the official repositories?

paulromano:
They didn't use `--enable-fortran2003` when configuring HDF5 until 15.04 so unfortunately there's nothing we can do about trusty. There's a chance we would be able to use MPICH or OpenMPI from the repositories though, although I seem to remember running into issues trying to do that?

nelsonag:
My previous push which just changed the number of cores/processes used just finished successfully in 30 minutes (it used to timeout at ~48 min).  I am now re-running that exact same commit on Travis to get a baseline total test time without having to build MPICH and HDF5.  I am doing this because my next push (ready to go locally) will be one which reduces the nuclides on the default InputSet a good amount (addressing #262).  With these changes, I've taken our slowest tests and reduced their total runtime by 40%.  That should give us some extra margin.

After that I'll see if I can use Trusty's MPICH or OpenMPI.  If it doesn't work the first try or two I probably wont pursue further - I imagine any debugging will be a serious PITA without a local Trusty install to work with.

paulromano:
Pretty sure the problem is that the packages in Ubuntu provide `mpif90.mpich`, `mpicc.mpich`, `mpirun.mpich`, etc., not the standard mpicc, mpif90, mpirun names. This would be done so that MPICH and OpenMPI packages don't conflict.

nelsonag:
Ah yes.  Welp I took one more stab at it and failed.  Since this is a pretty minor benefit, I'm ok with just giving up.  With that, this is ready for review with the aim of closing both #262 and #562

paulromano:
@nelsonag If you want, you can just move the branch back a few commits to not show any of the failed mpich efforts. Something like `git reset --hard HEAD~4` should do, and then force push.

nelsonag:
Oh that just makes me so nervous! Incorporated though.

paulromano:
Good job with changing history :) My one concern was about a potential drop in coverage by removing many nuclides in H-M but it looks like it really had no effect on physics coverage which is good. One thing you may wish to address (not in this PR) is the fact that coverage is quite poor in mgxs_header.F90 (63%). If you want to see for yourself, run
```bash
cd tests
./run_tests.py -s -C hdf5-debug_coverage
```
which will spit out a coverage directory. You'll need to have gcov and lcov installed.

nelsonag:
I'll look in to it.  Its probably driven by less tests on the
angle-dependent extended type.  It should be tested, but clearly not as
much as it should be.

PR #739, opened 2016-10-13
Treatment of delayed neutrons in multi-group mode

samuelshaner:
This PR adds the explicit treatment of delayed neutrons to the multi-group mode. In order to achieve this functionality, several modifications and additions needed to be made:
- **Global constant for the number of delayed groups**: Similar to the number of energy groups, a constant has been added to `global.F90` specifying the number of delayed groups of the input cross sections.
- **`delayed_groups` attribute added to `XSdata` and `MGXSLibrary` objects** 
- **Cross sections for `chi-delayed`, `chi-prompt`, `delayed-nu-fission`, and `prompt-nu-fission`**: `chi` and `nu-fission` have been broken up into separate prompt and delayed components in order for the fission neutrons to be sampled as either prompt or delayed and the appropriate spectrum to be used in sampling fission neutrons. The user can input cross sections with or without separate prompt and delayed components.
- **`beta` cross section added as `XSdata` attribute**: The delayed neutron fraction, beta, has been added. If `beta` and `nu-fission` are input, `beta` will break up `nu-fission` into separate prompt and delayed components.
- **Physics treatment of delayed neutrons**: The `create_fission_sites(...)` subroutine in `physics_mg.F90` has been modified to match the procedure in `physics.F90` in treating prompt and delayed emissions separately.
- **Tallies with energy out filters**: The separate mg and ce versions of the `score_fission_eout` subroutine in `tallies.F90` have been combined to reduce code redundancy.
- **Treatment of which nu-fission/chi multi-group cross section types to use**: OpenMC allows the user to input chi and a nu-fission vector or a nu-fission matrix (which allows a chi matrix to be extracted). When adding delayed neutrons, this procedure got very complicated. Instead of having a bunch of different case, certain cross-section types just take precedence over others. For instance, if a user inputs a nu-fission matrix, any values input to chi will be overridden with chi contained in the nu-fission matrix. With regards to delayed neutrons, any prompt or delayed chi or nu-fission cross sections will override values input from chi and nu-fission. The process of inputting MG cross sections is still somewhat confusing so if more clarity is needed or if others have better ideas, I'm all ears.

I'll also note that options for inputting a delayed-nu-fission or prompt-nu-fission matrix have been implemented, but in order to keep this PR as small as possible, that addition will be added in a future PR.

This PR builds on the great work by @nelsonag in PR #719 and will significantly reduce in size once that PR is merged.


src/input_xml.F90
```diff
@@ -4661,15 +4653,25 @@ subroutine read_mg_cross_sections_header(libraries)
...
     end if
 
-    allocate(rev_energy_bins(energy_groups + 1))
-    allocate(energy_bins(energy_groups + 1))
+    if (attribute_exists(file_id, "delayed_groups")) then
+      ! Get neutron delayed group count
+      call read_attribute(num_delayed_groups, file_id, "delayed_groups")
+    else
+      num_delayed_groups = 0
+      call write_message("WARNING: delayed_groups element not provided so &
```
nelsonag: 1) is "element" the right term?
2) I would think most users would not use delayed_groups.  In that case, why warn about it?

src/mgxs_data.F90
```diff
@@ -62,23 +61,6 @@ subroutine read_mgxs()
     allocate(micro_xs(n_nuclides_total))
 !$omp end parallel
 
-    ! Find out if we need fission & kappa fission
```
nelsonag: Why was this block removed?

src/tally.F90
```diff
@@ -1230,20 +1247,21 @@ subroutine score_general_mg(p, t, start_index, filter_index, i_nuclide, &
           else
             score = p % last_wgt
           end if
+
           if (i_nuclide > 0) then
-            score = score * nucxs % get_xs('inv_vel', p_g, UVW=p_uvw) / &
-                 matxs % get_xs('total', p_g, UVW=p_uvw) * flux
+            score = score * nucxs % get_xs('inverse-velocity', p_g, UVW=p_uvw) &
+                 / nucxs % get_xs('total', p_g, UVW=p_uvw) * flux
```
nelsonag: why did you convert from matxs to nucxs?

src/tally.F90
```diff
@@ -1440,15 +1472,421 @@ subroutine score_general_mg(p, t, start_index, filter_index, i_nuclide, &
...
 
       case (SCORE_KAPPA_FISSION)
+
+        ! make sure the correct energy is used
+        if (t % estimator == ESTIMATOR_TRACKLENGTH) then
+          p_g = p % g
+        else
+          p_g = p % last_g
+        end if
+
```
nelsonag: Good catch, thanks.

tests/test_mg_basic/results_true.dat
```diff
@@ -1,2 +1,2 @@
 k-combined:
-1.160653E+00 6.227102E-02
+5.788739E-01 1.945182E-02
```
nelsonag: Do we understand these changes?

openmc/mgxs/library.py
```diff
@@ -925,48 +926,87 @@ def get_xsdata(self, domain, xsdata_name, nuclide='total', xs_type='macro',
...
+                                              nuclide=[nuclide],
+                                              subdomain=subdomain)
+
+        if 'delayed-nu-fission' in self.mgxs_types:
+            mymgxs = self.get_mgxs(domain, 'delayed-nu-fission')
+            xsdata.set_delayed_nu_fission_mgxs(mymgxs, xs_type=xs_type,
+                                               nuclide=[nuclide],
+                                               subdomain=subdomain)
+
+        if 'beta' in self.mgxs_types:
```
samuelshaner: Nope, I'll add that to the documentation.

openmc/mgxs_library.py
```diff
@@ -48,6 +50,8 @@ class XSdata(object):
         to a single temperature at 294K.
     energy_groups : openmc.mgxs.EnergyGroups
         Energy group structure
+    delayed_groups : int
```
samuelshaner: Yeah.

openmc/mgxs_library.py
```diff
@@ -48,6 +50,8 @@ class XSdata(object):
...
+        Num delayed groups
```
samuelshaner: Yeah

openmc/mgxs_library.py
```diff
@@ -67,22 +71,7 @@ class XSdata(object):
         is "angle".
     num_polar : int
         Number of equal width angular bins that the polar angular domain is
-        subdivided into. This only applies when :attr:`XSdata.representation`
-        is "angle".
-    use_chi : bool
-        Whether or not a chi vector or nu-fission matrix was used.
-    vector_shape : iterable of int
```
samuelshaner: I just found that with delayed groups the lines between get blurry on what is a matrix shape. Whereas before vector shape noted just `groups` and matrix noted `groups x groups`, with delayed groups there would need to be shape terms for `energy_groups`, `energy_groups x energy_groups`, `delayed_groups`, `delayed_groups x energy_groups`, and `delayed_groups x energy_groups x energy_groups`. I'm go another PR on deck for a surface mgxs class which would add shapes for `mesh_surfaces` and `energy_groups x mesh_surfaces`. And then there is the `pn_matrix_shape` as well.

I was just thinking it is starting to get bloated with several different shape presets and might be best to just do them individually. How about a dictionary with all the different shape options and then selecting the ones available for the particular mgxs of interest when they are needed?

openmc/mgxs_library.py
```diff
@@ -99,47 +88,45 @@ class XSdata(object):
...
-    vector_shape :
-        total, absorption, fission, kappa_fission, chi, nu_fission (if chi is
-        provided), inverse_velocities
-    matrix_shape :
-        multiplicity_matrix, nu_fission (if chi is not provided)
-    pn_matrix_shape :
-        scatter_matrix
+        Group-wise fission production cross section vector (i.e., if ``chi`` is
+        provided), or is the group-wise fission production matrix.
+    prompt_nu_fission : dict of numpy.ndarray
```
samuelshaner: With this PR, yes. I have another PR with the code all written and ready to go which has a `prompt_nu_fission matrix` and `delayed_nu_fission matrix` that can replace the respective `chi`s. To keep the PRs digestible, I decided to break them up. I'm happy to add them to this PR if that would make things easier.

openmc/mgxs/library.py
samuelshaner: Nope, I'll add that to the documentation.

samuelshaner: Nope, I'll add that to the documentation

openmc/mgxs_library.py
```diff
@@ -99,47 +88,45 @@ class XSdata(object):
...
+        Group-wise prompt fission production cross section vector.
+    delayed_nu_fission : dict of numpy.ndarray
+        Group-wise delayed fission production cross section vector.
+    beta : dict of numpy.ndarray
+        Delayed-group-wise delayed neutron fraction cross section vector.
+    decay_rate : dict of numpy.ndarray
+        Delayed-group-wise decay rate cross section vector.
```
samuelshaner: I just kept it the same as the other cross sections. If 6 delayed group data is used, the decay rates for a material with multiple fissioning species will change with temperature as the cross sections change.

openmc/mgxs_library.py
```diff
@@ -1311,6 +1915,11 @@ def energy_groups(self, energy_groups):
         check_type('energy groups', energy_groups, openmc.mgxs.EnergyGroups)
         self._energy_groups = energy_groups
 
+    @delayed_groups.setter
+    def delayed_groups(self, delayed_groups):
+        check_type('delayed groups', delayed_groups, int)
```
samuelshaner: yeah, good catch.

src/input_xml.F90
samuelshaner: Yeah, probably best not warn about it since most people won't use it. My mindset is just so tailored towards transients now that I worry that someone would try to run a simulation but not set the delayed groups, so all the delayed neutron cross sections would be lumped integrated over delayed groups. I'll remove this warning for now.

src/mgxs_data.F90
samuelshaner: I removed the `get_kfiss` and `get_fiss` so that the all xs are initialized whether the user input them or not. If fission and kappa-fission are not set, they will be ZERO.

src/tally.F90
samuelshaner: Not sure. I changed it back. Also I changed `total` for `absorption` to be consistent with the other scores.

tests/test_mg_basic/results_true.dat
samuelshaner: Hmm...not really. Let me dig into that a bit since this is a pretty big difference!

openmc/mgxs_library.py
```diff
@@ -609,14 +657,17 @@ def set_chi_delayed(self, chi_delayed, temperature=294.):
         """
 
         # Get the accepted shapes for this xs
+        shapes = [self.xs_shapes["[G']"], self.xs_shapes["[DG][G']"]]
+
+        # Get the accepted shapes for this xs
         if self.representation is 'isotropic':
             shapes = [(self.energy_groups.num_groups,),
-                      (self.delayed_groups, self.energy_groups.num_groups)]
+                      (self.num_delayed_groups, self.energy_groups.num_groups)]
```
nelsonag: Shouldnt this block be removed since you use `self.xs_shapes` up above?

openmc/mgxs_library.py
```diff
@@ -876,12 +890,15 @@ def set_delayed_nu_fission(self, delayed_nu_fission, temperature=294.):
...
         """
 
-       # Get the accepted shapes for this xs
+        # Get the accepted shapes for this xs
+        shapes = [self.xs_shapes["[DG][G]"]]
+
+        # Get the accepted shapes for this xs
         if self.representation is 'isotropic':
-            shapes = [(self.delayed_groups, self.energy_groups.num_groups,)]
+            shapes = [(self.num_delayed_groups, self.energy_groups.num_groups,)]
```
nelsonag: Same here, why is this needd if you have the above?

openmc/mgxs_library.py
nelsonag: Thanks, it looks pretty cool this way and the string key is self explanatory too, which I like

src/tally.F90
nelsonag: Well this is an analog estimator, and so a collision was sampled based on the total macroscopic xs.  So shouldnt we be factoring out the total xs probability (since the sampling effectively incorporates that)?  I suggest you check to make sure the inverse velocity scores produced by your new code produce consistent and sensical answers compared to the other estimators/nuclidic or not/survival biasing or not estimators.  I had all those comparisons running when I originally set these score lines.

openmc/mgxs_library.py
samuelshaner: Yep, that can be removed.

openmc/mgxs_library.py
samuelshaner: Yeah

nelsonag:
I'm just making sure I didnt miss anything - I think the only thing I'm waiting for at this point is for the test results to be understood/fixed.

samuelshaner:
@nelsonag - I looked into things briefly last night, but was still getting drastically different answers for the mg tests. Given the large difference in answers, something is certainly different than before. Later this week I plan on taking a deep dive into finding out what the issue is and also doing some much needed verification of the treatment and tallying of delayed neutron parameters in mg mode so don't expect to see much movement on this PR for at least a few days.

samuelshaner:
@nelsonag - I found what the issue was for the large discrepancy in the test results. I wasn't reading in any "nu-fission" angle-dependent cross sections so the nu-fission xs were all zero for all the angle-dependent cross sections for those tests. The new test results are now much more sensible. I'm gonna do some verification this evening then I think this is ready to go.

samuelshaner:
I did some testing and verification on the cross-sections for the new mgxs that were added in this PR and things look to be working as expected. I think this PR is ready to be merged or further reviewed.

nelsonag:
I'd like to understand why the test results are still different.  Did you change the test library to incorporate delayed data and thus the only difference is statistics?

samuelshaner:
The only differences should be due to statistics. Previously in the `sample_fission_energy` functions in `mgxs_header.F90`, only one random number was needed for the outgoing fission energy. Now an additional random number is needed to choose whether the neutron is prompt or which delayed group it is in. Thus the neutron histories will change. 

That said no mg mode tests use delayed groups. I'm going to add some delayed group terms to `1d_mgxs.h5` and update the tests so those parts of code are tested.

samuelshaner:
I removed `nu-fission` and added `prompt-nu-fission` and `delayed-nu-fission` to the `uo2_ang` and `uo2_iso` materials in `1d_mgxs.h5` so that the tests actually test some of the code for setting delayed multi-group xs and encounter delayed neutrons.

nelsonag:
Looks good to me @samuelshaner; I will give this one day to allow anyone else to chime in and if no action happens then I will merge.

nelsonag:
Thanks @samuelshaner!

PR #741, opened 2016-10-17
Natural element expansion for wo and uranium enrichment option

samuelshaner:
This PR addresses a lingering `NotImplementError` by adding support for expanding the nuclides of an element when the element is added to a material as a weight percent. This is super helpful when specifying structural materials that are often reported as weight percents of a bunch of elements. In doing so, the bloated `input_xml.F90` file was reduced in size by replacing a large case statement with a look-up into a global array. I'm no expert on Fortran data structures, so if anyone has a recommendation for a more elegant way to do this, let me know. 

Additionally, an enrichment option for elements in xml and the Python API was added. The enrichment option is only supported for uranium. The relative fraction of U-234, U-235, and U-238 are computed using the same formulas found in CASMO 4 and the BEAVRS benchmark, namely W_{U-234} is equal to 0.008 \* W_{U-235} with U-238 making up the balance. I feel like this item has been discussed before, so I'm curious to see whether there is a back story to why this isn't already in the code.    

Tests have been added for element expansion for wo and for enrichment and the documentation has been updated.


smharper:
@samuelshaner, yeah, we probably should have had a quick way to handle enrichment ages ago.  Thanks for taking the time to add it now!  That said, I don't think it makes sense to have an `enrichment` attribute that is only valid for one element.  I'd propose leaving enrichment out of the Fortran code entirely, and just sticking it in some Python class.  Maybe we have a class `openmc.model.EnrichedUranium` which takes an enrichment in the constructor.  I can see us in the near future following that line of thought to add `openmc.model.Water` which accepts pressure and temperature, `openmc.BoratedWater` which takes ppm of boron, `openmc.model.UO2`, etc.

samuelshaner:
Gotcha. I'll take enrichment out of the Fortran code and consider making a class `openmc.model.EnrichedUranium`. I like your idea of the models for materials. I'm looking forward to seeing how those models are implemented for the materials you mentioned.

smharper:
Hmm, it's hard to think of a clean way to make a class like `EnrichedUranium` interact with `Material`.  On first thought, you might want it to inherit from `Material`:
```python
class EnrichedUranium(Material):
    def __init__(self, enrichment, material_id=None, name='', temperature=None):
        ...
```
But then, how do you handle the fact that the user might want the uranium to be specified in either ao (most convenient with UO2) or wo (U-Zr alloys)?  Do you stick an ao/wo flag in the constructor?  That would be kind of messy.  Maybe we want each of these models to be their own object with an `add_to_material` method:
```python
class EnrichedUranium(object):
    def __init__(self, enrichment):
        ...

    def add_to_material(mat, fraction=1.0, fraction_type='ao'):
        if fraction_type == 'ao':
            mat.add_nuclide('U234', fraction*self.mole_frac_234, 'ao')
            mat.add_nuclide('U235', fraction*self.mole_frac_235, 'ao')
            ...
```

samuelshaner:
Looking at it from the users perspective, I really think it is important to have something simple - like an optional argument to a function. Having a separate model makes sense for complicated things like interpolating in steam tables to get the density for water at a specific temperature and pressure, but seems a bit cumbersome to do something as simple as setting the uranium enrichment.

What about just having an optional argument to `material.add_element(..., enrichment=None)` where enrichment is only used for uranium and the element is force expanded to U-234, U-235, and U-238 with the correct percent densities if an enrichment is given?

nelsonag:
Does EnrichedUranium have to be a class? What if it was just some form of a
Material factory function?

smharper:
@samuelshaner, yeah you're right.  A whole separate class is probably over-complicating things.  I'll be happy with a optional `enrichment` argument to `add_element`.

Also, I imagine that the w_234 = 0.08% * w_235 is an approximation only valid at low enrichments.  Imagine if you stuck a mixture of just 235 and 234 in a centrifuge; clearly the fraction of 234 will rise.  I'd suggest raising a warning if the user asks for an enrichment greater than 6% (chosen arbitrarily by assuming that the 0.08% approx is valid for all our <5% reactors).

I like the way you think @nelsonag.  Maybe a factory method for `Material` makes the most sense for things like borated water.

samuelshaner:
Good suggestions @smharper - I added a warning if an enrichment of over 5 % is given. As a real-world justification, there is only one commercial enrichment plant in the world (located in Russia) licensed to enrich over 5 %, so any material over 5 % is likely going to be from down-blended HEU anyway and thus the isotopics are going to be a little weird.

paulromano:
@smharper Your physical intuition is correct -- U234 does get preferentially enriched as you go to higher enrichments. Just taking a quick peek at some HEU benchmarks from ICSBEP shows that w_234 > 0.01 w_235 for HEU.

paulromano:
While we're on the topic, let me propose something radical and see how people react. Right now, we have implementations of natural elements on both the Python and Fortran side, i.e. you can create an `openmc.Element` in the Python API and there is an equivalent `<element>` in the XML side that gets auto-expanded into nuclides by the Fortran code. What if we only had `openmc.Element`s that always get auto-expanded on the Python side. This would enable us to remove all that crap code on the Fortran side. It will probably require some changes in `openmc.mgxs` to handle elements correctly but probably nothing too bad.

wbinventor:
Whoa...be careful! we're fast approaching the 50/50 mark between lines of Fortran and Python in the codebase...

paulromano:
Actually we're already there:
screenshot from 2016-10-19 10-21-28

But really, this does kind of go with my vision for the future -- put all the fancy features that make OpenMC such a fun and easy code to use in the Python API and have the Fortran code (or C++ one day) be lean and mean.

wbinventor:
I didn't realize we were already well over that threshold! Haha, I was just kidding...I wholeheartedly agree with you on that vision, we can be proud rather than ashamed of our Python-heavy codebase!

paulromano:
Proud indeed, and you were the one that started it all, Mr. Boyd!

wbinventor:
Not true, there was a sizable codebase to begin with. Regardless, sorry for the distraction from how we should make elements out of nothing!

samuelshaner:
I'm in favor! The expansion is already (mostly) redundant in both Fortran and Python. If element expansion is only supported in the Python API, it should be much more readable that what is currently in Fortran.

One question I have is how would different cross section libraries be treated in order to expand into the appropriate nuclides in that library? Would the Python API access the `OPENMC_CROSS_SECTIONS` environment variable or would the user have to set the cross section library?

paulromano:
Either of those are viable options. Setting the library explicitly is a little more transparent, whereas looking at the environment variable is perhaps more convenient. The latter might get tricky though; what if the user sets Settings.cross_sections? Then the material needs to know that the environment variable has been overridden.

samuelshaner:
The presence of the `cross_sections` in the settings file complicate things. When trying to expand an element, several pieces of data are needed:

- The natural abundances of the isotopes in an element. Currently this is in `openmc/data/data.py`.
- The atomic masses of the isotopes (for expanding by wo). Currently this is in `openmc/data/mass.mas12`.
- Of the natural isotopes, which ones are supported in the requested cross section library. The isotopes are listed in the `cross_sections.xml` file.

Currently, in the Fortran element expansions, the supported isotopes are hard coded in. This is error prone as can be seen with the nndc cross sections. The nndc cross sections don't have Neon (or at least its not in the nndc cross sections I have), but all cross section libraries hard coded in OpenMC say they have all the natural Neon isotopes.

To get around this, we could consider doing the following:

- Moving the `cross_sections` attribute from the settings file to the materials file. This breaks backward compatibility though, so we would have to consider this carefully.
- When elements are expanded, the natural abundances will provide the isotopes that are physically present in the element.
- The natural isotopes will then be checked against the isotopes in the `cross_sections.xml` file. If *some* isotopes are not present, the missing isotopes abundance fraction will be combined with the next closest natural isotope, deferring the the isotope with more neutrons if it is a tie (e.g. Tungsten has W180 combined in W182 and Tantalum has Ta180 combined in Ta181 in many libraries). If *all* isotopes are not present, the A=0 version of the isotope will be used (e,g. Carbon, Zinc, Platinum, Osmium, and Thallium for some libraries). If the A=0 version of the isotope is not present, an error will be output saying there are no supported natural isotopes in your library for the element of interest.

nelsonag:
I have three thoughts on the matter:

1) I agree we should eventually make the general move @paulromano is
talking about doing: less Fortran, more Python.

2) I do however worry about replacing XML functionality when doing so.
That worry is because I wonder if we, the development team, are
underestimating our users reliance on the XML.  We're all in on the Python,
but that's because we obviously know how to program in it. I don't think
that's true for the standard user. While more features have been added to
the pyapi we have yet to take an existing feature away from the XML in
favor of the pyapi, so careful thought is required. Or maybe a survey on
the users group with a 1- 2- week window to answer.

3) As for the specific question on element expansion only in Python and how
to deal with cross_sections.xml: how about the Python just sticks to
'nature' and converts elements in to all the naturally occurring isotopes.
Then the Fortran raises a warning if one of the isotopes isn't in the data
set but IS a valid isotope? An error would be too much because then it
would be a PITA to get the code to run if you have a model with a missing
nuclide.  I guess I'm arguing that the Fortran do Sam's last bullet on the
previous comment.

Thoughts?

paulromano:
I hear you @nelsonag on the concern about users not necessarily using the Python API and instead writing XML directly. I have mixed feelings about this. On one hand, I can understand that some people might feel more comfortable just sticking to the XML and not having to write a program just to create inputs. In fact, it even took me a bit of time to embrace the Python API, not because it wasn't cool but because initially when @wbinventor put it together it didn't really provide any *features* beyond what was available with XML. That is clearly not the case in the present day where there are a lot of things you can do on the Python side that you simply can't do with just XML. That set of things that can only be done in Python will continue to grow as well (e.g. depletion!). In my opinion, this is why we should be, and essentially are, promoting use of the Python API over just writing XML. This is why our entire workshop at PHYSOR focused on using the Python API. So, to come back to the original point, in this case I'm ok removing capability on the XML side because it gives one more reason for users to "take the plunge" and start using the Python API. I completely understand that this will cause pain for some people, not only in having to learn a programming language (albeit one that most consider very easy to learn) but moreover in having to set up their system with Python and its third-party package dependencies. But, at the end of the day, I do think the pain is worth the gain.

After thinking about the implementation issue a bit, I'm leaning away from relying on the cross sections file because it changes behavior in a non-obvious manner. Instead, I'm thinking the user should just specify what isotopes, if any, are absent:
```Python
def add_element(self, element, ..., absent_isotopes=None):
    """
    ...
    absent_isotopes : list of str
        Isotopes that are not present when expanding the element into
        its naturally-occurring isotopes. [description of how the missing
        abundance is grouped with another nuclide]
    ...
    """
```
so for example, if you wanted to use elemental Oxygen with ENDF/B-VII.1 data, you'd call:
```Python
water.add_element('O', absent_isotopes=['O18'])
```
This would avoid the issues with finding the list of cross sections and would give the user full control.

samuelshaner:
I'm fully on-board with giving capabilities to the Python side that are not also supported on the XML side.

As for the element expansion, I would argue that element isotopic expansion should be completely abstracted away from the user. While it would be non-obvious what isotopes an element is being expanded in to, it abstracts away something that most users are probably not aware of. If the user wants full control over the isotopics, they can add them individually. When creating their high-strength steel material with an ENDF_BVII0 library, how many users are going to be aware that tungsten has W-180 and W-182 lumped together? When they get a:

    ERROR: Could not find nuclide W180 in cross_sections data file!

how are they going to know how to properly lump W-180 in to W-182? If we put in the logic that I described above, it should handle the proper expansion for all the major cross section libraries and avoid any user confusion about knowing how to treat absent isotopes. Furthermore, it should avoid accidentally listing absent isotopes that are actually present when, for instance, someone wants to run a problem with a multiple cross section libraries and compare the results. This means that the materials file would now how control over the `cross_sections.xml` file though, which is a departure from the current implementation of the settings file having control over it. That said, it seems more natural to me for the materials file to have control over the cross sections file anyways.

samuelshaner:
Based on the discussion, I've moved the element expansion entirely to Python API. As @paulromano noted, the element expansion needs to happen at "XML export time" rather than "material creation time". The logic that @paulromano described above has been implemented. The documentation has also been updated to reflect the removal of the capability to add elements to a `materials.xml` file. 

Additionally, the `cross_sections` and `multipole_library` are now set in the `materials.xml` instead of the `settings.xml`. The functionality of setting these elements in `settings.xml` still exists, but deprecation warnings alert the user that these elements be moved to the `materials.xml` file. 

`test_natural_element` has also been removed. `test_element_wo` performs the same tests, namely using natural elements, and `test_natural_element` was still in the xml format so it needed updating anyway.

openmc/element.py
```diff
@@ -102,9 +123,160 @@ def expand(self):
...
+        # Create a list of the isotopes in this element
         isotopes = []
-        for isotope, abundance in sorted(NATURAL_ABUNDANCE.items()):
-            if re.match(r'{}\d+'.format(self.name), isotope):
-                nuc = openmc.Nuclide(isotope)
-                isotopes.append((nuc, abundance))
+        for nuclide, abundance in zip(nuclides, abundances):
+            pct = float('{:2.8g}'.format(percent*abundance))
+            isotopes.append((nuclide, pct, percent_type))
+
```
paulromano: I feel like a lot of this code could be simplified. A few suggestions:

- Don't create `Nuclide` objects until the very last for loop; you can simply keep track of the names.
- Replace `nuclides` and `abundances` lists with a dictionary that maps nuclide names to abundances, e.g. `{'U234': ..., 'U235': ...}`
- Use `mutual_nuclides`, `absent_nuclides`, and `library_nuclides` for looping purposes, e.g. adopting this with the previous suggestions would turn

    ```Python
    if len(absent_nuclides) == 0:
        for nuclide, abundance in sorted(NATURAL_ABUNDANCE.items()):
            if re.match(r'{}\d+'.format(self.name), nuclide):
                nuc = openmc.Nuclide(nuclide)
                nuc.scattering = self.scattering
                nuclides.append(nuc)
                abundances.append(abundance)
    ```
    into
    ```Python
    if len(absent_nuclide) == 0:
        for nuclide in mutual_nuclides:
            abundances[nuclide] = NATURAL_ABUNDANCE[nuclide]
    ```

src/input_xml.F90
```diff
@@ -5628,17 +4632,206 @@ subroutine generate_rpn(cell_id, tokens, output)
   end subroutine generate_rpn
 
 !===============================================================================
-! NORMALIZE_AO normalizes the atom or weight percentages for each material
+! ASSIGN_NUCLIDE_DENSITIES Read in the nuclide densities from materials.xml
+! and assign to the corresponding nuclides.
+!===============================================================================
+
+  subroutine assign_nuclide_densities()
```
paulromano: I'm confused as to why nuclide densities are being read separately from the rest of the materials.xml file. Can you explain?

smharper:
Two comments:  First, test_wo doesn't actually need to run OpenMC.  You may as well just create a material and make sure it spits out the right number densities.  Maybe something like [this](https://github.com/smharper/openmc/blob/borated_water/tests/test_borated_water/test_borated_water.py)

Second, how would you feel about expressing enrichment in percent rather than fraction e.g. 5 rather than 0.05.  It's a little annoying (and maybe error prone?) to remember how many zeros you need to write.  CASMO uses percent.

openmc/element.py
```diff
@@ -102,9 +123,160 @@ def expand(self):
...
+                abundances[i] /= sum_abundances
+
+        # Create a list of the isotopes in this element
         isotopes = []
-        for isotope, abundance in sorted(NATURAL_ABUNDANCE.items()):
-            if re.match(r'{}\d+'.format(self.name), isotope):
-                nuc = openmc.Nuclide(isotope)
-                isotopes.append((nuc, abundance))
+        for nuclide, abundance in zip(nuclides, abundances):
+            pct = float('{:2.8g}'.format(percent*abundance))
```
samuelshaner: When I change it to `pct = percent * abundance` the `test_element_wo` and `test_enrichment` tests fail for Python 3. Any thoughts on how to avoid that? Perhaps cut off the amount when writing to the xml file?

openmc/nuclide.py
```diff
@@ -87,9 +89,22 @@ def name(self, name):
...
-            msg = 'Unable to set scattering for Nuclide to {0} ' \
-                  'which is not "data" or "iso-in-lab"'.format(scattering)
+        if not scattering in ['data', 'iso-in-lab', None]:
+            msg = 'Unable to set scattering for Nuclide to {0} which ' \
+                  'is not "data", "iso-in-lab", or None'.format(scattering)
             raise ValueError(msg)
 
         self._scattering = scattering
+
+    def get_element(self):
```
samuelshaner: I'll just remove this.

src/input_xml.F90
samuelshaner: Looks like that was something that just needed to be done when performing that natural element expansion in fortran for elements input as weight percents. I'll combine this into `read_materials_xml`.

src/input_xml.F90
```diff
@@ -5628,17 +4632,206 @@ subroutine generate_rpn(cell_id, tokens, output)
...
-    real(8)        :: x               ! atom percent
-    logical        :: percent_in_atom ! nuclides specified in atom percent?
-    logical        :: density_in_atom ! density specified in atom/b-cm?
+    integer                 :: i               ! index in materials array
+    integer                 :: j               ! index over nuclides in material
+    real(8)                 :: sum_percent     ! summation
+    real(8)                 :: awr             ! atomic weight ratio
+    real(8)                 :: x               ! atom percent
+    logical                 :: percent_in_atom ! nuclides specified in atom percent?
+    logical                 :: density_in_atom ! density specified in atom/b-cm?
```
samuelshaner: I added some other variables (with long initializers) in there in a previous commit and then took them out without readjusting the spaces. I'll remove the extra spaces.

tests/test_enrichment/test_enrichment.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Use surface half-spaces to define regions
+        fuel_pin.region = -fuel_or
+        cladding.region = +fuel_or & -clad_or
+        water.region = +clad_or & +left & -right & +bottom & -top
+
+        # Register Materials with Cells
+        fuel_pin.fill = fuel
+        cladding.fill = clad
+        water.fill = hot_water
```
samuelshaner: Will do.

samuelshaner:
@smharper - the suggestion on the test sounds good. I'll change to that format for `test_element_wo`.

No preference on the fraction vs. percent. I'll change to percent to be consistent with CASMO.

openmc/element.py
```diff
@@ -102,9 +123,160 @@ def expand(self):
...
+
+                # Adjust the abundances for the absent nuclides
+                for nuclide, abundance in sorted(NATURAL_ABUNDANCE.items()):
+                    if re.match(r'{}\d+'.format(self.name), nuclide) and \
+                       int(nuclide[name_len:]) in absent_nuclides:
+
+                        # Get index to the nearest nuclide
+                        a = int(nuclide[name_len:])
+                        i = min(list(range(len(nuclides_a))), key=lambda j: \
+                                abs(nuclides_a[j] - a))
```
samuelshaner: The elements and associated isotopes of concern are:

**Some isotopes lumped together**
- **O**: Some libraries lump O-18 with O-16 and some lump O-18 and O-17 with O-16.
- **Ta**: Some libraries lump Ta-180 together with Ta-181.
- **W**: Some libraries lump W-180 with W-182, but keep in mind W-181 is not a natural isotope.

**All isotopes lumped together**
- **V**: Some libraries lump V-50 and V-51 as V-0.
- **C**: All libraries lump C-12 and C-13 together as C-0.
-  **Zn**: Some libraries lump all the isotopes together as Zn-0.
- **Ga**: Some libraries lump Ga-69 and Ga-71 together as Ga-0.
- **Os**: Some libraries lump all the isotopes together as Os-0.
- **Pt**: Some libraries lump all the isotopes together as Pt-0.
- **Tl**: Some libraries lump all the isotopes together as Tl-0.

I didn't realize O-18 was lumped with O-16 instead of O-17. I'm just going to break up the "some isotopes lumped together" cases individually and print an error if it is a case other than those.

samuelshaner:
@smharper - I created two tests like the one you mentioned but they are failing on travis. Do you (or anyone else) know why a test with that format would fail on travis?

smharper:
@samuelshaner, I don't know, but I know how you can find out.  Make a new branch (so you don't clog the PR with unneeded commits) and merge in my [verbose_travis](https://github.com/smharper/openmc/tree/verbose_travis) branch.  I'd recommend adding `-R enrichment|element_wo` to lines 8 and 10 of tests/travis.sh so it quickly just runs the tests you're worried about.

smharper:
@samuelshaner: By the way, I've forgotten if it's automatic or not, but if it's not, you can ask Travis to test everything you push to your own repo.  For OpenMC, it will run a light version of the tests---just the hdf5-debug tests.

samuelshaner:
I was able to get the element_wo and enrichment tests working, but there seems to be a discrepancy between python 2 vs 3 when the expanded nuclides are output to the xml file. Options I have in mind include:

- Cutting off the expanded nuclide ao or wo at some value, like 14 decimal places.
- Cutting off the values output in the `material.get_nuclide_xml()` method at a certain number of digits, like 14. This will cause many other test to fail though and change answers in almost everyones code.

Any thoughts?

paulromano:
@samuelshaner Seems like the first option is easier to swallow for now.

smharper:
@samuelshaner, I think the reason for your `test_triso` failures is that `Element.expand` iterates over the dictionary `abundances` and the iteration order is non-deterministic.  I.e. isotopes can appear in different orders.  Use an ordered dictionary or sort the outputs.

Also, rather than messing with the precision of the underlying float values, change the string output in `Material._get_nuclide_xml`.   I would recommend `xml_element.set("ao", '{:.14f}'.format(nuclide[1]))`

samuelshaner:
Turns out the issue with `test_triso` was just the dictionary ordering so the round the atomic amount - good catch @smharper!

I believe all other items have been addressed.

src/input_xml.F90
```diff
@@ -2417,73 +2422,6 @@ subroutine read_materials_xml(libraries, library_dict, material_temps)
         end do INDIVIDUAL_NUCLIDES
       end if
 
-      ! =======================================================================
-      ! READ AND PARSE <element> TAGS
-
-      ! Get pointer list of XML <element>
-      call get_node_list(node_mat, "element", node_ele_list)
```
paulromano: Instead of completely ignoring `<element>`, we ought to abort on inputs that still have `<element>`s within a material.

paulromano:
@samuelshaner Looking good. Once those last two comments are addressed I think this is ready to be merged.

samuelshaner:
Just addressed those last two comments.

PR #742, opened 2016-10-18
Fixed the placement of a right parenthesis in xsdata path

tjlaboss:
Fixed the placement of a right parenthesis in xsdata path

This bug cause the --xsdata flag (for Serpent XS) not to work.


nelsonag:
Looks good to me, thanks @tjlaboss

PR #750, opened 2016-11-03
Specify the default source energy distribution in docs

smharper:
Just filling a small gap in our docs.

### DIFF
--- a/docs/source/usersguide/input.rst
+++ b/docs/source/usersguide/input.rst
@@ -528,7 +528,8 @@ attributes/sub-elements:
     sub-elements/attributes are those of a univariate probability distribution
     (see the description in :ref:`univariate`).
 
-    *Default*: Watt spectrum with :math:`a` = 0.988 MeV and :math:`b` = 2.249 MeV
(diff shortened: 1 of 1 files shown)


docs/source/usersguide/input.rst
```diff
@@ -528,6 +528,8 @@ attributes/sub-elements:
     sub-elements/attributes are those of a univariate probability distribution
     (see the description in :ref:`univariate`).
 
+    *Default*: Watt spectrum with :math:`a` = 0.988 MeV and :math:`b` = 2.249 MeV
```
smharper: Whoops, so it is.  Relatedly, is it possible to make one of these `*Default*` statements with a line-break in it?  When I try with a line-break, it makes the statement bold.

samuelshaner: I think so. You can look at the cell temperature sub-element for an example.

smharper: Figured it out.  Thanks for the pointer

samuelshaner:
Looks good!

PR #754, opened 2016-11-08
Additions to openmc.data.Function1D and openmc.data.DataLibrary

nelsonag:
This PR adds two additional classes to openmc.data.Function1D and the ability to initialize an openmc.data.DataLibrary object from a cross_sections.xml file.

Function1D additions
-`openmc.data.Piecewise`: A piecewise function which allows one to define a combination of callable functions, each with an interval of applicability.  
-`openmc.data.Combination`: A function which is a combination of other callabe functions, where the method of combination is simply provided as a `numpy.ufunc` operator for each callable.

An example of these two guys:

```python
func1 = lambda x: 1.
func2 = lambda x: 2.

f = openmc.data.Piecewise([func1, func2], [1.])
g = openmc.data.Combination([f, func1], [np.add, np.add])
x = np.linspace(-10, 10, 100)

plt.plot(x, f(x), label='Piecewise')
plt.plot(x, g(x), label='Combination')
plt.legend(loc='best')
plt.ylim(0, 4)
plt.show()
plt.close()
```
Produces:
piecewise

The openmc.data.DataLibrary loading from an xml constructor (`openmc.data.DataLibrary.from_xml`) is as its name implies - you initialize a DataLibrary object from a cross_sections.xml file itself.

Why am I doing this? Well, I am putting together a PR for plotting macroscopic xs from openmc.Material objects. These items are a key component of that work.

### DIFF
--- a/openmc/data/function.py
+++ b/openmc/data/function.py
@@ -427,9 +427,9 @@ class Combination(EqualityMixin):
         self.operations = operations
 
     def __call__(self, x):
-        ans = np.zeros_like(x)
-        for op, func in zip(self.operations, self.functions):
-            ans = op(ans, func(x))
+        ans = self.functions[0](x)
+        for i, operation in enumerate(self.operations):
+            ans = operation(ans, self.functions[i + 1](x))
         return ans
 
     @property
@@ -448,7 +448,7 @@ class Combination(EqualityMixin):
     @operations.setter
     def operations(self, operations):
         cv.check_type('operations', operations, Iterable, np.ufunc)
-        length = len(self.functions)
+        length = len(self.functions) - 1
         cv.check_length('operations', operations, length, length_max=length)
         self._operations = operations
 
--- a/openmc/data/function.py
+++ b/openmc/data/function.py
@@ -407,7 +407,7 @@ class Combination(EqualityMixin):
     Parameters
     ----------
     functions : Iterable of Callable
-        Functions which are to be added together
+        Functions to combine according to operations
     operations : Iterable of numpy.ufunc
         Operations to perform between functions; note that the standard order
         of operations will not be followed, but can be simulated by
@@ -418,9 +418,12 @@ class Combination(EqualityMixin):
     Attributes
     ----------
     functions : Iterable of Callable
-        Functions which are to be added together
+        Functions to combine according to operations
     operations : Iterable of numpy.ufunc
-        Operations to perform between functions
+        Operations to perform between functions; note that the standard order
+        of operations will not be followed, but can be simulated by
+        combinations of Combination objects. The operations parameter must have
+        a length one less than the number of functions.
 
     """
 
--- a/openmc/data/function.py
+++ b/openmc/data/function.py
@@ -410,7 +410,9 @@ class Combination(EqualityMixin):
         Functions which are to be added together
     operations : Iterable of numpy.ufunc
         Operations to perform between functions; note that the standard order
-        of operations (i.e., PEMDAS) will not be followed.
+        of operations will not be followed, but can be simulated by
+        combinations of Combination objects. The operations parameter must have
+        a length one less than the number of functions.
 
 
     Attributes
--- a/openmc/data/function.py
+++ b/openmc/data/function.py
@@ -428,8 +428,8 @@ class Combination(EqualityMixin):
 
     def __call__(self, x):
         ans = np.zeros_like(x)
-        for i, operation in enumerate(self.operations):
-            ans = operation(ans, self.functions[i](x))
+        for op, func in zip(self.operations, self.functions):
+            ans = op(ans, func(x))
         return ans
 
     @property
@@ -488,7 +488,7 @@ class Sum(EqualityMixin):
         self._functions = functions
 
 
-class Piecewise(EqualityMixin):
+class Regions1D(EqualityMixin):
     """Piecewise composition of multiple functions.
 
     This class allows you to create a callable object which is composed
@@ -499,9 +499,9 @@ class Piecewise(EqualityMixin):
     functions : Iterable of Callable
         Functions which are to be combined in a piecewise fashion
     breakpoints : Iterable of float
-        The breakpoints between each function. The functions
-        in the functions attribute must be provided in order of
-        increasing intervals.
+        The values of the dependent variable that define the domain of
+        each function. The *i*th and *(i+1)*th values are the limits of the
+        domain of the *i*th function. Values must be monotonically increasing.
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


openmc/data/function.py
```diff
@@ -397,6 +397,62 @@ def from_hdf5(cls, dataset):
...
+    """
+
+    def __init__(self, functions, operations):
+        self.functions = functions
+        self.operations = operations
+
+    def __call__(self, x):
+        ans = np.zeros_like(x)
+        for i, operation in enumerate(self.operations):
+            ans = operation(ans, self.functions[i](x))
```
paulromano:
```Python
for operation, function in zip(self.operations, self.functions):
    ans = operator(ans, function(x))
```
would be more Pythonic

openmc/data/function.py
```diff
@@ -432,6 +488,63 @@ def functions(self, functions):
         self._functions = functions
 
 
+class Piecewise(EqualityMixin):
```
paulromano: In GND terminology, this would be a `Regions1D`. Perhaps we should consider adopting that naming scheme? See [here](https://www.oecd-nea.org/science/wpec/sg38/Meetings/2015_May/generalDataContainers.pdf) for a draft spec of data containers (note that it is more than a year out of date).

openmc/data/function.py
```diff
@@ -397,6 +397,62 @@ def from_hdf5(cls, dataset):
...
+
+    @property
+    def operations(self):
+        return self._operations
+
+    @operations.setter
+    def operations(self, operations):
+        cv.check_type('operations', operations, Iterable, np.ufunc)
+        length = len(self.functions)
+        cv.check_length('operations', operations, length, length_max=length)
```
paulromano: I guess I find it a little counter-intuitive that you need to supply N operations for N functions. To me the natural way of specifying would be N-1 operations for N functions. For example, if you want just two functions to be multiplied, right now you'd have to specify `operators = [np.add, np.multiply]` rather than `operators = [np.multiply]`.

openmc/data/function.py
```diff
@@ -432,6 +488,63 @@ def functions(self, functions):
...
+    of multiple other callable objects, each applying to a specific interval
+
+    Parameters
+    ----------
+    functions : Iterable of Callable
+        Functions which are to be combined in a piecewise fashion
+    breakpoints : Iterable of float
+        The breakpoints between each function. The functions
+        in the functions attribute must be provided in order of
+        increasing intervals.
```
paulromano: Maybe change to something like: "The values of the dependent variable that define the domain of each function. The *i*th and *(i+1)*th values are the limits of the domain of the *i*th function. Values must be monotonically increasing."

openmc/data/library.py
```diff
@@ -22,6 +22,12 @@ class DataLibrary(EqualityMixin):
     def __init__(self):
         self.libraries = []
 
+    def __getitem__(self, material):
+        for library in self.libraries:
+            if material in library['materials']:
+                return library
+        return None
```
paulromano: I'm guilty of this too, but the risk of having this as a special method is that no users will ever know about it. It might be better to just have this as a normal method with an accompanying docstring.

openmc/data/library.py
```diff
@@ -78,3 +84,33 @@ def export_to_xml(self, path='cross_sections.xml'):
...
+
+    @classmethod
+    def from_xml(cls, path):
+        """Read cross section data library from an XML file.
+
+        Parameters
+        ----------
+        path : str
+            Path to XML file to read.
+
```
paulromano: Needs a Returns clause

openmc/data/function.py
nelsonag: I've struggled with this too. I prefer N-1 operations as you suggest deep down. I only went with this way to allow the first function to start as a negative.  That is, if `[np.add, np.multiply]` is `f(x) = g(x) * h(x)`, then `[np.subtract, np.multiply]` would be `f(x) = -g(x) * h(x)`

Let me know your thoughts on that. Tonight I'm going to take a stab at an implementation that follows standard order of operations (I'm thinking recursive Piecewise defs will help me) and will address the N-1 operations item at the same time if you prefer that route.

paulromano: I can see the motivation, but think the N operation is approach is confusing. It means that you always have one unary operator and N-1 binary operators. Always having N-1 binary operators just seems cleaner and more intuitive.

nelsonag:
Ok, sounds good.  I'm on the fence anyways, and am not standing on firm
ground as I have no plot types in my macroscopic plotter that requires
np.subtract on the first function.

nelsonag:
Ok, I've given up on how to get order of operations in. With that said, all comments are addressed.

openmc/data/function.py
```diff
@@ -397,6 +397,64 @@ def from_hdf5(cls, dataset):
...
+    """Combination of multiple functions with a user-defined operator
+
+    This class allows you to create a callable object which represents the
+    combination of other callable objects by way of a series of user-defined
+    operators connecting each of the callable objects.
+
+    Parameters
+    ----------
+    functions : Iterable of Callable
+        Functions which are to be added together
```
paulromano: "to be added" needs to be generalized

openmc/data/function.py
```diff
@@ -397,6 +397,64 @@ def from_hdf5(cls, dataset):
...
+        combinations of Combination objects. The operations parameter must have
+        a length one less than the number of functions.
+
+
+    Attributes
+    ----------
+    functions : Iterable of Callable
+        Functions which are to be added together
+    operations : Iterable of numpy.ufunc
+        Operations to perform between functions
```
paulromano: The description from the parameters section should probably be copied here

openmc/data/library.py
```diff
@@ -22,6 +22,22 @@ class DataLibrary(EqualityMixin):
     def __init__(self):
         self.libraries = []
 
+    def get_by_materials(self, value):
+        """Access a library entry by passing a value of the 'materials'
+        attribute of the library entry.
+
+        Returns
+        -------
+        library : dict
```
paulromano: dict or None

openmc/data/library.py
```diff
@@ -78,3 +94,38 @@ def export_to_xml(self, path='cross_sections.xml'):
...
+
+        Parameters
+        ----------
+        path : str
+            Path to XML file to read.
+
+        Returns
+        -------
+        data : openmc.data.DataLibrary
+            openmc.data.DataLibrary object initialized from the provided XML
```
paulromano: I prefer the simpler "Data library object initialized ..." since the type is already clear.

nelsonag:
Thanks for the quick comments @paulromano, all resolved.

PR #757, opened 2016-11-11
Added ability to plot nuclear data for materials, elements, and nuclides

nelsonag:
This PR expands the PyAPI to allow the user to plot macroscopic/microscopic cross sections for an `openmc.Material`, `openmc.Element`, or `openmc.Nuclide` object.  These plotting routines also substitute S(a,b) data for MT=2 at energies below the S(a,b) threshold.  For Materials, this happens if there is an S(a,b) dataset in the material, for the rest the sab table name is input as an optional argument.

The user can request plots which contain lines for the following types:
`total, scatter, elastic, inelastic, fission, absorption, capture, nu-fission, nu-scatter, slowing-down power`, or an integer corresponding to an MT.

The user can also request the lines made above be divided by another of the same type.  Why would you do this? Well, `nu-fission` / `fission` is `nu` (similar for scatter), and `slowing-down power` / `absorption` is the moderating ratio.

As an example, the following code:
```python
uo2 = openmc.Material(material_id=1, name='UO2 fuel at 2.4% wt enrichment')
uo2.set_density('g/cm3', 10.29769)
uo2.add_element(u, 1., enrichment=0.024)
uo2.add_element(o, 2.)
fig = uo2.plot_xs(['total', 'absorption', 'capture', 'scatter', 'fission'])
plt.savefig('uo2.png')
plt.close()
```
produces: 
uo2

and
```python
borated_water = openmc.Material(material_id=4, name='Borated water at 975 ppm')
borated_water.set_density('g/cm3', 0.740582)
borated_water.add_nuclide(b10, 8.0042e-6)
borated_water.add_nuclide(b11, 3.2218e-5)
borated_water.add_nuclide(h1, 4.9457e-2)
borated_water.add_nuclide(h2, 7.4196e-6)
borated_water.add_nuclide(o16, 2.4672e-2)
borated_water.add_nuclide(o17, 6.0099e-5)
borated_water.add_s_alpha_beta('c_H_in_H2O')
fig = borated_water.plot_xs(['slowing-down power'], divisor_types=['absorption'])
plt.savefig('mod_ratio.png')
plt.close()
```
produces:
mod_ratio


openmc/material.py
```diff
@@ -606,6 +608,294 @@ def get_nuclide_densities(self):
...
+            Temperature in Kelvin to plot. If not specified, a default
+            temperature of 294K will be plotted. Note that the nearest
+            temperature in the library for each nuclide will be used as opposed
+            to using any interpolation.
+        cross_sections : str, optional
+            Location of cross_sections.xml file. Default is None.
+
+        Returns
+        -------
+        unionE : numpy.array
```
samuelshaner: I would prefer something like `energy_grid` that is a little more descriptive and follows the under_line syntax.

openmc/nuclide.py
```diff
@@ -93,3 +96,293 @@ def scattering(self, scattering):
             raise ValueError(msg)
 
         self._scattering = scattering
+
+    def plot_xs(self, types, divisor_types=None, temperature=294., axis=None,
+                Erange=(1.E-5, 20.E6), sab_name=None, cross_sections=None,
```
samuelshaner: ditto on Erange syntax.

openmc/nuclide.py
```diff
@@ -93,3 +96,293 @@ def scattering(self, scattering):
...
+                **kwargs):
+        """Creates a figure of continuous-energy cross sections for this
+        nuclide
```
samuelshaner: Can this fit on the previous line? It looks like it would still be less than 80 characters.

openmc/plot_data.py
```diff
@@ -0,0 +1,74 @@
...
+                                875, 891),
+                 'unity': (UNITY_MT,),
+                 'slowing-down power': (2, 4, 11, 16, 17, 22, 23, 24, 25, 28,
+                                        29, 30, 32, 33, 34, 35, 36, 37, 41, 42,
+                                        44, 45, 152, 153, 154, 156, 157, 158,
+                                        159, 160, 161, 162, 163, 164, 165, 166,
+                                        167, 168, 169, 170, 171, 172, 173, 174,
+                                        175, 176, 177, 178, 179, 180, 181, 183,
+                                        184, 190, 194, 196, 198, 199, 200, 875,
+                                        891, XI_MT)}
```
samuelshaner: Can you add the damage cross section to this (MT=444)?

openmc/plot_data.py
```diff
@@ -0,0 +1,74 @@
...
+                    'inelastic': (False,) * len(PLOT_TYPES_MT['inelastic']),
+                    'fission': (False,), 'absorption': (False,),
+                    'capture': (False,), 'nu-fission': (True,),
+                    'nu-scatter': (True,) * len(PLOT_TYPES_MT['nu-scatter']),
+                    'unity': (False,),
+                    'slowing-down power':
+                        (True,) * len(PLOT_TYPES_MT['slowing-down power'])}
+
+# Types of plots to plot linearly in y
+PLOT_TYPES_LINEAR = ['nu-fission / fission', 'nu-scatter / scatter']
```
samuelshaner: Can you add `nu-fission / absorption` to this?

openmc/nuclide.py
```diff
@@ -93,3 +96,293 @@ def scattering(self, scattering):
...
+            temperature in the library for each nuclide will be used as opposed
+            to using any interpolation.
+        sab_name : str, optional
+            Name of S(a,b) library to apply to MT=2 data when applicable.
+        cross_sections : str, optional
+            Location of cross_sections.xml file. Default is None.
+
+        Returns
+        -------
+        E : numpy.array
```
samuelshaner: Can you make this `energy_grid` to be consistent with the `element.calculate_xs()` method?

openmc/nuclide.py
```diff
@@ -93,3 +96,293 @@ def scattering(self, scattering):
...
+        if cross_sections is None:
+            cross_sections = os.environ.get('OPENMC_CROSS_SECTIONS')
+
+        # If a cross_sections library is present, check natural nuclides
+        # against the nuclides in the library
+        if cross_sections is not None:
+            library = openmc.data.DataLibrary.from_xml(cross_sections)
+        else:
+            raise ValueError("cross_sections or OPENMC_CROSS_SECTIONS "
+                             "environmental variable must be set")
```
samuelshaner: Lines 249-260 are now used in several places in the python API. Could you make a general function somewhere so we can avoid code replication?

openmc/element.py
```diff
@@ -254,6 +257,203 @@ def expand(self, percent, percent_type, enrichment=None,
...
+            Ediv, data_div = self.calculate_xs(divisor_types, temperature,
+                                               sab_name, cross_sections)
+
+            # Create a new union grid, interpolate data and data_div on to that
+            # grid, and then do the actual division
+            Enum = E[:]
+            E = np.union1d(Enum, Ediv)
+            data_new = np.zeros((len(types), len(E)))
+
+            for l in range(len(types)):
```
samuelshaner: According to PEP 8, `l` should never be used for a variable name.

openmc/plot_data.py
nelsonag: - [x]

openmc/nuclide.py
nelsonag: - [x] I put it in to the actual DataLibrary.from_xml routine.

openmc/nuclide.py
nelsonag: - [x]

openmc/nuclide.py
nelsonag: - [x]

openmc/nuclide.py
nelsonag: - [x]

openmc/material.py
nelsonag: - [x]

openmc/material.py
```diff
@@ -606,6 +608,294 @@ def get_nuclide_densities(self):
...
+            divide some types.
+        temperature : float, optional
+            Temperature in Kelvin to plot. If not specified, a default
+            temperature of 294K will be plotted. Note that the nearest
+            temperature in the library for each nuclide will be used as opposed
+            to using any interpolation.
+        axis : matplotlib.axes, optional
+            A previously generated axis to use for plotting. If not specified,
+            a new axis and figure will be generated.
+        Erange : tuple of floats, optional
```
nelsonag: - [x]

openmc/material.py
```diff
@@ -606,6 +608,294 @@ def get_nuclide_densities(self):
...
+                x = nuc_densities[n]
+                sum_percent += x * awrs[n]
+            sum_percent = 1. / sum_percent
+            density = -density * sum_percent * \
+                sc.Avogadro / sc.value('neutron mass in u') * 1.E-24
+        nuc_densities = density * nuc_densities
+
+        nuclides = OrderedDict()
+        n = -1
+        for nuc in nucs:
```
nelsonag: - [x] It definitely would. Thanks.

openmc/element.py
nelsonag: - [x]

openmc/element.py
```diff
@@ -254,6 +257,203 @@ def expand(self, percent, percent_type, enrichment=None,
         for nuclide, abundance in abundances.items():
             nuc = openmc.Nuclide(nuclide)
             nuc.scattering = self.scattering
-            isotopes.append((nuc, percent*abundance, percent_type))
+            isotopes.append((nuc, percent * abundance, percent_type))
 
         return isotopes
+
+    def plot_xs(self, types, divisor_types=None, temperature=294., axis=None,
+                Erange=(1.E-5, 20.E6), sab_name=None, cross_sections=None,
```
nelsonag: - [x]

nelsonag:
Thanks @samuelshaner!  Most of the comments are incorporated.  I am very often guilty of camelcase vs unders, usually when I have something with energy, I Just heard using E!

What are your thoughts about a general method in openmc.plots or wherever it makes sense, that takes a nuclide, material, or element, and does type dispatch on plot_xs and calculate_xs?  This would have a non-OO implementation though as these would not be type-bound.

samuelshaner:
The PR does a foray into plotting directly in the Python API, which I don't think has been implemented before. In the long run, I imagine more plotting capabilities will be added to the Python API so I would favor an approach that uses a generic `plot_xs` method that takes in a generic object (material, element, or nuclide) and plots the requested xs types. This method could either be in `plots.py` or in a `plotter.py` file. Personally I would prefer to see this implemented in a `plotter.py` file that grows to include more Python API plotting capabilities in the future. This would have a non-OO implementation, but it should greatly reduce the amount of repeated code and produce the desired functionality.

nelsonag:
@samuelshaner Ok, all set. I created plotter.py and moved/condensed this new code to this area. I would prefer calculate_xs stay with the nuclide/material/element types, but I instead opted to move them to plotter instead. What drove me over the edge was I'd need to completely re-do the types scheme and how it is split in to mts, ops, and yields in calculate_xs in order to avoid a circular dependency.

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+        elif data_type == 'element':
+            ylabel = 'Elemental Cross Section [b]'
+        elif data_type == 'material':
+            ylabel = 'Macroscopic Cross Section [1/cm]'
+    ax.set_ylabel(ylabel)
+    ax.legend(loc='best')
+    ax.set_xlim(energy_range)
+    if this.name is not None:
+        title = 'Cross Section for ' + this.name
+        ax.set_title(title)
```
paulromano: Just pass string directly instead of defining title variable

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+
+# MTs to combine to generate associated plot_types
+PLOT_TYPES_MT = {'total': (2, 3,),
+                 'scatter': (2, 4, 11, 16, 17, 22, 23, 24, 25, 28, 29, 30,
+                             32, 33, 34, 35, 36, 37, 41, 42, 44, 45, 152,
+                             153, 154, 156, 157, 158, 159, 160, 161, 162,
+                             163, 164, 165, 166, 167, 168, 169, 170, 171,
+                             172, 173, 174, 175, 176, 177, 178, 179, 180,
+                             181, 183, 184, 190, 194, 196, 198, 199, 200,
+                             875, 891),
```
paulromano: A more rigorous way to handle this would be to use openmc.data.endf.SUM_RULES (e.g., look at `openmc.data.IncidentNeutron.get_reaction_components()`). Can you look into refactoring?

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+        sab_tab = sabs[nuclide[0].name]
+        temp_E, temp_xs = calculate_xs(nuclide[0], types, T, sab_tab,
+                                       cross_sections)
+        E.append(temp_E)
+        xs.append(temp_xs)
+
+    # Condense the data for every nuclide
+    # First create a union energy grid
+    energy_grid = E[0]
+    for n in range(1, len(E)):
```
paulromano: Maybe just `for grid in E[1:]`?

nelsonag: - [x]

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+                    if yield_check:
+                        found_it = False
+                        for prod in nuc[mt].products:
+                            if prod.particle == 'neutron' and \
+                                prod.emission_mode == 'total':
+                                func = openmc.data.Combination(
+                                    [nuc[mt].xs[nucT], prod.yield_],
+                                    [np.multiply])
+                                funcs.append(func)
+                                found_it = True
```
nelsonag: - [x]

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+        nuc = openmc.data.IncidentNeutron.from_hdf5(lib['path'])
+        # Obtain the nearest temperature
+        if strT in nuc.temperatures:
+            nucT = strT
+        else:
+            data_Ts = nuc.temperatures
+            for t in range(len(data_Ts)):
+                # Take off the "K" and convert to a float
+                data_Ts[t] = float(data_Ts[t][:-1])
+            min_delta = np.finfo(np.float64).max
```
nelsonag: - [x]

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+        temperature in the library for each nuclide will be used as opposed
+        to using any interpolation.
+    sab_name : str, optional
+        Name of S(a,b) library to apply to MT=2 data when applicable.
+    cross_sections : str, optional
+        Location of cross_sections.xml file. Default is None.
+
+    Returns
+    -------
+    energy_grid : numpy.array
```
nelsonag: - [x]

openmc/plotter.py
nelsonag: - [x]

openmc/plotter.py
```diff
@@ -0,0 +1,617 @@
...
+        divide some types.
+    temperature : float, optional
+        Temperature in Kelvin to plot. If not specified, a default
+        temperature of 294K will be plotted. Note that the nearest
+        temperature in the library for each nuclide will be used as opposed
+        to using any interpolation.
+    axis : matplotlib.axes, optional
+        A previously generated axis to use for plotting. If not specified,
+        a new axis and figure will be generated.
+    energy_range : tuple of floats
```
nelsonag: - [x]

openmc/material.py
```diff
@@ -606,6 +606,95 @@ def get_nuclide_densities(self):
...
+        sum_percent = np.sum(nuc_densities)
+        nuc_densities = nuc_densities / sum_percent
+        if not density_in_atom:
+            sum_percent = 0.
+            for n, nuc in enumerate(nucs):
+                x = nuc_densities[n]
+                sum_percent += x * awrs[n]
+            sum_percent = 1. / sum_percent
+            density = -density * sum_percent * \
+                sc.Avogadro / sc.value('neutron mass in u') * 1.E-24
```
nelsonag: - [x]

openmc/material.py
```diff
@@ -606,6 +606,95 @@ def get_nuclide_densities(self):
...
+        sum_percent = 0.
+
+        awrs = []
+        n = -1
+        for nuclide in nuclides.items():
+            n += 1
+            lib = library.get_by_material(nuclide[0])
+            if lib is not None:
+                nuc = openmc.data.IncidentNeutron.from_hdf5(lib['path'])
+                awrs.append(nuc.atomic_weight_ratio)
```
nelsonag: - [x]

openmc/material.py
```diff
@@ -606,6 +606,95 @@ def get_nuclide_densities(self):
...
+        if sum_density:
+            density = np.sum(nuc_densities)
+        percent_in_atom = np.all(nuc_density_types == 'ao')
+        density_in_atom = density > 0.
+        sum_percent = 0.
+
+        awrs = []
+        n = -1
+        for nuclide in nuclides.items():
+            n += 1
```
nelsonag: - [x]

openmc/plotter.py
nelsonag: I removed the explicit amounts of numbers here and replaced with SUM_RULES info when I could.

nelsonag:
The reasons calculate_xs routiens are in plotter instead of nuclide, element, and material are:
1) I dont expect anyone to need calculate_xs except for plotting
2) calculate_xs requires converting from the plot types in to MTs, yields, and operations. Knowing how to do that means pointing back to some reference somewhere that contains the info needed on how to do that. 
  a. Right now that info is in plotter.py.  I think thats the best place to leave it because there really isn't any other use for them besides plotting.  Since they are in plotter, moving the calculate_xs routines to the nuclide (e.g.) would create a circular dependency between nuclide and plotter.
  b. I could do the conversion between plot types and MTs, yields and ops upstream (in plot_xs perhaps), but then if a user would want to use calculate_xs directly, they'd need to input that stuff directly and I think thats pretty confusing.

So, my arguments hinge on one thing: I dont see a strong need for calculate_xs outside of plotting, thus the above. If there is another need, then OK, I guess its worth it to move it.

openmc/plotter.py
```diff
@@ -0,0 +1,588 @@
...
+            yields.append(PLOT_TYPES_YIELD[line])
+            ops.append(PLOT_TYPES_OP[line])
+        else:
+            # Not a built-in type, we have to parse it ourselves
+            cv.check_type('MT in types', line, Integral)
+            cv.check_greater_than('MT in types', line, 0)
+            mts.append((line,))
+            yields.append((False,))
+            ops.append(())
+    print(mts)
```
samuelshaner: Look like you might have used this for debugging.

openmc/plotter.py
```diff
@@ -0,0 +1,588 @@
...
+    else:
+        plot_func = ax.loglog
+    # Plot the data
+    for i in range(len(data)):
+        if data_type == 'nuclide':
+            to_plot = data[i](E)
+        else:
+            to_plot = data[i, :]
+        if np.sum(to_plot) > 0.:
+            plot_func(E, to_plot, label=types[i])
```
samuelshaner: I was trying to plot `'nu-fission / absorption'` but the plot wasn't showing up because some of the values at the highest energies were NANs. I'm not sure what the best general solution is, but I found that adding `to_plot = np.nan_to_num(to_plot)` fixed it by replacing the NANs with 0.

samuelshaner:
Just tested out plotting for some nuclides, elements, and materials and really like this new capability! As soon as my last comments are addressed, this PR has my approval.

openmc/plotter.py
```diff
@@ -0,0 +1,588 @@
...
+        by energy_grid
+
+    """
+
+    # Load the library
+    library = openmc.data.DataLibrary.from_xml(cross_sections)
+
+    # Expand elements in to nuclides with atomic densities
+    nuclides = this.expand(100., 'ao', enrichment=enrichment,
+                           cross_sections=cross_sections)
```
samuelshaner: `100.` should be changed to `1.`. Otherwise, the element cross section will be 100x too big!

zr

nelsonag:
Aw shoot you're right. Looks like we should change the name of the percent
parameter, as it's not a percentage but a fraction.

openmc/plotter.py
nelsonag: - [x]

openmc/plotter.py
nelsonag: Hmm good point and suggestion.

- [x]

openmc/plotter.py
nelsonag: I'll hold off on this sweeping change in this PR, though the specific find above is fixed.  I would need to spend more time thinking about it than I have at the moment.  (Materials uses percent, e.g., and im not sure its always normalized to 1 right now).

openmc/material.py
```diff
@@ -606,6 +606,85 @@ def get_nuclide_densities(self):
 
         return nuclides
 
+    def get_nuclide_atom_densities(self, cross_sections=None):
+        """Returns all nuclides in the material and their atomic densities in
+        units of atom/b-cm
+
+        Parameters
+        ----------
+        cross_sections : str, optional
```
paulromano: This argument is unused now; please remove.

openmc/plotter.py
```diff
@@ -0,0 +1,588 @@
...
+PLOT_TYPES = ['total', 'scatter', 'elastic', 'inelastic', 'fission',
+              'absorption', 'capture', 'nu-fission', 'nu-scatter', 'unity',
+              'slowing-down power', 'damage']
+
+# Special MT values
+UNITY_MT = -1
+XI_MT = -2
+
+# MTs to combine to generate associated plot_types
+_INELASTIC = [mt for mt in openmc.data.SUM_RULES[3] if mt not in [5, 27]]
```
paulromano: Why do you exclude MT=5? It normally does have a neutron in the exit channel, oftentimes with a tabulated yield.

openmc/plotter.py
```diff
@@ -0,0 +1,588 @@
...
+                                prod.emission_mode == 'total':
+                                func = openmc.data.Combination(
+                                    [nuc[mt].xs[nucT], prod.yield_],
+                                    [np.multiply])
+                                funcs.append(func)
+                                break
+                        else:
+                            for prod in nuc[mt].products:
+                                if prod.particle == 'neutron' and \
+                                    prod.emission_mode == 'prompt':
```
paulromano: Looks like you could easily combine this block with the previous one by changing the check on the previous one to `prod.emission_mode in ('prompt', 'total')`

openmc/data/library.py
```diff
@@ -112,6 +114,18 @@ def from_xml(cls, path):
...
+        # OPENMC_CROSS_SECTIONS environment variable
+        if path is None:
+            path = os.environ.get('OPENMC_CROSS_SECTIONS')
+
+        # Check to make sure there was an environmental variable.
+        if path is None:
+            raise ValueError("Either path or OPENMC_CROSS_SECTIONS "
+                             "environmental variable must be set")
+
+        check_type('path', path, str)
```
paulromano: Please add an import: `from six import string_types` and then change str to string_types.

openmc/plotter.py
```diff
@@ -0,0 +1,588 @@
...
+                    'scatter': (False,) * len(PLOT_TYPES_MT['scatter']),
+                    'elastic': (False,),
+                    'inelastic': (False,) * len(PLOT_TYPES_MT['inelastic']),
+                    'fission': (False,), 'absorption': (False,),
+                    'capture': (False,), 'nu-fission': (True,),
+                    'nu-scatter': (True,) * len(PLOT_TYPES_MT['nu-scatter']),
+                    'unity': (False,),
+                    'slowing-down power':
+                        (True,) * len(PLOT_TYPES_MT['slowing-down power']),
+                    'damage': (False,)}
```
nelsonag: Oh that was because I thought I would be smarter than the code and explicitly give values of false or true for each reaction. Then when I saw that the runtime penalty wasn't so bad I didn't bother but left the generality in. Since its trivial to put it back in when needed I will generalize this.

openmc/material.py
nelsonag: - [x]

openmc/plotter.py
nelsonag: - [x]

openmc/data/library.py
nelsonag: - [x]

openmc/plotter.py
nelsonag: - [x]

openmc/plotter.py
nelsonag: I erroneously assumed it didnt have a neutron in the outgoing channel. Now I include it and test for the presence of a neutron product.

openmc/plotter.py
nelsonag: Looks like I was comparing to a line that was close to 600, but there is another at around line 300. fixed now.

openmc/plotter.py
```diff
@@ -0,0 +1,537 @@
...
+                        pw_funcs = openmc.data.Regions1D(
+                            [sab_sum, nuc[mt].xs[nucT]],
+                            [sab_Emax])
+                        funcs.append(pw_funcs)
+                    else:
+                        funcs.append(nuc[mt].xs[nucT])
+                elif mt == 5 and mt in nuc:
+                    # Only consider the (n,misc) products if neutrons are
+                    # included in the outgoing channel since (n,misc) is only
+                    # explicitly needed for scatter cross sections.
```
paulromano: In every case I know of, MT=5 has a neutron in the exit channel. I checked all nuclides in ENDF/B-VII.1, B-VIII.0 beta, and JEFF 3.2 and there are no cases where MT=5 doesn't have a neutron. So, you probably don't need a special block for MT=5.

openmc/plotter.py
```diff
@@ -0,0 +1,537 @@
...
+                            if prod.particle == 'neutron' and \
+                                prod.emission_mode in ('total', 'prompt'):
+                                func = openmc.data.Combination(
+                                    [nuc[mt].xs[nucT], prod.yield_],
+                                    [np.multiply])
+                                funcs.append(func)
+                                break
+                        else:
+                            # Assume the yield is 1
+                            funcs.append(nuc[mt].xs[nucT])
```
paulromano: Did you encounter any cases where this `else` clause is used? Every reaction with neutrons in the outgoing channel should list a neutron product.

openmc/plotter.py
nelsonag: You're right, I checked. I will remove the special block.

openmc/plotter.py
nelsonag: Yeah, in ENDV/B-VII.0 from MCNP, NNDC ENDF/B-VII.1 and JEFF 3.2, MT=18 of U-234 uses this line.  MT=18 is a summed product on the python side, but at this moment there was no product pulled from total_nu.  I added that in and modified this code to read it, and then reorganized this block so we first look for a total product (which now works for all fissions) and failing that, sum the yields from delayed and prompt products.

PR #762, opened 2016-11-28
[HOTFIX] Removed use of Filter.type property

wbinventor:
This minimal PR fixes a few bugs where the old `Filter.type` property was still used by `openmc.filter`. In addition, the MGXS Part II Jupyter notebook is fixed to use the NNDC HDF5 data to plot continuous energy and multi-group cross sections.


nelsonag:
One quick note: the new plot in the MGXS Part II notebook should now use
the plotting capability added in #757.  Instead of having you re-do that
now, how about you remove these changes here and instead I include it in a
future PR which will add MGXS plotting capability just like the CE plotting
added in #757.  The interface will change *slightly* with that PR anyways
(i.e., 'cross_sections' will become `ce_cross_sections` - admittedly you
would leave these as default for this notebook).  This PR will be submitted
once #758 is accepted.

wbinventor:
Good point @nelsonag, that example is an excellent spot to showcase your new plotting routines. I've reverted the notebook in lieu of an update from you in the near future.

openmc/filter.py
```diff
@@ -1324,13 +1323,11 @@ def get_pandas_dataframe(self, data_size, **kwargs):
...
         # them as necessary to account for other filters.
         lo_bins = np.repeat(self.bins[:-1], self.stride)
-        hi_bins = np.repeat(self.bins[1:], self.stride)
         tile_factor = data_size / len(lo_bins)
         lo_bins = np.tile(lo_bins, tile_factor)
-        hi_bins = np.tile(hi_bins, tile_factor)
 
         # Add the new angle columns to the DataFrame.
-        df.loc[:, self.type + ' low'] = lo_bins
+        df.loc[:, 'polar low'] = lo_bins
```
smharper: It looks like an oversight that there are no `polar high` and `azimuthal high` columns.  Please add those the the DataFrame

openmc/filter.py
```diff
@@ -1324,13 +1323,14 @@ def get_pandas_dataframe(self, data_size, **kwargs):
         # Extract the lower and upper angle bounds, then repeat and tile
         # them as necessary to account for other filters.
         lo_bins = np.repeat(self.bins[:-1], self.stride)
-        hi_bins = np.repeat(self.bins[1:], self.stride)
+        hi_bins = np.repeast(self.bins[1:], self.stride)
```
smharper: Typo `repeast -> repeat`

smharper:
Any reason @paulromano is also assigned to this PR?  I'm happy to merge after that typo fix unless you also want him to take a look.

wbinventor:
Fixed the typo. No reason @paulromano has to necessarily review this, I just included you both since I wasn't sure how quickly either of you would get around to it.

PR #763, opened 2016-11-28
Track Density-Weighted Average MGXS

wbinventor:
This PR is a re-opened version of #732. This PR fixes an issue in the `MGXS.get_subdomain_avg_xs(...)` method, among a few other minor bug fixes. Currently, this method will take the geometric average of the MGXS for a bunch of distribcells or mesh cells. However, the geometric average does not preserve global reactivity.

Instead, a _track density-weighted (TD-weighted)_ average MGXS should be computed. Fortunately, this is as simple as substituting `Tally.summation(...)` for `Tally.average(...)` in the `get_subdomain_avg_xs(...)` method. In effect, this computes the  average MGXS from tally sums which are equivalent to what one would obtain if an explicit cell or material tally were specified for the subdomains of interest. And as my thesis broadly confirms, the TD-weighted average MGXS does indeed preserve global reactivity.

I made these changes a few months ago, and ran into some issues when merging with the develop branch earlier today due to #723 and #725. As a result, this PR is likely not ready for merging - I'm not yet sure I've fixed all of the issues. I haven't been able to run my scripts which use TD-weighted MGXS since something with the new `MeshFilter` is breaking when loading from a statepoint. I'll post here again once I have a chance to sort that out.


wbinventor:
This PR is ready for review since it is now up-to-date and the tests should all pass.

smharper:
There are several places where you replaced `np.arange` with `[i for i in range(...)]`.  Any particular reason why?

wbinventor:
@smharper I switched to using list comprehensions several months ago to resolve a very subtle bug I came across. To be honest, I can't recall what the bug was exactly. I do know that it had something to do with incorrect slicing of NumPy arrays with the output of `np.arange`, which I only noticed when doing specialized subdomain averaging over distribcell MGXS with the code for my thesis.

smharper:
Fair enough.  No comments on any of the code.  Merging!

PR #779, opened 2017-01-18
Calculate atomic weight with openmc.data

wbinventor:
This PR introduces a new `openmc.data.atomic_weight(element)` method to compute the atomic weight of naturally occurring elements.

### DIFF
--- a/openmc/data/data.py
+++ b/openmc/data/data.py
@@ -200,7 +200,7 @@ def atomic_weight(element):
     """
     weight = 0.
     for nuclide, abundance in NATURAL_ABUNDANCE.items():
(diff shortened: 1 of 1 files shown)


openmc/data/data.py
```diff
@@ -179,6 +179,32 @@ def atomic_mass(isotope):
...
+    Returns
+    -------
+    float or None
+        Atomic weight of element in atomic mass units. If the element listed does
+        not exist, None is returned.
+
+    """
+    weight = 0.
+    for nuclide, abundance in NATURAL_ABUNDANCE.items():
+        if re.match(element, nuclide):
```
wbinventor: Done

PR #782, opened 2017-02-07
mesh.Mesh.__eq__()

tjlaboss:
In `openmc/mesh.py`, the method `Mesh.__eq__()` contains the following statement:

`elif self._lower_left != mesh2.lower_left and self._upper_right != mesh2.upper_right:`

This results in a ValueError: "The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()". Using `.all()` for the comparison of these lists fixes the error.


openmc/mesh.py
```diff
@@ -75,8 +75,8 @@ def __eq__(self, mesh2):
             return False
 
         # Check lower left / upper right
-        elif self._lower_left != mesh2.lower_left and \
-             self._upper_right != mesh2.upper_right:
+        elif (self._lower_left != mesh2.lower_left).all() and \
```
wbinventor: Can you apply this same fix to the comparisons of the `width` and `dimension` attributes above?

nelsonag: What about using the EqualityMixin class?  (See the openmc.data package)

tjlaboss: @wbinventor Done.

nelsonag: What about using the EqualityMixin class?  (See the openmc.data package)

nelsonag: (oops)

wbinventor: @nelsonag my understanding is that EqualityMixin has been mostly used in the data submodule, while the array/list comparisons throughout the rest of the API use this more standard approach. Do you believe we should eventually convert everything over to use EqualityMixin? If so, what would be the benefit?

nelsonag:
I view there is no reason it should be only in OpenMC.data and not
elsewhere. The benefit would simply be avoiding the definition of __eq__
and __ne__ everywhere. At the same time in not sure I see a downside to
using it either.

paulromano:
Completely agree with @nelsonag. I don't think there was ever an intention that the mixin only be used in openmc.data since it is generally applicable.

wbinventor:
@nelsonag I see, I didn't notice `EqualityMixin` take over the API last fall! Very nice.

@tjlaboss can you make the `openmc.Mesh` class subclass from the `openmc.mixin.EqualityMixin` class? Then you should be able to remove the `__eq__` method altogether.

tjlaboss:
@nelsonag , @wbinventor : Done, see Pull Req. https://github.com/mit-crpg/openmc/pull/784

PR #778, opened 2017-01-16
Incorporation of angle-dependent MGXS capabilities in `openmc.MGXS`

nelsonag:
This PR adds the necessary coding to `openmc.MGXS` so that it can be used to create polar and azimuthal angle-dependent multi-group cross sections.  This checks off the last box in Issue #597, allowing us to close that issue when this is merged [finally].

A few bug fixes encountered in the testing of this code are also included 1) the multiplicity matrix wasn't being read correctly from H5 files, and 2) the mu chosen from a histogram scattering representation was off by one mu bin.

PR #787, opened 2017-02-16
Modified openmc.MGXS to allow users to convert between representations and scattering formats

nelsonag:
This PR started as a sanity check to make sure that the options for data representations in OpenMC produce the consistent results.  Doing this required modifying the `openmc.MGXS` API to allow for converting between angle and isotropic data representations, as well as converting between the histogram, tabular, and Legendre scattering formats.  

Well, good news, it all matches very tightly! The eigenvalues are pretty much dead on, and the images of the distributions are shown below.  These figures take P2 data, convert to Tabular data with 33 points, and then convert that tabular data to a 32-bin histogram.
p1 - t33

p1 - t33 - h32

Anyways, I'm submitting this PR because 1) I will probably incorporate this functionality in to some tests in the future(to ensure this current consistency sticks), and 2) I thought this would be useful for the users in the pursuit of exploring data representations.

I also want to ask reviewers if they think it would be worthwhile to just rely on this functionality for the users to convert Legendre to tabular representations instead of the code in the initialization routines of `mgxs_header.F90`which converts from Legendre to Tabular (by default).  

On a final note, I also have a script to convert from MGXS to CE data and it actually gets answers pretty close, but I don't quite have the time right now to get it all the way.  Since I can't see a user doing that I dont want to make much investment in it.  But if someone wants it, let me know and I'll send it.


openmc/mgxs/mgxs.py
```diff
@@ -252,7 +252,7 @@ def __deepcopy__(self, memo):
             clone._name = self.name
             clone._rxn_type = self.rxn_type
             clone._by_nuclide = self.by_nuclide
-            clone._nuclides = copy.deepcopy(self._nuclides)
+            clone._nuclides = copy.deepcopy(self._nuclides, memo)
```
wbinventor: Is this necessary in the event that `_nuclides` is a list of `Nulicde` objects instead of strings?

openmc/mgxs_library.py
```diff
@@ -14,9 +14,13 @@
 
 # Supported incoming particle MGXS angular treatment representations
 _REPRESENTATIONS = ['isotropic', 'angle']
+# Supported scattering angular distribution representations
 _SCATTER_TYPES = ['tabular', 'legendre', 'histogram']
+# List of MGXS dimension types
```
wbinventor: how about "indexing schemes" in place of "dimension types"?

wbinventor: Also, add a blank line in between each of these lists.

openmc/mgxs_library.py
```diff
@@ -348,7 +398,7 @@ def num_delayed_groups(self, num_delayed_groups):
     @representation.setter
     def representation(self, representation):
 
-        # Check it is of valid type.
+        # Check it is of valid value.
```
wbinventor: Might consider removing this comment altogether, I think it is self explanatory.

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
         self._inverse_velocity[i] = inverse_velocity.get_xs(
             nuclides=nuclide, xs_type=xs_type, subdomains=subdomain)
 
+    def convert_representation(self, target_representation, num_polar=None,
+                               num_azimuthal=None):
+        """Produce a new XSdata object with the same data, but converted to the
+        new representation
```
wbinventor: perhaps clarify as "new angular distribution representation"

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+
+        xsdata = copy.deepcopy(self)
+
+        # First handle the case where the current and requested
+        # representations are the same
+        if target_representation == self.representation:
+            # Check to make sure the num_polar and num_azimuthal values match
+            if target_representation == 'angle':
+                if num_polar != self.num_polar or num_azimuthal != self.num_azimuthal:
+                    raise NotImplementedError("XCannot translate between "
```
wbinventor: Is this something you plan to implement? Or for which you see a reasonable way to implement? If not, might consider turning this into a `ValueError` or something else.

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+                       'prompt_nu_fission', 'delayed_nu_fission',
+                       'kappa_fission', 'chi', 'chi_prompt', 'chi_delayed',
+                       'beta', 'decay_rate', 'inverse_velocity']:
+                # Get the original data
+                orig_data = getattr(self, '_' + xs)[i]
+                if orig_data is not None:
+                    if target_representation == 'isotropic':
+                        # Since we are going from angle to isotropic, the
+                        # current data is just the average over the angle bins
+                        new_data = orig_data.mean(axis=(0, 1))
```
wbinventor: I don't think this will conserve reaction rates since it isn't flux-averaged, correct? I'm not opposed to you using a simple arithmetic average here, but you should clearly state the caveat in the docstring that you are not using flux-weighting and hence reaction rates / reactivity will not be conserved.

wbinventor: The `openmg.mgxs.MGXS.get_condensed_xs(...)` method is designed to conserve reaction rates by collapsing both the reaction rate and flux tallies separately, and the combining them to compute the coarse group constants. You might consider doing the same if the angle-dependent constants were computed from `openmc.mgxs`.

openmc/mgxs_library.py
```diff
@@ -2099,6 +2429,107 @@ def get_by_name(self, name):
                 result = xsdata
         return result
 
+    def convert_representation(self, target_representation, num_polar=None,
+                               num_azimuthal=None):
+        """Produce a new MGXSLibrary object with the same data, but converted
+        to the new representation
```
wbinventor: ditto

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+        else:
+            check_greater_than('target_order', target_order, 0)
+
+        xsdata = copy.deepcopy(self)
+        xsdata.scatter_format = target_format
+        xsdata.order = target_order
+
+        # Reset and re-generate XSdata.xs_shapes with the new scattering format
+        xsdata._xs_shapes = None
+        xsdata.xs_shapes
```
wbinventor: Is this necessary?

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+                                np.sum((l + 0.5) *
+                                       eval_legendre(l, mu_fine[imu]) *
+                                       orig_data[..., l]
+                                       for l in range(self.num_orders))
+                        new_data[..., h_bin] = simps(table_fine, mu_fine)
+
+                    # Remove the very small values resulting from numerical
+                    # precision issues
+                    new_data[..., np.abs(new_data) < 1.E-10] = 0.
+
```
wbinventor: Add blank lines in between each block in this massive `for-if-else` code block.

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+        xsdata = copy.deepcopy(self)
+        xsdata.scatter_format = target_format
+        xsdata.order = target_order
+
+        # Reset and re-generate XSdata.xs_shapes with the new scattering format
+        xsdata._xs_shapes = None
+        xsdata.xs_shapes
+
+        for i, temp in enumerate(xsdata.temperatures):
```
wbinventor: Can this `for` loop over temperatures instead be performed implicitly with NumPy/SciPy vectorization?

openmc/mgxs_library.py
```diff
@@ -1634,6 +1685,267 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+
+        Returns
+        -------
+        openmc.XSdata
+            Multi-group cross section data with the same data as in self, but
+            represented as specified in :param:`target_format`.
+
+        """
+
+        from scipy.interpolate import interp1d
```
wbinventor: I haven't yet fully reviewed this method, but I'm curious if it might be made a bit more modular (and digestible, maintainable and extensible) by replacing it with `to_tabular(...)`, `to_legendre(...)` and `to_histogram(...)` that collectively provide the same functionality?

openmc/mgxs_library.py
```diff
@@ -2099,6 +2429,107 @@ def get_by_name(self, name):
...
+
+        Returns
+        -------
+        openmc.MGXSLibrary
+            Multi-group Library with the same data as self, but represented as
+            specified in :param:`target_representation`.
+
+        """
+
+        check_value('target_representation', target_representation,
```
wbinventor: Can lines 2459-2465 be removed since the`XSdata.convert_representation(...)` method also performs the same error checking?

openmc/mgxs_library.py
```diff
@@ -2099,6 +2429,107 @@ def get_by_name(self, name):
...
+        Returns
+        -------
+        openmc.MGXSLibrary
+            Multi-group Library with the same data as self, but with the
+            scatter format represented as specified in :param:`target_format`
+            and :param:`target_order`.
+
+        """
+
+        check_value('target_format', target_format, _SCATTER_TYPES)
```
wbinventor: Ditto, consider removing lines 2496-2501.

openmc/mgxs_library.py
```diff
@@ -2099,6 +2429,107 @@ def get_by_name(self, name):
...
+            library.xsdatas[i] = \
+                xsdata.convert_scatter_format(target_format, target_order)
+
+        return library
+
+    def convert_to_continuous_energy(self, h5_filename='ce_mgxs.h5',
+                                     library_filename='ce_mgxs.xml'):
+        """Converts the MGXSLibrary object to an equivalent
+        library of openmc.data.IncidentNeutron objects
+
```
wbinventor: If you are going to introduce this, perhaps add a more detailed explanatory paragraph to the docstring detailing how this can be supplied to a CE MC simulation.

openmc/mgxs_library.py
```diff
@@ -2099,6 +2429,107 @@ def get_by_name(self, name):
...
+        Parameters
+        ----------
+        h5_filename : str
+            HDF5 file to write with all the files
+        library_filename : str
+            cross_sections.xml file describing the HDF5 file
+
+        """
+
```
wbinventor: Add error checking for `h5_filename` and `library_filename`

wbinventor:
Also to your PR summary, I'm not particularly keen on accepting lots of new features and functions without accompanying documentation and/or examples and/or tests (even if only minimal). The angle-dependent MGXS introduced in #778, let alone MG mode in general, are not thoroughly documented or tested right now, so I'd be hesitant to merge this PR without some new examples and/or tests. That said, @paulromano and @smharper probably have more relevant opinions on this than I do at this point.

openmc/mgxs/mgxs.py
nelsonag: I agree, no, a standard assignment should be fine

-[x]

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: well it can be either angular or isotropic and this refers to the MGXS representation not really the scattering format.  either way I clarified a little bit.

- [x]

openmc/mgxs_library.py
nelsonag: I dont personally have plans to do this, but there is no reason one could not do it by assuming a histogram representation and using that to convert to a new grid.

Either way, I'm fine with `ValueError`

- [x]

openmc/mgxs_library.py
nelsonag: I am not doing flux-weighted anything here, right. That data isn't available here in the data library itself (vice `openmc.mgxs.Library` object where it is available).  I explain in the docstring now to be clear.

I can implement a function similar to `get_condensed_xs(...)`, though I would prefer to wait until a need is noted.

openmc/mgxs_library.py
nelsonag: I agree they can be separated but I'm not sure the benefit is worth the effort since everything is pretty well segregated in to its own if/else block.

openmc/mgxs_library.py
nelsonag: Not with the current format - the xs are lists of numpy arrays where the index of the list is temperatures.  This is just like the CE data.

openmc/mgxs_library.py
nelsonag: No, actually. its unnecessary i just checked.  I swore it was required before....

openmc/mgxs_library.py

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: Agree, it can be.

openmc/mgxs_library.py

openmc/mgxs_library.py
nelsonag: meh, its gone unless i hear otherwise (but if it comes back, thats a good idea)

openmc/mgxs/mgxs.py
wbinventor: I wasn't questioning your implementation, just was curious if that was the reasoning. I think it should probably remain as is.

openmc/mgxs_library.py
wbinventor: Yeah I don't see a reason to implement flux-weighting unless or until a practical use case arises where this is needed to ensure accurate downstream multi-group calculations.

openmc/mgxs_library.py
wbinventor: Ok vectorization isn't necessary for this PR of course. But my NumPy-ify everything OCD nature thinks it might be worth considering a single NumPy array for all temperatures in the future :-)

openmc/mgxs_library.py
nelsonag: OBE

nelsonag:
@wbinventor a few things:
- What about MG mode in general isn't documented or tested?  Do you mean the Python side or the Fortran side?  
- I added a test of these features. It runs through *all* the conversion paths,  runs them, and stores the keff.

nelsonag:
The test suite is now butting up against the time limit with these tests.  I can reduce the number of neutrons to be statistically non-informative, but that is something that bothers me in general.  Any opinions on the value of testing all possible paths through this new python code vs reducing histories to very small numbers?

openmc/mgxs_library.py
```diff
@@ -1634,6 +1681,286 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+    def convert_representation(self, target_representation, num_polar=None,
+                               num_azimuthal=None):
+        """Produce a new XSdata object with the same data, but converted to the
+        new representation (isotropic or angle-dependent).
+
+        This method cannot be used to change the number of polar or
+        azimuthal bins of an XSdata object that already uses an angular
+        representation. Finally, this method simply uses an arithmetic mean to
+        convert from an angular to isotropic representation; no flux-weighting
+        is applied and therefore the correctness of the solution is not
```
wbinventor: Considering making this statement of caution a tad bit stronger: "correctness of the solution is not guaranteed" -> "reactions rates will not be preserved".

openmc/mgxs_library.py
```diff
@@ -2099,6 +2444,76 @@ def get_by_name(self, name):
...
```
wbinventor: ditto

tests/test_mg_convert/test_mg_convert.py
```diff
@@ -0,0 +1,206 @@
...
+        # Create an initial uniform spatial source distribution
+        bounds = [-5, -5, -5, 5, 5, 5]
+        uniform_dist = openmc.stats.Box(bounds[:3], bounds[3:])
+        settings_file.source = openmc.source.Source(space=uniform_dist)
+
+        settings_file.export_to_xml()
+
+    def _run_openmc(self):
+        # Run multiple conversions to compare results
+        cases = [['legendre', 2], ['legendre', 0],
```
wbinventor: I really like the way you designed this test to parametrically explore all of the conversion pathways - it's too bad we don't have more spare time on our Travis build! I wouldn't be opposed to commenting out a subset of the options below to relax the runtime requirements for this test.

wbinventor:
@nelsonag I'm no longer so queasy adding these features since your new test ensures that the code "works."

As to your question, the various MG features are thoroughly documented with respect to the XML tags and Python API docstrings. However, I think that without explicit examples, the new advanced features in the PyAPI won't be used (or maintained or even understood) except by super users (perhaps I'm wrong). These features could be expounded upon within a few new Jupyter notebooks, though I'm not sure how large we want to allow our collection of notebooks to expand.

openmc/mgxs_library.py

openmc/mgxs_library.py

nelsonag:
@wbinventor I didn't have the heart to cut any of the cases in the test; the poor little ones did nothng wrong.  Instead, I ended up bringing the histories down to the same as any other test.

nelsonag:
@wbinventor I will open an issue to create example notebooks better showcasing MG-mode.  I'll keep it separate from here simply to keep this in a smaller digestable chunk

wbinventor:
Thanks @nelsonag. I still need to thoroughly review the logic of your code to convert scattering representations. I'll do so later today.

openmc/mgxs_library.py
```diff
@@ -1634,6 +1681,285 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+                    # Like the tabular -> legendre path above, this code will
+                    # be written to utilize the vectorized integration
+                    # capabilities instead of having an isotropic and
+                    # angle representation path.
+                    for h_bin in range(xsdata.num_orders):
+                        mu_fine = np.linspace(mu[h_bin], mu[h_bin + 1], _NMU)
+                        new_data[..., h_bin] = \
+                            norm * simps(interp(mu_fine), mu_fine)
+
+                    # Remove the very small values resulting from numerical
```
wbinventor: This line of code is repeated four times in different blocks. Could it instead be applied to `new_data` once at the very end, regardless of the self and target representations?

openmc/mgxs_library.py
```diff
@@ -1634,6 +1681,285 @@ def set_inverse_velocity_mgxs(self, inverse_velocity, temperature=294.,
...
+
+                elif target_format == 'tabular':
+                    mu = np.linspace(-1, 1, xsdata.num_orders)
+                    # Evaluate the legendre on the mu grid
+                    for imu in range(len(mu)):
+                        new_data[..., imu] = \
+                            np.sum((l + 0.5) * eval_legendre(l, mu[imu]) *
+                                   orig_data[..., l]
+                                   for l in range(self.num_orders))
+                elif target_format == 'histogram':
```
wbinventor: Add a blank line between `if-else` blocks here to correspond to the convention below (and to make more readable).

openmc/mgxs_library.py

nelsonag:
Thanks @wbinventor, all set!

nelsonag:
Nnnnooo it's not

wbinventor:
Looks like Travis failed. I'll merge once it passes.

nelsonag:
This one looks like it will work

wbinventor:
Good to go! Nice PR @nelsonag!

PR #793, opened 2017-02-19
Consistent Multi-Group Scattering Matrices

wbinventor:
This PR introduces a new formulation for computing multi-group scattering matrices with the `openmc.mgxs` module. These changes attempt to eliminate reaction rate imbalance between total MGXS and scattering matrices. This work addresses instabilities observed by @geogunow with CMFD and linear/quadratic source approximations in OpenMOC when solving 3D problems.

Presently, scattering matrices are computed out of necessity with analog tally estimators, while the total cross section is computed with a track-length estimator (by default). As a result of this mixture of tally estimators, neutron balance will not be conserved since it is not necessarily true that the total MGXS is equal to the sum of the absorption XS and the scattering matrix in each energy group. This is particularly relevant for void (*e.g.*, helium gap) regions with miniscule cross sections, where the statistical inefficiency of the analog tally estimator used for the scattering matrix is particularly problematic for some applications (*i.e.*, CMFD + linear source MOC).

This PR addresses this with a new formulation of the scattering matrix which is computed from the product of the (a) scattering cross section and the (b) group-to-group scattering probability matrix. Reaction rates are exactly preserved (with the total cross section) since (a) is computed using a tracklength estimator, and simply convolved with (b). While (b) is necessarily computed with an analog estimator, it is normalized to unity across outgoing energy groups which ensures that the product (a) x (b) preserves neutron balance with the total and absorption cross sections.

From a design perspective, this PR introduces the following new classes to the `openmc.mgxs` module:

* `ConvolvedMGXS` (abstract)
* `ScatteringProbabilityMatrix`
* `NuScatteringProbabilityMatrix`
* `ConsistentScatteringMatrixXS`
* `ConsisistentNuScatteringMatrixXS`

I introduced the new `ConvolvedMGXS` class in `openmc.mgxs`, which is simply a `MGXS` computed as a convolution of two or more "sub"-`MGXS` objects. While this may have been overkill, I thought it would provide a platform to build upon in the future, though I am certainly open to alternative ideas. The new (`Nu`)`ScatteringProbabilityMatrix` class encapsulates the group-to-group scattering probabilities. The `ConsistentScatteringMatrixXS` class defines the new scattering formulation as a product of a `ScatterXS` (tracklength) and `ScatterProbabilityMatrix` (analog) objects.

The current scattering matrix formulation and class were left intact. Note that the new classes do not (yet) support Legendre moments greater than zero. The new classes do support the use of a transport correction just as do the current scattering matrices.

Although @geogunow has yet to thoroughly evaluate this new scattering matrix in OpenMOC, I have done some simple sanity checks to confirm that it is roughly consistent with the current (original) scattering matrix formulation. For example, the following plot shows the discrepancy/imbalance between the total, absorption and scattering matrix cross sections as a function of realization for the current and new approaches (for a helium gap in a PWR fuel pin in the fast group). As expected, the imbalance for the new approach is very nearly zero (to within machine precision), while the current approach exhibits an imbalance that slowly converges to zero in the limit of infinite realizations.

It is worth noting that the new formulation does *not* accelerate the convergence of the scattering matrix statistical uncertainties. Indeed, the uncertainties are roughly the same as those for the current scattering matrix as shown below:

@nelsonag would you be interested in reciprocating and reviewing this PR as you find the time?


nelsonag:
@wbinventor great move. Yes I'll review. This is something that should take
some of the low hanging variance "fruit" that my dissertation beat down
with brute force.  I also love that you did something like this without
having to touch Fortran!

My initial thoughts, without having looked at any code, so apply salt, are:
- once this is proven out, including a FOM, it seems like it should become
default by openmc.mgxs.Library when a (nu)scatter matrix is chosen
- could this be implemented as an option on OpenMC.mgxs.Library instead of
a new mgxs_type? Maybe there isn't a benefit of that besides allowing for
it to be a default.
- the nu part of nu-scatter is still analog.  Did you make a similar plot
to those in the PR for nu-scatter? Does it look as good? It may be that the
additional variance due to the multiplication isn't that big, so it all
works just fine.  What you've done is obviously in the direction of
goodness, but part of me thinks that most users would want nu-scattering
and thus we should have an idea if the benefit is there or not.

wbinventor:
@nelsonag thanks for the quick comments last night. While these changes certainly don't even approach the level of rigor you went to in your thesis and with NDPP, they at least provide a "quick fix" to a related problem. And as for Fortran, well I think my love for all things Python has been readily apparent for some time now :-) I've tried to answer your initial questions below:

**Default scattering matrix**
I had thought about this as well, but decided against it for now since I thought it would be best for the default scattering matrix formulation to support higher order Legendre moments. The consistent formulation unfortunately can't support higher order moments since these scores require analog tallies. However, whether to make `ScatterMatrixXS` or `ConsistentScatterMatrixXS` the default is simply a value judgement. If you differ with my reasoning then I'm happy to revise accordingly.

**A `Library` option**
If we decide to support both the current and new formulations, then I agree that it would be nice to toggle this on from the `Library`. However, I didn't implement this since it seemed like it would interfere with the generality of the `MGXS_TYPES` and `MGXS.get_mgxs(...)` method.

**Nu is still analog**
Very good point, I actually forgot that `NuScatterXS` uses an analog estimator by default. To resolve, this I pushed some changes this morning which makes the new `ConsistentNuScatterMatrixXS` a product of (1) `ScatterXS` (tracklength), (2) `NuScatterProbabilityMatrix` (analog) and (3) `MultiplicityMatrixXS` (analog). As shown below, this new formulation exhibits the same (consistent) behavior as what I already showed for the `ConsistentScatterMatrixXS`. Note that this plot is for the fast group for the helium gap in a PWR fuel pin which has a negligible multiplicity, and hence neutron balance (total - absorb - scatter = 0) should be exactly preserved.

wbinventor:
Also, I've updated the MGXS tests to include results from the following new `MGXS` subclasses:

* `ScatterProbabilityMatrix`
* `NuScatterProbabilityMatrix`
* `ConsistentScatterMatrixXS`
* `ConsistentNuScatterMatrixXS`

This should resolve the Travis failures.

wbinventor:
It's unclear to me why Travis is still failing, since the tests all pass for me on both my workstation and laptop.

docs/source/pythonapi/index.rst
```diff
@@ -127,6 +127,7 @@ respectively.
    openmc.get_hexagonal_prism
    openmc.get_rectangular_prism
 
+>>>>>>> upstream/develop
```
paulromano: Merge crud leftover

openmc/mgxs/mgxs.py
```diff
@@ -3611,7 +3623,6 @@ def __init__(self, domain=None, domain_type=None,
                                         num_azimuthal)
         self._rxn_type = 'scatter'
 
-
```
paulromano: Classes should be separated by two lines

openmc/mgxs/mgxs.py
```diff
@@ -4891,36 +4902,42 @@ def xs_tally(self):
...
     necessary data to compute multi-group cross sections from a
     :class:`openmc.StatePoint` instance. The derived multi-group cross section
-    can then be obtained from the :attr:`NuFissionMatrixXS.xs_tally` property.
+    can then be obtained from the :attr:`ScatterProbabilityMatrix.xs_tally`
+    property.
 
     For a spatial domain :math:`V`, incoming energy group
     :math:`[E_{g'},E_{g'-1}]`, and outgoing energy group :math:`[E_g,E_{g-1}]`,
-    the fission production is calculated as:
+    the group-to-group scatterin probabilities are calculated as:
```
paulromano: scattering

openmc/mgxs/mgxs.py
```diff
@@ -4974,7 +4991,7 @@ class NuFissionMatrixXS(MatrixMGXS):
         The tally estimator used to compute the multi-group cross section
     tallies : collections.OrderedDict
         OpenMC tallies needed to compute the multi-group cross section. The keys
-        are strings listed in the :attr:`NuFissionMatrixXS.tally_keys`
+        are strings listed in the :attr:`MultiplicityMatrixXS.tally_keys`
```
paulromano: Looks like this should be ScatterProbabilityMatrix

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+            self._rxn_rate_tally = self.tallies['scatter']
+            self._rxn_rate_tally.sparse = self.sparse
+        return self._rxn_rate_tally
+
+    @property
+    def xs_tally(self):
+
+        if self._xs_tally is None:
+            energyout_bins = \
+                [self.energy_groups.get_group_bounds(i) for i in range(self.num_groups, 0, -1)]
```
paulromano: This would look better as:
```Python
energyout_bins = [self.energy_groups.get_group_bounds(i)
                  for i in range(self.num_groups, 0, -1)]
```

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+    minimum, one needs to set the :attr:`NuScatterProbabilityMatrix.energy_groups`
+    and :attr:`NuScatterProbabilityMatrix.domain` properties. Tallies for the
+    appropriate reaction rates over the specified domain are generated
+    automatically via the :attr:`NuScatterProbabilityMatrix.tallies` property,
+    which can then be appended to a :class:`openmc.Tallies` instance.
+
+    For post-processing, the :meth:`MGXS.load_from_statepoint` will pull in the
+    necessary data to compute multi-group cross sections from a
+    :class:`openmc.StatePoint` instance. The derived multi-group cross section
+    can then be obtained from the :attr:`ScatterProbabilityMatrix.xs_tally`
```
paulromano: Missing Nu here

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+@add_metaclass(ABCMeta)
+class ConvolvedMGXS(MGXS):
+    """An abstract convolution of multiple multi-group cross sections for some
+    energy group structure within some spatial domain.
+
+    This class can be used for both OpenMC input generation and tally data
+    post-processing to compute spatially-homogenized and energy-integrated
+    multi-group cross sections for multi-group neutronics calculations.
+
+    NOTE: Users should instantiate the subclasses of this abstract class.
```
paulromano: Use note directive, as in:
```rst
.. note:: Users should...
```

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+        for mgxs in self.mgxs:
+            filters.extend(mgxs.filters)
+        return filters
+
+    @property
+    def tally_keys(self):
+        keys = []
+        for mgxs in self.mgxs:
+            for tally_key in mgxs.tallies:
+                keys += ['{} : {}'.format(mgxs.hdf5_key, tally_key)]
```
paulromano: Use `keys.append(...)` rather than adding a list

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+    def nuclides(self, nuclides):
+        cv.check_iterable_type('nuclides', nuclides, string_types)
+
+        self._nuclides = nuclides
+        for mgxs in self.mgxs:
+            mgxs.nuclides = nuclides
+
+    @MGXS.estimator.setter
+    def estimator(self, estimator):
+        raise ValueError('Unable to assign a tally estimator a ConvolvedMGXS')
```
paulromano: Unable to assign a tally estimator *to* a ConvolvedMGXS?

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+    .. math::
+
+       \langle \sigma_{s,g'\rightarrow g} \phi \rangle &= \int_{r \in V} dr
+       \int_{4\pi} d\Omega' \int_{E_{g'}}^{E_{g'-1}} dE' \int_{4\pi} d\Omega
+       \int_{E_g}^{E_{g-1}} dE \; \sigma_s (r, E'
+       \rightarrow E, \Omega' \cdot \Omega) \psi(r, E', \Omega')\\
+       \langle \phi \rangle &= \int_{r \in V} dr \int_{4\pi} d\Omega
+       \int_{E_g}^{E_{g-1}} dE \; \psi (r, E, \Omega) \\
+       \sigma_{s,g'\rightarrow g} &= \frac{\langle
+       \sigma_{s,,g'\rightarrow g} \phi \rangle}{\langle \phi \rangle}
```
paulromano: Extra comma here

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+
+    For post-processing, the :meth:`MGXS.load_from_statepoint` will pull in the
+    necessary data to compute multi-group cross sections from a
+    :class:`openmc.StatePoint` instance. The derived multi-group cross section
+    can then be obtained from the :attr:`ConsistentScatterMatrixXS.xs_tally`
+    property.
+
+    For a spatial domain :math:`V`, incoming energy group
+    :math:`[E_{g'},E_{g'-1}]`, and outgoing energy group :math:`[E_g,E_{g-1}]`,
+    the Legendre scattering moments are calculated as:
```
paulromano: No Legendre moments yet, right?

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+
+    .. math::
+
+       \langle \sigma_{s,g'\rightarrow g} \phi \rangle &= \int_{r \in V} dr
+       \int_{4\pi} d\Omega' \int_{E_{g'}}^{E_{g'-1}} dE' \int_{4\pi} d\Omega
+       \int_{E_g}^{E_{g-1}} dE \; \sigma_s (r, E'
+       \rightarrow E, \Omega' \cdot \Omega) \psi(r, E', \Omega')\\
+       \langle \phi \rangle &= \int_{r \in V} dr \int_{4\pi} d\Omega
+       \int_{E_g}^{E_{g-1}} dE \; \psi (r, E, \Omega) \\
```
paulromano: Shouldn't it be E_{g'} and E_{g'-1}?

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+                    self._tallies['scatter-1'].nuclides.append('total')
+
+        return self._tallies
+
+    @property
+    def scatter_xs(self):
+        return self.mgxs[0]
+
+    @property
+    def probability_matrix(self):
```
paulromano: This property is not documented in the docstring (same goes for scatter_xs)

openmc/mgxs/mgxs.py
paulromano: Also, these equations don't really demonstrate that the group-to-group scattering xs is calculated as a product of the g' scattering rate and the probability of g' -> g.

openmc/mgxs/mgxs.py
```diff
@@ -5011,13 +5028,1153 @@ class NuFissionMatrixXS(MatrixMGXS):
...
+                if self.domain_type == 'mesh':
+                    domain_filter = filter_type(self.domain)
+                else:
+                    domain_filter = filter_type(self.domain.id)
+
+                # Create each Tally needed to compute the multi group cross section
+                self._tallies['scatter-1'] = openmc.Tally(name=self.name)
+                self._tallies['scatter-1'].estimator = 'analog'
+                self._tallies['scatter-1'].scores = [self.scores[-1]]
+                self._tallies['scatter-1'].filters = [domain_filter, *self.filters[-1]]
```
paulromano: This is the line that is causing Travis to fail. If this is really what you intended, this type of argument unpacking didn't become acceptable until Python 3.5.

wbinventor:
To be fair, this was Kord's idea based on his discussions with @geogunow about the impact that small reaction rate imbalances can have on MOC linear source and CMFD solvers.

I didn't implement higher order Legendre moments for this new formulation since we do not support tracklength tally estimators for `'scatter-N'` scores for N > 0. I don't understand why this is the case, and haven't looked at the Fortran source to determine the reason. Do you want to enlighten me @paulromano?

openmc/mgxs/mgxs.py
wbinventor: Good catch, fixed this here and elsewhere

openmc/mgxs/mgxs.py
wbinventor: Fixed this here and elsewhere

openmc/mgxs/mgxs.py
wbinventor: Fixed this here and elsewhere

openmc/mgxs/mgxs.py
wbinventor: Well I didn't think it made sense to allow for this since each of the "sub"-`MGXS` in the convolution may accept different estimators. This is indeed the case for both subclasses of `ConvolvedMGXS` implemented here - `ScatterProbabilityMatrixXS` is a convolution of `ScatterXS` and `ScatterProbabilityMatrixXS` which support analog or tracklength and analog only estimators, respectively. Would you prefer that `ConvolvedMGXS` attempt to assign the tally estimators to each "sub"-`MGXS` to occur at their level instead?

openmc/mgxs/mgxs.py
wbinventor: Nope, no Legendre moments. I left this in here for the moment since I was just hacking this together and didn't know if we might figure out a way to add in Legendre moments later. I'll remove this if we settle on simply supporting the first Legendre moment only.

openmc/mgxs/mgxs.py
paulromano: Oh no, I agree it doesn't make sense. I was just saying that you need a "to" to be grammatical correct, haha.

openmc/mgxs/mgxs.py
wbinventor: Yeah, these equations are identical to those in the docstrings for (`Nu`)`ScatterMatrixXS`, save for my removal of the Legendre moment subscript. I can re-write them as the product of the scattering rate and group-to-group probabilities if you would prefer to be completely transparent about how the calculation is performed.

paulromano:
Re: scatter-N, you need the scattering angle so it can only be done post-collision, and hence analog. I guess there's no obvious way to support Legendre moments unless we were to allow tracklength scatter-N tallies...

openmc/mgxs/mgxs.py
wbinventor: Nice catch!

openmc/mgxs/mgxs.py
wbinventor: Oh of course haha!

nelsonag:
I again admit to not having a chance to read the code, but isn't what
you're doing just the following?:

Sig_{s,"TL",g'->g}=Sig_{s,TL,g'->all} × Sig_{s,analog,g'->g} / sum_{g'}[
Sig_{s,analog,g'->g}]


If so, can't you recast for the following generic moment?:

Sig_{s,"TL",g'->g,l}=Sig_{s,TL,g'->all,0} × Sig_{s,analog,g'->g,l} /
sum_{g'}[Sig_{s,analog,g'->g,0}]

In either case you are scaling the analog estimator so it preserves the
total rate a TL estimator would give you, seems it can apply to the other
Legendres orders as well.

wbinventor:
@nelsonag I think you are right, I will try implementing that this morning. If this does indeed perform as expected, what do you and @paulromano think about altogether replacing the current (`Nu`)`ScatterMatrixXS` with the new formulation?

openmc/mgxs/mgxs.py
wbinventor: I don't believe so - it can be either E' or E, but we've used simply E in all other equations which are not dependent on the outgoing energy.

wbinventor:
Also @paulromano I do agree that "mgxs.py" has grown far too large (7000+ lines!). Much of this is of course the redundant docstrings. However, I agree that we should probably break it up into multiple individual files. Shall I leave this for a future PR?

nelsonag:
@wbinventor I think the current scatter matrix classes should stay.

The MG solver of OpenMC is "immune" to the effect you're trying to solve if
the Library is created correctly (and the way we do in the example notebook
on purpose).  If you provide total and absorption mgxs when generating the
library, then the value of the scattering xs doesn't matter, and it only
provides the energy-angle distribution given a scatter [come to think of
it, I do this exactly for the reasons you're trying to solve here]. So
while both my way (tallying total w/ a TL estimator) and your way (tallying
scatter w/ a TL estimator) have the same computational expense on the CE
side, the proposed method adds a less-than-obvious "hack" to the process
that impedes understanding and so id like to not leave it as the preferred
path.

Did that make sense?

nelsonag:
@wbinventor I'm still leery of the nu-scatter issue. This could be
resolved similar to what NDPP does by creating an array of xs-weighted nu
values as a function of incoming energy, then you don't need analog
tallying for nu-scatter.  This would incur a small error if the nu data is
not using the same interpolation type as the xs values.

@paulromano any recollection about the interpolation issue? Also don't we
do some sort of cheap pre-processing somewhere else, like for
kappa-fission? If not, any issue with adding this as supplemental data to
the hdf5 libraries?

openmc/mgxs/mgxs.py
paulromano: I'm talking about the limits of integration on energy. Written as is, you have the g' -> g rate divided by the group g flux instead of the group g' flux.

wbinventor:
@nelsonag I recognize that OpenMC's MG solver is immune to this for the reasons you stated, my I think that most folks who use `openmc.mgxs` are not generating MGXS for OpenMC but rather faster deterministic codes. Although some of these other codes may take the approach you mention, others do not - at least in the case of OpenMOC and some other MOC codes, the absorption XS isn't needed. I'm not sure about our friends at ANL, but it sounds like they have run into the same issue as we have with OpenMOC, so I assume that they don't back out the total scattering XS from the total XS and absorption XS.

By supporting both formulations, we'll make things confusing for users. I personally think it would be best to simplify and converge to a single scattering matrix implementation if possible (which appears to be the case with your suggested approach for Legendre moments). Nevertheless, I do recognize that this new formulation is a bit more convoluted than the current implementation. However, I think this is only true for developers and not for users, and I tend to be more concerned about users (we have quite a few more users than contributors to `openmc.mgxs` at MIT anyway).

Also, adding in this new feature has definitely exposed the challenges of extending `openmc.mgxs`. We will have of more redundant code blocks if we support both scattering formulations, and I don't see a straightforward method to get around this. Perhaps I just don't want to fall into the same trap that I've ranted and raved about on your recent PRs :-)

openmc/mgxs/mgxs.py
wbinventor: Oh of course that makes sense, sorry.

paulromano:
On one hand, I do feel that one should be able to get consistent cross sections by default. On the other hand, it's not immediately obvious to me that doing it with the new "consistent" classes will be more efficient than just using analog estimators. It seems that the scattering matrix (rates or probabilities) is the hardest thing to resolve, and that ends up being analog in either case. I'm not against making the new classes default, but it would be nice to see a real comparison between the two on FOM basis first.

paulromano:
@wbinventor Yes, I'd leave breaking up the mgxs.py module for a separate PR.

wbinventor:
@paulromano you are exactly right that the scattering probability matrix is "the hardest thing to resolve." This is what my second plot showed - the statistical uncertainties for this new formulation are basically the same as those for the current one, *i.e.*, the consistent approach doesn't gain anything in terms of efficiency. However, having precise resolution on the scattering probabilities doesn't seem to be as important as consistency from a reaction rate preservation standpoint.

openmc/mgxs/mgxs.py
```diff
@@ -4891,6 +4903,1151 @@ def xs_tally(self):
...
+        if self._xs_tally is None:
+            energyout_bins = [self.energy_groups.get_group_bounds(i)
+                              for i in range(self.num_groups, 0, -1)]
+            norm = self.rxn_rate_tally.summation(
+                filter_type=openmc.EnergyoutFilter, filter_bins=energyout_bins)
+
+            # Remove the AggregateFilter summed across energyout bins
+            norm._filters = norm._filters[:2]
+
+            # Compute the group-to-group probailities
```
nelsonag: probabilities

nelsonag:
I'd agree that reaction rate preservation is more important - if you dont get the right number of neutrons it doesn't matter their direction or energy.  

@wbinventor I think that the reason for OpenMC's MG mode is not convenience, its education and a deeper understanding of the worth of approximations (as well as flexibility, relating to education).  Therefore I am intrinsically against anything that gets in the way of that.  Having a scattering matrix which isnt really a scattering matrix but an adjusted one does get in the way of that goal.  So while I totally agree that what you're trying to do with this PR is useful, awesome, and a great idea, I approach it with hesitation. I think the deterministic users you speak of could be 'de-confused' if we had templates for the mgxs-types in the library.  I.e., if you want a library for MGMC its best to use this set of types, if for OpenMOC, these are the best, etc.

Also, perhaps your hesitation to have these 5 new classes plus the previous two be because that leads to 7.  Couldn't this instead be just two extended ScatteringMatrix types (nu and not nu), which simply replace the current sig{s,l,g'->g}*phi_{g'}/phi_{g'} with the equation I have listed a few comments ago? (With this you'd have 4 total classes for scattering vice 7 - we can even make this 2 if th nu-flag simply becomes a parameter of the scattering matrix)

wbinventor:
I've added in support for higher order Legendre moments per @nelsonag's suggestion. From a few preliminary tests, the new formulation seems to produce very nearly the same results as the existing code. However, @bforget mentioned to me in passing today that he thinks this new approach may be flawed; perhaps he can elaborate here, or we will discuss it together later this week. I had already mostly implemented this concept, and decided to push since the results "look" good.

As for the software design, code bloat, etc., perhaps we can cogitate on this a bit before moving forward. I like the idea of having a `ConvolvedMGXS` class because it does simplify some things, but perhaps it's not worth the effort.

nelsonag:
I'm curious to hear @bforget thoughts!

PR #802, opened 2017-02-25
Assortment of bug fixes

nelsonag:
This PR fixes the following issues I found when writing some new example notebooks:

1. Extended `openmc.run` to also have an output more which only prints the summary (see the update mgxs-part-iv notebook for it in practice)
2. Fixed the generation of cells from a mesh (previously there was an incorrect ordering of universes in the lattice)
3. `openmc.MGXSLibrary.xs_shapes` for delayed groups had wrong ordering (fixed), this required changes in `openmc.MGXSLibrary` and `openmc.Plotter`
4. Fixed error in multi-group plots which only plotted one group not all (compare new to old mgxs-part-iv)
5. Fixed issue with setting max order to 0 for Legendre scattering (not the default: Legendres converted to tabular w/in OpenMC)
6. Removed superfluous xs assignments in MG mode which only slowed things down.

### DIFF
--- a/openmc/executor.py
+++ b/openmc/executor.py
@@ -5,7 +5,7 @@ import sys
 
 from six import string_types
 
-summary_indicator = "TIMING STATISTICS"
+_summary_indicator = "TIMING STATISTICS"
 
 
 def _run(command, output, cwd):
@@ -25,7 +25,7 @@ def _run(command, output, cwd):
         if output == 'full':
             # If user requested output, print to screen
             print(line, end='')
-        elif output == 'summary' and summary_indicator in line:
+        elif output == 'summary' and _summary_indicator in line:
             # If they requested a summary, look for the start of the summary
             storage_flag = True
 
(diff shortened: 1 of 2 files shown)


openmc/executor.py
```diff
@@ -5,21 +5,33 @@
 
 from six import string_types
 
+summary_indicator = "TIMING STATISTICS"
```
wbinventor: Make this private with a preceding underscore

openmc/executor.py
```diff
@@ -44,7 +56,7 @@ def plot_geometry(output=True, openmc_exec='openmc', cwd='.'):
 
 
 def run(particles=None, threads=None, geometry_debug=False,
-        restart_file=None, tracks=False, output=True, cwd='.',
+        restart_file=None, tracks=False, output="full", cwd='.',
```
wbinventor: Use single quotes around `'full'`, just to be consistent with the convention across the rest of the API.

openmc/mesh.py
```diff
@@ -189,21 +189,21 @@ def from_hdf5(cls, group):
 
     def cell_generator(self):
         """Generator function to traverse through every [i,j,k] index
-        of the mesh.
+        of the mesh in the same order that would be done by the Fortran
```
wbinventor: Any reason to add this comment?

openmc/executor.py
nelsonag: - [x]

openmc/executor.py
nelsonag: - [x]

openmc/mesh.py
nelsonag: Not particularly no, I've removed it.

- [x]

PR #807, opened 2017-02-28
Consistent Multi-Group Scattering Matrices

wbinventor:
This PR is to replace #793 since I apparently damaged my that branch and can't figure out how to repair it. The code is the same, but I have additionally updated the code to reflect @nelsonag's consolidation of the (nu-)scatter MGXS in #803.


nelsonag:
What happened to the @bforget concern?

bforget:
Sorry for missing the last post, I have a tendency to delete the OpenMC e-mails when I get behind! I think my worries were unfounded, I just have a hard time wrapping my head around current collapsing (or higher order moment collapsing) of a cross-section in a volume.  I think Will mentioned you had closed the loop and re-inserted high order multigroup data and were able to reproduce the continuous energy results? If that's the case, it's hard to argue against that!

wbinventor:
Well that is partially true - OpenMC's MG mode code can indeed *use* the higher order moments, and @nelsonag's prior work has shown the the current scattering moment formulation can well reproduce CE results. However, I have *not* taken the time and effort to confirm that this new formulation can be used in MG mode to reproduce the CE results. @nelsonag is this something that you could try in some of your scripts?

wbinventor:
Also I can't figure out why the test suite is failing, since all tests pass on my machine and for the life of me I can't get the Travis failure log to load in my browser.

paulromano:
Amazon S3 is having problems which apparently Travis relies on. Try restarting tests once their outage is over.

openmc/mgxs/mgxs.py
```diff
@@ -3938,9 +3959,9 @@ def get_xs(self, in_groups='all', out_groups='all',
         (3rd dimension), nuclides (4th dimension), and moments/histograms
         (5th dimension).
 
-        NOTE: The scattering moments are not multiplied by the :math:`(2l+1)/2`
-        prefactor in the expansion of the scattering source into Legendre
-        moments in the neutron transport equation.
+        .. note:: The scattering moments are not multiplied by the
+        :math:`(2l+1)/2` prefactor in the expansion of the scattering source
+        into Legendre moments in the neutron transport equation.
```
paulromano: To get this to render properly, it need to be written as:
```rst
        .. note:: The scattering moments are not multiplied by the
                  :math:`(2\ell+1)/2` prefactor in the expansion of the
                  scattering source into Legendre moments in the neutron
                  transport equation.
```
(Also note I've changed `l` to `\ell`)

openmc/mgxs/mgxs.py
```diff
@@ -4528,36 +4549,42 @@ def xs_tally(self):
...
     necessary data to compute multi-group cross sections from a
     :class:`openmc.StatePoint` instance. The derived multi-group cross section
-    can then be obtained from the :attr:`NuFissionMatrixXS.xs_tally` property.
+    can then be obtained from the :attr:`ScatterProbabilityMatrix.xs_tally`
+    property.
 
     For a spatial domain :math:`V`, incoming energy group
     :math:`[E_{g'},E_{g'-1}]`, and outgoing energy group :math:`[E_g,E_{g-1}]`,
-    the fission production is calculated as:
+    the group-to-group scatterin probabilities are calculated as:
```
nelsonag: "scatterin" -> "scattering"

openmc/mgxs/mgxs.py
```diff
@@ -4621,7 +4648,7 @@ class NuFissionMatrixXS(MatrixMGXS):
         The tally estimator used to compute the multi-group cross section
     tallies : collections.OrderedDict
         OpenMC tallies needed to compute the multi-group cross section. The keys
-        are strings listed in the :attr:`NuFissionMatrixXS.tally_keys`
+        are strings listed in the :attr:`MultiplicityMatrixXS.tally_keys`
```
nelsonag: ScatterProbabilityMatrixXS vice MultiplicityMatrixXS

openmc/mgxs/mgxs.py
```diff
@@ -4655,83 +4682,98 @@ class NuFissionMatrixXS(MatrixMGXS):
 
     """
 
-    def __init__(self, domain=None, domain_type=None, groups=None,
+    def __init__(self, domain=None, domain_type=None, groups=None, nu=False,
                  prompt=False, by_nuclide=False, name='', num_polar=1,
```
nelsonag: should remove the prompt argument

openmc/mgxs/mgxs.py
wbinventor: Oh interesting, do all of the ``.. note::` blocks need to be indented as such?

openmc/mgxs/mgxs.py
wbinventor: Fixed

openmc/mgxs/mgxs.py
wbinventor: Done

wbinventor:
All of the tests appear to fail on the Travis server, for some reason unbeknownst to me. I forget if we have less-than-hacky way to report an extended version of the Travis error log? If so, it might actually be worth documenting that in the developer's guide.

paulromano:
@wbinventor same reason as before -- syntax error on line 5345 in mgxs.py. Did you lose updates when you resubmitted this PR??

wbinventor:
Dang, I'm not sure what happened to my other branch and didn't take the time to de-mangle it. It's clear that a few of the changes didn't make it into this branch, but I've fixed the list dereference issue now, so hopefully it passes Travis.

wbinventor:
Looks like Travis passes now.

paulromano:
Looks good to me. @nelsonag, are you satisfied?

nelsonag:
No I havent really started yet. I thought it was mentioned in the last PR
that we wanted to understand what this did to the convergence rate?
(i.e., downstream Multi-Group keff vs upstream batches of the CE solver).
I'm particularly interested in how this does for nuscatter, as thats the
more important one and the noisier one.

openmc/mgxs/mgxs.py
```diff
@@ -4809,64 +4849,953 @@ class Chi(MGXS):
...
+                 num_azimuthal=1, nu=False):
+        super(ConsistentScatterMatrixXS, self).__init__(
+            domain, domain_type, groups, by_nuclide=by_nuclide,
+            name=name, num_polar=num_polar, num_azimuthal=num_azimuthal)
+
+        self.nu = nu
+
+        # Initialize each MGXS used by the convolution
+        self._mgxs = \
+            [ScatterXS(), ScatterProbabilityMatrix(), MultiplicityMatrixXS()]
```
nelsonag: It looks like this scheme requires four different analog tallies (`nu-scatter-0` and `scatter-0` with energy in/out filters for multiplicity, and `nu-scatter-pN` [or with histogram bins] with energy/inout filters for the scatter probability matrix, `flux` w/ energyin filter) and two tracklength tallies (`flux` and `scatter-0` both with an energyin filter).  

If you coded this to directly deal with the tally data instead of going through the pre-existing MGXS classes then you could drop the analog `nu-scatter-0` with energy in/out filters, and the analog `flux`.  
I think this could be done with:
`psuedo_tl_nu-scatter[gin, gout, n] = tl_scatter-0[gin] / tl_flux[gin] *  (analog_nu-scatter-pN[gin, gout, 0] / analog_scatter-0[gin, gout]) * (analog_nu-scatter-pN[gin, gout, n] / sum_{gout=1:G}(analog_nu-scatter-pN[gin, gout, 0]))`
(of course the above only works for Legendre scattering, not histogram, but thats only a minor tweak)

That is a big deal for OpenMC runtime performance.  In addition, doing so obviates the need for the generality of the Convolved class, meaning I think this can just be an extended type of `ScatterMatrixXS`, which I'm also obviously a fan of! ;-)

openmc/mgxs/mgxs.py
```diff
@@ -4809,64 +4849,953 @@ class Chi(MGXS):
         Whether or not the MGXS is merged from one or more other MGXS
     hdf5_key : str
         The key used to index multi-group cross sections in an HDF5 data store
+    mgxs : list of openmc.mgxs.MGXS
+        A list of MGXS to combine to compute this multi-group cross section
 
     """
 
     def __init__(self, domain=None, domain_type=None, groups=None,
```
wbinventor: Yes, Python's Method Resolution Order [strategy](https://www.google.com/search?client=ubuntu&channel=fs&q=python+MRO&ie=utf-8&oe=utf-8) will select the `get_xs(...)` implementation from the `ScatterMatrixXS` parent class.

openmc/mgxs/mgxs.py
```diff
@@ -4809,64 +4849,953 @@ class Chi(MGXS):
...
+
+            if self.nu:
+                self._xs_tally *= self.multiplicity_matrix.xs_tally
+
+            self._xs_tally -= correction
+            self._compute_xs()
+
+        return self._xs_tally
+
+    @ScatterMatrixXS.correction.setter
```
wbinventor: That's right. I actually realized that these inherited properties weren't implemented correctly, and will be pushing the fix in a moment.

wbinventor:
As requested, I ran two case studies to determine how much this new formulation improves the stability of the eigenvalues calculated by a deterministic multi-group transport code. I generated a 2-group MGXS library - with both the current and new scattering formulations - for the "pincell" example input with 1000 batches of 1000 particles / batch. The tallies were stored to statepoint files every 5 batches. OpenMOC was run twice for each statepoint file for the two different scattering formulations, for a total of 400 simulations for each case study without and with scattering multiplication.

The OpenMOC eigenvalue bias with respect the OpenMC's "most converged" eigenvalue reported after 1000 batches (1.164869 +/- 0.000879) was plotted for normal scattering:



as well as for nu-scattering with multiplication:



While the results for normal scattering are quite compelling, those with scattering multiplication don't demonstrate the exact same trends for both formulations. I find this quite surprising since I would have hypothesized that (1) the noise from the analog multiplicity matrix would have a negligible impact and (2) the two would converge with enough histories. Any thoughts? Is there a bug?

Btw, these case studies are super easy and fast to run. Just in case posterity is interested, the inputs can be downloaded from my Dropbox:

https://www.dropbox.com/sh/5nko882ctq69y9z/AADl3XwtOPXT4IYJzLtGYuqAa?dl=0

openmc/mgxs/mgxs.py
```diff
@@ -4809,64 +4849,953 @@ class Chi(MGXS):
...
+        avg_xs = \
+            super(ConsistentScatterMatrixXS, self).get_subdomain_avg_xs(subdomains)
+
+        for i, mgxs in enumerate(avg_xs._mgxs):
+            avg_xs._mgxs[i] = mgxs.get_subdomain_avg_xs(subdomains)
+
+        return avg_xs
+
+
+class NuFissionMatrixXS(MatrixMGXS):
```
wbinventor: Yes, I didn't change `NuFissionMatrixXS`

nelsonag:
Hmm, I looked at the script and there is nothing obvious there. I'm about to leave for the day, so I can't follow up on this, but I wanted to run some form of a 1 and/or 2-group infinite medium problem with it (to remove the importance of the angular distributions) and see if those variations on the bias follow with the variations in the multiplicity.

By the way, I'm running your script now (without changing anything) and it seems as if the Consistent case has a keff value of 0.14ish.  Any obvious reason why that would happen?  (I just checked out your branch minutes ago)

wbinventor:
@nelsonag yes I'm neglected to mention that the openmoc.materialize module must be slightly modified to make it look for "consistent (nu-)scattering matrix" in an MGXS library. Since scattering matrices can't be found, the code is instead using an all zeroes scattering matrix for each material (and should be reporting a warning each time). An easy fix can be found on my branch "remove-opencg" and fork of OpenMOC, which is PR'ed as mit-CRPG/OpenMOC#327 (namely, here: https://github.com/mit-crpg/OpenMOC/pull/327/files#diff-00b800a88bb508e841d3edbb11e78ae1R388).

Separately, I realized that some of the code I had working to support Legendre moments on my old branch is no longer here, and I need to add it back in here, so this isn't ready to merge yet anyway.

wbinventor:
I've fixed the Legendre moments. I do plan to take a stab at simplifying this formulation as @nelsonag suggests by: 

(1) introducing a new `ScatterMatrixXS.consistency` boolean property to toggle this formulation on/off
(2) keeping `ScatterProbabilityMatrix` since it can be useful on its own
(3) removing `ConvolvedMGXS`

In other news, @nelsonag you may be interested to know that I just took the prerogative to merge in mit-crpg/OpenMOC#327. You should pull and re-install OpenMOC in order to run the batch-by-batch scripts I shared earlier.

wbinventor:
Alright so I refactored this PR as requested and the new code is now *much* simpler. I removed the `ConvolvedMGXS` and `ConsistentScatterMatrixXS` classes. The `ScatterMatrixXS` class now has a `formulation` property which can take on values of `'simple'` (default) and `'consistent'`.

Even better, by making these changes I must have fixed an issue in the "nu" treatment for my original `ConsistentScatterMatrixXS` class. You can now see the batch-by-batch stability of the OpenMOC eigenvalue with nu-scattering matrices is greatly improved with the new `'consistent'` formulation just as I demonstrated for normal scattering matrices without multiplication last night:



@nelsonag and @paulromano this is ready for final review. I think the results shown for a 2D pin cell are pretty clear that this is a slam dunk, but feel free to test this out on an infinite medium if you wish to further test this without angular effects @nelsonag.

openmc/mgxs/mgxs.py
```diff
@@ -3525,6 +3552,36 @@ class ScatterMatrixXS(MatrixMGXS):
...
     in the above relation, the `nu` parameter can be set to `True`.
 
+    An alternative form of the scattering matrix is computed when the
+    `formulation` property is set to 'consistent' rather than the default
+    of 'simple'. This formulation computes the scattering matrix multi-group
+    cross section as the product of the scatter cross section and
+    group-to-group scattering probabilities.
+
+    Unlike the default 'simple' formulation, the 'consistent' formulation
+    is computed from the groupwise cattering cross section which uses a
```
nelsonag: `cattering` -> `scattering`

openmc/mgxs/mgxs.py
```diff
@@ -3694,70 +3758,252 @@ def histogram_bins(self):
...
+            if self.tallies is None:
+                msg = 'Unable to get xs_tally since tallies have ' \
+                      'not been loaded from a statepoint'
+                raise ValueError(msg)
+
+            # Use super class method
+            if self.formulation == 'simple':
+                self._xs_tally = MGXS.xs_tally.fget(self)
+
+            else:
```
nelsonag: Nice!

openmc/mgxs/mgxs.py
```diff
@@ -3694,70 +3758,252 @@ def histogram_bins(self):
...
+                # Compute groupwise scattering cross section
+                self._xs_tally = self.tallies['scatter'] / \
+                                 self.tallies['flux (tracklength)']
+
+                # Compute scattering probability matrix
+                energyout_bins = [self.energy_groups.get_group_bounds(i)
```
nelsonag: Shouldn't there be a special branch around here for histogram scattering?  i.e., the scattering probability won't be in scatter-0, it has to be found from all of the histogram bins (see my `mgxs_library.py` convert_scatter_format for a guide

openmc/mgxs/mgxs.py
```diff
@@ -4528,41 +4776,45 @@ def xs_tally(self):
         return self._xs_tally
 
 
-class NuFissionMatrixXS(MatrixMGXS):
-    r"""A fission production matrix multi-group cross section.
+class ScatterProbabilityMatrix(MatrixMGXS):
```
nelsonag: What is the use case of this class?

nelsonag:
Oh, @wbinventor, one more.  Didn't the last PR have a consistent `TransportXS` formulation? Why did you choose against that here? (If it wasn't in the last, then I guess the question becomes why not do it as well)

openmc/mgxs/mgxs.py
wbinventor: Gotcha

openmc/mgxs/mgxs.py
wbinventor: I'll take a look into implementing this, perhaps next week.

openmc/mgxs/mgxs.py
wbinventor: I left this class intact since it could be useful for educational or illustrative purposes. I'm refactoring it by removing support for higher order Legendre moments so that it doesn't encapsulate so much redundant code, and I really can't think of a use case for higher order moments in this case.

wbinventor:
@nelsonag I've addressed your two questions below:

**Test results changed**
The test inputs changed since there are more tallies - and they are merged / exported in a different order - to support the new `ScatterProbabilityMatrix` and consistent scattering matrix options. I need to dig into the other results to determine why they have changed - I agree that there shouldn't be any changes but rather only new results for the new formulation.

**"Consistent" Transport XS**
I actually never implemented nor proposed this before. I think what you are implying is that the (total / flux) could be computed using tracklength estimators, while the transport correction (scatter_p1 / flux) could be separately computed from analog estimators? I'm going to ask that this not be a requirement for this PR for two reasons. 

First, this is really addressing a different problem altogether from the global reaction rate consistency that this PR addresses. The transport correction is subtracted from both sides of the transport equation and hence it does not impact balance. However, using a mixture of tracklength and analog estimators *might* improve the convergence of the transport-corrected total cross section.

Second, our current transport correction sucks, and new formulations better suited for MC are currently being explored as a research topic at MIT. I'll propose to let that work take the place of - or at least complement - our current transport correction rather than hack in a confusing mixture of estimators which won't really improve matters (in terms of bias due to iso-in-lab scattering).

wbinventor:
@nelsonag I pushed a refactored and greatly slimmed down version of the new `ScatterProbabilityMatrix` class. I also reviewed the changed test results and it appears that none of the results changed per se - there are only additions to the test results for the new consistent scattering matrices and the scatter probability matrix. So I think we are good to go on this front.

I'll take a stab at adding support for angular histograms at some point once I have time to wrap my mind around what this entails.

nelsonag:
Hey @wbinventor I was just having trouble using your Dropbox script to try and run a case where 'total' is switched to 'transport' and I can't turn off the `correction = None` line in the two scripts without encountering an error.  Does this make sense/seem related to you?:

```
File "build-xml.py", line 141, in <module>
    mgxs_lib.add_to_tallies_file(tallies_file, merge=True)
  File "/home/nelsonag/git/openmc/openmc/mgxs/library.py", line 553, in add_to_tallies_file
    for tally in mgxs.tallies.values():
  File "/home/nelsonag/git/openmc/openmc/mgxs/mgxs.py", line 507, in tallies
    self._tallies[key].filters.append(add_filter)
  File "/home/nelsonag/git/openmc/openmc/checkvalue.py", line 321, in append
    check_type(self.name, item, self.expected_type)
  File "/home/nelsonag/git/openmc/openmc/checkvalue.py", line 33, in check_type
    raise TypeError(msg)
TypeError: Unable to set "tally filters" to "[EnergyoutFilter
	Bins           =	[  0.00000000e+00   6.25000000e-01   2.00000000e+07]
]" which is not one of the following types: "Filter, CrossFilter, AggregateFilter"
```

wbinventor:
@nelsonag pull my latest commit which should fix a small bug in the `ScatterMatrixXS.filters` property for the transport correction term.

nelsonag:
Thanks @wbinventor, that may have fixed that problem but it seems to have exposed another bug. If you set `total` to `transport`, and obviously take off the `correction`, you end up with a tally in tallies.xml that has a `flux` score with an energyout filter. Its probably from `ScatterMatrixXS` since `TransportXS` looks fine to me. Any thoughts on this?

wbinventor:
@nelsonag I've corrected the issue that led to a flux tally with an energyout filter.

wbinventor:
@nelsonag I have added in support for histogram scattering in 4be0e58. Let me know if this meets your expectations.

nelsonag:
Thanks @wbinventor, it works!  I did run cases with nu-transport and transport using your script.  The resultant figures are attached.

With just `transport`
keff-bias_transport

With `nu-transport` (and `consistent nu-scatter matrix` or `nu-scatter matrix`)
keff-bias_nutransport

I dont see a need to hold up this PR to implement a consistent transport option, but it looks to me (in the little time I have right now to dive in) that there is either a need for consistent transport, or 2) no driving need for consistent scatter matrix (unless one is doing Pn scattering, which for our audience right now means their using MG mode, which means they dont need it either).  Thoughts?

wbinventor:
Thanks for sharing your results with the transport correction @nelsonag. I've updated `TransportXS` to now use a "consistent" correction term with that used for the scattering matrix. In particular, the transport correction term is now computed as the ratio of a `'scatter-p1'` and `'flux'` tallies, both computed with analog estimators, for both `TransportXS` and `ScatterMatrixXS`. Previously, the `TransportXS` tried to be "smart" and just use a single tracklength flux estimator for both the total MGXS and transport correction terms, which in retrospect was not so smart since it was inconsistent with the way the transport correction was computed using all analog estimators for `ScatterMatrixXS`. 

In any case, I think the new results speak for themselves. Here is the case without scattering multiplication:



and with scattering multiplication:



Thanks for pushing further into the transport correction, I'm glad we fixed this discrepancy in this PR.

wbinventor:
Fyi the test suite failed with the following error:

```bash
$ git clone --depth=50 https://github.com/mit-crpg/openmc.git mit-crpg/openmc
Cloning into 'mit-crpg/openmc'...
remote: Counting objects: 14824, done.
remote: Compressing objects: 100% (6153/6153), done.
remote: Total 14824 (delta 10196), reused 12293 (delta 8225), pack-reused 0
Receiving objects: 100% (14824/14824), 12.28 MiB | 18.26 MiB/s, done.
Resolving deltas: 100% (10196/10196), done.
$ cd mit-crpg/openmc
0.59s$ git fetch origin +refs/pull/807/merge:
remote: Counting objects: 339, done.
remote: Compressing objects: 100% (126/126), done.
remote: Total 339 (delta 243), reused 303 (delta 212), pack-reused 0
Receiving objects: 100% (339/339), 101.96 KiB | 0 bytes/s, done.
Resolving deltas: 100% (243/243), completed with 36 local objects.
From https://github.com/mit-crpg/openmc
 * branch            refs/pull/807/merge -> FETCH_HEAD
$ git checkout -qf FETCH_HEAD
3.07s$ git submodule update --init --recursive
fatal: No url found for submodule path 'src/xml/fox' in .gitmodules
```

I'll take a look at our ".travis.yml" file and see if this is an easy thing that I can fix.

nelsonag:
Do you need to merge upstream/develop?

nelsonag:
I obviously didn't mean to start a new review, but the comments still stand.  Phone browsers suffer with long PRs and barely respond to input.

nelsonag:
So here is what's left:

1. Fix Fox issue. Looks like you're on it now.
2. Histogram question from a few minutes ago
3. This is not for this PR, but I urge you to at least write down, in an issue, the fixes you'd like to make now that all this knowledge is fresh in your head: is there anything in the CMFD xs generation for OpenMC based on what we learned here?

wbinventor:
1. I think I fixed the Fox issue since the tests just passed. For some reason when I merged with develop it didn't remove the "src/xml" directory but I removed it myself in https://github.com/mit-crpg/openmc/pull/807/commits/67ff99511db109fdd3fadc31af80f13b25c4437c
2. Support for histograms were added in https://github.com/mit-crpg/openmc/commit/4be0e58be807fecaaa63e6badcd81ea18902e5dc
3. I'm not sure what you mean about MGXS generation for OpenMC CMFD, which I've never used. Do you mean OpenMOC? If that's what you meant, there are no real outstanding issues other (a) than that our transport correction is no good and (b) we can always improve our spatial self-shielding models (the focus of my thesis).

nelsonag:
2. But don't you have to change rxn_rate_tally for the histogram case?
3. I mean create_cmfd_tally in cmfd_input.F90 in OpenMC and related code in
cmfd_data.F90. are those the best scores types? Are the MGXS calculation
procedures the right/best ones to use?

wbinventor:
1. I don't think so, but maybe I'm missing something. The `ScatterMatrixXS.rxn_rate_tally` property actually throws a `NotImplementedError` when `formulation` is set to `'consistent'`. The `ScatterMatrixXS.xs_tally` property should cleanly accommodate the histogram format within the scattering probability matrix.

2. I'm not at all too familiar with the tallies used for CMFD (nor am I aware of anyone using CMFD these days). From my cursory look of "cmfd_data.F90" I do think that it is using the "inconsistent" formulation of the scattering matrix. I'll open an issue and record this once this PR is merged.

nelsonag:
1. You should try and run a histogram case with a consistent nu-scatter matrix.  The current mgxs-library-iv would help.
2. Thank you.
3. Speaking of item 1, I now have to add a new one: the MGXS-part-iv notebook doesnt work if you switch `nu-scatter matrix` to `consistent nu-scatter matrix` (without even changing to histogram yet).  It fails for multiple reasons, first the library correctness, then if you work around that it fails when tryign to get the mgxs using names which arent right anymore (i.e., it should use consistent nu-scatter matrix vice nu-scatter matrix), and thats as far as I got.

wbinventor:
@cjosey the issues you mention were rooted in a very subtle issue with tally merging. I've fixed this by introducing a new `Tally._consolidate_moment_scores()` method which is called by `Tally.merge(...)`, and will remove redundantly defined moment scores - *i.e.*, remove `'scatter-2'` if `'scatter-P3'` is already a score. This eliminates issues which can arise in `openmc.mgxs` whereby a `Tally` cannot be properly "sliced" if it contains two or more of the same scores, which happens when a `'scatter-PN'` score is expanded under-the-hood at runtime.

While the "mgxs-part-iv.ipynb" notebook now runs as is, I have not yet tested out the histogram option for the consistent scattering matrix. I'll do so later when I have more time. I'm glad I had this notebook of yorus to help debug this issue :-P

wbinventor:
@nelsonag I've confirmed that the new consistent formulation "works" with histogram binning. I say that in quotes because I haven't fully put it through the gauntlet as you might want to do, but I at least confirmed that a `MuFilter` is added into the group-to-group scatter probability matrix, and this additional filter is successfully propagated through the chain of tally arithmetic used to compute the scatter matrix itself. In addition, I confirmed that the `MGXS.get_pandas_dataframe(...)`, `MGXS.get_condensed_xs(...)` methods work. However, I discovered that the `ScatterMatrixXS.print_xs(...)` method fails for histogram binning for both simple and consistent formulations as noted in #828. One caveat is that the values for the simple and consistent formulations with histogram binning don't match very well - I'm not sure if this is due to either (1) poor statistics or (2) an incorrect implementation. Perhaps you can take a look and let me know if you see anything astray.

nelsonag:
@wbinventor the current mgxs-part-iv still doesn't work when you switch `nu-scatter matrix` to `consistent nu-scatter matrix` and switch 

```
mgxs_lib.legendre_order = 3
``` 
to
```
mgxs_lib.scatter_format = 'histogram'
mgxs_lib.histogram_bins = 2
```

The following error is received:
```
  File "mgxs-part-iv.py", line 384, in <module>
    mgxs_lib.load_from_statepoint(sp)
  File "/home/nelsonag/git/openmc/openmc/mgxs/library.py", line 596, in load_from_statepoint
    mgxs.load_from_statepoint(statepoint)
  File "/home/nelsonag/git/openmc/openmc/mgxs/mgxs.py", line 4129, in load_from_statepoint
    self.tallies[self.rxn_type].scores = [self.rxn_type]
KeyError: 'nu-scatter'
```

wbinventor:
@nelson thanks for noticing that, I fixed that bug with my most recent commit. However, I the eigenvalue bias with the "consistent nu-scatter matrix" option with the histograms is quite large. I'm not sure if this is an issue with my implementation, or is to be expected. But at least more of the bugs have been worked out now.

wbinventor:
After a little thought I figured this one out. The scattering probability matrix needed to be normalized to the sum across mu filter bins. With that fixed I think (hope!) this PR is ready to go.

openmc/mgxs/mgxs.py
```diff
@@ -4125,8 +4125,9 @@ def load_from_statepoint(self, statepoint):
                     self.tallies[tally_key].scores = \
                         [score_prefix + '{}'.format(i)
                          for i in range(self.legendre_order + 1)]
-        elif self.scatter_format == 'histogram':
-            self.tallies[self.rxn_type].scores = [self.rxn_type]
+#        elif self.scatter_format == 'histogram':
```
nelsonag: Remove this commented out block

openmc/mgxs/mgxs.py
```diff
@@ -3838,13 +4134,15 @@ def load_from_statepoint(self, statepoint):
...
-                     for i in range(self.legendre_order + 1)]
-        elif self.scatter_format == 'histogram':
-            self.tallies[self.rxn_type].scores = [self.rxn_type]
+            for tally_key, tally in self.tallies.items():
+                if 'scatter-P' in tally.scores[0]:
+                    score_prefix = tally.scores[0].split('P')[0]
+                    self.tallies[tally_key].scores = \
+                        [score_prefix + '{}'.format(i)
+                         for i in range(self.legendre_order + 1)]
+#        elif self.scatter_format == 'histogram':
```
nelsonag: Remove this commented out block

nelsonag:
@wbinventor all set after the above comment! Since I'm only working on this sporadically this week, someone else can feel free to merge if they'd like once the commented out code is removed.

openmc/mgxs/mgxs.py
wbinventor: Done

nelsonag:
Thanks @wbinventor for the awesome PR!

PR #818, opened 2017-03-05
Addition of new notebooks to showcase Multi-Group mode features

nelsonag:
This PR adds new example notebooks for multi-group mode (covering all the bullets in #791). This PR also moves the example notebooks to a top level location in the documentation as discussed in #791.


docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+   "execution_count": 13,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# OpenMC simulation parameters\n",
+    "batches = 1000\n",
+    "inactive = 100\n",
+    "particles = 5000\n",
```
smharper: This takes a few minutes to run.  I think we should go for fewer batches just to demo the functionality.  Maybe 150 with 50 inactive.  That's enough to get +/- 100 pcm and a decent looking plot of pin power.

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "\n",
+    "# We dont need a tallies.out file, as we will be working with the StatePoint\n",
+    "# file instead, so tell OpenMC not to write the file.\n",
+    "settings_file.output = {'tallies': False}\n",
+    "\n",
+    "# Tell OpenMC this is a multi-group problem\n",
+    "settings_file.energy_mode = 'multi-group'\n",
+    "\n",
+    "# Set the verbosity to 6 so we dont see output for every batch\n",
+    "settings_file.verbosity = 6\n",
```
smharper: I'd also recommend a verbosity where we can see the simulation progress.

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+    "zircaloy = openmc.Material(name='Zircaloy')\n",
+    "zircaloy.set_density('g/cm3', 6.55)\n",
+    "zircaloy.add_nuclide(zr90, 7.2758e-3)\n",
+    "\n",
+    "# borated water\n",
+    "water = openmc.Material(name='Borated Water')\n",
+    "water.set_density('g/cm3', 0.740582)\n",
+    "water.add_nuclide(h1, 4.9457e-2)\n",
+    "water.add_nuclide(o16, 2.4732e-2)\n",
+    "water.add_nuclide(b10, 8.0042e-6)\n"
```
smharper: This material declaration can be simplified a lot.  You can pass a string directly to `add_nuclide` rather than first creating a Nuclide, and we can take advantage of the new `add_element(enrichment=...)` feature.
```python
fuel.add_element('U', 1.0, enrichment=1.6)
fuel.add_element('O', 2.0)
```
And with the `add_element` functionality, I see no reason to use only one nuclide of Zr
```python
zircaloy.add_element('Zr', 1)
```

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Create the surface used for each pin\n",
+    "pin_surf = openmc.ZCylinder(surface_id=1, x0=0, y0=0, R=0.54, name='pin_surf')\n",
```
smharper: The `x0=0` and `y0=0` are unnecessary.

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Create cylinders for the fuel and clad\n",
+    "fuel_outer_radius = openmc.ZCylinder(x0=0.0, y0=0.0, R=0.39218)\n",
+    "clad_outer_radius = openmc.ZCylinder(x0=0.0, y0=0.0, R=0.45720)\n",
```
smharper: The `x0=0` and `y0=0` are unnecessary.

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: I've taken the strategy to be more verbose to add clarity instead of leaving some stuff unsaid

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: Your suggestion here is fine (so long as the plot does come out decent of course), but since this will likely come up again for mg-mode-part-iii, I want to take the opportunity to state that we do need enough to get close and have the code and its features be attractive.  

In general I don't see a huge downside to having many batches/particles: its not like these get run every time a commit is made like our test suite on Travis, and the user will be following the steps one-by-one and can modify as needed.

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: I took the view here that since this is not the first of many examples, that the user has already seen the iterations and will just be scrolling past anyways.  Displaying progress is only for us, the ones making the notebook, not for those learning from it.  Since I'd rather run more particles to get attractive results but not want to require 30 pages of printouts, this was the compromise.  Thoughts?

docs/source/examples/mg-mode-part-ii.ipynb
nelsonag: No general objections; I want to do whatever is clearest to the user vice more pythonic or condensed.  With that in mind does that change the view on this (or should mine be corrected)?

nelsonag:
@smharper thanks for the review so far!  To answer your big question, most of the time just a how-to will suffice.  For these MG notebooks, however, I think a little of both is in order.  Not many users are familiar with non-Legendre (or non-isotropic for some) scattering or angle-dependent cross sections.  So I think the onus is on us to show not only how to do it, but that there might be a reason to care.  

Also, I want to be clear that my comments on the line items in your review are meant to start the discussion, not to say "I'm not doing it because I'm right".  I don't think we as a team have had a clear discussion on what the exact intent of the notebooks are and so this seems like a good time as any to see what the consensus is.

As for the location of the notebooks, thats not a bad idea and I don't mind either way.  Is there a way to have the notebook rst files refer to locations outside of the docs directory?

smharper:
RE notebook rst files: I think we can refer to something outside of the docs with a sufficient number of `..`s

docs/source/examples/mg-mode-part-i.ipynb
wbinventor: These default values are used elsewhere in the API for the sake of clarity.

docs/source/examples/mg-mode-part-i.ipynb
wbinventor: I agree with @smharper that we should use fewer batches/particles at the risk of the results not looking quite as good (that's just MC). In fact, we might want to devise and enforce a rule that no notebook can take more than say, a 90 seconds to run, or something like that (and one or more of the current notebooks may already break this rule). I think this is important for the following reasons:

(1) The longer it takes to run a notebook, the larger the burden on developers who *should* be running the notebooks before submitting PRs to confirm that they aren't affected.
(2) When developers do run the notebooks, they will be inclined to reduce the number of batches/particles anyway, and may accidentally commit these slight adjustments. The parameters should be designed such that there is little incentive for a developer to change them.
(3) IMO we should aim to include the notebooks as part of the test suite someday (even if they aren't run on Travis) to better automate QA for our development workflow.

Just my two cents.

docs/source/examples/mg-mode-part-ii.ipynb
wbinventor: I agree with @smharper, everyone would be using these features if they knew about them from our examples.

wbinventor:
I agree that it would also be nice to raise the profile of the notebooks by re-locating them to the "examples" directory, if at all possible.

docs/source/examples/mg-mode-part-i.ipynb
wbinventor: Actually, that's not quite right, the default values are used in some places and not in others.

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+    "surfaces['Assembly x-'] = openmc.XPlane(x0=pin_pitch)\n",
+    "surfaces['Assembly x+'] = openmc.XPlane(x0=length - pin_pitch)\n",
+    "surfaces['Assembly y-'] = openmc.YPlane(y0=pin_pitch)\n",
+    "surfaces['Assembly y+'] = openmc.YPlane(y0=length - pin_pitch)\n",
+    "\n",
+    "# Create cylinders for the fuel and clad\n",
+    "surfaces['Fuel Radius'] = openmc.ZCylinder(R=fuel_rad)\n",
+    "surfaces['Clad Radius'] = openmc.ZCylinder(R=clad_rad)\n",
+    "\n",
+    "# Set surfaces for Control Rods\n",
```
wbinventor: no need to capitalize "control rods"

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Set half-spaces for geometry building\n",
+    "half_spaces = {}\n",
```
wbinventor: Perhaps rename `half_spaces` as `regions`?

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+    "universes = {}\n",
+    "cells = {}\n",
+    "\n",
+    "for name, mat, in zip(['Fuel Pin', 'Gd Pin'],\n",
+    "                      [materials['Fuel'], materials['Gad']]):\n",
+    "    universes[name] = openmc.Universe(name=name)\n",
+    "    cells[name] = openmc.Cell(name=name)\n",
+    "    cells[name].fill = mat\n",
+    "    cells[name].region = half_spaces['Fuel']\n",
+    "    universes[name].add_cell(cells[name])\n",
```
wbinventor: Add blank lines between each `Universe` instantiation

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+    "    cells[name + ' Clad'] = openmc.Cell(name=name + ' Clad')\n",
```
wbinventor: Could this block include a second nested loop over fuel, clad and water to simplify matters?

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Set constants for the problem and assembly dimensions\n",
```
wbinventor: This entire code block as well as the following block to instantiate `half_spaces` could be greatly simplified by using `openmc.get_rectangular_prism(...)` to instantiate rectangular boxes. We should probably use this throughout our examples to simplify box creation.

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+    "iso_mgxs_lib.domains = [mesh]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we will set the scattering treatment that we wish to use.\n",
+    "\n",
+    "In the `mg-mode-part-ii` notebook, the cross sections were generated with a typical P3 scattering expansion in mind. Now, however, we will use a more advanced technique: OpenMC will directly provide us a histogram of the change-in-angle (i.e., $\\mu$) distribution.\n",
```
wbinventor: Would it be possible to link to this other notebook directly?

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+    "iso_mgxs_lib.histogram_bins = 11"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Ok, we made our isotropic library with histogram-scattering!\n",
+    "\n",
+    "Now why dont we go ahead and create a library to do the same, but with angle-dependent MGXS. That is, we will avoid making the isotropic flux weighting approximation and instead just store a cross section for every polar and azimuthal angle pair.\n",
```
wbinventor: typo: "dont" -> "don't"

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 25,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Move the StatePoint File\n",
+    "ce_spfile = './ce_statepoint.h5'\n",
```
wbinventor: Ditto consider renaming to "statepoint_ce" and "summary_ce"

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+      "  new_tally._mean = data['self']['mean'] / data['other']['mean']\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Allow the API to create our Library, materials, and geometry file\n",
+    "iso_mgxs_file, materials_file, geometry_file = iso_mgxs_lib.create_mg_mode()\n",
+    "\n",
+    "# Tell the materials file what we want to call the multi-group library\n",
+    "materials_file.cross_sections = './mgxs.h5'\n",
```
wbinventor: Ditto I don't think the "./" is necessary here and elsewhere

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+    "\n",
+    "This will be performed by first obtaining the one-group fission rate tally information from our state point files. After we have this information we will re-shape the data to match the original mesh laydown. We will then normalize, and finally create side-by-side plots of all."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 40,
+   "metadata": {
+    "collapsed": false,
+    "scrolled": f
```
wbinventor: Instead of, or in addition to, can you plot the reaction rate errors between the two MG calculations and the "reference" CE calculation? This would better illustrate the points about the gradient in the isotropic case than the fission rates themselves. Of course this might not be very useful if we decide to reduce the number of histories simulated in these examples.

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: I did remove the unnecessary superfluous `surface_id` parameter, the users likely dont need to use that.

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: I still hesitate to go too far with this; I will bring it down to the levels @smharper suggests, but I dont want to go much further.  These notebooks are about the users not the developers. If we want these to be tests for developers, they should be tests, but the will be different code in the end.

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This Notebook illustrates the usage of OpenMC's multi-group calculational mode with the Python API. This example notebook creates and executes the 2-D [C5G7](https://www.oecd-nea.org/science/docs/2003/nsc-doc2003-16.pdf) benchmark model using the `openmc.MGXSLibrary`class to create the supporting data library on the fly."
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "from IPython.display import Image\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "uo2_xsdata.set_total([1.779490E-01, 3.298050E-01, 4.803880E-01, 5.543670E-01,\n",
+    "                      3.118010E-01, 3.951680E-01, 5.644060E-01], temperature=294.)\n",
+    "\n",
+    "uo2_xsdata.set_absorption([8.02480E-03, 3.71740E-03, 2.67690E-02, 9.62360E-02,\n",
+    "                           3.00200E-02, 1.11260E-01, 2.82780E-01], temperature=294.)\n",
+    "\n",
+    "# The scattering matrix is ordered with incoming groups as rows and outgoing groups as columns\n",
+    "# (i.e., above the diagonal is up-scattering).\n",
+    "scatter_matrix = \\\n",
+    "    [[[1.275370E-01, 4.237800E-02, 9.437400E-06, 5.516300E-09, 0.000000E-00, 0.000000E-00, 0.000000E-00],\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "      [0.000000E-00, 3.254520E-01, 1.639500E-03, 1.598200E-09, 0.000000E-00, 0.000000E-00, 0.000000E-00],\n",
+    "      [0.000000E-00, 0.000000E-00, 4.531880E-01, 2.614200E-03, 0.000000E-00, 0.000000E-00, 0.000000E-00],\n",
+    "      [0.000000E-00, 0.000000E-00, 0.000000E-00, 4.571730E-01, 5.539400E-03, 0.000000E-00, 0.000000E-00],\n",
+    "      [0.000000E-00, 0.000000E-00, 0.000000E-00, 1.604600E-04, 2.768140E-01, 9.312700E-03, 9.165600E-09],\n",
+    "      [0.000000E-00, 0.000000E-00, 0.000000E-00, 0.000000E-00, 2.005100E-03, 2.529620E-01, 1.485000E-02],\n",
+    "      [0.000000E-00, 0.000000E-00, 0.000000E-00, 0.000000E-00, 0.000000E-00, 8.494800E-03, 2.650070E-01]]]\n",
+    "scatter_matrix = np.array(scatter_matrix)\n",
+    "scatter_matrix = np.rollaxis(scatter_matrix, 0, 3)\n",
+    "mox43_xsdata.set_scatter_matrix(scatter_matrix)\n",
+    "mox43_xsdata.set_fission([7.62704E-03, 8.76898E-04, 5.69835E-03, 2.28872E-02,\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "mox7_xsdata.set_chi([5.87910E-01, 4.11760E-01, 3.39060E-04, 1.17610E-07,\n",
+    "                     0.000000E-00, 0.000000E-00, 0.000000E-00])\n",
+    "\n",
+    "mox87_xsdata = openmc.XSdata('mox87', groups)\n",
+    "mox87_xsdata.order = 0\n",
+    "mox87_xsdata.set_total([1.830450E-01, 3.367050E-01, 5.005070E-01, 6.061740E-01,\n",
+    "                        5.027540E-01, 9.210280E-01, 9.552310E-01])\n",
+    "mox87_xsdata.set_absorption([9.48620E-03, 4.65560E-03, 3.62400E-02, 1.32720E-01,\n",
+    "                             2.08400E-01, 6.58700E-01, 6.90170E-01])\n",
+    "scatter_matrix = \\\n",
```
nelsonag: Better idea, with just the uo2_data, I now show how one would write a library, but then I end up saying 'but im just going to use a library I already made, called `c5g7.h5` and skip over those other data sets.

- [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "scatter_matrix = np.array(scatter_matrix)\n",
+    "scatter_matrix = np.rollaxis(scatter_matrix, 0, 3)\n",
+    "water_xsdata.set_scatter_matrix(scatter_matrix)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "So far we have not yet set the data for individual materials but have not yet combined them together into a library; we will do that next and then write that library to disk."
```
nelsonag: - [x] OBE

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Instantiate some Macroscopic Data using a name that is the same as that which was used in the\n",
+    "# MGXS Library earlier\n",
+    "uo2_data = openmc.Macroscopic('uo2')\n",
```
nelsonag: Good idea, done.

(but not the cleaner names - the library I am using had the shorter names)

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Instantiate a Materials collection, register all Materials, and export to XML\n",
+    "materials_file = openmc.Materials(materials.values())\n",
+    "\n",
+    "# Set the location of the cross sections file\n",
+    "materials_file.cross_sections = './mgxs.h5'\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "     [m, n, g, o, o, g, o, o, f, o, o, g, o, o, g, n, m],\n",
+    "     [m, n, n, o, o, o, o, o, o, o, o, o, o, o, n, n, m],\n",
+    "     [m, n, n, o, o, o, o, o, o, o, o, o, o, o, n, n, m],\n",
+    "     [m, n, g, o, o, g, o, o, g, o, o, g, o, o, g, n, m],\n",
+    "     [m, n, n, n, o, o, o, o, o, o, o, o, o, n, n, n, m],\n",
+    "     [m, n, n, g, n, o, o, o, o, o, o, o, n, g, n, n, m],\n",
+    "     [m, n, n, n, n, g, n, n, g, n, n, g, n, n, n, n, m],\n",
+    "     [m, n, n, n, n, n, n, n, n, n, n, n, n, n, n, n, m],\n",
+    "     [m, m, m, m, m, m, m, m, m, m, m, m, m, m, m, m, m]]\n",
+    "# Create a containing cell and universe\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "cells['MOX Assembly'] = openmc.Cell(name='MOX Assembly')\n",
+    "cells['MOX Assembly'].fill = lattices['MOX Assembly']\n",
+    "universes['MOX Assembly'] = openmc.Universe(name='MOX Assembly')\n",
+    "universes['MOX Assembly'].add_cell(cells['MOX Assembly'])\n",
+    "    \n",
+    "# Instantiate the Reflector Lattice\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "universes['Reflector Assembly'].add_cell(cells['Reflector Assembly'])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": true
+   },
+   "source": [
+    "Lets now create the core layout in a 3x3 lattice wher each lattice position is one of the assemblies we just defined.\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+   "source": [
+    "root_universe.plot(center=(0., 0., 0.), width=(3 * 21.42, 3 * 21.42), pixels=(500, 500),\n",
+    "                   color_by='material')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "OK, it looks pretty good, lets go ahead and write the file"
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "# Create root Universe\n",
+    "root_universe = openmc.Universe(name='root universe', universe_id=0)\n",
+    "root_universe.add_cell(root_cell)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Before we commit to the geometry, lets view it using the Python API's plotting capability"
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "inactive = 100\n",
+    "particles = 5000\n",
+    "\n",
+    "# Instantiate a Settings object\n",
+    "settings_file = openmc.Settings()\n",
+    "settings_file.batches = batches\n",
+    "settings_file.inactive = inactive\n",
+    "settings_file.particles = particles\n",
+    "\n",
+    "# We dont need a tallies.out file, as we will be working with the StatePoint\n",
```
nelsonag: - [x] OBE, no reason for this here, who cares about the file being present or not

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "\n",
+    "# Export to \"settings.xml\"\n",
+    "settings_file.export_to_xml()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Lets go ahead and execute the simulation! You'll notice that the output for multi-group mode is exactly the same as for continuous-energy. The differences are all under the hood."
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "openmc.run()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Results Visualization\n",
+    "\n",
+    "Now that we have run the simulation, lets look at the fission rate and flux tallies that we tallied."
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The previous Notebook in this series used multi-group mode to perform a calculation with previously defined cross sections.  However, there are times where you would like multi-group data is only a given in some circumstances; in others one would need to calculate cross sections for the specific application (or at least verify the use of cross sections from another application).  \n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "from IPython.display import Image\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+    "# Specify multi-group cross section types to compute\n",
+    "mgxs_lib.mgxs_types = ['total', 'absorption', 'nu-fission', 'fission',\n",
+    "                       'nu-scatter matrix', 'multiplicity matrix', 'chi']"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we must specify the type of domain over which we would like the `Library` to compute multi-group cross sections. The domain type corresponds to the type of tally filter to be used in the tallies created to compute multi-group cross sections. At the present time, the `Library` supports \"material\" \"cell\", \"universe\", and \"mesh\" domain types. In this simple example, we wish to compute multi-group cross sections only for each material and therefore will use a \"material\" domain type.\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+    "\n",
+    "**Note:** By default, the `Library` class will instantiate `MGXS` objects for each and every domain (material, cell, universe, or mesh) in the geometry of interest. However, one may specify a subset of these domains to the `Library.domains` property."
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+   "cell_type": "code",
+   "execution_count": 25,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Move the StatePoint File\n",
+    "ce_spfile = './ce_statepoint.h5'\n",
+    "os.rename('statepoint.' + str(batches) + '.h5', ce_spfile)\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+      "/home/nelsonag/git/openmc/openmc/tallies.py:1837: RuntimeWarning: invalid value encountered in true_divide\n",
+      "  new_tally._mean = data['self']['mean'] / data['other']['mean']\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Create a MGXS File which can then be written to disk\n",
+    "mgxs_file = mgxs_lib.create_mg_library(xs_type='macro', xsdata_names=['fuel', 'zircaloy', 'water'])\n",
+    "\n",
+    "# Write the file to disk using the default filename of `mgxs.h5`\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
+    "\n",
+    "# moderator\n",
+    "water_mg = openmc.Material(name='Water')\n",
+    "water_mg.add_macroscopic(water_macro)\n",
+    "\n",
+    "# Finally, instantiate our Materials object\n",
+    "materials_file = openmc.Materials((fuel_mg, zircaloy_mg, water_mg))\n",
+    "\n",
+    "# Set the location of the cross sections file\n",
+    "materials_file.cross_sections = './mgxs.h5'\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1684 @@
...
...imq2lkZnueFG\nLeQkIlmqWn+dyyRhS8caY9KFswphBZ1a1Z8tOZr8LOTkJTne2h3HMcN9/AuIXoOpMcaYsFrkZMY8\naPjlZQDgU8A24Cz3sRV4OpaFMsYYk7y85Dj2VNXTg57fLiLfxapAxhhjkpuXGscuEalZdUZEDgJ2\nxa5IxhhjkpmXGscVwLMiEpirfDPwy9gVyRhjTDKLGDhEJAMYoKr7ikgrAFW1rrjGGNOMRWyqUtVq\n4Gr3560WNIwxxnjJcXwoIr8TkR4i0i7wiHnJXO5U7k+KyCvxuqYxxpjwvASOi3GmUZ+EM0fVTMDT\n6DoReUpE1onInDrbx4jIAhFZLCI3RTqHqi5R1Uu8XM8YY0zseZlWvSmrno8HHgCeDWxwp2Z/EDga\nKAami8hbQCbw9zqvv1hV/U0Ub4wxJqa8jBy/SkTaBD1vKyK/9nJyVZ0EbKqzeSSw2K1JlAMvAier\n6mxVPbHOw3PQEJHLAqPb169f7/VlxhhjfPLSVHWpqtZM3q6qm4GmzD7bDVgR9LzY3RaSiLQXkUeA\nYSJyc7jjVPUxVR2hqiOKioqaUDxjjDGReBnHkSEiou5siG5TU04Trhlq0eKwMy2q6kacsSTGGGOS\ngJcaxwfASyJypIgcAUwA3m/CNYuB4KXIugOrmnC+GiIyVkQe27JlSzROZ4wxJgQvgeNG4BPgSpze\nVR8Dv2/CNacD/USkj4jkAOcAbzXhfDVUdaKqXta6deuGDzbGGNMoXnpVVQMPuw9fRGQCMBroICLF\nwK2q+qSIXI1Tk8kEnlLVuX7PHeZ6tnSsMcbEWIMLOYlIP5xusoOAmknhVXWP2Bat8WwhJ2OM8Seq\nCznhrL3xMFAJHI4zJuO5xhfPGGNMKvMSOFqo6sc4tZOfVPU24IjYFqtxLDlujDGx5yVwlLqz5C4S\nkatF5FSgY4zL1SiWHDfGmNjzEjh+A+QD/w8YDpyPrcdhjDHNlpdeVdPdH7cDF8W2OE1jvaqMMSb2\nwgYOd+LBsFT1pOgXp2lUdSIwccSIEU2ZEsUYY0wEkWoco3DmlJoAfE3oqUKMMcY0M5ECR2ecqc/P\nBX4OvANMiNZgPWOMMakpbHJcVatU9X1V/SVwALAY+ExErolb6Xyy7rjGGBN7EXtViUiuiJwG/Adn\nnqr7gNfiUbDGsO64
```
nelsonag: - [x]

nelsonag: I end on a general one now
 - [x]

docs/source/examples/mg-mode-part-ii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
nelsonag: I note these are the default values now, so the users know going forward.

- [x]

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1460 @@
...
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "from IPython.display import Image\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: Looks a bit complicated without much savings to handle the differences.

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: It could be, save for regions that use a problem boundary. Sadly thats all but one region.  I wont use two different ways to do that here. When we put a boundary condition option in to get_*_prism, then this can be changed. Sadly I'm hitting the road today for a week and havent decided if my computer is coming or not, so if this is the last item I suggest it goes to a new PR.

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: Yep! markdown for the win! (It wont work in the actual notebook hosted live by Jupyter, but it will on our docs)

nelsonag:
Ok, I'm about to push a commit. I have to add that I wasnt able to locate these guys outside the `docs/` directory. Just adding a relative path in the `*.rst` file wasnt right, and neither was adding the path to the `conf.py` file, surprisingly. As I mentioned above, I am running out of here shortly on travel and so if this is still desired for this PR, it will have to wait a bit (and so consider pulling it out from this PR)

docs/source/examples/mg-mode-part-i.rst
```diff
@@ -0,0 +1,13 @@
+.. _notebook_mg_mode_part_i:
+
+=====================================
+Multi-Group Mode Part I: Introduction
```
wbinventor: I like your explicit and clear naming convention - could you do something similar and rename the M(D)GXS notebook links as "Multi-Group (Delayed) Cross Section Generation: Part XXX"?

wbinventor:
When I build the documentation I see the following warning:

```bash
/home/wbinventor/Documents/NSE-CRPG-Codes/openmc/docs/source/examples/mgxs-part-iv.rst:: WARNING: document isn't included in any toctree
/home/wbinventor/Documents/NSE-CRPG-Codes/openmc/docs/source/pythonapi/openmoc_compatible.rst:: WARNING: document isn't included in any toctree
```

The MGXS Part IV notebook and .rst files need to be removed.

docs/source/examples/mg-mode-part-iii.ipynb
wbinventor: I've added in an option to specify the boundary type on a prism in #824, but I don't think its necessary to use this for this PR.

wbinventor:
I think it would be great to include links to the Jupyter notebooks from *both* the Python API page, as well as from the new dedicated page you've created here. Thoughts?

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1661 @@
...
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# OpenMC simulation parameters\n",
+    "batches = 1000\n",
+    "inactive = 100\n",
```
wbinventor: Reduce the number of inactive batches to 10 - it doesn't even take this many batches to converge the Shannon entropy for this simple fuel assembly. This will help speed things up for this notebook, even if only slightly.

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1502 @@
...
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# OpenMC simulation parameters\n",
+    "batches = 2000\n",
+    "inactive = 200\n",
```
wbinventor: Ditto, I doubt it takes more than about 20 batches to converge the Shannon entropy for this problem, though you could check that to convince yourself of this if you wish. Reduce the inactive batches to 20 or 30 to shave off a little runtime for this notebook.

wbinventor:
I tested all of the notebooks and all of the successfully ran to completion. I still maintain that the number of histories should be reduced for the "mg-mode-part-ii.ipynb" notebook (and perhaps also "mg-mode-part-iii.ipynb"). It took 25 minutes to run Part II on my laptop (and over 10 mins to run Part III). When I reduced the number of batches from 1000 to 200 (and inactives from 100 to 10) for Part II the keff bias was reported as:

```bash
Continuous-Energy keff = 1.165915
Multi-Group keff = 1.162943
P0 bias [pcm]: 380.9
P3 bias [pcm]: 297.1
Mixed Scattering bias [pcm]: 52.0
```

Likewise, I reduced the number of batches from 2000 to 500 (and inactives from 200 to 20) for Part III the keff bias was reported as:

```bash
Isotropic to CE Bias [pcm]: 1262.6
Angle to CE Bias [pcm]: 179.4
```

Although these aren't quite as close as what you obtain with 4-5x more histories, I think that this more than a sufficient number of histories to showcase the various scattering and angular formulations. While I agree with @nelsonag that the notebooks are written for the user, they are *also* for developers to ensure that new code will successfully run within a more realistic workflow than that which our test suite provides. A nearly half hour runtime for a single notebook with little benefit (*i.e.*, the same features are showcased either with more or fewer histories) is a large burden for manual testing. Thoughts @nelsonag @smharper @paulromano?

nelsonag:
I still maintain that these are only for users. As for us the developers, these technically only need to be run once every release. 

If you want it more than that, as a developer aid, then the right way is to put them in the test suite with their own harness. When you do that there are changes you have to make anyways, like removing notebook specific commands where applicable, so you also have the opportunity to modify the history count. 

My main point here is that developer needs are separable from user needs.

smharper:
More than the runtime, I'm disturbed by the fact that the mixed scattering run gives less bias than P3.  I don't want a user to look at this and think that mixed is "supposed" to be more accurate than P3.  Maybe we should run a problem that's easier to converge like a homogeneous medium with a vacuum BC and more energy groups.  I hope that would show the expected convergence behavior.

wbinventor:
@smharper you certainly cannot draw any conclusions about the relative competitive advantages of P0 vs. P3 vs. mixed scattering from the results here - either those @nelsonag ran or the results I gave for fewer particles. Even with lots of histories, the 2-sigma uncertainty on the continuous eigenvalue is much larger than the variation for three MG mode eigenvalues. In addition, there is most likely cancellation of error going on between the (a) energy condensation, (b) angular condensation and (c) scattering approximations. From everything I've seen before, (a) and (b) largely cancel each other for the 2-group structure used here for PWR models.

In any case, regardless of the number of histories used, I agree with @smharper that we don't want anyone using our notebook results to justify their use of particular approximations. We should add caveats to each notebook warning not to use our notebooks as a basis for particular selection of approximations (perhaps my MGXS generation notebooks need this as well).

nelsonag:
@smharoer that sounds like what the final note in the second notebook is
trying to combat against.

wbinventor:
Oh sorry I overlooked your note about this at the end of the notebooks @nelsonag. I think that it is a fine warning, no need to elaborate IMO.

smharper:
Yeah, I don't know if this is a great problem to use if we want to demonstrate the difference between various scattering approximations.  Look at the change to the biases in the 2nd notebook caused by the last commit:
```
-      "P3 bias [pcm]: -156.7\n",
-      "P0 bias [pcm]: 68.0\n"
+      "P3 bias [pcm]: 31.4\n",
+      "P0 bias [pcm]: 158.4\n"
```

As best I can tell, that huge swing in bias is only caused by a change in B10 content.  So there could be a lot of tricky cancellation of error in this problem.

wbinventor:
I think the question is whether we want to show the *impact* of the different scattering approximations, or the *existence* of these various approximations as features in the API. I think that the notebook is meant to show the latter as evident from our preceding discussion, though if it showed both that would be nice. This would be simple to do by adding an assembly-width water reflector to one side of the fuel assembly with vacuum boundary conditions.

smharper:
Okay, if we focus this as just a how-to then the lattice problem in part 2 is totally fine by me... as long as we make it super duper clear that it's only a rough how-to.  It just seems a bit misleading to me that we show "bias" figures at all.  When I see something like that, I get focused on the details of which method performed "better" on that bias figure.

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Create Geometry and set root Universe\n",
+    "geometry = openmc.Geometry()\n",
+    "geometry.root_universe = root_universe\n",
```
paulromano: These days I prefer `geometry = openmc.Geometry(root_universe)`

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,707 @@
...
+   "source": [
+    "# Create a 7-group structure with arbitrary boundaries (the specific boundaries are unimportant)\n",
+    "groups = openmc.mgxs.EnergyGroups(np.logspace(-5, 7, 8))\n",
+    "\n",
+    "uo2_xsdata = openmc.XSdata('uo2', groups)\n",
+    "uo2_xsdata.order = 0\n",
+    "\n",
+    "# When setting the data let the object know you are setting the data for a temperature of 294K.\n",
+    "uo2_xsdata.set_total([1.77949E-1, 3.29805E-1, 4.80388E-1, 5.54367E-1,\n",
+    "                      3.11801E-1, 3.95168E-1, 5.64406E-1], temperature=294.)\n",
```
paulromano: This appears in the C5G7 spec as transport cross section -- might be good to mention that. Also, probably good to point out that values are given in order of increasing group / decreasing energy.

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "                        1.78084E-2, 8.30348E-2, 2.16004E-1], temperature=294.)\n",
+    "\n",
+    "uo2_xsdata.set_nu_fission([2.005998E-2, 2.027303E-3, 1.570599E-2, 4.518301E-2,\n",
+    "                           4.334208E-2, 2.020901E-1, 5.257105E-1], temperature=294.)\n",
+    "\n",
+    "uo2_xsdata.set_chi([5.87910E-1, 4.11760E-1, 3.39060E-4, 1.17610E-7,\n",
+    "                    0.00000E-0, 0.00000E-0, 0.00000E-0], temperature=294.)\n",
+    "\n",
+    "# The scattering matrix is ordered with incoming groups as rows and outgoing groups as columns\n",
+    "# (i.e., above the diagonal is up-scattering).\n",
```
paulromano: Wouldn't it be down-scattering about the diagonal?

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "# Generate 2-D C5G7 Problem Input Files"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To build the actual 2-D model, we will first begin by creating the `materials.xml` file.\n",
+    "\n",
+    "First we need to define materials that will be used in the problem. In other notebooks, either `openmc.Nuclides` or `openmc.Elements` were created at the equivalent stage. We can do that in multi-group mode as well.  However, multi-group cross-sections are sometimes provided as macroscopic cross-sections; the C5G7 benchmark data are macroscopic. In this case, we can instead use `openmc.Macroscopic` objects to in-place of `openmc.Nuclides` or `openmc.Elements`.\n",
```
paulromano: `openmc.Nuclide` and `openmc.Element` .. if you really want to make it plural, put the -s outside of backticks.

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "\n",
+    "`openmc.Macroscopic`, unlike `openmc.Nuclides` and `openmc.Elements` objects, do not need to be provided enough information to calculate number densities, as no number densities are needed.\n",
```
paulromano: Same thing here, no s is class names

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,707 @@
...
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# For every cross section data set in the library, assign an openmc.Macroscopic object to a material\n",
+    "materials = {}\n",
+    "for xs in ['uo2', 'mox43', 'mox7', 'mox87', 'fiss_chamber', 'guide_tube', 'water']:\n",
+    "    materials[xs] = openmc.Material(name=xs)\n",
+    "    materials[xs].set_density('macro', 1.)\n",
+    "    materials[xs].add_macroscopic(openmc.Macroscopic(xs))"
```
paulromano: Maybe just use the shorthand `add_macroscopic(xs)`

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1661 @@
...
+    "template_x = np.array([5, 8, 11, 3, 13, 2, 5, 8, 11, 14, 2, 5, 8,\n",
+    "                       11, 14, 2, 5, 8, 11, 14, 3, 13, 5, 8, 11])\n",
+    "template_y = np.array([2, 2, 2, 3, 3, 5, 5, 5, 5, 5, 8, 8, 8, 8,\n",
+    "                       8, 11, 11, 11, 11, 11, 13, 13, 14, 14, 14])\n",
+    "\n",
+    "# Initialize an empty 17x17 array of the lattice universes\n",
+    "universes = np.empty((17, 17), dtype=openmc.Universe)\n",
+    "\n",
+    "# Fill the array with the fuel pin and guide tube universes\n",
+    "universes[:,:] = fuel_pin_universe\n",
```
paulromano: `universes[:, :]`

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1661 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Instantiate a 2-group EnergyGroups object\n",
+    "groups = openmc.mgxs.EnergyGroups()\n",
```
paulromano: `groups = openmc.mgxs.EnergyGroups([0., 0.625, 20.0e6])`

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1661 @@
...
+    "# Instantiate our Macroscopic Data\n",
+    "fuel_macro = openmc.Macroscopic('fuel')\n",
+    "zircaloy_macro = openmc.Macroscopic('zircaloy')\n",
+    "water_macro = openmc.Macroscopic('water')\n",
+    "\n",
+    "# Now re-define our materials to use the Multi-Group macroscopic data\n",
+    "# instead of the continuous-energy data.\n",
+    "# 1.6 enriched fuel UO2\n",
+    "fuel_mg = openmc.Material(name='UO2')\n",
+    "fuel_mg.add_macroscopic(fuel_macro)\n",
```
paulromano: Again, you don't really need to create a `Macroscopic` object above. Would be shorter to just pass the string here

docs/source/examples/mg-mode-part-ii.ipynb
```diff
@@ -0,0 +1,1661 @@
...
...BCgs4isAG5S1SdF5CqckkwmME5Vv4/03CGuZ0vHGmNMnDW5\nkJOIDMTpJjsEqJ8UXlX7xze06NlCTsYYE5mYLuSEs/bGI0ANcCjOmIxnog/PGGNMS+YlceSr6oc4\npZNlqvpX4LD4hhUdaxw3xpj485I4Kt1Zcn8QkatE5CSgS5zjioo1jhtjTPx5SRy/AQqAXwN7Aedi\n63EYY0yr5aVX1TT3x63AhfENp3msV5UxxsRfyMThTjwYkqqeEPtwmkdVJwITS0pKmjMlijHGmDDC\nlThG4cwpNQH4iuBThRhjjGllwiWOnXCmPj8TOAt4C5gQq8F6xhhjWqaQjeOqWquq76jq+cC+wCLg\nExG5OmHRRci64xpjTPyF7VUlIrkicjLwLM48VfcDryQisGhYd1xjjIm/cI3jTwPDgLeBm1V1Tqhj\njTHGtB7h2jjOBcqBQcCvRerbxgVQVW0X59iMMcakoJCJQ1W9DA5MKTaOwxhj4q/FJYdwrI3DGGPi\nL60ShzHGmPizxGGMMSYiljiMMcZExBKHMcaYiKRV4rCR48YYE39plTisV5UxxsRfWiUOY4wx8WeJ\nwxhjTEQscRhjjImIJQ5jjDERscRhjDEmIpY4jDHGRCStEoeN4zDGmPhLq8Rh4ziMMSb+0ipxGGOM\niT9LHMYYYyJiicMYY0xELHEYY4yJiCUOY4wxEbHEYYwxJiKWOIwxxkTEEocxxpiIiKomO4aYE5Et\nwIJkxxEHnYENyQ4iTtL13uy+Wp50vbem7quPqhZ7OVFWbOJJOQtUtSTZQcSaiExPx/uC9L03u6+W\nJ13vLZb3ZVVVxhhjImKJwxhjTETSNXE8luwA4iRd7wvS997svlqedL23mN1XWjaOG2OMiZ90LXEY\nY4yJE0scxhhjImKJwxhjTERaVeIQkUNEZIqIPCoihyQ7nlgSkV3d+3pJRH6V7HhiRUT6i8iTIvJS\nsmOJhXS7H590ff9B+n5uiMiB7j09ISKfR/LaFpM4RGSciKwTkTkB248RkQUiskhEbmziNApsBfKA\nFfGKNVKxuDdVnaeqlwOnAykxeClG97VEVS+Ob6TNE8l9toT78YnwvlLu/RdOhO/NlPzcCCbCv9kU\n92/2JvB0RBdS1RbxAA4C9gTm+G3LBBYD/YEcYBYwBNjN/WX4P7oAGe7rugL
```
paulromano: Personally I wouldn't even both mentioning the setting within settings.xml; mentioning the attribute of `openmc.Settings` is enough (if someone is advanced enough to be doing that, let's hope they're using the Python API)

docs/source/examples/mg-mode-part-iii.ipynb
```diff
@@ -0,0 +1,1502 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 40,
+   "metadata": {
+    "collapsed": false,
+    "scrolled": false
+   },
+   "outputs": [
+    {
+     "dat
```
paulromano: You really should use a diverging colormap so you can see what's greater than one and what's less than one.

paulromano: Would be good to add a `plt.colorbar`

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

docs/source/examples/index.rst
```diff
@@ -0,0 +1,26 @@
...
+
+.. toctree::
+    :maxdepth: 1
+
+    post-processing
+    pandas-dataframes
+    tally-arithmetic
+    mgxs-part-i
+    mgxs-part-ii
+    mgxs-part-iii
```
nelsonag: part-iv is replaced by mg-mode-part-ii

paulromano: If that's the case, then you should remove mgxs-part-iv.ipynb and mgxs-part-iv.rst

docs/source/examples/mg-mode-part-i.rst
nelsonag: - [x]

docs/source/examples/index.rst
```diff
@@ -0,0 +1,26 @@
+.. _examples:
+
+=========================
+Example Jupyter Notebooks
```
nelsonag: - [x]

docs/source/examples/index.rst
```diff
@@ -0,0 +1,26 @@
...
+=========================
+
+The following series of Jupyter Notebooks provide examples for usage of OpenMC
+features via the `Python API`_.
```
nelsonag: - [x]

docs/source/examples/index.rst
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
```diff
@@ -0,0 +1,926 @@
...
+    "import openmc\n",
+    "\n",
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will now create the multi-group library using data directly from Appendix A of the C5G7 benchmark documentation. All of the data below will be created at 294K, consistent with the benchmark.\n",
```
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: Whoops, yes.

-[x]

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-ii.ipynb
nelsonag: - [x]

docs/source/examples/mg-mode-part-iii.ipynb
nelsonag: - [x]

nelsonag: - [x]

nelsonag:
@paulromano just finished dispositioning all of your comments. 
For @smharper and @wbinventor you'll be pleased to see I reduced the bins. I didnt go as far as you suggested, but close.  These things take only a few minutes in total to run on my machine.

docs/source/examples/mg-mode-part-i.ipynb
paulromano: I don't see this one changed.

smharper:
Sorry for the sporradic review, @nelsonag.  I just got back from travel and I'm still playing the catch-up game.

I'm pretty happy with most of the content of the notebooks, but I'm still a bit concerned about the runtime and the pcm bias figures.  It takes me 11 minutes to run the part 2 notebook which is a non-trivial addition to development workload.  What do we gain by running so many particles?  We don't need so many neutrons to show how to set up the inputs.  We probably don't even need so many neutrons to show plots that have the right general characteristics.  If we are using those neutrons because we want to resolve the differences between the scattering models, then I think we are using the wrong geometry and number of groups.  Similarly, I don't like the idea of presenting pcm bias figures because I'm worried that users will read those bias figures as a proof of scattering model accuracy despite the warning at the bottom.  Thoughts, @wbinventor, @paulromano?  Am I being a little too cautious there?

Other than those two quibbles, the notebooks look pretty good.  They have a nice ramp-up in complexity and by the third notebook there is a compelling case for the flexibility of MGXS + MG.

docs/source/examples/mg-mode-part-i.ipynb
nelsonag: Its in the markdown text above the code block

nelsonag:
I'm happy enough with the results of mg-mode-ii at 600 batches, and so I'll be pushing that soon.

docs/source/examples/mg-mode-part-i.ipynb
paulromano: My bad, thanks for pointing that out!

smharper:
Merged.  Thanks, @nelsonag!

PR #834, opened 2017-03-18
Added version information to MGXS Library format and files

nelsonag:
This PR simply adds filetype and version information to the MGXS Library files and the format specification.


wbinventor:
Looks good to me!

PR #854, opened 2017-04-10
Cloning of geometry and materials

wbinventor:
This PR introduces new `clone()` methods to the `Material`, `Surface`, `Cell`, `Universe`, `Lattice` and `Geometry` classes. These new methods are essentially a wrapper around the `__deepcopy__()` method, except they assign a unique ID to the clone (whereas `__deepcopy__()` retains the original ID). Furthermore, any portion of the CSG tree below the clone is also cloned, and hence has unique IDs. 

This feature is intended to support @GiudGiud and @tjlaboss with their research. Based on their use cases, I think this could be a more broadly useful feature. Of course @paulromano will note that we have previously discussed on at least one occasion whether the `__deepcopy__()` methods should assign unique IDs as the new `clone()` methods do. I still stand by my original view that `__deepcopy__()` should return a genuine deep copy, but of course I won't be surprised if this PR re-opens this conversation :-)

### DIFF
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -504,6 +504,20 @@ class Cell(object):
 
         return universes
 
+    def clone(self):
+        """Create a copy of this cell with a new unique ID, and clones
+        the cell's region and fill."""
+
+        clone = copy.deepcopy(self)
+        clone.id = None
+
+        if self.region is not None:
+            clone.region = self.region.clone()
+        if self.fill is not None:
+            clone.fill = self.fill.clone()
+
+        return clone
+    
     def create_xml_subelement(self, xml_element):
         element = ET.Element("cell")
         element.set("id", str(self.id))
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -505,13 +505,13 @@ class Cell(object):
 
         return universes
 
-    def clone(self, memoize=None):
+    def clone(self, memo=None):
         """Create a copy of this cell with a new unique ID, and clones
         the cell's region and fill.
 
         Parameters
         ----------
-        memoize : defaultdict(dict) or None
+        memo : defaultdict(dict) or None
             A nested dictionary of previously cloned objects. This parameter
             is used internally and should not be specified by the user.
 
@@ -522,25 +522,26 @@ class Cell(object):
 
         """
 
-        if memoize is None:
-            memoize = defaultdict(dict)
+        if memo is None:
+            memo = defaultdict(dict)
 
         # If no nemoize'd clone exists, instantiate one
-        if self.id not in memoize['cells']:
+        if self.id not in memo['cells']:
             clone = deepcopy(self)
             clone.id = None
             if self.region is not None:
-                clone.region = self.region.clone(memoize)
+                clone.region = self.region.clone(memo)
             if self.fill is not None:
                 if self.fill_type == 'distribmat':
-                    clone.fill = [fill.clone(memoize) for fill in self.fill]
+                    clone.fill = [fill.clone(memo) if fill is not None else None
+                                  for fill in self.fill]
                 else:
-                    clone.fill = self.fill.clone(memoize)
+                    clone.fill = self.fill.clone(memo)
 
             # Memoize the clone
-            memoize['cells'][self.id] = clone
+            memo['cells'][self.id] = clone
 
-        return memoize['cells'][self.id]
+        return memo['cells'][self.id]
     
     def create_xml_subelement(self, xml_element):
         element = ET.Element("cell")
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict, Iterable
+from collections import OrderedDict, Iterable, defaultdict
 from copy import deepcopy
 from math import cos, sin, pi
 from numbers import Real, Integral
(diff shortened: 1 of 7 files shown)


openmc/region.py
```diff
@@ -221,6 +222,12 @@ def apply_operator(output, operator):
         # at the end
         return output[0]
 
+    @abstractmethod
+    def clone(self):
+        """Create a copy of this region - each of the surfaces in the
+        region's nodes will be cloned and will have new unique IDs."""
+        return False
```
paulromano: The correct thing to do in this situation is to raise [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError).

openmc/geometry.py
```diff
@@ -497,3 +498,11 @@ def determine_paths(self):
 
         # Recursively traverse the CSG tree to count all cell instances
         self.root_universe._determine_paths()
+
+    def clone(self):
+        """Create a copy of this geometry with new unique IDs for all of its
+        enclosed materials, surfaces, cells, universes and lattices."""
+
+        clone = deepcopy(self)
+        clone.root_universe = deepcopy(self.root_universe)
```
paulromano: `clone.root_universe = self.root_universe.clone()`?

wbinventor:
Thanks for the quick review @paulromano. I certainly overlooked the memoization issue. My latest commits add in memoization, as well as expand the docstrings with boths `Parameters` (for memoization) and `Returns` blocks.

openmc/geometry.py
wbinventor: Oops, of course :-)

openmc/region.py
wbinventor: I agree. I stubbed this out when implementing this and forgot to properly implement it.

wbinventor:
Also before I forget, I thought I might mention to @nelsonag that perhaps this functionality could be used to simplify the process of creating mesh geometries for MG mode.

nelsonag:
The mesh geometry generator uses a lattice vice cuboidal cells, so it's
already pretty simple. What did you have in mind?

wbinventor:
Ah, my recollection must have failed me then, I thought there was some more involved cell/universe copying involved.

openmc/cell.py
```diff
@@ -504,6 +505,43 @@ def get_all_universes(self):
 
         return universes
 
+    def clone(self, memoize=None):
```
paulromano: I'd prefer the argument be called `memo` to match the interface for `deepcopy`.

wbinventor: Sure, good point.

openmc/cell.py
```diff
@@ -504,6 +505,43 @@ def get_all_universes(self):
...
+
+        # If no nemoize'd clone exists, instantiate one
+        if self.id not in memoize['cells']:
+            clone = deepcopy(self)
+            clone.id = None
+            if self.region is not None:
+                clone.region = self.region.clone(memoize)
+            if self.fill is not None:
+                if self.fill_type == 'distribmat':
+                    clone.fill = [fill.clone(memoize) for fill in self.fill]
```
paulromano: It's possible to have `None` in the fill list, so this should be:
```Python
clone.fill = [fill.clone(memo) if fill is not None else None
              for fill in self.fill]

Comment (wbinventor) [inline openmc/cell.py]: I forgot that `None` could be used to fill a distribmat.

**Comment by paulromano** [comment]
You could probably simplify the implementation by just using the objects as keys directly in the `memo` dictionary. It looks like all our classes are hashed properly except for surfaces. If we add a `__hash__` method to `Surface`, then you should just be able to do:
```Python
if self not in memo:
    ...
return memo[self]
```

wbinventor:
I liked your idea @paulromano and revised the `memo` parameters.

GiudGiud:
Could there be an option not to clone the materials but keep the old materials?
While it is less general, it will save memory and save time (2s out of 70 @#833 , I checked on a degenerate assembly) when loading MGXS from openmc to openmoc

openmc/material.py
```diff
@@ -799,6 +799,35 @@ def get_nuclide_atom_densities(self):
...
+
+        Returns
+        -------
+        clone : openmc.Material
+            The clone of this material
+
+        """
+
+        if memo is None:
+            memo = dict
```
paulromano: `memo = {}`

openmc/lattice.py
```diff
@@ -414,6 +415,57 @@ def find(self, point):
...
+            clone = deepcopy(self)
+            clone.id = None
+
+            if self.outer is not None:
+                clone.outer = self.outer.clone(memo)
+
+            # Clone all unique universes in the lattice
+            univ_clones = self.get_unique_universes()
+            for univ_id in univ_clones:
+                univ_clones[univ_id] = univ_clones[univ_id].clone(memo)
```
paulromano:
```
univ_clones = {univ_id: u.clone(memo) for univ_id, u in
               self.get_unique_universes()}
```
would be a bit more concise

paulromano: Come to think of it, why does this need to be done up front? You can just call `Universe.clone(memo)` below directly and it will be handled correctly.

openmc/material.py
wbinventor: Oops, of course

openmc/lattice.py
wbinventor: I've revised the loop to simply call `Universe.clone(memo)` now. I didn't do this before, since I didn't memoize in the first place.

paulromano:
Thanks @wbinventor!

PR #860, opened 2017-04-11
Bug fix for consistent scattering matrix with nuclides

wbinventor:
This PR fixes a bug that arises when tallying multi-group scattering matrices by nuclides for the `'consistent'` formulation. In this case, the current implementation produces an array that is length N**2 rather than N for N nuclides in the nuclides dimension of the tally results array. The reason is that tally arithmetic performs a tensor product between the `AggregateNuclide`s and `Nuclides` in the two derived tally operands used for the consistent formulation, when in fact it should use an entrywise product.

I've fixed this by over-riding the nuclides in the tally operands to be identical to force tally arithmetic to use an entrywise product (the default when the two operands have identical nuclides). A cleaner approach might be to call the `Tally.hybrid_product(...)` method and enforce an entrywise product over nuclides with the `nuclide_product='entrywise'` parameter, but unfortunately this didn't work for me. Rather than debugging the `Tally.hybrid_product(...)` method (which can be quite challenging), I've taken the more straightforward approach given in this PR.


nelsonag:
Looks good to me, thanks @wbinventor!

PR #864, opened 2017-04-20
Check for void materials in tracklength tallies

smharper:
Sometimes in a tracklength tally we would use `p % material` without checking for a `MATERIAL_VOID`.  This PR fixes that bug (and trims back some of the tally.F90 code to 80 columns).


wbinventor:
This looks straightforward enough to me. Thanks for the bug fix @smharper!

PR #877, opened 2017-05-11
Fix for OpenMOC compatibility for generic planar surfaces

wbinventor:
This short PR fixes two issues that I discovered in the OpenMOC compatibility module while helping @tjlaboss convert an OpenMC model of the TREAT reactor for modeling with OpenMOC. The primary issue is that the halfspace convention used for generic planar surfaces differs between OpenMC and OpenMOC. The second bug fix included in this PR fixes an issue when converting a rectangular lattice from OpenMOC into OpenMC. @paulromano do you have time for a speedy review of this code?

### DIFF
--- a/openmc/universe.py
+++ b/openmc/universe.py
@@ -76,10 +76,17 @@ class Universe(object):
             return False
         elif self.name != other.name:
             return False
-        elif self.cells != other.cells:
+        elif len(self.cells) != len(other.cells):
             return False
-        else:
-            return True
+
+        # Check each cell
+        for cell_id in self.cells:
+            if cell_id not in other.cells:
+                return False
+            if self.cells[cell_id] != other.cells[cell_id]:
+                return False
+
+        return True
 
     def __ne__(self, other):
         return not self == other
--- a/openmc/universe.py
+++ b/openmc/universe.py
@@ -76,17 +76,10 @@ class Universe(object):
             return False
         elif self.name != other.name:
             return False
-        elif len(self.cells) != len(other.cells):
+        elif dict.__eq__(self.cells, other.cells):
             return False
-
-        # Check each cell
-        for cell_id in self.cells:
-            if cell_id not in other.cells:
-                return False
-            if self.cells[cell_id] != other.cells[cell_id]:
-                return False
-
-        return True
+        else:
+            return True
 
     def __ne__(self, other):
         return not self == other
--- a/openmc/universe.py
+++ b/openmc/universe.py
@@ -76,7 +76,7 @@ class Universe(object):
             return False
         elif self.name != other.name:
             return False
-        elif dict.__eq__(self.cells, other.cells):
+        elif dict.__ne__(self.cells, other.cells):
             return False
         else:
             return True
--- a/openmc/lattice.py
+++ b/openmc/lattice.py
@@ -56,11 +56,11 @@ class Lattice(object):
             return False
         elif self.name != other.name:
             return False
-        elif self.pitch != other.pitch:
+        elif np.any(self.pitch != other.pitch):
             return False
         elif self.outer != other.outer:
             return False
-        elif self.universes != other.universes:
+        elif np.any(self.universes != other.universes):
             return False
         else:
             return True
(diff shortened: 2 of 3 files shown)


wbinventor:
I should note for @tjlaboss and @bforget that with these minor bug fixes, the TREAT model can be exported to OpenMOC in all its gory heterogeneous detail.

paulromano:
It seems to me that the half-space convention is not quite reversed, but just that the surface is [defined](https://github.com/mit-crpg/OpenMOC/blob/develop/src/Surface.h#L370) as `Ax + By + Cz + D = 0` rather than `Ax + By + Cz - D = 0` as in OpenMC. Also, [this](https://mit-crpg.github.io/OpenMOC/methods/constructive_solid_geometry.html#arbitrary-plane) looks out of date.

paulromano:
Shouldn't the solution just be to reverse the sign on the D coefficient when converting between models?

wbinventor:
Thanks for the advice about reversing the D coefficient. Just reverted to that alternative approach instead.

openmc/universe.py
```diff
@@ -76,10 +76,17 @@ def __eq__(self, other):
...
             return False
-        else:
-            return True
+
+        # Check each cell
+        for cell_id in self.cells:
+            if cell_id not in other.cells:
+                return False
+            if self.cells[cell_id] != other.cells[cell_id]:
+                return False
```
paulromano: What was the motivation for the changes in this file?

wbinventor: Well tbh I cannot fully defend these changes and am happy to remove them. But during the process of debugging our issue with TREAT, I did a lot of comparisons of the OpenMC geometry before and after conversion to OpenMOC (i.e, the original compared to a version converted to OpenMOC and *back* to OpenMC). Although the geometries were identical (as evidenced by plots and tallied results), this equality method returned false for a few of the universes. The comparison of the dictionaries of cells returned false, while this new more elaborate comparison verified that the cells were identical. Any idea what would cause this behavior when comparing dictionaries of cells?

paulromano: I was curious about this so I dug a little bit. [Here](https://github.com/python/cpython/blob/master/Objects/dictobject.c#L2692-L2737) is the Python `dict.__eq__` implementation. It checks the length of each, then compares keys/values (much the same as you are doing here). However, `Universe.cells` is an `OrderedDict` -- the check for equality on those also [accounts for the order](https://github.com/python/cpython/blob/master/Lib/collections/__init__.py#L291). So, my guess is that when you were converting back and forth, the order of cells ended up different. If you want to do an order-insensitive equality check here, you could just do:
```Python
dict.__eq__(self.cells, other.cells)
```
Can you check if that works for the TREAT model conversion?

wbinventor: Yes, your much simpler approach does indeed work, thanks! Do note that in the process, I also discovered bugs in the `Lattice.__eq__(...)` method - namely, when comparing the `pitch`, `lower_left` and `universes` attribute collections - and fixed them with the `numpy.any()` method.

nelsonag:
Does the EqualityMixin module help here? Any lessons learned to incoporate
therein?

wbinventor:
I'm not sure - @paulromano should we make `Lattice` and `Universe` inherit `EqualityMixIn`?

paulromano:
No, the equality mixin won't help because it will still do a normal `__eq__` check on all the attributes, which won't work for the ordered dictionaries we use here. We could in theory change the mixin so that it does the less rigorous check if the objects being compared are `OrderedDict`s though.

openmc/universe.py
```diff
@@ -76,7 +76,7 @@ def __eq__(self, other):
             return False
         elif self.name != other.name:
             return False
-        elif self.cells != other.cells:
+        elif dict.__eq__(self.cells, other.cells):
```
paulromano: `not dict.__eq__` or `dict.__ne__`?

wbinventor: Geez, this is embarrassing! My mind must have been in another place.

PR #878, opened 2017-05-15
Function to generate square cylinder region

samuelshaner:
This PR implements a `get_square_cylinder(...)` function in the Python API to generate a square cylinder region similar to the functions used to generate rectangular and hexagonal prisms.  As a check, I made a pin cell with a square cylinder fuel pin:

plot_1

The documentation has been updated to note this additional function.

### DIFF
--- a/docs/source/pythonapi/base.rst
+++ b/docs/source/pythonapi/base.rst
@@ -102,7 +102,6 @@ respectively.
 
    openmc.get_hexagonal_prism
    openmc.get_rectangular_prism
-   openmc.get_square_cylinder
 
 .. _pythonapi_tallies:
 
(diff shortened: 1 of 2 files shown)


paulromano:
@samuelshaner What do you think about just having an optional argument on `get_rectangular_prism` called `corner_radius` that would achieve the same effect?

samuelshaner:
I'm fine with that. I'll add it to `get_hexagonal_prism` as well.

samuelshaner:
I've updated the `get_rectangular_prism` and `get_hexagonal_prism` functions to now take an optional `corner_radius` argument. Here are some plots of rounded rectangular and hexagonal prisms pin cells:

d03916cc-396e-11e7-9bf8-5434082cb347
plot_1

paulromano:
Maybe let's make `corner_radius` the last argument of those functions to preserve the order of the existing arguments?

samuelshaner:
I'm fine with that. I just pushed that edit.

paulromano:
Thanks @samuelshaner!

PR #882, opened 2017-05-19
Fix Multipole module

liangjg:
The multipole python api module has several bugs that affect the loading of wmp library or xs evaluation from multipole data.
```python
import openmc.data
nuc_wmp = openmc.data.WindowedMultipole.from_hdf5("./wmp/092235.h5")

(sigT, sigA, SigF) = nuc_wmp(E=0.001, T=300)
AttributeError: module 'numpy' has no attribute 'erf'

(sigT, sigA, SigF) = nuc_wmp(E=2240, T=300)
sigT, sigA, SigF
(array(-4954886358.87597),
 array(-171244680.68258008),
 array(-1001055628.4242938))
```
For the second example where the xs are nonsense, it was found the python code to evaluate xs has inconsistencies with the Fortran code.

Another fix is for reading or evaluating non-fissionable multiple data for both fortran and python codes. If processing a nonfissionable nuclide, the code is likely crashing since residues and curvefit data for fission xs do not exist. That didn't happen because currently all the nuclides in our wmp library are fissionable (fissionable = 1, even though they are actually not!). I've been generating more wmp libraries which are mostly non-fissionable, these modifications have been verified using those libraries.

### DIFF
--- a/openmc/data/multipole.py
+++ b/openmc/data/multipole.py
@@ -1,8 +1,8 @@
 from numbers import Integral, Real
+from math import exp, erf, pi, sqrt
 
 import h5py
 import numpy as np
-from math import exp, erf, pi
 from six import string_types
 
 from . import WMP_VERSION
@@ -102,7 +102,7 @@ def _broaden_wmp_polynomials(E, dopp, n):
         The value of each Doppler-broadened curvefit polynomial term.
 
     """
-    sqrtE = np.sqrt(E)
+    sqrtE = sqrt(E)
     beta = sqrtE * dopp
     half_inv_dopp2 = 0.5 / dopp**2
     quarter_inv_dopp4 = half_inv_dopp2**2
@@ -124,7 +124,7 @@ def _broaden_wmp_polynomials(E, dopp, n):
     factors[0] = erf_beta / E
     factors[1] = 1.0 / sqrtE
     factors[2] = (factors[0] * (half_inv_dopp2 + E)
-                  + exp_m_beta2 / (beta * np.sqrt(pi)))
+                  + exp_m_beta2 / (beta * sqrt(pi)))
 
     # Perform recursive broadening of high order components. range(1, n-2)
     # replaces a do i = 1, n-3.  All indices are reduced by one due to the
@@ -528,8 +528,8 @@ class WindowedMultipole(EqualityMixin):
         # Bookkeeping
 
         # Define some frequently used variables.
-        sqrtkT = np.sqrt(K_BOLTZMANN * T)
-        sqrtE = np.sqrt(E)
+        sqrtkT = sqrt(K_BOLTZMANN * T)
+        sqrtE = sqrt(E)
         invE = 1.0 / E
         dopp = self.sqrtAWR / sqrtkT
 
@@ -537,7 +537,7 @@ class WindowedMultipole(EqualityMixin):
         # the 1-based vs. 0-based indexing.  Similarly startw needs to be
         # decreased by 1.  endw does not need to be decreased because
         # range(startw, endw) does not include endw.
-        i_window = int(np.floor((sqrtE - np.sqrt(self.start_E)) / self.spacing))
+        i_window = int(np.floor((sqrtE - sqrt(self.start_E)) / self.spacing))
         startw = self.w_start[i_window] - 1
         endw = self.w_end[i_window]
 
@@ -621,7 +621,7 @@ class WindowedMultipole(EqualityMixin):
             # At temperature, use Faddeeva function-based form.
             for i_pole in range(startw, endw):
                 Z = (sqrtE - self.data[i_pole, _MP_EA]) * dopp
-                w_val = _faddeeva(Z) * dopp * invE * np.sqrt(pi)
+                w_val = _faddeeva(Z) * dopp * invE * sqrt(pi)
                 if self.formalism == 'MLBW':
                     sigT += ((self.data[i_pole, _MLBW_RT] *
                               sigT_factor[self.l_value[i_pole]-1] +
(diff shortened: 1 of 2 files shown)


liangjg:
@smharper @cjosey could anyone of you take a look? Thanks.

openmc/data/multipole.py
```diff
@@ -77,7 +77,7 @@ def _faddeeva(z):
     if np.angle(z) > 0:
         return wofz(z)
     else:
-        return -np.conj(wofz(z))
+        return -np.conj(wofz(np.conj(z)))
```
paulromano: Should we be using `z.conjugate()` instead?

openmc/data/multipole.py
```diff
@@ -128,7 +129,7 @@ def _broaden_wmp_polynomials(E, dopp, n):
     # Perform recursive broadening of high order components.  range(1, n-4)
     # replaces a do i = 1, n=3.  All indices are reduced by one due to the
     # 1-based vs. 0-based indexing.
-    for i in range(1, n-4):
+    for i in range(1, n-2):
```
paulromano: Change comment above too

openmc/data/multipole.py
```diff
@@ -405,9 +406,9 @@ def curvefit(self, curvefit):
             cv.check_type('curvefit', curvefit, np.ndarray)
             if len(curvefit.shape) != 3:
                 raise ValueError('Multipole curvefit arrays must be 3D')
-            if curvefit.shape[2] != 3:  # One each for sigT, sigA, sigF
+            if curvefit.shape[2] not in (2, 3):  # sigT, sigA (and maybe sigF)
                 raise ValueError('The third dimension of multipole curvefit'
-                                 ' arrays must have a length of 3')
+                                 ' arrays must have a length of 2 or 2 or 3')
```
paulromano: 2 or 3

openmc/data/multipole.py
```diff
@@ -112,7 +112,8 @@ def _broaden_wmp_polynomials(E, dopp, n):
         erf_beta = 1.0
         exp_m_beta2 = 0.0
     else:
-        erf_beta = np.erf(beta)
+        from scipy.special import erf
```
cjosey: In my opinion, this import shouldn't be nested in the middle of the function like this.

openmc/data/multipole.py
```diff
@@ -434,7 +433,10 @@ def from_hdf5(cls, group_or_filename):
             group = group_or_filename
         else:
             h5file = h5py.File(group_or_filename, 'r')
-            version = h5file['version'].value[0].decode()
+            try:
```
cjosey: This is odd.  Why is this try/except block necessary?  Is H5PY interpreting it differently depending on the version?  If so, it might be worthwhile to only put that exception in the except block so that other errors can properly throw.

openmc/data/multipole.py
paulromano: Thus far we've kept scipy as an optional dependency, so scipy imports have only appeared within a function like this. I'm personally in favor of just making scipy a required dependency but I'm not sure if everyone agrees.

nelsonag:
Yes, definitely make it required; it is in every package manager's main
repos and likely to be installed already on a cluster.

openmc/data/multipole.py
liangjg: I found this `erf()` in the `else` branch is not always called so made the `import` inside. If making it outside is better, we have another choice: [math package also provides this error function ](https://docs.python.org/3/library/math.html#math.erf).

openmc/data/multipole.py
liangjg: @cjosey I felt confusing here too. All the libraries I generated using openw broke when running `h5file['version'].value[0].decode()` as the `h5file['version'].value` is a string `v0.2` while it is a array `[v0.2]` for the existing libraries. I tested on h5py `2.5.0` and `2.6.0`. Do you know why? Have you changed the openw?

liangjg: @cjosey did you generate the library using old hdf5 interface which writes string as a array in [this commit](https://github.com/mit-crpg/WHOPPER/commit/5727e8721e9e57e3a177793614c97e61c404ec7f)?

cjosey:
Ah, I see what the issue was.  All the published libraries were made with the v0.1 file format and I converted them with a python script.  It appears the python script wrote it as a 1d array of strings as opposed to a string.  I really wish I had the time to revise the WMP libraries.

As for the performance of the if statement, I imagine it will be architecture-dependent.   On x86-64, without the if statement, it might be possible for those lines to collapse into vector operations.  With the if statement, it might not and there may be some instruction cache misses.  However, the difference is going to be on the order of a few clock cycles per cross section evaluation at most, so I think it is fine to leave it as is.

openmc/data/multipole.py
paulromano: Yes, we should be using `math.erf`. Can you make that change and also use `math.sqrt`, `math.exp`, and `math.pi` in this function? i.e., just have `from math import exp, erf, pi` at the top of the module. The numpy versions of those functions should only be used if the argument is a vector (in this case, it's all scalars)

openmc/data/multipole.py
```diff
@@ -434,7 +433,10 @@ def from_hdf5(cls, group_or_filename):
...
+                version = h5file['version'].value.decode()
+            except:
```
paulromano: I would make this specifically `except AttributeError:`

openmc/data/multipole.py
```diff
@@ -2,6 +2,7 @@
 
 import h5py
 import numpy as np
+from math import exp, erf, pi
```
paulromano: Per PEP8, standard-library imports should come first, so this should be moved up with `from numbers ...`. Also, I forgot to mention that `sqrt` should also come from `math`. Can you change that too?

PR #959, opened 2018-01-24
Transition surface implementation to C++

smharper:
This PR brings us a little further down the road of #603.  It removes virtually all of the surface-related Fortran code and replaces it with C++.

Except for periodic boundary conditions, the translation of our OO Fortran surface code to C++ was fairly straightforward.  The new C++ classes and methods almost identically match the old Fortran ones.  For interoperability, Fortran retains the `Surface` class which only contains the neighbor lists and a pointer to the C++ object.  I've tried to move as much as possible to C++ so even the IDs now need to be accessed by a type-bound method which wraps C++ code.

I've been pretty thorough at testing the `sense`, `normal`, and `distance` methods of each surface type and so far I've found that this new code precisely reproduces the old version.  In theory, this could be slower due to extra function call overhead (that will disappear as more code gets converted to C++), but it looks negligible in my timing tests.

Implementation notes:
* The Fortran surface code had a lot of redundancy e.g. with the distance functions of `SurfaceXCylinder`, `SurfaceYCylinder`, and `SurfaceZCylinder`.  Here we're able to cut that down with templated functions.
* We use the `select type` constructs in the periodic BC code, and they are rather difficult to translate to C++ cleanly.  Apparently that sort of runtime type information encourages bad code.  I tried channeling my inner Stroustrup and writing the periodic BC code without using `dynamic_cast` to inspect surface types, but the end result in this case essentially works out to implicit type checking with things like the `normal()` method.  There are probably much cleaner ways to implement periodic BCs, but I think this code will do fine for now.
* This is the first of our C++ code to deal with XML and HDF5.  Accordingly, I've made an xml_interface.h and hdf5_interface.h.  They're small and limited now, but I'm sure we can grow them later as needed.  For example, I haven't touched the details of MPI I/O in hdf5_interface.h since surfaces only go in the summary.h5 file which is written from the master process.
* Note that I needed to change the `MPI` definition to `OPENMC_MPI` because `MPI` is defined by mpi.h

@paulromano, I'm not attached to many of the design decisions here so do feel free to suggest major changes.  At some point I think I would like to make a surface namespace and make the change
```C++
// Old:
int n_surfaces{0};
Surface **surfaces_c;
std::map<int, int> surface_dict;
// New:
std::vector<std::shared_ptr<Surface>> surfaces_c;
std::map<int, std::weak_ptr<Surface>> surface_dict;
```
but at the moment it feels like we might be breaking some contracts due to Fortran holding ownership of some `Surface` pointers without a destructor.

### DIFF
--- a/src/surface.cpp
+++ b/src/surface.cpp
@@ -1053,10 +1053,12 @@ extern "C" void
 read_surfaces(pugi::xml_node *node)
 {
   // Count the number of surfaces.
-  int n_surfaces{0};
   for (pugi::xml_node surf_node = node->child("surface"); surf_node;
        surf_node = surf_node.next_sibling("surface")) {
     n_surfaces++;
+  }
+  if (n_surfaces == 0) {
+    fatal_error("No surfaces found in geometry.xml!");
   }
 
   // Allocate the array of Surface pointers.
@@ -1137,7 +1139,14 @@ read_surfaces(pugi::xml_node *node)
       // Downcast to the PeriodicSurface type.
       Surface *surf_base = surfaces_c[i_surf];
       PeriodicSurface *surf = dynamic_cast<PeriodicSurface *>(surf_base);
-      //TODO: check for null pointer
+
+      // Make sure this surface inherits from PeriodicSurface.
+      if (surf == NULL) {
+        std::string err_msg{"Periodic boundary condition not supported for "
+                            "surface "};
+        err_msg += std::to_string(surf_base->id);
+        err_msg += ". Periodic BCs are only supported for planar surfaces.";
+      }
 
       // See if this surface makes part of the global bounding box.
       struct BoundingBox bb = surf->bounding_box();
--- a/src/surface.cpp
+++ b/src/surface.cpp
@@ -1,12 +1,12 @@
 #include "surface.h"
 
 #include <array>
+#include <cmath>
 #include <sstream>
 
 #include "error.h"
 #include "hdf5_interface.h"
 #include "xml_interface.h"
-#include <iostream>
 
 
 namespace openmc {
--- a/src/surface.cpp
+++ b/src/surface.cpp
@@ -1,11 +1,15 @@
+#include "surface.h"
+
 #include <array>
-#include <math.h>  // For fabs
+#include <sstream>
 
 #include "error.h"
 #include "hdf5_interface.h"
 #include "xml_interface.h"
+#include <iostream>
 
-#include "surface.h"
+
+namespace openmc {
 
 //==============================================================================
 // Helper functions for reading the "coeffs" node of an XML surface element
@@ -14,7 +18,7 @@
 int word_count(const std::string &text)
 {
   bool in_word = false;
-  int count{0};
+  int count {0};
   for (auto c = text.begin(); c != text.end(); c++) {
     if (std::isspace(*c)) {
       if (in_word) {
@@ -35,10 +39,9 @@ void read_coeffs(pugi::xml_node surf_node, int surf_id, double &c1)
   std::string coeffs = get_node_value(surf_node, "coeffs");
   int n_words = word_count(coeffs);
   if (n_words != 1) {
-    std::string err_msg{"Surface "};
-    err_msg += std::to_string(surf_id);
-    err_msg += " expects 1 coeff but was given ";
-    err_msg += std::to_string(n_words);
+    std::stringstream err_msg;
+    err_msg << "Surface " << surf_id << " expects 1 coeff but was given "
+            << n_words;
     fatal_error(err_msg);
   }
 
@@ -56,10 +59,9 @@ void read_coeffs(pugi::xml_node surf_node, int surf_id, double &c1, double &c2,
   std::string coeffs = get_node_value(surf_node, "coeffs");
   int n_words = word_count(coeffs);
   if (n_words != 3) {
-    std::string err_msg{"Surface "};
-    err_msg += std::to_string(surf_id);
-    err_msg += " expects 3 coeffs but was given ";
-    err_msg += std::to_string(n_words);
+    std::stringstream err_msg;
+    err_msg << "Surface " << surf_id << " expects 3 coeffs but was given "
+            << n_words;
     fatal_error(err_msg);
   }
 
@@ -77,10 +79,9 @@ void read_coeffs(pugi::xml_node surf_node, int surf_id, double &c1, double &c2,
   std::string coeffs = get_node_value(surf_node, "coeffs");
   int n_words = word_count(coeffs);
   if (n_words != 4) {
-    std::string err_msg{"Surface "};
-    err_msg += std::to_string(surf_id);
-    err_msg += " expects 4 coeffs but was given ";
-    err_msg += std::to_string(n_words);
+    std::stringstream err_msg;
+    err_msg << "Surface " << surf_id << " expects 4 coeffs but was given "
+            << n_words;
     fatal_error(err_msg);
   }
 
@@ -99,10 +100,9 @@ void read_coeffs(pugi::xml_node surf_node, int surf_id, double &c1, double &c2,
   std::string coeffs = get_node_value(surf_node, "coeffs");
   int n_words = word_count(coeffs);
   if (n_words != 10) {
-    std::string err_msg{"Surface "};
-    err_msg += std::to_string(surf_id);
-    err_msg += " expects 10 coeffs but was given ";
-    err_msg += std::to_string(n_words);
+    std::stringstream err_msg;
+    err_msg << "Surface " << surf_id << " expects 10 coeffs but was given "
+            << n_words;
     fatal_error(err_msg);
   }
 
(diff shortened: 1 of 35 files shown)


paulromano:
Awesome PR. I'll start taking a look when I get a chance.

src/surface.cpp
```diff
@@ -0,0 +1,1239 @@
...
+
+bool
+Surface::sense(const double xyz[3], const double uvw[3]) const
+{
+  // Evaluate the surface equation at the particle's coordinates to determine
+  // which side the particle is on.
+  const double f = evaluate(xyz);
+
+  // Check which side of surface the point is on.
+  if (fabs(f) < FP_COINCIDENT) {
```
paulromano: C++ actually provides floating-point overloads for `abs`, so I would just use that instead of the C-like `fabs`

paulromano:
One other question -- what MPI implementation are you using that defines a `MPI` macro in `mpi.h`?

src/surface.h
```diff
@@ -0,0 +1,419 @@
...
+extern "C" const int BC_REFLECT{2};
+extern "C" const int BC_PERIODIC{3};
+
+//==============================================================================
+// Constants that should eventually be moved out of this file
+//==============================================================================
+
+extern "C" double FP_COINCIDENT;
+const double INFTY{std::numeric_limits<double>::max()};
+const int C_NONE{-1};
```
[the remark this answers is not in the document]
smharper: Sure, space before the brace it is.

src/surface.h
```diff
@@ -0,0 +1,419 @@
...
+
+//==============================================================================
+// Global variables
+//==============================================================================
+
+int n_surfaces{0};
```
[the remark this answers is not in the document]
smharper: I'm going to put `n_surfaces` in the openmc namespace.  Will it still count as "global" for default initialization?  Also I realized that I had forgotten to declare this guy as `extern "C"` even though it's used in Fortran.  That's fixed now.

src/surface_header.F90
```diff
@@ -1,1073 +1,212 @@
...
-    real(8) :: A, B, C, D, E, F, G, H, J, K
-  contains
-    procedure :: evaluate => quadric_evaluate
-    procedure :: distance => quadric_distance
-    procedure :: normal => quadric_normal
-  end type SurfaceQuadric
+  end type Surface
 
-  integer(C_INT32_T), bind(C) :: n_surfaces  ! # of surfaces
+  integer(C_INT), bind(C) :: n_surfaces  ! # of surfaces
```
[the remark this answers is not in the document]
smharper: Nope, that was a mistake.

src/tracking.F90
```diff
@@ -404,86 +404,35 @@ subroutine cross_surface(p)
...
-          ! Move particle along normal vector based on distance
-          p % coord(1) % xyz(:) = p % coord(1) % xyz(:) - d*xyz
-        end select
-      end select
+      ! Get a pointer to the partner periodic surface.  Offset the index to
+      ! correct for C vs. Fortran indexing.
+      surf2 => surfaces(surf % i_periodic() + 1)
+
+      ! Adjust the particle's location and direction.
+      rotational = surf2 % periodic_translate(surf, p % coord(1) % xyz, &
```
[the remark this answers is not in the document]
smharper: Yes, there is a reason.  We want to put the particle exactly at e.g. `x0` of the opposite surface.  `x0` is of course a private variable so the particles coords need to be set by the opposite surface.  You could fake it a bit with the `evaluate` routine but it doesn't give bitwise reproducibility.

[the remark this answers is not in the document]
paulromano: Sounds like this could be solved by through a `friend` function?

src/surface.h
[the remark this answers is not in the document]
paulromano: Yes, global and namespace variables both have static storage duration and are default initialized.

smharper:
Thanks for the feedback, @paulromano.  It's made a lot of the code cleaner.  Regarding the MPI define, I am using MPICH-3.2.  It's checked in the very first set of statements in mpicxx.h:
```
#ifdef MPI
#error "You cannot define MPI; that name is reserved for the MPI namespace"
#endif
```

paulromano:
Darn, you're right. Can we change it to `USE_MPI` and `USE_MPIF08` then?

src/tracking.F90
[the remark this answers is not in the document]
smharper: It could be solved through a `friend`, but that would require a dynamic/reinterpret cast which I think will be less clean overall.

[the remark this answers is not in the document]
paulromano: Fair enough. We'll leave it be and revisit it another time.

smharper:
Could do `USE_MPI`, but I'm a little worried it might collide with someone else's defines.  `OPENMC_MPI` pretty much kills any chance of collision.

src/surface.cpp
```diff
@@ -0,0 +1,1235 @@
+#include "surface.h"
+
+#include <array>
+#include <sstream>
+
+#include "error.h"
+#include "hdf5_interface.h"
+#include "xml_interface.h"
+#include <iostream>
```
paulromano: Group this with other standard library includes. Also, you are missing `#include <cmath>` for `abs` and `sqrt`.

src/surface.h
```diff
@@ -0,0 +1,427 @@
...
+class Surface
+{
+public:
+  int id;                    //!< Unique ID
+  //int neighbor_pos[],        //!< List of cells on positive side
+  //    neighbor_neg[];        //!< List of cells on negative side
+  int bc;                    //!< Boundary condition
+  std::string name{""};      //!< User-defined name
+
+  Surface(pugi::xml_node surf_node);
```
paulromano: Declare this constructor `explicit` (as well as the single-argument constructors for all derived classes).

tests/unit_tests/test_capi.py
```diff
@@ -29,11 +29,11 @@ def pincell_model():
...
-    files = ['geometry.xml', 'materials.xml', 'settings.xml', 'tallies.xml',
-             'statepoint.10.h5', 'summary.h5', 'test_sp.h5']
-    for f in files:
-        if os.path.exists(f):
-            os.remove(f)
+    #files = ['geometry.xml', 'materials.xml', 'settings.xml', 'tallies.xml',
+    #         'statepoint.10.h5', 'summary.h5', 'test_sp.h5']
+    #for f in files:
+    #    if os.path.exists(f):
+    #        os.remove(f)
```
paulromano: ??

src/surface.h
```diff
@@ -0,0 +1,427 @@
...
+  //!   this surface.
+  //! @param uvw[3] A direction that will be rotated for systems with rotational
+  //!   periodicity.
+  //! @return true if this surface and its partner make a rotationally-periodic
+  //!   boundary condition.
+  virtual bool periodic_translate(PeriodicSurface *other, double xyz[3],
+                                  double uvw[3]) const = 0;
+
+  //! Get the bounding box for this surface.
+  virtual struct BoundingBox bounding_box() const = 0;
```
paulromano: remove struct

src/surface.h
```diff
@@ -0,0 +1,427 @@
...
+public:
+  SurfaceXPlane(pugi::xml_node surf_node);
+  double evaluate(const double xyz[3]) const;
+  double distance(const double xyz[3], const double uvw[3], bool coincident)
+         const;
+  void normal(const double xyz[3], double uvw[3]) const;
+  void to_hdf5_inner(hid_t group_id) const;
+  bool periodic_translate(PeriodicSurface *other, double xyz[3], double uvw[3])
+       const;
+  struct BoundingBox bounding_box() const;
```
paulromano: remove struct

src/surface.h
```diff
@@ -0,0 +1,427 @@
...
+public:
+  SurfaceYPlane(pugi::xml_node surf_node);
+  double evaluate(const double xyz[3]) const;
+  double distance(const double xyz[3], const double uvw[3],
+                  bool coincident) const;
+  void normal(const double xyz[3], double uvw[3]) const;
+  void to_hdf5_inner(hid_t group_id) const;
+  bool periodic_translate(PeriodicSurface *other, double xyz[3], double uvw[3])
+       const;
+  struct BoundingBox bounding_box() const;
```
paulromano: and this one and all the others

paulromano:
I think we're good to go after the latest round of comments is resolved.

src/surface.cpp
smharper: How do you find these missing includes.  Do you have a compiler flag that makes them obvious, or are you just operating from memory?

paulromano: Actually using `-Ddebug=on` on my local machine with gcc 6.3 caught it (complained about ambiguous overloads of `abs`). I guess gcc 4.8 on Travis is not smart enough.

smharper: Ah, I'm working with 5.3.0 so I reckon it's about time to move up a few version numbers.

tests/unit_tests/test_capi.py
smharper: Whoops!  Debugging purposes.

src/surface.cpp
smharper: Also, I was justing using std::cout for debugging so I've entirely removed this include now.

paulromano:
@smharper Thanks for bringing us a little closer to #603!

PR #1006, opened 2018-05-16
Checkvalue bug

hanzhuoran:
Fix a bug in checkvalue.py
`.join()` can take only str as arguments, not numpy.int32.

### DIFF
--- a/openmc/checkvalue.py
+++ b/openmc/checkvalue.py
@@ -265,7 +265,7 @@ def check_filetype_version(obj, expected_type, expected_version):
         if this_version[0] != expected_version:
             raise IOError('{} file has a version of {} which is not '
                           'consistent with the version expected by OpenMC, {}'
-                          .format(this_filetype, '.'.join(this_version),
+                          .format(this_filetype, '.'.join(str(v) for v in this_version),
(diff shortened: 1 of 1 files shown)


openmc/checkvalue.py
```diff
@@ -265,7 +265,7 @@ def check_filetype_version(obj, expected_type, expected_version):
         if this_version[0] != expected_version:
             raise IOError('{} file has a version of {} which is not '
                           'consistent with the version expected by OpenMC, {}'
-                          .format(this_filetype, '.'.join(this_version),
+                          .format(this_filetype, '.'.join(str(v) for v in this_version),
```
smharper: We try to roughly keep our code within 80 columns wide (just so it can be viewed in a narrow text editor window).  Split this call onto another line.  Probably the best way is to change
```python
.format(this_filetype, '.'.join(str(v) for v in this_version),
        expected_version))
```
to
```python
.format(this_filetype,
        '.'.join(str(v) for v in this_version),
        expected_version))
```

smharper:
Merging.  Congrats on your first PR @hanzhuoran!

PR #1029, opened 2018-07-30
Fix indexing error with openmc_cell_set_temperature

smharper:
It looks like two different conventions were used for indexing in `openmc_cell_set_temperature`.  The indexing of a cell's `material` array expected that `instance` was 1-based, but the indexing of `sqrtkT` was 0-based.  This PR switches to a consistent 0-based convention.


nelsonag:
This reminds me that we should probably have a few tests (CE and MG) that
use multi-temperatures.  I don't think we have any.  I can tackle that
(also could be decent low-hanging fruit for new users), but not for a few
weeks.

PR #1045, opened 2018-08-16
Add Material objects to C++

smharper:
This PR is small in terms of migrating functionality, but it will probably be helpful to do small, fast PRs like these given our pace on the C++ rewrite.  This PR adds a basic `Material` class to C++.  Right now the only data it owns is the `id`, but that's enough for C++ to also take control of the cell-material mapping.  This change is probably a pre-requisite for moving the rest of geometry.F90 and plot.F90 over to C++.

While I was at it, I also switched most of my old code to our newly-established style for pointers and references.

@paulromano, I also switched your new default arguments to the `get_node_value` function to `true`.  That's by far the more common case needed for all the XML inputs I've dealt with.  Let me know if you want it to stay `false`.  (Also, I tried to figure out where you needed `false` in #1042, but I couldn't find it.)

### DIFF
--- a/src/cell.cpp
+++ b/src/cell.cpp
@@ -429,7 +429,7 @@ Cell::contains_complex(Position r, Direction u, int32_t on_surface) const
 //==============================================================================
 
 extern "C" void
-read_cells(pugi::xml_node *node)
+read_cells(pugi::xml_node* node)
 {
   // Count the number of cells.
   for (pugi::xml_node cell_node: node->children("cell")) {n_cells++;}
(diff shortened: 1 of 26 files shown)


src/xml_interface.h
```diff
@@ -17,7 +17,7 @@ check_for_node(pugi::xml_node node, const char *name)
 }
 
 std::string get_node_value(pugi::xml_node node, const char *name,
-  bool lowercase=false, bool strip=false);
+                           bool lowercase=true, bool strip=true);
```
paulromano: I would still prefer these guys to be false by default (i.e., no extra work by default). For any kind of numeric data we are reading from XML, there is generally no reason to lowercase and strip the string. Taking a look at where we are using `get_node_value` on the develop branch, most cases actually use the `false` default.

src/cell.cpp
```diff
@@ -467,35 +544,40 @@ read_cells(pugi::xml_node *node)
...
 
-  int32_t* cell_fill_ptr(Cell *c) {return &c->fill;}
-
-  int32_t cell_n_instances(Cell *c) {return c->n_instances;}
+  //TODO: off-by-one
+  int32_t cell_material(Cell* c, int i)
+  {
+    int32_t mat = c->material[i-1];
+    if (mat == C_NONE) return F90_NONE;
+    if (mat == MATERIAL_VOID) return MATERIAL_VOID;
```
paulromano: This is starting to look a little messy. Is there really a need to have a separate value indicating "no material was specified" as opposed to "void material". Maybe we should adopt the following behavior:
1. `<cell>` has `material` attribute but no `fill`: single material or distributed materials and `Cell.fill = -1`
2. `<cell>` has `fill` attribute but no `fill`: universe/lattice and `Cell.material = -1`
3. `<cell>` has neither `material` nor `fill`: void material with both `Cell.material = -1` and `Cell.fill = -1`.

If we want to check "Is this material filled with a universe/lattice" (for example, `geometry_aux.cpp:23`), that would just be `Cell.fill >= 0`.

src/cell.cpp
```diff
@@ -429,21 +446,20 @@ Cell::contains_complex(Position r, Direction u, int32_t on_surface) const
...
 
-  // Allocate the vector of Cells.
-  global_cells.reserve(n_cells);
-
   // Loop over XML cell elements and populate the array.
+  global_cells.reserve(n_cells);
   for (pugi::xml_node cell_node: node->children("cell")) {
     global_cells.push_back(new Cell(cell_node));
   }
+  global_cells.shrink_to_fit();
```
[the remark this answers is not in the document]
smharper: I was just thinking about the wording of the `reserve` and `shrink_to_fit` requirements in the standards.  `reserve` could reserve more space than needed, whereas `shrink_to_fit` I think is a more strong request for shrinking to the smallest capacity.  This is a really small issue though so I'll remove the `shrink_to_fit` if you think it's unnecessary clutter.

src/xml_interface.h
smharper: Good point, most of our data is numeric.  I've switched back to `false`.

src/cell.cpp
smharper: Much cleaner, thanks!

smharper: Note, I went with something slightly different than what you've suggested here.  Case 3 is prevented (at least when reading from XML) by a `fatal_error`.  For case 2, `Cell.material` is just left as an empty vector.

src/cell.cpp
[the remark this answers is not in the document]
paulromano: In practice, `reserve(n)` will make the capacity exactly `n` if it's greater than the current capacity. I've confirmed with a little program:
```c++
#include <iostream>
#include <vector>

int main() {
  for (int n = 0; n < 100; ++n) {
    std::vector<int> x;
    x.reserve(n);
    std::cout << "n = " << n << ", capacity = " << x.capacity() << '\n';
  }
}
```
Using g++, icpc, and clang, all give the same results (capacity == n for all n).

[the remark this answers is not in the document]
paulromano: So, in your case, `shrink_to_fit` will be a no-op; might as well remove it.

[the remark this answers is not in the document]
smharper: Good to know.  I'll cut out the unneeded ones.

PR #1046, opened 2018-08-16
fixed bug with trigger, model and statepoint

tokamaster:
Name of the statepoint file generated when using the trigger feature depends on the final number of batches needed to get a result with an error below the threshold defined with trigger and the number of digits of the maximum number of batches admissible.
If we define a maximum number of batches of 1000000, and 100 batches were needed, statepoint file name will follow like this: statepoint.0000100.h5

This is fine when we try to run the simulation without defining a model, but when a model is defined, the name of the expected statepoint file is: statepoint.100.h5.

This causes an error since model.py tries to read a file which doesn’t exist.

@makeclean

### DIFF
--- a/openmc/model/model.py
+++ b/openmc/model/model.py
@@ -218,10 +218,23 @@ class Model(object):
 
         openmc.run(**kwargs)
 
+        # in order to accomodate the fact that you may start with only 2
+        # batches but set max_batches to be 10000, the format that openmc
+        # actually writes would be in this case 00001
+        
         n = self.settings.batches
+        m = self.settings.trigger_max_batches
+               
         if self.settings.statepoint is not None:
             if 'batches' in self.settings.statepoint:
                 n = self.settings.statepoint['batches'][-1]
 
+
+        if self.settings.trigger_active:
+            if self.settings.trigger_max_batches is not None:
+                len_s_n = len(str(self.settings.batches))
+                n = str().zfill(len(str(self.settings.trigger_max_batches))-len_s_n)
+                n += str(self.settings.batches)
+                
         with openmc.StatePoint('statepoint.{}.h5'.format(n)) as sp:
             return sp.k_combined


smharper:
Hi, @tokamaster.  Thanks for pointing out this bug and taking the time to fix it!

Your changes fix the error, but unfortunately the code will read the k_eff from a statepoint that is not necessarily the most recent.  For example, if you set `settings.batches = 50` but the triggers aren't satisfied until batch 87, this code will pick up the statepoint from batch 50 rather than batch 87.

I have a recommendation for fixing that.  We can stick a regex inside openmc.executor that watches for the output text of the form "Creating statepoint ..." and then use that to figure out what was the final statepoint written.  Here is a way you can modify executor.py to return the name of the last statepoint:

```diff
--- a/openmc/executor.py
+++ b/openmc/executor.py
@@ -1,4 +1,5 @@
 from collections.abc import Iterable
+import re
 import subprocess
 from numbers import Integral
 
@@ -11,14 +12,22 @@ def _run(args, output, cwd):
     p = subprocess.Popen(args, cwd=cwd, stdout=subprocess.PIPE,
                          stderr=subprocess.STDOUT, universal_newlines=True)
 
+    # Compile a regex object to identify state point writes
+    regex_prog = re.compile('Creating state point (.+)...')
+
     # Capture and re-print OpenMC output in real-time
     lines = []
+    last_statepoint = None
     while True:
         # If OpenMC is finished, break loop
         line = p.stdout.readline()
         if not line and p.poll() is not None:
             break
 
+        # If a statepoint was written, capture its name
+        match = regex_prog.search(line)
+        if match is not None: last_statepoint = match.group(1)
+
         lines.append(line)
         if output:
             # If user requested output, print to screen
@@ -29,6 +38,8 @@ def _run(args, output, cwd):
         raise subprocess.CalledProcessError(p.returncode, ' '.join(args),
                                             ''.join(lines))
 
+    return last_statepoint
+
 
 def plot_geometry(output=True, openmc_exec='openmc', cwd='.'):
     """Run OpenMC in plotting mode
@@ -209,4 +220,4 @@ def run(particles=None, threads=None, geometry_debug=False,
     if mpi_args is not None:
         args = mpi_args + args
 
-    _run(args, output, cwd)
+    return _run(args, output, cwd)
```

As a bonus, that will also work when the user specifies an alternate output directory.

tokamaster:
You are completely right. I noticed there is always a statepoint generated with the minimum number of batches (e.g., your example would generate one for 50 batches and one for 87 batches). Not sure how essential it is to generate two statepoints (unless they're required to predict the number of batches necessary to satisfy error conditions).
I forgot this is the reason I get the scores from the last statepoint generated in my code (I check it sweeping from trigger_max_batches to batches -i.e., higher to lower-).

Thanks, I will work on it.

smharper:
Hey @tokamaster, do you have any questions about how to move forward on this issue?  There's no rush, but I just want to check in and make sure you aren't fighting unnecessarily with something we can help you with.

alex-lyons:
I would like to work on this issue. The problem with the statepoint filename format seems fairly clear.
Can you assign this to me?

paulromano:
Hi @alex-lyons I just invited you to the openmc-dev organization so that I'm able to assign you on an issue. I think the right way forward on this is to create a separate issue (that I'll assign you on) and close this pull request since you won't be able to make changes on it. Then, once you're ready with a fix, you can submit a new pull request.

alex-lyons:
Hi Paul,

Thanks – yes, that seems the best approach.

I’m looking at issue #1 of the list from UKAEA: “Correct the trigger functionality such that fixed source simulations can automatically terminate when tallies have reached an error limit. There is an existing suggestion for this feature on the Github pull request list… pull/1046”. That particular pull request seemed more to do with setting a consistent statepoint filename than anything to do with including error counts in tallies, but I guess its a good place to start.

Alex.

paulromano:
Closing this old PR; @alex-lyons will be working on a new branch to fix it and that will come as a new PR. I've created an issue to track this (#1473)

PR #1054, opened 2018-08-20
Accommodate the new WMP Library

liangjg:
[A new, simplified windowed multipole library](https://github.com/mit-crpg/WMP_Library) has been generated using the vectoring fitting technique. This PR makes changes in openmc to accommodate the new library.
The changes of library format include (as shown in [docs/source/io_formats/data_wmp.rst](https://github.com/openmc-dev/openmc/compare/develop...liangjg:wmp_v1.0?expand=1#diff-e9b14cb6a603a03ab065771152821080)):
- removed `formalism`, `l_value`, `pseudo_K0RS` datasets
- renamed `start_E`, `end_E` to `E_min`, `E_max`
- put `w_start`, `w_end` into one dataset `windows`

The openmc core, python api, documentation and tests are updated accordingly.


liangjg:
I also tried to run ICSBEP benchmarks to validate the wmp library against to ace library, here are the results for about 250 cases ( need running 40 hours more to finish all 400+ cases):
- keff differences: statistically random, mostly within 3 times of standard deviation, within 30 pcm
- calculation rates: wmp slows down by 10~40% compared to ace
keff_comparison
calc_rate_diff_inact

liangjg:
@smharper I like the idea to have a merged nuclear data library. It is much more convenient for the users. The multipole data is light and it is straightforward to squeeze it into current library. My concerns are:
- Which temperatures for the point-wise cross section should we include in the default merged library? The multipole data can do the Doppler broadening to any temperature (0K-3000K), but it only accounts for RRR or energy region below the first threshold if there is a threshold reaction within RRR. The users may still need temperature-dependent data for URR or some threshold reactions where Doppler broadening matters. 
- If a user needs to use his own point-wise data, the multipole data should be transfered/merged again.
- The current wmp library contains all 423 nuclides from ENDF/B VII.1, but the user may use other evaluation libraries. The cross sections may be inconsistent? Or there are nuclides with multipole data missing?

I mean we need to consider more about different user cases. I may suggest we try to implment this later or in another PR.

paulromano:
Also was going to say -- we can start a separate issue to discuss merging of the multiple data.

smharper:
Thanks, @liangjg!

liangjg:
@paulromano @smharper The cluster is highly loaded, I'm still waiting for all the icsbep jobs to finish (I expect to post the final results this evening). To answer the question about the cases with keff differences higher than 3 standard deviations, I compared the outlier cases and did not find anything abnormal. The wmp cross sections used in these cases show good accuracy compare with ACE libraries.  Then I tried to run openmc with different random number seeds with ACE nuclear data and found, interestingly, it is normal to have keff difference beyond 3 standard deviation between 2 individual runs. Here is the results of 120 runs (seeds are odd numbers from 1 to 239) for icsbep-pu-sol-therm-003-case-7 model:
icsbep-pu-sol-therm-003-case-7
There are quite a few cases that have keff higher or lower than another one by 3+ standard deviations.
To consider more, as we are comparing two random numbers, the uncertainty of the difference should be a combination of their individual uncertainties. A more reasonable standard deviation to be used should be ```\sqrt(STD1^2+STD2^2)```, i.e. 8.5 pcm in this case. However,  the largest difference, 35 pcm is still much larger than 3*8.5=25.5pcm. I guess we need to be more tolerant about the 3 standard deviation criteria in the future.

liangjg:
Here are the final results of icsbep.
keff_comparison
8 out of 400 cases have keff differences larger than 3 standard deviations.
calc_rate_diff_inact
The calculation rates with WMP are slower than ACE by 20-40% for thermal spectrums. For fast spectrums, the performance decrease is modest.

PR #1075, opened 2018-09-22
Initial CAD-based geometry implementation

pshriwise:
This PR adds support for CAD-based DAGMC geometries in OpenMC. This can be enabled by setting dagmc to `true` in the `settings.xml` and a geometry file named "dagmc.h5m" in the run directory is required, a `.h5m` file being a specialized hdf5 file used in [MOAB](http://sigma.mcs.anl.gov/moab-library/).

As of now, DAGMC simulations rely on the presence of a `materials.xml` file where cell materials and densities are specified by id and value in the `.h5m`. (more on that the material markup [here](https://svalinn.github.io/DAGMC/usersguide/codes/dag-mcnp.html)). Material temperatures have not yet been addressed.

Key changes are:

 - Addition of general `Surface` and `Cell` classes along with new `DAGSurface`/`DAGCell` subclasses (the original Surface/Cell class implementations are now `CSGSurface`/`CSGCell`)

 - A C++ cad loading function exposed to the Fortran input_xml subroutines to replace the CSG geometry with the CAD versions

 - Small changes in tracking to avoid searching all cells when crossing a surface.

 - Addition of a MOAB/DAGMC installation to CI (when needed) and a CAD build in the test matrix

It's a fair amount of line changes - mostly on the C++ side. The tracking algorithm hardly needed modification thanks to the recent move of Surfaces and Cells to C++. I tried to keep it to a minimum for the initial implementation and some test addition in CI, but I can try to break this up into smaller chunks if requested.


pshriwise:
I almost forgot to mention - this implementation is not yet threadable, but will be upon the next release of MOAB thanks to a couple of fixes to the interface there.

include/openmc/position.h
```diff
@@ -57,9 +58,7 @@ struct Position {
   }
 
   // Data members
-  double x = 0.;
-  double y = 0.;
-  double z = 0.;
+  union{ double xyz[3]; struct{ double x,y,z = 0.0; }; };
```
smharper: I'm a little concerned about this union since we'd be relying on technically undefined behavior.  It's hard to imagine that any compiler would make it so that `&(p.y) != &(p.xyz[1])`, but I'd still like to stick within the standards if possible.  Maybe we should just leave `Position` unchanged and then fill a different 'double xyz[3]` local variable to use when calling `ray_fire`.  Do you see noticeable performance drop with that data copying?

pshriwise:
Thanks for the comments @paulromano and @smharper! Taking care of them straight away.

include/openmc/position.h
pshriwise: Fair enough. I was thinking that because these were the same type it wouldn't matter and would make interfacing with DAGMC easier, but if it makes the code more vulnerable I'll remove it.

pshriwise: I, as of yet, have no evidence that performing that data copy hurts performance significantly, so until I do I'll use the local var. :)

include/openmc/cell.h
```diff
@@ -130,21 +133,54 @@ class Cell
...
+  virtual std::pair<double, int32_t>
+  distance(Position p, Direction u, int32_t on_surface) const = 0;
+
+  //! Write all information needed to reconstruct the cell to an HDF5 group.
+  //! @param group_id An HDF5 group id.
+  virtual void to_hdf5(hid_t group_id) const = 0;
+
+};
+ 
+class CSGCell : public Cell
```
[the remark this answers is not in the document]
pshriwise: Yes. Definitely. Thanks!

CMakeLists.txt
```diff
@@ -36,6 +39,19 @@ if(MPI_ENABLED AND mpif08)
...
+# DAGMC Geometry Support - need DAGMC/MOAB
+#===============================================================================
+if(dagmc)
+  find_package(DAGMC REQUIRED)
+  if(NOT DAGMC_FOUND)
+    message(FATAL_ERROR "Could not find DAGMC installation")
+  endif()
+  add_definitions(-DDAGMC)
+  link_directories(${DAGMC_LIBRARY_DIRS})
+  include_directories(${DAGMC_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
pshriwise: Sounds good. Updating these now. I'm fairly sure we'll need the `link_directories` line due to the variables set by DAGMC's CMake file though.

include/openmc/cell.h
```diff
@@ -130,21 +133,54 @@ class Cell
...
 protected:
   bool contains_simple(Position r, Direction u, int32_t on_surface) const;
   bool contains_complex(Position r, Direction u, int32_t on_surface) const;
 };
 
+#ifdef DAGMC
+class DAGCell : public Cell
+{
+ public:
+  moab::DagMC *dagmc_ptr_;
```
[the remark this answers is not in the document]
pshriwise: Bummed I missed this one because I prefer that layout too... Fixing.

Our model in the past has been non-owned pointers which are cleaned up elsewhere at the end of the run. I'm open to other options though!

src/cell.cpp
```diff
@@ -585,6 +591,46 @@ Cell::contains_complex(Position r, Direction u, int32_t on_surface) const
...
+//==============================================================================
+#ifdef DAGMC
+DAGCell::DAGCell() : Cell{} {};
+
+std::pair<double, int32_t> DAGCell::distance(Position p, Direction u, int32_t on_surface) const {
+  moab::ErrorCode rval;
+  moab::EntityHandle vol = dagmc_ptr_->entity_by_id(3, id_);
+  moab::EntityHandle hit_surf;
+  double dist;
+  rval = dagmc_ptr_->ray_fire(vol, p.xyz, u.xyz, hit_surf, dist);
```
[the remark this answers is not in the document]
pshriwise: True. I did that originally and then decided on `union` because it would be more clear (in a way). 

Another option is to force alignment in that struct, but, as @smharper alluded to, I should look at performance first to make sure it matters.

tests/regression_tests/dagmc/settings.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<settings>
+
+  <run_mode>eigenvalue</run_mode>
+  <dagmc>true</dagmc>
+  <batches>15</batches>
+  <inactive>5</inactive>
+  <particles>10000</particles>
```
smharper: This test is pretty slow.  Let's cut down the runtime with something like
```xml
  <batches>5</batches>
  <inactive>0</inactive>
  <particles>100</particles>
```

tests/regression_tests/dagmc/tallies.xml
```diff
@@ -0,0 +1,31 @@
...
+
+  <tally id="2">
+    <filters>1 2</filters>
+    <scores>total scatter nu-scatter absorption fission nu-fission</scores>
+  </tally>
+
+  <tally id="3">
+    <filters>1 2 3</filters>
+    <scores>scatter nu-scatter nu-fission</scores>
+  </tally>
```
smharper: No need to test all of these tallies.  Probably just a single tally with filter 1 and a `total` score will do.

src/dagmc.cpp
```diff
@@ -0,0 +1,136 @@
...
+    if (DAG->has_prop(vol_handle, "mat")){
+      std::string mat_value;
+      rval = DAG->prop_value(vol_handle, "mat", mat_value);
+      MB_CHK_ERR_CONT(rval);
+      to_lower(mat_value);
+      
+      if (mat_value == "void" || mat_value == "vacuum") {
+	c->material_.push_back(MATERIAL_VOID);
+      } else {
+	c->material_.push_back(std::stoi(mat_value));
```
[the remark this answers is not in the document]
pshriwise: Odd. These lines look fine in my editor, but appear incorrectly here. Going to look into this...

include/openmc/cell.h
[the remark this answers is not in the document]
pshriwise: This may be the wrong idea, but what I'm thinking now is that these attributes of the `Cell` class should have virtual getter/setter functions (maybe a few resize functions as well) which get called instead of accessing the class attributes directly. This would do a few things: 

  - Provide some protection for the attributes of these classes.

  - Maintain the use of a general `Cell` class throughout the geometry kernel (nothing new here I guess).

  - Allow some protection for improper use of the class depending on the simulation type. For example, access to anything other than the first cell instance in a `CADCell` should return an error or warning, likely indicating that is is being accessed by some CSG-specific part of the code when it shouldn't be. 

Thoughts?

tools/ci/travis-install.py
```diff
@@ -48,22 +49,93 @@ def install(omp=False, mpi=False, phdf5=False):
...
+    if dagmc:
+        build_dagmc()
+        cmake_cmd.append('-Ddagmc=ON')
+        
     # Build and install
     cmake_cmd.append('..')
     print(' '.join(cmake_cmd))
     subprocess.check_call(cmake_cmd)
-    subprocess.check_call(['make', '-j'])
+    subprocess.check_call(['make', '-j2'])
```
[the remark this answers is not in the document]
pshriwise: I was having problems with memory usage on Travis. This was meant to be a temporary fix. I'll look into a more permanent solution soon.

Here's what I'm seeing when I use -j: https://travis-ci.org/pshriwise/openmc/jobs/435138577

pshriwise:
I've removed the geometry read if the simulation used CAD geometry by checking for an attribute of the `/geometry` group. I'm in the process of supporting DAGMC versions for `Cell`, `Region`, `Surface` etc. but am planning on submitting that as a separate PR. Happy to include it here though.

include/openmc/cell.h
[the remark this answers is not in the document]
pshriwise: @paulromano still going back and forth on this in my head. Would it be better to just do the type conversion from `Cell` to `CSG/DAGCell` explicitly in the code?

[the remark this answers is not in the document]
paulromano: I was mostly thinking about members like `region` and `rpn` that are very CSG-specific. Those we could probably move into `CSGCell` without needing getters/setters. Were you thinking about other members?

cmake/Modules/FindDAGMC.cmake
```diff
@@ -0,0 +1,18 @@
...
+#
+# Once done this will define
+#
+#  DAGMC_FOUND - system has DAGMC
+#  DAGMC_INCLUDE_DIRS - the DAGMC include directory
+#  DAGMC_LIBRARIES - Link these to use DAGMC
+#  DAGMC_DEFINITIONS - Compiler switches required for using DAGMC
+
+find_path(DAGMC_CMAKE_CONFIG NAMES DAGMCConfig.cmake
+          HINTS ${DAGMC_ROOT}
```
paulromano: If you use a `DAGMC_ROOT` environment variable (as I tried to do), this doesn't work because it only looks for a cmake variable of that name. Can you add `$ENV{DAGMC_ROOT}` here too?

pshriwise: Definitely. I'l make sure that gets propagated into main version of this file in the DAGMC repo too.

include/openmc/cell.h
```diff
@@ -130,21 +132,57 @@ class Cell
   //! \param on_surface The signed index of a surface that the coordinate is
   //!   known to be on.  This index takes precedence over surface sense
   //!   calculations.
+  virtual bool
+  contains(Position p, Direction u, int32_t on_surface) const = 0;
```
paulromano: Position r

include/openmc/cell.h
```diff
@@ -130,21 +132,57 @@ class Cell
...
+
+  //! Find the oncoming boundary of this cell.
+  virtual std::pair<double, int32_t>
+  distance(Position p, Direction u, int32_t on_surface) const = 0;
```
paulromano: Position r

include/openmc/dagmc.h
```diff
@@ -0,0 +1,21 @@
+
+#ifdef DAGMC
+
+#ifndef OPEN_DAGMC_H
+#define OPEN_DAGMC_H
```
paulromano: Should be OPENMC_DAGMC_H

include/openmc/dagmc.h
```diff
@@ -0,0 +1,21 @@
...
+
+namespace openmc {
+
+extern moab::DagMC* DAG;
+
+extern "C" void load_dagmc_geometry();
+extern "C" void free_memory_dagmc();
+
+}
+#endif // DAGMC_H
```
paulromano: OPENMC_DAGMC_H

include/openmc/surface.h
```diff
@@ -104,12 +108,41 @@ class Surface
...
+//==============================================================================
+#ifdef DAGMC
+class DAGSurface : public Surface
+{
+public:
+  moab::DagMC* dagmc_ptr_;
+  DAGSurface();
+  double evaluate(Position r) const;
+  double distance(Position r, Direction u,
+                  bool coincident) const;
```
paulromano: Combine onto previous line

tests/regression_tests/dagmc/__init__.py
```diff
@@ -0,0 +1 @@
+import pytest
```
paulromano: Remove this file?

CMakeLists.txt
```diff
@@ -471,12 +488,17 @@ endif()
 # target_link_libraries treats any arguments starting with - but not -l as
 # linker flags. Thus, we can pass both linker flags and libraries together.
 target_link_libraries(libopenmc ${ldflags} ${HDF5_LIBRARIES} pugixml
-                      faddeeva xtensor)
+  faddeeva xtensor ${DAGMC_LIBRARIES})
```
paulromano: Remove this one here (already covered below)

CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ option(debug    "Compile with debug flags"                       OFF)
 option(optimize "Turn on all compiler optimization flags"        OFF)
 option(coverage "Compile with coverage analysis flags"           OFF)
 option(mpif08   "Use Fortran 2008 MPI interface"                 OFF)
+option(dagmc    "Enable support for DAGMC (CAD) geometry"        OFF)
```
paulromano: Let's add a check below that throws an error if a user has dagmc=on and openmp=on (until the MOAB thread safety issues are resolved).

tests/regression_tests/dagmc/test.py
```diff
@@ -0,0 +1,11 @@
+from tests.testing_harness import CADTestHarness
+import os
+import pytest
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get('DAGMC') != 'y',
+    reason="DAGMC CAD geometry is not enabled.")
+
+def test_cad():
```
paulromano: test_dagmc?

tests/testing_harness.py
```diff
@@ -312,3 +312,15 @@ class HashedPyAPITestHarness(PyAPITestHarness):
     def _get_results(self):
         """Digest info in the statepoint and return as a string."""
         return super()._get_results(True)
+
+
+class CADTestHarness(TestHarness):
```
paulromano: I don't think we need a special test harness for this test. Do we?

tools/ci/travis-install.py
```diff
@@ -2,6 +2,7 @@
 import shutil
 import subprocess
 
+MOAB_VERSION= '5.0'
```
paulromano: space before =

CMakeLists.txt
pshriwise: Maybe a warning? Threads still work with CSG geometry and users may want to run either geom type with the same executable. 

I can force num_threads = 1 if it's a dagmc run and give a message to the user that this is happening.

tools/ci/travis-install.py
pshriwise: Removing that variable altogether. MOAB install is done in a different place now.

CMakeLists.txt
paulromano: Ah, good point. In that case a warning is probably fine.

tests/testing_harness.py
pshriwise: Once I add expected result files, no we won't.

paulromano: Gotcha. Are you planning on doing that as part of this PR?

paulromano:
I think I'm satisfied once the latest round of comments are resolved (and we come to some resolution on where to store data members of `Cell`, `CSGCell` and `DAGCell`). Thanks for the hard work on this @pshriwise!

include/openmc/cell.h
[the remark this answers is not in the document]
smharper: I agree with @pshriwise, that it'd be tough to move any of these members to `CSGCell` without a bunch of type-casting.  I think the really proper way to handle this issue is to have a `CSGGeometryKernel` that unconditionally keeps cells of the type `CSGCell` and a `DAGMCGeometryKernel` with `DAGCell`s, but that's a big enough change that I'm happy to leave it out of this PR.

tests/testing_harness.py
pshriwise: Definitely. In the process of switching machines, but should have that done soon.

CMakeLists.txt
pshriwise: Ok I'll update that. Can't wait to remove it later :)

tests/regression_tests/dagmc/test.py
```diff
@@ -0,0 +1,11 @@
+from tests.testing_harness import CADTestHarness
+import os
+import pytest
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get('DAGMC') != 'y',
+    reason="DAGMC CAD geometry is not enabled.")
```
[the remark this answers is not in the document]
pshriwise: Would it be possible to expose that through a property of the openmc python module like `__dagmc_enabled__`?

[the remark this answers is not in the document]
pshriwise: Possible --> appropriate 

I've tried exposing a function `dagmc_enabled` in the capi and using it to set that property of the OpenMC module, which we can then use in testing. I don't know if 

a) it's an appropriate use of the capi to do this

b) if this is going to provide consistent behavior. My understanding at this point is that the python capi interface relies on the last-build version of `libopenmc`. Is that right?

include/openmc/cell.h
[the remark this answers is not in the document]
pshriwise: Are we ok with punting this to another PR after some discussion on the design?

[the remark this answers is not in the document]
pshriwise: After looking at this this morning, there is only one place in the code which requires a type-cast to get to the `region_` attribute. This isn't too bad, but as @smharper said, probably not something to let become commonplace.

tests/regression_tests/dagmc/test.py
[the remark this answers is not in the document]
pshriwise: For the sake of more info, here's what that looks like:
https://github.com/pshriwise/openmc/compare/dagmc...pshriwise:dagmc_capi

src/geometry_aux.cpp
```diff
@@ -159,7 +159,8 @@ neighbor_lists()
   write_message("Building neighboring cells lists for each surface...", 6);
 
   for (int i = 0; i < cells.size(); i++) {
-    for (auto token : cells[i]->region_) {
+    CSGCell* c= (CSGCell*)cells[i];
+    for (auto token : c->region_) {
```
smharper: We prefer the new C++ style casts to C-style casts like these.  The proper one to use in this case would be a `dynamic_cast`, and then you should make sure the cast was successful by checking if the `dynamic_cast` gave you back a null pointer.  I don't think the mess is worth it though.  I think we should just leave `region_` and `rpn_` as members of the `Cell` class and make a more thorough cleanup later (outside of this PR).

tests/regression_tests/dagmc/test.py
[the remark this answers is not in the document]
smharper: Yeah, that's a good solution!  Just call it `dagmc_enabled` though; I don't think the double underscores are needed here.  Also, you can skip the function call by having something like
```c++
#ifdef DAGMC
extern "C" constexpr bool dagmc_enabled {true};
#else
extern "C" constexpr bool dagmc_enabled {false};
#endif

```

src/geometry_aux.cpp
pshriwise: Yeah I think you're right. I'll remove the commits changing this. I'd rather leave those as part of `Cell` than have to do casting at all.

pshriwise:
Added that attribute. I'll test it locally to make sure it looks ok when rendered with sphinx.

I re-added the `__init__.py` file so the `test.py` file won't clash with others in the test suite. This seems to be a common practice in the regression tests dir. Am I missing something? Definitely not an expert on pytest.

pshriwise:
I think this is pretty close now @smharper @paulromano! 

I'll comb through it again sometime today for good measure. Any further comments are appreciated too.

paulromano:
I'm good to go; thanks @pshriwise for this awesome new feature! The ball is in @smharper's court now.

include/openmc/dagmc.h
```diff
@@ -0,0 +1,28 @@
...
+namespace openmc {
+
+extern moab::DagMC* DAG;
+
+extern "C" void load_dagmc_geometry();
+extern "C" void free_memory_dagmc();
+
+}
+
+#endif // OPENMC_DAGMC_H
```
smharper:
```c++
#endif // DAGMC
```

include/openmc/dagmc.h
```diff
@@ -0,0 +1,28 @@
...
+
+#endif
```
smharper:
```c++
#endif // OPENMC_DAGMC_H
```

src/settings.cpp
```diff
@@ -396,6 +408,13 @@ void read_settings_xml()
 #endif
   }
 
+#ifdef _OPENMP
+  if (dagmc) {
```
smharper:
```c++
  if (dagmc && omp_get_max_threads() > 1) {
```
to avoid spurious warnings

tools/ci/travis-install.py
```diff
@@ -48,23 +47,26 @@ def install(omp=False, mpi=False, phdf5=False):
...
 
+    if dagmc:
+        cmake_cmd.append('-Ddagmc=ON')
+        
     # Build and install
     cmake_cmd.append('..')
     print(' '.join(cmake_cmd))
     subprocess.check_call(cmake_cmd)
-    subprocess.check_call(['make', '-j'])
+    subprocess.check_call(['make', '-j4'])
```
smharper: Just to be clear, do I understand correctly that you'll get memory errors if you just use `-j` rather than `-j4` here?

pshriwise: That's correct. It's not clear why at this point.

PR #1089, opened 2018-10-08
Voxel script

pshriwise:
This PR updates the `openmc-voxel-to-silovtk` in the following ways:

  - support for silo has been removed. The Python package supporting this format is no longer maintained

  - a test for the conversion script has been added to regression tests for Python versions 3.5, 3.6 (vtk is not yet available in Python 3.7)

  - renames script to `open-voxel-to-vtk`


scripts/openmc-voxel-to-vtk
```diff
@@ -0,0 +1,58 @@
...
+import h5py
+import vtk
+
+
+def main():
+    # Process command line arguments
+    parser = ArgumentParser()
+    parser.add_argument('voxel_file', help='Path to voxel file')
+    parser.add_argument('-o', '--output', action='store',
+                        default='plot', help='Path to output SILO or VTK file.')
```
paulromano: remove "SILO or"

tests/regression_tests/plot/test.py
```diff
@@ -18,6 +18,15 @@ def __init__(self, plot_names):
...
         openmc.plot_geometry(openmc_exec=config['exe'])
 
+        # TEMP: this should always be checked once vtk is added
+        # to Python3.7
+        try:
+            import vtk
+            call(['../../../scripts/openmc-voxel-to-vtk'] +
+                 glob.glob('plot_4.h5'))
+        except:
+            pass
```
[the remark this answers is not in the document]
paulromano: Actually `subprocess.run` wasn't added until 3.5, so this would break the script for 3.4. However, you can use [check_call](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) which is equivalent to `run(..., check=True)`.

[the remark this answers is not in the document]
paulromano: And let's make the except clause more specific: `except ImportError` so that actual errors don't pass silently.

tests/regression_tests/plot/test.py
```diff
@@ -56,6 +66,14 @@ def _get_results(self):
...
 def test_plot():
-    harness = PlotTestHarness(('plot_1.ppm', 'plot_2.ppm', 'plot_3.ppm',
-                               'plot_4.h5'))
+    expected_plots = ['plot_1.ppm', 'plot_2.ppm', 'plot_3.ppm', 'plot_4.h5']
+    # TEMP: this should always be checked once vtk is added
+    # to Python3.7
+    try:
+        import vtk
+        expected_plots.append('plot.vti')
+    except:
```
paulromano: except ImportError

tests/regression_tests/plot/test.py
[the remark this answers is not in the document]
pshriwise: I'll move to `check_call` in the next commit. Thanks!

Good point, will become a proper skip check now that it's going to be its own test though.

tests/regression_tests/plot_voxel/test.py
```diff
@@ -0,0 +1,65 @@
...
+import openmc
+import pytest
+
+from tests.testing_harness import TestHarness
+from tests.regression_tests import config
+
+
+pytestmark = pytest.mark.skipif(
+    importlib.util.find_spec('vtk') is None,
+    reason="vtk module is not installed.")
```
paulromano: pytest makes this slightly simpler: see [here](https://docs.pytest.org/en/documentation-restructure/how-to/skipping.html#skipping-on-a-missing-import-dependency)

paulromano:
All set to merge. Since this PR is not quite "trivial", I'm going to wait until the 36 hour review period is up.

tools/ci/travis-install.sh
```diff
@@ -26,5 +26,10 @@ python tools/ci/travis-install.py
 # Install Python API in editable mode
 pip install -e .[test]
 
+# conditionally install vtk
+if [ $TRAVIS_PYTHON_VERSION -ne 3.7 ]; then
```
paulromano: Just noticed that this doesn't actually work (primarily because `-ne` is an integer operator) and Travis is just skipping the vtk tests. I would change this to (combining with previous line):
```bash
if [[ $TRAVIS_PYTHON_VERISON == 3.7 ]]; then
    pip install -e .[test]
else
    pip install -e .[test,vtk]
fi
```

pshriwise: Whoops. Nice catch. Thanks!

PR #1103, opened 2018-10-23
Moving CMFD to Python through C API - Replication and performance tests

shikhar413:
This PR pushes the CMFD code I have rewritten in Python through the C API, with the purpose of showing performance results and demonstrating that it replicates the code written currently in Fortran. The purpose of this PR is not to merge this branch into the develop branch, but to provide a way to compare the Fortran and Python code side-by-side and also to figure out any design changes that need to be made to `openmc/openmc/cmfd.py`. Once `openmc/openmc/cmfd.py` is reviewed, I will create a separate PR that removes Fortran CMFD source code completely and update the test suite to call CMFD through the C API instead of through Fortran.

While I will be describing at length some of the design details and performance results, here is  a summary of the main observations from this implementation:

- The CMFD code written in Python now relies on all CMFD of the input parameters being defined directly through Python through the `CMFDRun` class instead of creating a `cmfd.xml` file. It should work with both OpenMP-based and MPI-based parallelism
- CMFD equations are now solved through the scipy sparse matrix solver, which results in faster performance for medium-sized problems, but a noticeable performance hit for larger problems, when compared to a simple Fortran-based red-black GS solver. In the future, it may be instructive to include a C++ based matrix solver to potentially speed up CMFD calculations, since this portion of the code is where the majority of the time is spent in CMFD.
- The majority of variables used to populate the CMFD matrices are numpy arrays. To the best of my ability, I tried to define functions that populate / modify these arrays in a vectorized manner, but this form of implementation results in complicated code that runs faster but is not easily decipherable to someone unfamiliar with numpy vectorization or CMFD. To get around this, I have included vectorized and un-vectorized (for-loop based) implementations of functions such as `compute_dhat`, `compute_dtilde`, `build_matrices`, and `calc_fission_source` for easier readability

## General C API-based CMFD algorithm
The core of the CMFD algorithm can be found in the `run()` function of the CMFDRun class. To run CMFD the traditional way, `cmfd_run` should be set to true in `settings.xml` and a corresponding `cmfd.xml` with CMFD parameters should be defined. To run CMFD through the C API, `cmfd_run` should be false, and all CMFD parameters are defined in a separate python file with the following template:
```python
import openmc
from mpi4py import MPI

comm = MPI.COMM_WORLD

# Define CMFD mesh
cmfd_mesh = openmc.CMFDMesh()
#Define CMFD mesh parameters
...

# Initialize CMFDRun object
cmfd_run = openmc.CMFDRun()
# Set all runtime parameters (cmfd_mesh, tolerances, tally_resets, etc) through CMFDRun instance
cmfd_run.cmfd_mesh = cmfd_mesh
...

# Run CMFD
cmfd_run.run()
```

Examples of these different run configurations can be found in the `openmc/tests/cmfd_testing` directory. It should be noted that error checking for CMFD parameters happens when setting the instance variables of the CMFDRun class. In the final PR, the CMFD class in `openmc/cmfd.py` will be removed completely, but it is included for now so that CMFD xml files can still be created through the Python API. All other parts of the code that will be removed for the following PR are marked with `TODO remove`.

## Numpy vectorization tradeoffs
In order to populate the CMFD matrices that will be passed into the matrix solver, intermediate multidimensional Numpy array that define the multigroup cross sections and diffusion parameters are created. These Numpy arrays also need to be populated based on the CMFD tallies, and they are defined with bulk vectorized operations wherever possible. For functions such as `neutron_balance`, vectorized definitions simplify the function definitions so as not to rely on for loops any more, but other functions are much more difficult to understand because these Numpy arrays being defined do not lend themselves to straightforward vectorized definitions. In `compute_dhat` and `compute_dtilde`, the dhat and dtilde arrays are populated first by definining values at each array boundary, and then defining all inner values, since these quantities are defined differently where cell boundaries do and don't exist. While I tried to create some intermediate Numpy arrays to clarify these vectorized calculations, I did try to find a balance between creating too many arrays to reduce memory costs, since each intermediate array definition creates an additional copy of a Numpy array in memory. In a lot of cases, these vectorized functions becomes difficult to understand/follow, so I included a non-vectorized form of these definitions that relies on for loops and looks very similar to the current Fortran definitions. CMFD calls that utilize vectorized numpy definitions can be toggled on using the optional `vectorized=True` input argument (set to true by default) when calling CMFDRun.run(). As shown in the next section, vectorized runs perform faster than non-vectorized runs, but are much less readable. `calc_fission_source` and `build_matrices` are two more examples of such functions. In `calc_fission_source`, the 1-dimension phi vector that results from solving the CMFD equations has to be mapped back to a multidimensional cmfd_flux vector that accounts for the non-accelerated regions in a vectorized manner (`phi` contains only the regions that are defined as "accelerated" in the CMFD coremap). `build_matrices` tries to build CMFD matrices diagonal by diagonal in a vectorized manner as csr (compressed sparse row) matrices. I am very open to any suggestions to make these vectorized functions more readable.

## Performance results

I have created test cases for 1 group problems to see the effect of problem size on runtime. This is done for 1D, 2D, and 3D problems, with 100,000 particles, 15 inactive cycles, 20 total cycles, and CMFD feed turned on at batch 5. Vacuum boundary conditions are imposed on significant dimensions. The scripts can be found in `openmc/examples/cmfd_testing/`. All of these tests are conducted on one process, since the CMFD runs on a single process. While these tests also analyze the behavior of the number of threads on performance, there was no significant change in CMFD runtimes with this approach, since multithreading is not supported for CMFD calculations. I am currently gathering simulation results by running CMFD on the 2D BEAVRS problem to illustrate that 2-group CMFD also works, but these results are still pending. Here are the plots that compare performance when running CMFD using Fortran and Python, where the Python tests are also distinguished between using vectorized Numpy definitions, and using traditional for loops:

- These plots compare the total simulation run time for the 1d, 2d, and 3d cases respectively
totaltime-1d1grp
totaltime-2d1grp
totaltime-3d1grp

- These plots compare the total time in CMFD for the 1d, 2d, and 3d cases respectively
cmfdtime-1d1grp
cmfdtime-2d1grp
cmfdtime-3d1grp

- These plots compare the time building CMFD matrices for the 1d, 2d, and 3d cases respectively
buildmattime-1d1grp
buildmattime-2d1grp
buildmattime-3d1grp

- These plots compare the time solving CMFD matrices for the 1d, 2d, and 3d cases respectively
solvemattime-1d1grp
solvemattime-2d1grp
solvemattime-3d1grp

- Finally, runtimes for `calc_fission_source` as well as the cumulative time in `compute_dhat` and `compute_dtilde` are compared in Python when running both in vectorized manner and non-vectorized manner
fstime-1d1grp
dhattime-2d1grp
dhattime-3d1grp
dhattime-1d1grp
dhattime-2d1grp
dhattime-3d1grp

It should be noted that all times are plotted on logscale. From these plots, the following conclusions can be drawn:

- Solving CMFD matrices takes the bulk of the time in CMFD for more difficult problems
- The Python implementation is faster than Fortran for smaller problems, but the default Python matrix solver runs much slower for larger problems and makes the current Python implementation slower for larger problems. Working on a C++ based matrix solver that couples with CMFD could potentially speed up calculations, since solving CMFD matrices constitutes the majority of the time spent in CMFD.
- A vectorized approach in Numpy is faster than a non-vectorized for loop-based implementation, but the code is messier and more difficult to follow.

## Miscellaneous Implementation Details

- With the conversion of MPI code to C++, I am running into segfault issues when running 2D BEAVRS with a large number of particles. This issue is avoided by dynamically allocating the variable `temp_sites` in `eigenvalue.cpp` instead of statically allocating it, and I will address this in the next PR once I figure out how best to address this issue. I decided to include a temporary fix here just so that all test cases would run.
- Currently, all CMFD functions are contained within a single CMFDRun class, for ease of implementation instead of having to worry about passing a large set of variables from one subclass to another. Do let me know if there is opposition to such an approach.
- Indexing of the CMFD coremap is switched from 1 as non-accelerated and 2 as accelerated in Fortran to 0 as non-accelerated and 1 as accelerated in Python.
-With all matrix equations solved using Python's scipy sparse solver, Gauss Seidel tolerance variables (rtoli, atoli) are no longer needed.
- Instead of exposing the function `count_bank_sites` through the C API, which relies on MPI calls that could result in erratic behavior through the C API, this function was recreated in Python, resulting in a minor duplication of code.
- The variable `cmfd_src` is broadcasted to all processes in Fortran, but this is not necessary and is not done on the Python side
- MPI calls are made through Python using mpi4py. If mpi4py is not installed on the local machine, the `have_mpi` variable defined at the beginning of `openmc/cmfd.py` is set to False and MPI calls are not made.
- Since there is no way currently to control the output columns through the C API, there is no way to append the columns related to CMFD output, so all CMFD variables that need to be outputted at the end of each batch are displayed on separate lines.
- When write_matrices is set to true in the CMFDRun class, production and loss matrices as well as the CMFD flux vector are written both to file and as .npz/.npy files. These variables can be loaded in Python through numpy.load().
- Currently, numpy divide warnings are suppressed with `np.seterr(divide='ignore', invalid='ignore')`. These warnings appear as a result of having zero values in the divisor vector, and while I do handle this case by setting the output to zero for cases when the divisor is zero, numpy.divide will still carry out the divide, and hence the warning thrown. 
- Since the Fortran version of CMFD uses the k-effective value at each generation as an initial guess when entering the power iteration step of solving the CMFD equations, I created a `keff_temp` C API function that always returns the k-effective at each generation instead of the average k-effective. While this does not necessary impact the results since the final CMFD flux vector is normalized, there will be minor differences in the flux vector. I wanted to address this change here for exact replication, but in the final PR I will be using the keff function defined through the C API
- All CMFD-related errors throw an OpenMCError. This could instead be changed to a CMFDError type instead as well.
- While the Fortran version of CMFD supports running OpenMC in MG mode, this will result in an error when running the Python version, since supporting such an implementation would require exposing additional variables which seems unnecessary for such a specific use case.
- The current Python implementation version does not support plotting cmfd meshes, saving CMFD variables to statepoint, and running CMFD from statepoint, and this is all part of future work.


smharper:
@shikhar413, you should make a separate PR for the `temp_sites` dynamic memory thing.  I've found that you don't need the BEAVRS model to reproduce the bug, btw.  It looks like you can get it with any model if you crank the number of particles up high enough.

openmc/cmfd.py
```diff
@@ -518,3 +603,2717 @@ def export_to_xml(self):
...
+
+        """
+        # Get matrix index from coremap
+        matidx = ng*(self._coremap[i,j,k]) + g
+        return matidx
+
+    def _execute_power_iter(self, loss, prod):
+        """Main power iteration routine for the CMFD calculation
+
+        Para
```
smharper: Agreed that this vectorized code is pretty tough to read.  But I think we can try to make it a bit more readable.  In this case, we can clear things up a bit by making aliases for variables like `self._albedo[0]`, splitting one of these `np.where`s into an `if` block, and using a lot of tuples for indexing:
```python
        where_is_accel = np.where(is_accel)
        boundary = (0, where_is_accel[1], where_is_accel[2])
        boundary_grps = boundary + (slice(None),)
        D = self._diffcof[boundary_grps]
        dz = self._hxyz[boundary + (np.newaxis, 0)]
        if is_zero_flux_alb[0]:
            self._dtilde[boundary_grps + (0,)] = 2.0 * D / dz
        else:
            alb = self._albedo[0]
            self._dtilde[boundary_grps + (0,)] = ((2 * D * (1 - alb))
                 / (4 * D * (1 + alb) + (1 - alb) * dz))
```

smharper: Ah, whoops.  Something like `mask` is fine, but I'm hoping we can do this without creating any new arrays that are the full 3D size.  Would this or something close to it work?
```python
        nx, ny, nz = self._coremap.shape
        y_inds, z_inds = np.indices((ny, nz))
        boundary_accel = is_accel[(0, y_inds, z_inds)]
        boundary = (0, y_inds[boundary_accel], z_inds[boundary_accel])
```
Also, am I correct in thinking that `is_accel` is fixed and doesn't change over the course of a run?  In that case, we can also precompute these boundary indexing arrays at the start of the run and reuse them at each iteration.

shikhar413: This looks much better, thanks again. Also, you are correct, I will work on precomputing the boundary indexing arrays

shikhar413:
I've pushed some updates on making vectorized functions more readable. Previously, a code block like this one would populate all elements of the first row of the dtilde array as:

```python
# Define dtilde at left surface for all mesh cells on left boundary
self._dtilde[0,:,:,:,0] = np.where(is_accel[0,:,:,np.newaxis],
        np.where(is_zero_flux_alb[0], 2.0 * self._diffcof[0,:,:,:] / \
                 self._hxyz[0,:,:,np.newaxis,0],
                 (2.0 * self._diffcof[0,:,:,:] * \
                 (1.0 - self._albedo[0])) / \
                 (4.0 * self._diffcof[0,:,:,:] * \
                 (1.0 + self._albedo[0]) + \
                 (1.0 - self._albedo[0]) * \
                 self._hxyz[0,:,:,np.newaxis,0])), 0)

# Define dtilde at left surface for all mesh cells not on left boundary
self._dtilde[1:,:,:,:,0] = np.where(is_accel[1:,:,:,np.newaxis], \
        np.where(adj_reflector[1:,:,:,np.newaxis],
                 (2.0 * self._diffcof[1:,:,:,:] * \
                 (1.0 - ref_albedo[1:,:,:,:])) / \
                 (4.0 * self._diffcof[1:,:,:,:] * \
                 (1.0 + ref_albedo[1:,:,:,:]) + \
                 (1.0 - ref_albedo[1:,:,:,:]) * \
                 self._hxyz[1:,:,:,np.newaxis,0]), \
                 (2.0 * self._diffcof[1:,:,:,:] * neig_dc[1:,:,:,:]) / \
                 (neig_hxyz[1:,:,:,np.newaxis,0] * self._diffcof[1:,:,:,:] + \
                 self._hxyz[1:,:,:,np.newaxis,0] * neig_dc[1:,:,:,:])), 0.0)
```

Now, it would look like this:

```python
# Define dtilde at left surface for all mesh cells on left boundary
# Separate between zero flux b.c. and alebdo b.c.
boundary = self._first_x_accel
boundary_grps = boundary + (slice(None),)
D = self._diffcof[boundary_grps]
dx = self._hxyz[boundary + (np.newaxis, 0)]
if is_zero_flux_alb[0]:
    self._dtilde[boundary_grps + (0,)] = 2.0 * D / dx
else:
    alb = self._albedo[0]
    self._dtilde[boundary_grps + (0,)] = ((2.0 * D * (1.0 - alb))
         / (4.0 * D * (1.0 + alb) + (1.0 - alb) * dx))

# Define dtilde at left surface for all mesh cells not on left boundary
# Dtilde is defined differently for regions that do and don't neighbor
# reflector regions
boundary = self._notfirst_x_accel
boundary_grps = boundary + (slice(None),)
D = self._diffcof[boundary_grps]
dx = self._hxyz[boundary + (np.newaxis, 0)]
neig_D = neig_dc[boundary_grps]
neig_dx = neig_hxyz[boundary + (np.newaxis, 0)]
alb = ref_albedo[boundary_grps]
is_adj_ref = self._is_adj_ref_left
self._dtilde[boundary_grps + (0,)] = np.where(is_adj_ref,
     (2.0 * D * (1.0 - alb)) / (4.0 * D * (1.0 + alb) + (1.0 - alb) * dx),
     (2.0 * D * neig_D) / (neig_dx * D + dx * neig_D))
```

where `self._notfirst_x_accel` and `self._first_x_accel` and `self._is_adj_ref_left` are precomputed and stored as class instance variables in the function `_precompute_array_indices`.

In addition, I also precompute the rows and columns of CMFD CSR matrices so that the only thing that updates at each batch is the data entries that correspond to these row and columns.

Here are the performance results comparing the two vectorized approaches, where "Python (Vectorized, Updated)" refers to the updated code that was just pushed.

totaltime-1d1grp
totaltime-2d1grp
totaltime-3d1grp
cmfdtime-1d1grp
cmfdtime-2d1grp
cmfdtime-3d1grp
buildmattime-1d1grp
buildmattime-2d1grp
buildmattime-3d1grp
solvemattime-1d1grp
solvemattime-2d1grp
solvemattime-3d1grp
dhattime-1d1grp
dhattime-2d1grp
dhattime-3d1grp
fstime-1d1grp
fstime-2d1grp
fstime-3d1grp

These plots show that the updated approach is on par in terms of performance with the previous vectorized implementation, and any slight discrepancies can probably be attributed to the stochastic nature of how the time in CMFD is quite a small fraction of the total simulation time.

2D BEAVRS results of runs with updated vectorized functions are pending, the timing results for those runs should be more telling of performance for larger problems. Do let me know if there are other parts of the code right now that could be made more readable.

shikhar413:
Adding the results for a 2D BEAVRS run with 10 million particles, 150 inactives and 300 batches, using 60 cores. The Fortran implementation, un-vectorized Python implementation, and vectorized Python implementation show consistent results. Here are the timing results using the updated vectorized code and the initial vectorized code, respectively, and they show consistent timing results as well.

Updated vectorized:
```
=====================>     CMFD TIMING STATISTICS     <====================

   Time in CMFD                    =  7.70570E+02 seconds
     Building matrices             =  4.02047E-01 seconds
     Solving matrices              =  5.18417E+02 seconds
     Compute dhat, dtilde          =  6.87992E-01 seconds
     Compute fission source        =  7.74291E-02 seconds
```

Initial vectorized:
```
=====================>     CMFD TIMING STATISTICS     <====================

   Time in CMFD                    =  7.73975E+02 seconds
     Building matrices             =  7.61484E-01 seconds
     Solving matrices              =  5.24530E+02 seconds
     Compute dhat, dtilde          =  8.39190E-01 seconds
     Compute fission source        =  8.91912E-02 seconds
```

smharper:
@shikhar413, thanks for all the timing data.  We can probably just focus on the 3D timing cases + BEAVRS from here on out.  I'd like to see if we can get those 3D CMFD runtimes down.  Luckily, your plots show that all the expense is in the solver itself so we don't have to mess with all the data-handling too much.  Can you try moving just the linear solver to C++ through ctypes?  I'm hoping that will be a pretty minimal amount of C++ code that will still get us closer to the Fortran-only runtimes.  It looks like numpy gives a [ctypes interface](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ndarray.ctypes.html#numpy.ndarray.ctypes) for its arrays.  Not sure if it's easy to use or not, but I think it'd be good to try out.

smharper:
Found another good link on this topic [here](https://www.scipy-lectures.org/advanced/interfacing_with_c/interfacing_with_c.html#id6)

shikhar413:
I'll take a look into this. The code seems pretty straightforward, but the one part that I'm not sure about is creating the shared library that would link C and Python code. I'm guessing that this would require some modification within cmakelists?

smharper:
No changes to the cmakelists should be required (other than possibly adding a new source file).  OpenMC is already compiled as the libopenmc.so shared library.  We can stick the GS solver in with the rest of the C++ source and access it through ctypes just like we already do elsewhere in the C-API

shikhar413:
Pushing code for running the linear solver through C++ instead of the scipy solver, and it seems as though this approach is much faster, in many cases even faster than the Fortran implementation. The timing plots for the 3d 1 group case are below, followed by a table of timing results for 2d BEAVRS with 10 million particles, 150 inactive, and 300 batches, running on 60 cores.


buildmattime-3d1grp
solvemattime-3d1grp
cmfdtime-3d1grp
totaltime-3d1grp

| Solver                          | Time Building Matrices | Time Solving Matrices | Time in CMFD    | Total Time     |
|---------------------------------|------------------------|-----------------------|-----------------|----------------|
| Fortran                         | 4.1132E-02 sec         | 1.1511E+02 sec        | 1.3964E+02 sec  | 3.0956E+04 sec |
| Vectorized Numpy + Scipy Solver | 4.02047E-01 sec        | 5.18417E+02 sec       | 7.70570E+02 sec | 3.1064E+04 sec |
| Vectorized Numpy + C++ Solver   | 4.04467E-01 sec        | 1.19158E+02 sec       | 3.71095E+02 sec | 3.1384E+04 sec |

For the 2D case, the time in the solver is much more on par with the Fortran case, although the total time in CMFD is still slightly higher. Doing some initial profiling on the time spent in CMFD, it looks like the function `cmfd_reweight` is the next largest source of time spent in CMFD after solving the matrices, and my hunch is that the MPI calls through Python slow things down, where this effect is not present in the 1d case because all tests are run on a single core.

With regards to the C++ solver, it mirrors the current solvers written in `cmfd_solver.F90` very closely. I passed Numpy matrices to C++ using ctypes as suggested, and I stored [CSR representations](https://www.scipy-lectures.org/advanced/scipy_sparse/csr_matrix.html) of the loss matrix in C++ in order to implement the solver logic. I ran into issues of data being overridden when storing arrays such as `indices` and `indptr` simply as `int*`, so I stored values in a more persistent manner by saving them as type `std::vector<int>`. I should also note that I save nonvarying variables in the beginning of the simulation in `openmc_initialize_linsolver` and only pass varying data when calling `openmc_run_linsolver`

smharper:
This looks great @shikhar413! Since the CMFD time is two orders of magnitude smaller than the transport time for BEAVRS, I think we can leave cmfd_reweight in Python. If you feel good about the current code architecture, then go ahead and clean up the code (removing unused solvers, etc) and we can go into the final review process.

shikhar413:
Cool. I've cleaned up the code so that now CMFD  calls always invoke vectorized numpy routines and the C++ linear solver. There will be further code clean up once I start replacing Fortran code with the Python/C++ version of CMFD. Should I go ahead and start with that process (will close this PR and start a new PR, cherrypicking any of the updated that I made to this branch)?

smharper:
Yeah, go ahead. You can leave it in this branch/PR, too if you want, no difference to me.

shikhar413:
I created a separate branch that mirrors this one but doesn't keep the timing tests, I'll work on that one so that the commit history is a bit cleaner

PR #1116, opened 2018-11-02
Fix array IndexOutofbounds crash when a input material is never used for MG mode

liangjg:
@wuwenbin2006 reported this bug. Since #1016, we use `push_back` instead of `allocate` to initialize the `macro_xs` array. However, the array length should be equal to the number of materials, though we do not need to load the materials which we do not actually use in the problem.

### DIFF
--- a/src/mgxs_data.F90
+++ b/src/mgxs_data.F90
@@ -126,13 +126,9 @@ contains
 
       name = trim(mat % name) // C_NULL_CHAR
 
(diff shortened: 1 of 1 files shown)


nelsonag:
Given the simplicity of this PR, I will bypass the 36hr merge window

nelsonag:
Thanks @liangjg  and @wuwenbin2006!

PR #1133, opened 2018-12-04
Simplified depletion chain from CASL-ORIGEN

liangjg:
With the help of Jiankai @rockfool, I extracted this 255-nuclides simplified burnup chain used by CASL in VERA-ORIGEN ([Kang Seog Kim, "Specification for the VERA Depletion Benchmark Suite", CASL-U-2015-1014-000, Rev. 0, ORNL/TM-2016/53, 2016](https://info.ornl.gov/sites/publications/files/Pub61302.pdf)).

This PR adds a python dictionary `casl_chain.py` that includes all constituent nuclides in this chain, and a script `openmc-make-depletion-chain-casl` to generate the chain file for `openmc.deplete`, by reading the decay, neutron reaction and fission yield data from ENDF library.

@rockfool, you may upload some preliminary accuracy/performance results using this chain if you have any. Thanks.


paulromano:
@liangjg and @rockfool: Thanks for putting this together! A general question about the resulting chain -- when I run the script, I end up with a bunch of reactions that have `target="Nothing"`. When depleting, this means those reactions will result in a loss of mass. Would it be better to just exclude those reactions altogether so that the original target is not transmuted?

liangjg:
@paulromano you reminded me that whether or not keeping the `target="Nothing"` reactions in the chain have different meanings: keeping them means we will account for the neutron reactions but negelect the transmuted target nuclides (a loss of mass) while to remove `Nothing` means no neutron reactions will be accounted for (no transmution).
For example, for `Ho165`,
``` xml
  <nuclide name="Ho165" reactions="4">
    <reaction type="(n,2n)" Q="-8030400.0" target="Nothing"/>
    <reaction type="(n,3n)" Q="-14660000.0" target="Nothing"/>
    <reaction type="(n,4n)" Q="-23063000.0" target="Nothing"/>
    <reaction type="(n,gamma)" Q="6242400.0" target="Nothing"/>
  </nuclide>
```
Or
``` xml
<nuclide name="Ho165" reactions="0"/>
```
I'm not sure what the original intention was when they developed this simplified chain. But according to the chain plots in the report, I guess to account for the reaction is more likely faithful.
For example,  from `Ho165` to `Ho166` (`Ho166`  is not in the depletion chain), the blue lines are neutron reactions (the numbers below are the neutron capture cross sections).
image
image

@bforget any comments?

bforget:
I think we should leave the nothing and sum all of them in a dummy variable to keep track of lost mass.  I just spoke with Kord about this, and it seems like the common way of doing it when simplifying a depletion chain.

rockfool:
Thanks for your effort, Based on the CASL version deplete chain file, I used to apply VERA depletion pin cell benchmark to conduct the preliminary verification. 
This is the plot of pin cell model.
plot_1a
This is the preliminary results of keff for different depletion chain file. 
results-001

liangjg:
Thanks, @rockfool. It seems the simplified chain is as accurate as the full ENDF chain from the keff curve but halves the transport time.

PR #1156, opened 2019-02-06
Bug fixes for precursor decay rate tallies

smharper:
This PR fixes two bugs for tallies with `SCORE_DECAY_RATE` that do not have a DelayedGroupFilter.  In one case, the wrong nuclide index was used which results in a segfault.  In two cases, the gross delayed nu value was used where a nu value corresponding to just one precursor group should have been used.

With these changes, I'm now seeing that decay-rate values computed without a filter equal the sum of the decay-rate values that you get with a DelayedGroupFilter applied.


smharper:
Good catch on the non-fissionable nuclide segfault.  That should be fixed now.  I added some decay-rate scores to the tallies test.  While I was at it, I also tried to hit both the total material vs. single nuclide branch cases for more scores.

PR #1158, opened 2019-02-08
Increase MG tallies testing coverage

smharper:
This PR increases the amount of code covered by the mg_tallies regression test.  This will be helpful for me as I'm converting tallies over to C++ to make sure I don't get a regression from the F90 code.  There are also a few bugfixes here that became obvious when I increased the coverage.

I think more time needs to be spent checking the MG tally scores for errors, but I'd rather leave that until after I finish the C++ conversion.  (This PR doesn't include the change we talked about @nelsonag.)


smharper:
@nelsonag, it checks out for inverse velocity.  The decay-rate score looks to be off by about a factor of two.  And there are some other issues you can see in these tally results, too, like delayed-nu-fission + prompt-nu-fission do not add up to nu-fission.  I think the root cause here is that some scores are divided by absorption where they should be divided by total.  But I'm not sure, and I'd rather not spend the time figuring it out at the moment.  I'd prefer to get the MG tallies converted to C++ now so they don't hold up other parts of the rewrite, and then revisit the verification.

nelsonag:
Ok, the last one we dont have a status for is kappa-fission.  (I'm looking to have the expected pass-failure of each additional test included before accepting since the info isn't readily apparent from the hashed results of course.)  How does that one look?

smharper:
Kappa fission is good!

nelsonag:
Great @smharper thanks. I'm going to merge now due to the simplicity of this PR.

Issue #1223, opened 2019-04-29
Add method to retrieve cell temperature


aprilnovak:
For simulations with thermal-hydraulics feedback in Picard-style iteration loops, convergence of the coupled solve may be based on the change in a temperature norm between two successive iterations. If OpenMC runs first in the Picard iteration loop, the change in the norm for the very first iteration should compare the most recent T/H solve to the temperatures used in the OpenMC input file.

It would be helpful to add a new method with a signature like:

`extern "C" int openmc_cell_get_temperature(int32_t index, double& T, const int32_t* instance)`

that will allow a coupled driver to determine what temperatures were actually used as the initial condition in OpenMC.

The alternative would be to separately inform a multiphysics driver (i.e. through an XML input file) of the OpenMC initial condition, which 1) is error prone for the user and 2) requires additional development effort on the multiphysics driver side unless you're doing something simple, like setting a uniform temperature in the entire OpenMC model.

PR #1224, opened 2019-04-29
Add openmc_cell_get_temperature method. 

aprilnovak:
Added method to query cell's temperature to inform multiphysics drivers of the initial temperature used in an OpenMC simulation.

I'm fairly new to the python API inner workings, so I more or less followed the pattern in `capi/cell.py`. Apologies if I missed anything!

Refs #1223

### DIFF
--- a/docs/source/capi/index.rst
+++ b/docs/source/capi/index.rst
@@ -73,6 +73,17 @@ Functions
    :return: Return status (negative if an error occurred)
    :rtype: int
 
+.. c:function:: int openmc_cell_get_temperature(int32_t index, const int32_t* instance, double* T)
+
+   Get the temperature of a cell
+
+   :param int32_t index: Index in the cells array
+   :param int32_t* instance: Which instance of the cell. If a null pointer is passed, the temperature
+                             of the first instance is returned.
+   :param double* T: temperature of the cell
+   :return: Return status (negative if an error occurred)
+   :rtype: int
+
 .. c:function:: int openmc_cell_set_fill(int32_t index, int type, int32_t n, const int32_t* indices)
 
    Set the fill for a cell
--- a/openmc/capi/cell.py
+++ b/openmc/capi/cell.py
@@ -24,6 +24,10 @@ _dll.openmc_cell_get_fill.argtypes = [
     c_int32, POINTER(c_int), POINTER(POINTER(c_int32)), POINTER(c_int32)]
 _dll.openmc_cell_get_fill.restype = c_int
 _dll.openmc_cell_get_fill.errcheck = _error_handler
+_dll.openmc_cell_get_temperature.argtypes = [
+    c_int32, POINTER(c_int32), POINTER(c_double)]
+_dll.openmc_cell_get_temperature.restype = c_int
+_dll.openmc_cell_get_temperature.errcheck = _error_handler
 _dll.openmc_cell_set_fill.argtypes = [
     c_int32, c_int, c_int32, POINTER(c_int32)]
 _dll.openmc_cell_set_fill.restype = c_int
@@ -127,6 +131,23 @@ class Cell(_FortranObjectWithID):
         elif fill is None:
             indices = (c_int32*1)(-1)
             _dll.openmc_cell_set_fill(self._index, 1, 1, indices)
+
+    def get_temperature(self, instance=None):
+        """Get the temperature of a cell
+
+        Parameters
+        ----------
+        instance: int or None
+            Which instance of the cell
+
+        """
+
+        if instance is not None:
+          instance = c_int32(instance)
+
+        T = c_double()
+        _dll.openmc_cell_get_temperature(self._index, instance, T)
+        return T.value
 
     def set_temperature(self, T, instance=None):
         """Set the temperature of a cell
--- a/openmc/capi/cell.py
+++ b/openmc/capi/cell.py
@@ -143,7 +143,7 @@ class Cell(_FortranObjectWithID):
         """
 
         if instance is not None:
-          instance = c_int32(instance)
+            instance = c_int32(instance)
 
         T = c_double()
         _dll.openmc_cell_get_temperature(self._index, instance, T)
@@ -160,7 +160,11 @@ class Cell(_FortranObjectWithID):
             Which instance of the cell
 
         """
-        _dll.openmc_cell_set_temperature(self._index, T, c_int32(instance))
+
+        if instance is not None:
+            instance = c_int32(instance)
+
+        _dll.openmc_cell_set_temperature(self._index, T, instance)
 
 
 class _CellMapping(Mapping):
--- a/src/cell.cpp
+++ b/src/cell.cpp
@@ -922,6 +922,36 @@ openmc_cell_set_temperature(int32_t index, double T, const int32_t* instance)
   return 0;
 }
 
+extern "C" int
+openmc_cell_get_temperature(int32_t index, const int32_t* instance, double* T)
+{
+  if (index < 0 || index >= model::cells.size()) {
+    strcpy(openmc_err_msg, "Index in cells array is out of bounds.");
+    return OPENMC_E_OUT_OF_BOUNDS;
+  }
+
+  Cell& c {*model::cells[index]};
+
+  if (c.sqrtkT_.size() < 1) {
+    strcpy(openmc_err_msg, "Cell temperature has not yet been set.");
+    return OPENMC_E_UNASSIGNED;
+  }
+
+  if (instance) {
+    if (*instance >= 0 && *instance < c.n_instances_) {
+      double sqrtkT = c.sqrtkT_.size() == 1 ? c.sqrtkT_[0] : c.sqrtkT_[*instance];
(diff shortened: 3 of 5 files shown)


src/cell.cpp
```diff
@@ -922,6 +922,40 @@ openmc_cell_set_temperature(int32_t index, double T, const int32_t* instance)
   return 0;
 }
 
+extern "C" int
+openmc_cell_get_temperature(int32_t index, double* T, const int32_t* instance)
+{
+  if (index >= 0 && index < model::cells.size()) {
```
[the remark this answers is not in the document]
aprilnovak: Good idea, thanks!

openmc/capi/cell.py
```diff
@@ -114,6 +118,12 @@ def fill(self):
         else:
             raise NotImplementedError
 
+    @property
+    def temperature(self, instance=None):
```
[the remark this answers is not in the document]
aprilnovak: I like the idea of having two `capi` functions - it might be unexpected behavior for the user to call `openmc_cell_get_temperature` and receive an averaged temperature when they simply forgot to include an instance.

openmc/capi/cell.py
```diff
@@ -128,6 +132,19 @@ def fill(self, fill):
             indices = (c_int32*1)(-1)
             _dll.openmc_cell_set_fill(self._index, 1, 1, indices)
 
+    def get_temperature(self, instance=None):
```
aprilnovak: @pshriwise I'll change this to the two methods you mentioned in a bit, but is this what you had in mind by getters/setters with values of each cell instance?

pshriwise: Pretty close. I was thinking that getting temperatures would still be like accessing a property of `openmc.capi.cell`, so something like:

```python
@property
def temperatures(self):
    temps = POINTER(c_double)()
    n_temps = c_int()
    _dll.openmc_get_cell_temperatures(self._index, temps, n_temps)
    temp_array = as_array(temps, (n.value,))
    return temp_array
```

`openmc.capi.material.densities` is a good reference for this.

This also makes me realize I messed up my proposed capi functions above :)

`openmc_get_cell_temperature` should probably be `openmc_get_cell_temperatures` with a signature like

`openmc_get_cell_temperatures(int32_t index, double **T, int* n)`

My bad!

docs/source/capi/index.rst
```diff
@@ -73,6 +73,17 @@ Functions
    :return: Return status (negative if an error occurred)
    :rtype: int
 
+.. c:function:: int openmc_cell_get_temperature(int32_t index, double* T, const int32_t* instance)
```
paulromano: I would move `instance` before `T` since it is an [in] parameter and T is an [out] parameter

paulromano:
@pshriwise It sounds like you're suggesting that we expose the underlying array of temperatures for each cell instance directly as a numpy array through the Python API. I think for now I'd steer clear of that, leaving just a function to manipulate a specific cell instance (which is what @aprilnovak has here).

There's kind of a bigger issue right now that the model has to be set up carefully in order for cell temperature modifications to work. Specifically, if a cell is repeated in the geometry, the user is responsible for assigning a list of temperature values, even if they are all the same. Otherwise, OpenMC will only use a single value in the temperature vector for the cell, and it's not possible to customize the temperature of each instance of the cell.

pshriwise:
Ooh, that's a good point. I hadn't thought of that complication! In that case, yeah, the `get_temperature` method @aprilnovak's added looks good to me.

At some point, could we in theory expand the `cell:sqrtkT_` vector in a hypothetical `openmc_set_temperatures` function (assuming the correct number of temps is passed in) and return repeated values in a hypothetical `openmc_get_temperatures` function for the common temperature scenario?

src/cell.cpp
```diff
@@ -922,6 +922,39 @@ openmc_cell_set_temperature(int32_t index, double T, const int32_t* instance)
...
+  Cell& c {*model::cells[index]};
+
+  if (c.sqrtkT_.size() < 1) {
+    strcpy(openmc_err_msg, "Cell temperature has not yet been set.");
+    return OPENMC_E_UNASSIGNED;
+  }
+
+  if (instance) {
+    if (*instance >= 0 && *instance < c.sqrtkT_.size()) {
+      *T = c.sqrtkT_[*instance] * c.sqrtkT_[*instance] / K_BOLTZMANN;
```
pshriwise: Along the lines of [this morning's conversation](https://github.com/openmc-dev/openmc/pull/1224#issuecomment-488260908), if the cell instances are sharing a common temperature we'll want to return the first (and only) value in the vector as the cell temperature.

```cpp
if (*instance >= 0 && *instance < c.n_instances_) {
  double sqrtkT = c.sqrtkT_.size() == 1 ? c.sqrtkT_[0] : c.sqrtkT_[*instance];
  *T = sqrtkT * sqrtKT / K_BOLTZMAN;
} else { 
...
}
```

aprilnovak: Ah, okay, so multiple instances of a cell may exist if they either use unique temperatures + materials or they share a common temperature + material?

docs/source/capi/index.rst
aprilnovak: Makes sense. I just liked the symmetry with the `openmc_cell_set_temperature` method.

openmc/capi/cell.py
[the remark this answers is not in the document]
aprilnovak: Just checking that I'm up to date on where the conversations have taken us - @pshriwise, are you still wanting the `openmc_cell_get_temperature` method to not compute the average temp if no instance is passed (and have that instead be a separate method)?

src/cell.cpp
pshriwise: (Sorry for the delay)

Yup yup. For clarity, it is possible for cell instances can have either unique temperatures or unique materials or both, depending on the input.

openmc/capi/cell.py
```diff
@@ -128,6 +132,19 @@ def fill(self, fill):
...
+        """Get the temperature of a cell
+
+        Parameters
+        ----------
+        instance: int or None
+            Which instance of the cell
+
+        """
+        T = c_double()
+        _dll.openmc_cell_get_temperature(self._index, c_int32(instance), T)
```
pshriwise: `c_int32(instance)` isn't happy taking `None`. We'll need a check here for that:

```python
if instance is not None:
    instance = c_int32(instance)
_dll.openmc_cell_get_temperature(self._index, instance, T)
```

openmc/capi/cell.py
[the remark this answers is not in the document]
pshriwise: Yeah that was my thought. Or even just add `get_temperature` for now. Is there a current need for an averaged cell temperature?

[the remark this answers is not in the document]
aprilnovak: Not that I can think of - I originally included the average as some type of sensible calculation if an instance wasn't provided, but I'm not sure we need it either

openmc/capi/cell.py
aprilnovak: Apologies for the delay on my end, I'm new to this capi module. Would there not also be a problem with `instance` taking `None` in `set_temperature`? It's not clear to me why `get_temperature` would need this check but `set_temperature` wouldn't

openmc/capi/cell.py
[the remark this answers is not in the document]
pshriwise: Maybe best to leave it out for now then. If one is using a model with distributed cell temperatures, hopefully they'll know they can gather them all via this function and compute an average if needed.

openmc/capi/cell.py
pshriwise: Very fair question. `set_temperature` *does* need this check. It currently won't work without a value for `instance`. After confirming, would you mind adding a similar check there as well?

paulromano: On that topic, you should add a test somewhere in `tests/unit_tests/test_capi.py` for get_temperature, and perhaps we ought to add the `None` cases there too.

pshriwise:
This is looking pretty good to me. Once the fix from #1230 is added, I'd say it's ready to go!

pshriwise:
Will merge today if there are no further comments. Thanks @aprilnovak!

aprilnovak:
Thanks @pshriwise! Nothing further from me

openmc/capi/cell.py
```diff
@@ -128,6 +132,23 @@ def fill(self, fill):
...
+
+        Parameters
+        ----------
+        instance: int or None
+            Which instance of the cell
+
+        """
+
+        if instance is not None:
+          instance = c_int32(instance)
```
paulromano: Four spaces per PEP8

openmc/capi/cell.py
```diff
@@ -139,7 +160,11 @@ def set_temperature(self, T, instance=None):
             Which instance of the cell
 
         """
-        _dll.openmc_cell_set_temperature(self._index, T, c_int32(instance))
+
+        if instance is not None:
+          instance = c_int32(instance)
```
paulromano: Four spaces

paulromano:
I'm good to merge after the trivial style fixes

PR #1272, opened 2019-06-26
WIP New fission energy deposition mode: local photons

drewejohnson:
This PR is related to #1238, improving the indirect fission energy deposition. 

## 26 June 
The value of recoverable energy from fission is reduced by subtracting out the energy from `prompt_neutrons` and `delayed_neutrons` from the `fission_release` group. Slight modifications were made to `endf` functions to try and keep much of the tally logic the same
https://github.com/openmc-dev/openmc/blob/d10265ac2c46536a96eb6510ad16e103cbf9dd93/src/tallies/tally_scoring.cpp#L1051.
This PR introduces a new setting to control this behavior: `fiss_edep_mode` with two options: `local` [the previous default] and `local photons` [the new default].  `local photons` obtains this modified `fiss_q_recov_` function. The choice to make this the default stemmed from the fact that KERMA heating cross sections are distributed with `openmc`, and since this makes a noticeable improvement in the energy deposition, we should use them.

### Testing

I have done some very simple testing, just tallying energy deposition in a pin cell with and without this new feature. Statistics are lowish, but other tallies, like flux and fission rate, were identical. Below is the reported recoverable fission energy with and without this feature in fuel, cladding, and water.

| Material | Q_recov local | Q_recov local photons
| --- | --- | --- |
| Fuel| 1.17829e+08 +/- 139137 | 1.17816e+08 +/- 139134| 
| Cladding | 0 | 0 |
| Coolant | 0 | 0 |

## 28 June

Some status updates. Realized that, for some nuclide, the fission energy release data is not provided for/through the `Tabulated1D` class, but for/through the `Polynomial` class. 
Also, after some discussions, it may be possible to obtain the "energy deposition" through a combination of `heating` and `fission_q_recov` tallies. Care will have to be taken to use/modify nuclear data appropriately, because the heating coefficients also include recoverable fission energy, and we want to avoid double counting. 

## 1 July

PR #1278 would be used to take Q values directly from a `fission_q_recov` tally in each burnable materials, and across fissile isotopes. 

Below will be a list of remaining action items.
- [x] Allow neutrons to deposit fission energy over time [N/A if using heating, `Q` tallies]
- [ ] Additional documentation in theory / methodology
- [ ] Fix tallies regression test
- [ ] Update`Polynomial` interpolation class to work with updated `q_recoverable` data
- [ ] "Smart" energy deposition tally to include `heating` and `fission_q_recoverable`


drewejohnson:
This PR has been replaced by #1342

PR #1301, opened 2019-07-23
Allow user control over initial concentration for trace isotopes

### LINKED ISSUE #1288: Incorrect value read from openmc.deplete.ResultsList for fisrst simulation time step
I were testing deplete functionality both with old and new *.hdf5 cross-sections version and  when try to get `results = od.ResultsList('depletion_results.h5')
                                          results.get_atoms('1','nuclide_name')`
concentration for nuclide initially not pointed in materials.xml for first time step I received the same value 4.68084739e+02.

drewejohnson:
Closes #1288 by allowing the user to specify the value of `dilute_initial`  when initializing an `Operator`. This value is allowed to be zero to instruct Openmc not to add any trace isotopes into the problem. The default value of 1000 atoms / cm^3 is left in place, but some documentation is added in the `ResultsList.get_atoms` methods and io format documentation about why some values may be non-zero for these trace isotopes. 

## Some discussion 

I ran some comparison tests on a quarter PWR assembly over the weekend without depletion, with the current dilution value of 1E-3 atoms/cm^3, and a value of zero initial trace concentration. The error in reactivity compared to the reference case at time zero was _slightly_ greater [8.56 pcm vs. 8.04 pcm] when using a dilution of 1E-3 atoms/cm. For comparison, the uncertainty from the steady state calculation was 18.8 pcm.

However, when you look at k over time, the presence of the trace isotopes makes a noticeable effect. The two figures below show how the system eigenvalue and Xe concentration change due to these settings.

k-deplete

xe

By adding the trace isotopes, we appear to be helping the xenon equilibrium, as the build up of initial xenon causes a sharp drop in k during the first depletion step. Once the system reaches or further approaches this equilibrium, we have good but not complete agreement. 

I bring this up to highlight the effect of the default setting, whatever it is. The operator is able to handle having a value of 0 for dilution, but I'm not sure how the xenon equilibrium feature [#1245] would fare. Looking over the methodology proposed in the M&C paper, one would have zero-valued xenon absorption cross section initially, but xenon is introduced during the equilibrium search anyway. I think openmc will cause errors if you construct a tally for a nuclide that doesn't exist in the problem.

### DIFF
--- a/openmc/deplete/abc.py
+++ b/openmc/deplete/abc.py
@@ -10,11 +10,12 @@ from pathlib import Path
 from abc import ABC, abstractmethod
 from xml.etree import ElementTree as ET
 from warnings import warn
+from numbers import Real
 
 from numpy import nonzero, empty
 
 from openmc.data import DataLibrary, JOULE_PER_EV
-from openmc.checkvalue import check_type
+from openmc.checkvalue import check_type, check_greater_than
 from .chain import Chain
 
 OperatorResult = namedtuple('OperatorResult', ['k', 'rates'])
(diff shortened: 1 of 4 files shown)


openmc/deplete/abc.py
```diff
@@ -52,17 +54,21 @@ class TransportOperator(metaclass=ABCMeta):
...
+    dilute_initial : float, optional
+        Initial atom density [atoms/cm^3] to add for nuclides that are zero
+        in initial condition to ensure they exist in the decay chain.
+        Only done for nuclides with reaction rates.
+        Defaults to 1.0e3.
 
     Attributes
     ----------
     dilute_initial : float
         Initial atom density to add for nuclides that are zero in initial
```
pshriwise: The units of this value are specified elsewhere, let's do that here too.

Issue #1359, opened 2019-09-21
rotate_angle incorrect


gridley:
Greetings peeps,

Seems that rotate_angle isn't working correctly. IIRC, the polar coordinates phi=0, theta=pi/2 should be a unit vector that point along the x axis, but in OpenMC, rotate_angle returns a unit vector that points along the -y axis for these polar inputs, using the reference (polar) axis as +z.

Should be an easy fix!

Check out this tiny example program if you don't believe me.

```cpp
#include "math_functions.h"
#include <iostream>

using namespace std;
using namespace openmc;

// Is this the expected direction result?

int main()
{
  Direction z_axis = {0., 0., 1.};
  double phi = 0.0;
  Direction should_be_x = rotate_angle(z_axis, 0., &phi);
  cout << "Components of theta=pi/2, phi=0:" << endl;
  cout << "x = " << should_be_x.x << endl;
  cout << "y = " << should_be_x.y << endl;
  cout << "z = " << should_be_x.z << endl;
}
```
Gives:

```
gavin@gpad:~/code/openmc$ g++ test.cpp -Iinclude/openmc build/lib/libopenmc.so 
gavin@gpad:~/code/openmc$ ./a.out 
Components of theta=pi/2, phi=0:
x = 0
y = -1
z = 0
```

Or, maybe our semantics about polar coordinates are different.

smharper:
What application brought up this issue? Is it setting the angular distribution of a source for a fixed source calculation?

The only other use for `rotate_angle` that I know of is our collision physics. It works fine there because the azimuth reference direction is arbitrary in those cases. I get the argument for changing `rotate_angle` to make the fixed source setup match the conventions better, but I'm hesitant because it will break reproducibility of particle tracks from the current version. Maybe we could modify the way that source site sampling uses `rotate_angle`.

makeclean:
That might also explain an oddity we found for the use of an angular dependent neutron source, which didn't behave as we expected. We presumed it was an error in our method to produce the source and figured we'd come back to it. I think certainly for sources this convention break is problematic.

When you say reproducibility of tracks, you mean pre and post 'fix' right?  I'm not fully caffeined yet, but I don't why that would be a problem, isit the tests?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

gridley:
[...]

PR #1363, opened 2019-09-24
Rotate azimuthal distributions to match conventional coordinate systems

smharper:
This PR addresses #1359 and the use of `rotate_angle` for source distributions. Previously we only used `rotate_angle` for physics where the zero point for the azimuthal angle is arbitrary and essentially meaningless. But we also now use `rotate_angle` when sampling particles with a specified `PolarAzimuthal` distribution. In that case, we are using a global spherical coordinate system where established conventions say that mu=0, phi=0 should point in the direction of the positive x-axis (assuming the zenith is along the positive z-axis). This PR fixes `PolarAzimuthal` sampling to match that convention.

I decided to change `PolarAzimuthal` rather than `rotate_angle` because changes to `rotate_angle` would break track reproducibility across versions.


smharper:
@gridley, would you mind checking to see if this PR gives the behavior you're expecting?

gridley:
Hey @smharper, thanks for putting a fix on this. I tried pulling your code, but am maybe making some mistake with OpenMC? I wanted to simplify my input some to test this, and made an input representative of a tub of water with neutrons emitted with phi uniform between 0 and pi and mu likewise between -1 and 1 from roughly the origin. The following input does this:

```python
#!/usr/bin/env python3
import openmc as mc
import numpy as np

m1 = mc.model.borated_water(0)
mats = mc.Materials((m1,))
mats.export_to_xml()

box = mc.model.rectangular_prism(5.0, 20.0, boundary_type='vacuum')
wa_cell = mc.Cell(cell_id=1, name='water_zone', fill=m1)
wa_cell.region = box

uni = mc.Universe(cells=(wa_cell,) )
geom = mc.Geometry(uni)
geom.export_to_xml()

sett = mc.Settings()
sett.batches = 20
sett.inactive = 10
sett.particles = 5000
source_region = mc.stats.Box((-.1,-.1,-.1), (.1, .1, .1))
costheta_distr = mc.stats.Uniform(a=-1, b=1)
phi_distr = mc.stats.Uniform(a=0, b=np.pi)
whiteInCurrent = mc.stats.PolarAzimuthal(mu=costheta_distr, phi=phi_distr)
sett.source = mc.Source(space=source_region, angle=whiteInCurrent)
sett.run_mode = 'fixed source'
sett.export_to_xml()

nx = 100
mesh = mc.RegularMesh()
mesh.lower_left = -5., -20., -200.
mesh.dimension = nx, 4*nx, 1
mesh.width = 10., 40., 400.
mesh_filt = mc.MeshFilter(mesh)

t1 = mc.Tally(tally_id=1)
t1.scores = ['flux']
t1.filters = [mesh_filt]
t1.estimator = 'tracklength'
tallies = mc.Tallies([t1])
tallies.export_to_xml()
```

I assume that by not specifying neutron energy here, fission-ish spectrum neutrons are used?

Anyways, unfortunately, my mesh seems to only have tallies in the first cell. Do you know why this may happen? Moreover, the openmc-plot-mesh-tally seems to just hang up on my computer. Maybe for that reason?

smharper:
`mesh.width` sets the width of each mesh bin, not the overall mesh. You can instead specify `mesh.upper_right` to get something close to what I think you're looking for. As for the openmc-plot-mesh-tally, I think it's just being slow for that kind of fine mesh.

gridley:
Ah, thanks for the tip. Yes, that was pretty much what I was looking for.

For the mesh plotter, indeed, it it just slow on a 100x400 mesh. I was surprised; Serpent which I've been accustomed to can plot meshes like this lickety-spit. Guess that comes with compiling your plotting tools.

Anyways, after fixing that, I obtained the expected result below with neutrons emitted with phi in [0,pi]. The PR seems to work as expected.

correct_phi

paulromano:
@smharper Just wanted to ping you on this. I'd like to get this PR in for the next release

smharper:
@paulromano, sorry for the delay. I've been looking more closely at `rotate_angle` because we get sensible results for sources where the source `reference_uvw` is not along the z-axis. A change to `rotate_angle` would only need to modify the special if-case for z-axis aligned directions.

smharper:
I've modified it now so the fix is only applied for `reference_uvw`s along the z-axis. I've also left some code commented out and marked with a `TODO` so we can modify the behavior of `rotate_angle` directly in the future.

paulromano:
Awesome, thanks @smharper!

PR #1385, opened 2019-10-23
Fix for Mesh __repr__  methods

pshriwise:
I noticed that the `Mesh` classes in our Python API were failing to generate their string representations due to a missing `_type` attribute. Even doing something like:

```python
import openmc

mesh = openmc.RegularMesh()

print(mesh)

mesh = openmc.RectilinearMesh()

print(mesh)

mesh.x_grid = (-1.0, 0.0, 0.5, 10.0)

print(mesh)
```

would fail with 

```python
  File "mesh_test.py", line 5, in <module>
    print(mesh)
  File "/home/shriwise/opt/openmc/src/openmc/mesh.py", line 193, in __repr__
    string += '{0: <16}{1}{2}\n'.format('\tType', '=\t', self._type)
AttributeError: 'RegularMesh' object has no attribute '_type'
```
I was a little stuck with what to do about reporting the x/y/z points for `RectilinearMesh`, so I settled for the number of points in each dimension and their extents. The output now looks like:

```bash
RegularMesh
	ID             =	1
	Name           =	
	Dimensions     =	None
	Mesh Cells     =	None
	Width          =	None
	Origin         =	None
	Pixels         =	None

RectilinearMesh
	ID             =	2
	Name           =	
	Dimensions     =	3
	N X pnts:      =	None
	N Y pnts:      =	None
	N Z pnts:      =	None

RectilinearMesh
	ID             =	2
	Name           =	
	Dimensions     =	3
	N X pnts:      =	4
	X Min:         =	-1.0
	X Max:         =	10.0
	N Y pnts:      =	None
	N Z pnts:      =	None
```

### DIFF
--- a/openmc/mesh.py
+++ b/openmc/mesh.py
@@ -530,14 +530,24 @@ class RectilinearMesh(MeshBase):
         self._z_grid = grid
 
     def __repr__(self):
+        fmt = '{0: <16}{1}{2}\n'
         string = super().__repr__()
-        string += '{0: <16}{1}{2}\n'.format('\tDimensions', '=\t', self.n_dimension)
+        string += fmt.format('\tDimensions', '=\t', self.n_dimension)
         x_grid_str = str(self._x_grid) if not self._x_grid else len(self._x_grid)
+        string += fmt.format('\tN X pnts:', '=\t', x_grid_str)
+        if self._x_grid:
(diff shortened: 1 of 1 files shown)


smharper:
This looks good. Thanks @pshriwise!

PR #1399, opened 2019-11-05
TRISO geometry genetration improvement 

rockfool:
This PR is to improve the triso lattice creation when handling pebble bed reactor modeling via python api. My testing based on the model of HTR-PM reactor from Tsinghua University shows that the geometry.xml generation time is reduced from 2000 hours to 10 minutes. By the way, my testing manually applied the xml writing features in PR #1181 .


openmc/model/triso.py
```diff
@@ -827,10 +833,10 @@ def create_triso_lattice(trisos, lower_left, pitch, shape, background):
     triso_locations = {idx: [] for idx in indices}
     for t in trisos:
         for idx in t.classify(lattice):
-            if idx in sorted(triso_locations):
+            if idx in triso_locations:
```
pshriwise: Good call removing sorted here!

openmc/model/triso.py
```diff
@@ -827,10 +833,10 @@ def create_triso_lattice(trisos, lower_left, pitch, shape, background):
...
                 # Create copy of TRISO particle with materials preserved and
                 # different cell/surface IDs
-                t_copy = copy.deepcopy(t)
+                t_copy = t.deepcopy()
```
pshriwise: IIRC, the main reason that we need to use `deepcopy` here is to create a new outer surface object that contains the TRISO's fill universe. While I like the approach here, it might be best to be a little more explicit about that. This is what we found when doing it before anyway. Something like: 

```python
t_copy = copy.copy(t)
t_copy.id = None
t_copy.fill = t.fill
t_copy._surface = openmc.Sphere(r=t._surface.r,
                                x0=t._surface.x0,
                                y0=t._surface.y0,
                                z0=t._surface.z0)
t_copy.region = -t_copy._surface
```

I think would do the trick and require no changes to the test inputs.

rockfool:
> This looks good to me @rockfool! Tried it out along with the other TRISO-related PRs and it worked like a charm. Will merge tomorrow if there are no further comments.

Thanks so much for your comments. This is only minor modification based on your massive and creative work. Finally, openmc has the efficient capability to model the full core of triso-based reactor.

PR #1413, opened 2019-11-18
fix a typo bug in SIIntegrator class

rockfool:
This PR is to fix the typo bug existed in deplete/abc.py when I tried to test the SI-CE/LIIntegrator and SI-LEQIIntegrator.

### DIFF
--- a/openmc/deplete/abc.py
+++ b/openmc/deplete/abc.py
@@ -815,11 +815,11 @@ class SIIntegrator(Integrator):
         reset_particles = False
         if step_index == 0 and hasattr(self.operator, "settings"):
             reset_particles = True
-            self.operator.settings.particles *= self.n_stages
+            self.operator.settings.particles *= self.n_steps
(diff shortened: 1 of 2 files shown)


rockfool:
> Good fix! I was a little confused why this wasn't caught in our test suite, and it turns out the second conditional `hasattr(self.operator, "settings")` isn't true for the `DummyOperator`
> https://github.com/openmc-dev/openmc/blob/09fbc034d84e2d5d0d2307df6ebec602a6082f7a/tests/dummy_operator.py#L132-L136
> 
> I am fine with these changes as you have proposed, and I would also be fine with adding something to the `DummyOperator` to engage this behavior from the SIIntegrators, e.g.
> 
> ```python
> self.settings = Mock()
> self.settings.particles = 10
> ```
> 
> This is also something that can be added in a later PR as well. Let me know how you would like to proceed

Good suggestion. I added explicit settings in dummy_operator, which can help unit_tests to catch such kind of bug now. Thanks.

PR #1414, opened 2019-11-19
use relative distances for coincidence test in hex lattice

### LINKED ISSUE #1383: Change of materials file causes negative lattice distance error
So, I have encountered what may be perhaps the most mind-boggling bug I've ever encountered. I have input below which, when using one materials.xml file, runs fine, but using another materials.xml causes a geometry error. The only words I can muster at the sight of this are: what in tarnation??? This is on the develop branch, btw.

The input is very similar to the hex lattice example input, except with a larger lattice. I have not been able to reproduce this on the smaller lattice case, instead only on one with seven hexagonal rings. So, here is the MWE I've found:

settings.xml

```xml
<settings>
  <run_mode>eigenvalue</run_mode>
  <particles>10000</particles>
  <batches>500</batches>
  <inactive>400</inactive>
  <keff_trigger>
    <type>std_dev</type>
    <threshold>0.0005</threshold>
  </keff_trigger>
  <source strength="1.0">
    <space type="fission">
      <parameters>-1 -1 -1 1 1 1</parameters>
    </space>
  </source>
  <trigger>
    <active>true</active>
    <max_batches>100</max_batches>
  </trigger>
</settings>
```

geometry.xml

```xml
<geometry>
  <cell fill="5" id="1" name="Cell 1" region="-1 -3 2" universe="0" />
  <cell id="101" material="1" name="cell 2" region="-5" universe="1" />
  <cell id="102" material="2" name="cell 3" region="5" universe="1" />
  <cell id="500" material="2" name="cell 4" universe="3" />
  <hex_lattice id="5" n_rings="8">
    <pitch>24.0</pitch>
    <outer>3</outer>
    <center>0.0 0.0</center>
    <universes>
       1
      1 1
     1 1 1
    1 1 1 1
   1 1 1 1 1
  1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1
 1 1 1 1 1 1 1
  1 1 1 1 1 1
   1 1 1 1 1
    1 1 1 1
     1 1 1
      1 1
       1</universes>
  </hex_lattice>
  <surface boundary="vacuum" coeffs="0 0 187.5" id="1" type="z-cylinder" />
  <surface boundary="vacuum" coeffs="0.0" id="2" name="core bottom" type="z-plane" />
  <surface boundary="vacuum" coeffs="346.5" id="3" name="core top" type="z-plane" />
  <surface coeffs="0 0 3.9847549744358415" id="5" type="z-cylinder" />
</geometry>
```

Good materials.xml file that gives no geometry errors:

```xml
<materials>
  <material depletable="true" id="1" name="fuel">
    <density units="g/cc" value="4.5" />
    <nuclide ao="1.0" name="U235" />
  </material>
  <material id="2" name="moderator">
    <density units="g/cc" value="1.0" />
    <nuclide ao="1.99968852" name="H1" />
    <nuclide ao="0.00031148" name="H2" />
    <nuclide ao="0.9976206" name="O16" />
    <nuclide ao="0.000379" name="O17" />
    <nuclide ao="0.0020004" name="O18" />
    <sab name="c_H_in_H2O" />
  </material>
</materials>
```

Bad materials.xml file that leaks particles, somehow:

```xml
<materials>
  <material depletable="true" id="1" name="salt">
    <density units="g/cc" value="2.14674" />
    <nuclide ao="0.1818336162988115" name="U235" />
    <nuclide ao="0.7181663837011885" name="U238" />
    <nuclide ao="2.5725" name="Zr90" />
    <nuclide ao="0.5609999999999999" name="Zr91" />
    <nuclide ao="0.8575" name="Zr92" />
    <nuclide ao="0.869" name="Zr94" />
    <nuclide ao="0.14" name="Zr96" />
    <nuclide ao="29.1" name="Be9" />
    <nuclide ao="0.006499999999999284" name="Li6" />
    <nuclide ao="64.9935" name="Li7" />
    <nuclide ao="146.8" name="F19" />
  </material>
  <material id="2" name="graphite">
    <density units="g/cc" value="1.8" />
    <nuclide ao="0.988922" name="C12" />
    <nuclide ao="0.011078" name="C13" />
    <sab name="c_Graphite" />
  </material>
</materials>
```

Output with good materials.xml:

```
 ====================>     K EIGENVALUE SIMULATION     <====================

  Bat./Gen.      k            Average k
  =========   ========   ====================
        1/1    0.30112
        2/1    0.47196
        3/1    0.53538
        4/1    0.57733
        5/1    0.61784
        6/1    0.62776
        7/1    0.63802
        8/1    0.63641
...
```

Output with bad materials.xml:

```
 ====================>     K EIGENVALUE SIMULATION     <====================

  Bat./Gen.      k            Average k
  =========   ========   ====================
        1/1    0.14539
 WARNING: Particle 8086 had a negative distance to a lattice boundary
 WARNING: Particle 8087 had a negative distance to a lattice boundary
 WARNING: Particle 5710 had a negative distance to a lattice boundary
...
```

And the simulation goes on to run particles incorrectly until the max lost particle count goes too high. So.... yeah. That is going to be tough to debug.

gridley:
This fixes #1383 

Currently, it seems that for large hex lattices, particles get leaked in cells near the boundary. This issue can be reproduced with the input in #1383, where I had the impression that the materials were influencing the particle tracker's ability to move through a hex lattice. It ended up being the case that one of the material files simply had an optically thinner material dominating the lattice, which allowed particles to reach the lattice periphery more quickly. In fact, both configurations give this error after a few particle batches. This only happens if the cells in that periphery are defined using the "outer" keyword, which means that they are registered as invalid lattice indices in OpenMC, triggering a find_cell call.

I have found that particles get leaked in this case because find_cell ends up calling HexLattice::get_indices, which fails to put a particle in the correct cell for lattices of sufficiently large pitch or extent if the particle is crossing a lattice cell boundary far from the origin (>~100cm or so). Ordinarily, get_indices would determine that the particle is "on" a boundary if two distances calculated to cell centers are within FP_COINCIDENT; this fails for large lattices. For the hex lattices I was running, this difference ends up being around 5e-12 after the particle has been advanced to a lattice boundary, and would presumably be larger for larger lattices. This can be probably attributed to the less well-conditioned math of the hex lattices' non-orthogonal coordinates.

So, to fix this, I propose that in checking whether a particle is coincident with a lattice boundary, the relative distance to cell center is used rather than the actual distance.

This took me forever to figure out, and I can't believe it has such a tiny fix! BTW, I have tried re-ordering some arithmetic in hex lattice routines to get better results here, but haven't had much luck. I think this is pretty much unavoidable since non-orthogonal coordinates are intrinsically prone to larger numerical errors.

### DIFF
--- a/src/lattice.cpp
+++ b/src/lattice.cpp
@@ -895,8 +895,12 @@ HexLattice::get_indices(Position r, Direction u) const
       Position r_t = get_local_position(r, i_xyz);
       // calculate distance
       double d = r_t.x*r_t.x + r_t.y*r_t.y;
-      // check for coincidence
-      bool on_boundary = coincident(d, d_min);
+      // check for coincidence. Because the numerical error incurred
+      // in hex geometry is higher than other geometries, the relative
+      // coincidence is checked here so that coincidence is successfully
+      // detected on large hex lattice with particles far from the origin
+      // which have rounding errors larger than the FP_COINCIDENT thresdhold.
+      bool on_boundary = coincident(1.0, d_min/d);
       if (d < d_min || on_boundary) {
         // normalize r_t and find dot product
         r_t /= std::sqrt(d);


pshriwise:
Great find @gridley! I think I was recently running into this problem too.

gridley:
I'm glad to hear that this helped you out @pshriwise! :smile:

PR #1428, opened 2019-12-06
Review by PullRequest, Inc. - Part 3

PullRequestOpen:


openmc/data/decay.py
```diff
@@ -0,0 +1,484 @@
...
+                ex['alpha'] = ufloat(*pairs[11])
+                ex['recoil'] = ufloat(*pairs[12])
+                ex['SF'] = ufloat(*pairs[13])
+                ex['neutron'] = ufloat(*pairs[14])
+                ex['proton'] = ufloat(*pairs[15])
+                ex['neutrino'] = ufloat(*pairs[16])
+
+            items, values = get_list_record(file_obj)
+            spin = items[0]
+            if spin == -77.777:
```
PullRequestOpen: Would be nice as a constant or at least a comment explaining what this means.

openmc/data/decay.py
```diff
@@ -0,0 +1,484 @@
...
+
+    """
+    def __init__(self, ev_or_filename):
+        # Get evaluation if str is passed
+        if isinstance(ev_or_filename, Evaluation):
+            ev = ev_or_filename
+        else:
+            ev = Evaluation(ev_or_filename)
+
+        file_obj = StringIO(ev.section[8, 457])
```
PullRequestOpen: Buffer should be closed here as well.

openmc/data/decay.py
```diff
@@ -0,0 +1,484 @@
...
+        # Assign basic nuclide properties
+        self.nuclide = {
+            'name': ev.gnd_name,
+            'atomic_number': ev.target['atomic_number'],
+            'mass_number': ev.target['mass_number'],
+            'isomeric_state': ev.target['isomeric_state']
+        }
+
+        # Read independent yields
+        if (8, 454) in ev.section:
```
PullRequestOpen: Might be nice to leave a comment here as these numbers leave the reader with little context.

openmc/data/decay.py
```diff
@@ -0,0 +1,484 @@
...
+                decay_type = get_decay_modes(values[6*i])
+                isomeric_state = int(values[6*i + 1])
+                energy = ufloat(*values[6*i + 2:6*i + 4])
+                branching_ratio = ufloat(*values[6*i + 4:6*(i + 1)])
+
+                mode = DecayMode(self.nuclide['name'], decay_type, isomeric_state,
+                                 energy, branching_ratio)
+                self.modes.append(mode)
+
+            discrete_type = {0.0: None, 1.0: 'allowed', 2.0: 'first-forbidden',
```
PullRequestOpen: Nit: discrete implies natural numbers, not floats as the keys are here.

openmc/data/decay.py
```diff
@@ -0,0 +1,484 @@
...
+        self.nuclide = {
+            'name': ev.gnd_name,
+            'atomic_number': ev.target['atomic_number'],
+            'mass_number': ev.target['mass_number'],
+            'isomeric_state': ev.target['isomeric_state']
+        }
+
+        # Read independent yields
+        if (8, 454) in ev.section:
+            file_obj = StringIO(ev.section[8, 454])
```
PullRequestOpen: This buffer is never closed.

https://docs.python.org/3/library/io.html#io.IOBase.close

openmc/data/decay.py
```diff
@@ -0,0 +1,484 @@
...
+            self.energies, self.independent = get_yields(file_obj)
+
+        # Read cumulative yields
+        if (8, 459) in ev.section:
+            file_obj = StringIO(ev.section[8, 459])
```
PullRequestOpen: This buffer is never closed.

https://docs.python.org/3/library/io.html#io.IOBase.close

openmc/data/endf.py
```diff
@@ -0,0 +1,550 @@
...
+        Information about the projectile such as its mass.
+    reaction_list : list of 4-tuples
+        List of sections in the evaluation. The entries of the tuples are the
+        file (MF), section (MT), number of records (NC), and modification
+        indicator (MOD).
+
+    """
+    def __init__(self, filename_or_obj):
+        if isinstance(filename_or_obj, (str, PurePath)):
+            fh = open(str(filename_or_obj), 'r')
```
PullRequestOpen: This needs to be closed as well

openmc/data/endf.py
```diff
@@ -0,0 +1,550 @@
...
+        Path to ENDF-6 formatted file
+
+    Returns
+    -------
+    list
+        A list of :class:`openmc.data.endf.Evaluation` instances.
+
+    """
+    evaluations = []
+    with open(str(filename), 'r') as fh:
```
PullRequestOpen: This is a better way to interact with file handles

openmc/data/endf.py
```diff
@@ -0,0 +1,550 @@
...
+    def __repr__(self):
+        if 'zsymam' in self.target:
+            name = self.target['zsymam'].replace(' ', '')
+        else:
+            name = 'Unknown'
+        return '<{} for {} {}>'.format(self.info['sublibrary'], name,
+                                       self.info['library'])
+
+    def _read_header(self):
+        file_obj = io.StringIO(self.section[1, 451])
```
PullRequestOpen: Should be closed as well.

openmc/data/endf.py
```diff
@@ -0,0 +1,550 @@
...
+    -------
+    float
+        The number
+
+    """
+    return float(ENDF_FLOAT_RE.sub(r'\1e\2\3', s))
+
+
+if not _CYTHON:
+    float_endf = py_float_endf
```
PullRequestOpen: This should probably be on line 27

openmc/data/correlated.py
```diff
@@ -0,0 +1,462 @@
...
+        # Initialize list of distributions
+        energy_out = []
+        mu = []
+
+        # Read each outgoing energy distribution
+        for i in range(n_energy_in):
+            idx = ldis + loc_dist[i] - 1
+
+            # intt = interpolation scheme (1=hist, 2=lin-lin)
+            INTTp = int(ace.xss[idx])
```
PullRequestOpen: Uppercase variable name doesn't conform to snake_case naming style

openmc/data/correlated.py
```diff
@@ -0,0 +1,462 @@
...
+        n_discrete_lines = np.empty(len(self.energy_out), dtype=int)
+        offset_e = 0
+        offset_mu = 0
+
+        # Populate offsets and eout array
+        for i, d in enumerate(self.energy_out):
+            n = len(d)
+            offsets[i] = offset_e
+
+            if isinstance(d, Mixture):
```
PullRequestOpen: There are more `Univariate` types than `Mixture`, `Tabular`, and `Discrete`. Consider checking for those specific types or subclassing `CorrelatedAngleEnergy` to handle these specific cases separately, or you could have unintended side effects when given an unhandled `Univariate`. Conversely, each of those types could "know" how to calculate these values themselves, and you just call the relevant instance method.

openmc/data/data.py
```diff
@@ -0,0 +1,385 @@
...
+    gamma1_pi = 0.0
+    for n, I, J in zip(n1f, I1f, J1f):
+        gamma1_pi -= n * I * (7.1 - pi)**(I - 1) * (tau - 1.222)**J
+
+    # Compute the leading coefficient.  This sets the units at
+    #   1 [MPa] * [kg K / kJ] * [1 / K]
+    # = 1e6 [N / m^2] * 1e-3 [kg K / N / m] * [1 / K]
+    # = 1e3 [kg / m^3]
+    # = 1 [g / cm^3]
+    coeff = pressure / R_GAS_CONSTANT / temperature
```
PullRequestOpen: No chance that temperature or pressure are 0, right?

openmc/data/data.py
```diff
@@ -0,0 +1,385 @@
...
+    -------
+    float
+        Water density in units of [g/cm^3]
+
+    """
+
+    # Make sure the temperature and pressure are inside the min/max region 1
+    # bounds.  (Relax the 273.15 bound to 273 in case a user wants 0 deg C data
+    # but they only use 3 digits for their conversion to K.)
+    if pressure > 100.0:
```
PullRequestOpen: Should this be `>= 100.0`? The docstring above reads `Region 1 is limited to liquid water below 100 [MPa]` indicating that `100.0` is also out of the bounds.

openmc/data/data.py
```diff
@@ -0,0 +1,385 @@
...
+
+    metastable = int(state[2:]) if state else 0
+    return (ATOMIC_NUMBER[symbol], int(A), metastable)
+
+
+# Values here are from the Committee on Data for Science and Technology
+# (CODATA) 2014 recommendation (doi:10.1103/RevModPhys.88.035009).
+
+# The value of the Boltzman constant in units of eV / K
+K_BOLTZMANN = 8.6173303e-5
```
PullRequestOpen: Constants with global scope should be declared first

openmc/data/data.py
```diff
@@ -0,0 +1,385 @@
+import itertools
+from math import sqrt
+import os
+import re
+from warnings import warn
+
+
+# Isotopic abundances from Meija J, Coplen T B, et al, "Isotopic compositions
```
PullRequestOpen: Nice citation, this is very helpful.

openmc/data/data.py
```diff
@@ -0,0 +1,385 @@
...
+        Atomic mass of isotope in [amu]
+
+    """
+    if not _ATOMIC_MASS:
+
+        # Load data from AME2016 file
+        mass_file = os.path.join(os.path.dirname(__file__), 'mass16.txt')
+        with open(mass_file, 'r') as ame:
+            # Read lines in file starting at line 40
+            for line in itertools.islice(ame, 39, None):
```
PullRequestOpen: Is it worth pre-processing this file and creating something like  a `.csv` file that could be more easily read?

openmc/data/ace.py
```diff
@@ -0,0 +1,460 @@
...
+
+.. _MCNP: https://laws.lanl.gov/vhosts/mcnp.lanl.gov/
+.. _NJOY: http://t2.lanl.gov/codes.shtml
+.. _ENDF: http://www.nndc.bnl.gov/endf
+
+"""
+
+from pathlib import PurePath
+import struct
+import sys
```
PullRequestOpen: Unused import

openmc/data/ace.py
```diff
@@ -0,0 +1,460 @@
...
+                temperature = float(words[2])
+
+            datastr = ' '.join(lines[2:6]).split()
+            pairs = list(zip(map(int, datastr[::2]),
+                             map(float, datastr[1::2])))
+
+            datastr = '0 ' + ' '.join(lines[6:8])
+            nxs = np.fromstring(datastr, sep=' ', dtype=int)
+
+            n_lines = (nxs[1] + 3)//4
```
PullRequestOpen: The `+ 3` and  `/ 4` seem magical. Can an explanation be provided here?

openmc/data/ace.py
```diff
@@ -0,0 +1,460 @@
...
+    ----------
+    ascii_file : str
+        Filename of ASCII ACE file
+    binary_file : str
+        Filename of binary ACE file to be written
+
+    """
+
+    # Open ASCII file
+    ascii = open(str(ascii_file), 'r')
```
PullRequestOpen: `ascii` is a pre-defined variable in Python. I would recommend not redefining it in this function. Consider an alternate variable name such as `ascii_data` or `ascii_file`

openmc/data/ace.py
```diff
@@ -0,0 +1,460 @@
...
+            Tables from the file to read in.  If None, reads in all of the
+            tables. If str, reads in only the single table of a matching name.
+        verbose : str, optional
+            Whether to display what tables are being read. Defaults to False.
+
+        """
+
+        tables_seen = set()
+
+        lines = [ace_file.readline() for i in range(13)]
```
PullRequestOpen: 13 seems like a magic number used here and elsewhere. Can we move it to a constant?

openmc/data/ace.py
```diff
@@ -0,0 +1,460 @@
...
+    """
+
+    cv.check_type('zaid', zaid, int)
+    cv.check_value('metastable_scheme', metastable_scheme, ['nndc', 'mcnp'])
+
+    Z = zaid // 1000
+    mass_number = zaid % 1000
+
+    if metastable_scheme == 'mcnp':
+        if zaid > 1000000:
```
PullRequestOpen: Can `1000000` be moved to a variable? This is extremely easy to get wrong due to a typo / missing or adding an extra 0.

openmc/data/angle_distribution.py
```diff
@@ -0,0 +1,310 @@
...
+        # Incoming energy grid
+        energy = ace.xss[idx:idx + n_energies]*EV_PER_MEV
+        idx += n_energies
+
+        # Read locations for angular distributions
+        lc = ace.xss[idx:idx + n_energies].astype(int)
+        idx += n_energies
+
+        mu = []
+        for i in range(n_energies):
```
PullRequestOpen: Not a fan of all the magic numbers used in this loop. Consider using variables (or providing comments explaining values)

openmc/mgxs/mgxs.py
```diff
@@ -0,0 +1,5909 @@
...
+                estimator=tally.estimator, exact_filters=True)
+            sp_tally = sp_tally.get_slice(
+                tally.scores, filters, filter_bins, tally.nuclides)
+            sp_tally.sparse = self.sparse
+            self.tallies[tally_type] = sp_tally
+
+        self._loaded_sp = True
+
+    def get_xs(self, groups='all', subdomains='all', nuclides='all',
+               xs_type='macro', order_groups='increasing',
```
PullRequestOpen: I would also make `increasing` a constant, it is used 35 times.

openmc/mgxs/mgxs.py
```diff
@@ -0,0 +1,5909 @@
...
+        columns : Iterable of str
+            Names of the re-named and re-valued columns
+
+        """
+        # Override polar and azimuthal bounds with indices
+        if self.num_polar > 1 or self.num_azimuthal > 1:
+            # First for polar
+            bins = np.linspace(0., np.pi, self.num_polar + 1, True)
+            _df_column_convert_to_bin(df, 'polar low', 'polar bin', bins)
+            del df['polar high']
```
PullRequestOpen: Might be better to use `.pop()` here instead of `del` just in case `polar high` doesn't exist.

openmc/mgxs/mgxs.py
```diff
@@ -0,0 +1,5909 @@
...
+        self._sparse = False
+        self._loaded_sp = False
+        self._derived = False
+        self._hdf5_key = None
+        self._valid_estimators = ESTIMATOR_TYPES
+
+        self.name = name
+        self.by_nuclide = by_nuclide
+
+        if domain_type is not None:
```
PullRequestOpen: This check is not really needed. If `domain_type` is `None` it'll just remain `None`. This applies for the other checks below for `domain` and `energy_groups`.

openmc/mgxs/mgxs.py
```diff
@@ -0,0 +1,5909 @@
...
+            clone._tallies = OrderedDict()
+            for tally_type, tally in self.tallies.items():
+                clone.tallies[tally_type] = copy.deepcopy(tally, memo)
+
+            memo[id(self)] = clone
+
+            return clone
+
+        # If this object has been copied before, return the first copy made
+        else:
```
PullRequestOpen: Definitely a style decision but it would probably be easier to read if you immediately returned `existing` when you found it in the memoization cache.

openmc/mgxs/mgxs.py
```diff
@@ -0,0 +1,5909 @@
...
+                     'mesh': openmc.MeshFilter}
+
+# Supported domain classes
+_DOMAINS = (openmc.Cell,
+            openmc.Universe,
+            openmc.Material,
+            openmc.RegularMesh)
+
+# Supported ScatterMatrixXS angular distribution types
+MU_TREATMENTS = ('legendre', 'histogram')
```
PullRequestOpen: The string `legendre` is used ~108 times in this. I think you should make it a constant.

openmc/mgxs/mgxs.py
```diff
@@ -0,0 +1,5909 @@
...
+
+    def build_hdf5_store(self, filename='mgxs.h5', directory='mgxs',
+                         subdomains='all', nuclides='all',
+                         xs_type='macro', row_column='inout', append=True,
+                         libver='earliest'):
+        """Export the multi-group cross section data to an HDF5 binary file.
+
+        This method constructs an HDF5 file which stores the multi-group
+        cross section data. The data is stored in a hierarchy of HDF5 groups
+        from the domain type, domain id, subd
```
PullRequestOpen: This same conditional is copied three times. Maybe this can be factored out into it's own method?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+            The domain type for spatial homogenization
+        energy_groups : openmc.mgxs.EnergyGroups
+            The energy group structure for energy condensation
+        by_nuclide : bool
+            If true, computes cross sections for each nuclide in domain.
+            Defaults to False
+        name : str, optional
+            Name of the multi-group cross section. Used as a label to identify
+            tallies in OpenMC 'tallies.xml' file. Defaults to the empty string.
+        delayed_groups : list of int
```
PullRequestOpen: Question: Is this also optional (it is defaulted to `None`).  If so, it should be noted here similar to the other "optional" parameters.  If it is not optional, it is more standard to have non-optional parameters precede optional parameters.

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+
+    @property
+    def scores(self):
+        return ['delayed-nu-fission', 'delayed-nu-fission']
+
+    @property
+    def filters(self):
+        # Create the non-domain specific Filters for the Tallies
+        group_edges = self.energy_groups.group_edges
+        energyout = openmc.EnergyoutFilter(group_edges)
```
PullRequestOpen: Question: The relationship between `energyout` and `energyin` isn't 100% clear to me from these slices into `group_edges`.

It tracks that that `energyin` would be the last and first index of `group_edges`, but if that is the case I would believe that those indices would be excluded by `energyout`?

If this case is correct, a clarifying comment here (specifically clarifying how non-domain specific filters for the tallies relates to `group_edges` and the last/first index) would be helpful.

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+            if self.by_nuclide:
+                densities = self.get_nuclide_densities(nuclides)
+            else:
+                densities = self.get_nuclide_densities('sum')
+            if value == 'mean' or value == 'std_dev':
+                xs /= densities[np.newaxis, :, np.newaxis]
+
+        # Eliminate the trivial score dimension
+        xs = np.squeeze(xs, axis=len(xs.shape) - 1)
+        xs = np.nan_to_num(xs)
```
PullRequestOpen: Question: Is this `nan_to_num` call related to the `numpy.squeeze` call above to eliminate the trivial score dimension, or is it a general `nan_to_num` call for the whole multi-dimensional array?  As far as I know, the `squeeze` operation does not explicitly produce `NaN`'s that did not exist prior.

If `nan_to_num` is general for the matrix, I would recommend adding a comment here explaining the significance of eliminating `NaN`'s in the multi-dimensional array.  My first guess would be that calculations can not be run on `NaN`'s, but it is not 100% clear from context.

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+    filters : list of openmc.Filter
+        The filters in each tally used to compute the multi-group cross section
+    tally_keys : list of str
+        The keys into the tallies dictionary for each tally used to compute
+        the multi-group cross section
+    estimator : {'tracklength', 'analog'}
+        The tally estimator used to compute the multi-group cross section
+    tallies : collections.OrderedDict
+        OpenMC tallies needed to compute the multi-group cross section. The keys
+
```
PullRequestOpen: Question: Any reason `subdomains` has a max_depth of 3 here? Is that a sanity-check limitation, or is it something that has business or functional importance elsewhere in the logic?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+            # nuclides in the domain
+            if nuclides == 'sum' or nuclides == ['sum']:
+
+                # Retrieve the fission production tallies
+                delayed_nu_fission_in = self.tallies['delayed-nu-fission-in']
+                delayed_nu_fission_out = self.tallies['delayed-nu-fission-out']
+
+                # Sum out all nuclides
+                nuclides = self.get_nuclides()
+                delayed_nu_fission_in = delayed_nu_fission_in.summation\
```
PullRequestOpen: Nit: It's not strictly non-PEP8, but it's very uncommon to line break between the function name and the parenthesis of the function call.  Both for here and below.

There are two alternatives I would recommend:

Compress it to one line: 
`delayed_nu_fission_in.summation(nuclides=nuclides)`

Do a line break after the parenthesis, rather than before.  Example:

```
delayed_nu_fission_in.summation(
   nuclides=nuclides
)
```

openmc/mgxs/mdgxs.py
PullRequestOpen: Spelling / Typo: Did you mean "This is equal to the number" rather than "When the  This is equal to the number"?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+    domain_type : {'material', 'cell', 'distribcell', 'universe', 'mesh'}
+        The domain type for spatial homogenization
+    groups : openmc.mgxs.EnergyGroups
+        The energy group structure for energy condensation
+    by_nuclide : bool
+        If true, computes cross sections for each nuclide in domain
+    name : str, optional
+        Name of the multi-group cross section. Used as a label to identify
+        tallies in OpenMC 'tallies.xml' file.
+    delayed_groups : list of int
```
PullRequestOpen: It looks like `delayed_groups` has a default of `None`.  Is it a `optional` parameter for this method? If so, it should be listed here as as `optional` similar to the `name` and `num_polar` parameters.

If it is not `optional`, it should precede any optional parameters in the method parameter ordering.

openmc/mgxs/mdgxs.py
PullRequestOpen: Spelling / Typo: Did you mean "This is equal to the number" rather than "When the  This is equal to the number"?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+    rxn_rate_tally : openmc.Tally
+        Derived tally for the reaction rate tally used in the numerator to
+        compute the multi-group cross section. This attribute is None
+        unless the multi-group cross section has been computed.
+    xs_tally : openmc.Tally
+        Derived tally for the multi-group cross section. This attribute
+        is None unless the multi-group cross section has been computed.
+    num_subdomains : int
+        The number of subdomains is unity for 'material', 'cell' and 'universe'
+        domain types. When the  This is equal to the number of cell instances
```
PullRequestOpen: Typo: Did you mean "This is equal" rather than "When the  This is equal"?

openmc/mgxs/mdgxs.py
```diff
@@ -0,0 +1,2619 @@
...
+        return mdgxs
+
+    def get_xs(self, groups='all', subdomains='all', nuclides='all',
+               xs_type='macro', order_groups='increasing',
+               value='mean', delayed_groups='all', squeeze=True, **kwargs):
+        """Returns an array of multi-delayed-group cross sections.
+
+        This method constructs a 4D NumPy array for the requested
+        multi-delayed-group cross section data for one or more
+        subdomains (1st dimension), delayed groups (2nd demension),
```
PullRequestOpen: Thank you for clearly laying out the dimensionality of the arrays.  Super helpful for review!

openmc/mgxs/groups.py
```diff
@@ -0,0 +1,297 @@
...
+        if not isinstance(other, EnergyGroups):
+            return False
+        elif self.num_groups != other.num_groups:
+            return False
+        elif np.allclose(self.group_edges, other.group_edges):
+            return True
+        else:
+            return False
+
+    def __ne__(self, other):
```
PullRequestOpen: Is there a reason to override this method? From https://docs.python.org/3/reference/datamodel.html#object.__ge__:

By default, `__ne__()` delegates to `__eq__()` and inverts the result unless it is `NotImplemented`.

openmc/mgxs/library.py
```diff
@@ -0,0 +1,1469 @@
...
+
+        # Make directory if it does not exist
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+
+        full_filename = os.path.join(directory, filename + '.pkl')
+        full_filename = full_filename.replace(' ', '-')
+
+        # Load and return pickled Library object
+        pickle.dump(self, open(full_filename, 'wb'))
```
PullRequestOpen: This file is being opened and never closed.

Something like this should be better:


```
with open(full_filename, 'wb') as f:
    pickle.dump(self, f)
```

openmc/mgxs/library.py
```diff
@@ -0,0 +1,1469 @@
...
+        """
+
+        cv.check_type('filename', filename, str)
+        cv.check_type('directory', directory, str)
+
+        # Make directory if it does not exist
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+
+        full_filename = os.path.join(directory, filename + '.pkl')
```
PullRequestOpen: Not a huge deal but a lot of this code base uses string formatting and this is using string concat.

`filename + '.pkl'` vs `'{}.pkl'.format(filename)`

openmc/mgxs/library.py
```diff
@@ -0,0 +1,1469 @@
...
+        full_filename = full_filename.replace(' ', '-')
+
+        # Load and return pickled Library object
+        return pickle.load(open(full_filename, 'rb'))
```
PullRequestOpen: We should close this file handle as well.

openmc/lib/cell.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def fill(self):
+        fill_type = c_int()
+        indices = POINTER(c_int32)()
+        n = c_int32()
+        _dll.openmc_cell_get_fill(self._index, fill_type, indices, n)
+
+        if fill_type.value == 1:
+            if n.value > 1:
```
PullRequestOpen: To confirm, `n.value` should never equal 0?

openmc/model/funcs.py
```diff
@@ -0,0 +1,578 @@
...
+    -------
+    openmc.Material
+
+    """
+    # Perform any necessary unit conversions.
+    check_value('temperature unit', temp_unit, ('K', 'C', 'F'))
+    if temp_unit == 'K':
+        T = temperature
+    elif temp_unit == 'C':
+        T = temperature + 273.15
```
PullRequestOpen: This would be a great place for a constant. Boiling point of water in Kelvin and any other reference points in temperatures would make sense as a constant.

openmc/model/funcs.py
```diff
@@ -0,0 +1,578 @@
...
+        :class:`openmc.Universe`, like ``name="Fuel pin"``
+
+    Returns
+    -------
+    :class:`openmc.Universe`
+        Universe of concentric cylinders filled with the desired
+        items
+    """
+    if "cells" in kwargs:
+        raise SyntaxError(
```
PullRequestOpen: I would change this to be a `ValueError` as it's not really a syntax error and `SyntaxError`s are typically reserved for the Python interpreter.

openmc/model/triso.py
```diff
@@ -0,0 +1,1335 @@
...
+from random import uniform, gauss
+
+import numpy as np
+import scipy.spatial
+
+import openmc
+from openmc.checkvalue import check_type
+
+
+MAX_PF_RSP = 0.38
```
PullRequestOpen: Good use of constants!

openmc/model/triso.py
```diff
@@ -0,0 +1,1335 @@
...
+        except ValueError:
+            pass
+
+    if not domain:
+        raise ValueError('Could not map region {} to a container: supported '
+                         'container shapes are rectangular prism, cylinder, '
+                         'sphere, and spherical shell.'.format(region))
+
+    # Determine the packing fraction/number of spheres
+    volume = 4/3*pi*radius**3
```
PullRequestOpen: Might be nice to have the sphere volume computation as a function as it's used quite a bit.

openmc/stats/multivariate.py
```diff
@@ -0,0 +1,543 @@
...
+    y : openmc.stats.Univariate
+        Distribution of y-coordinates
+    z : openmc.stats.Univariate
+        Distribution of z-coordinates
+
+    """
+
+
+    def __init__(self, x, y, z):
+        super().__init__()
```
PullRequestOpen: Spatial's \_\_init\_\_ function doesn't do anything, so this line doesn't add anything to the method. Can be removed and ditto throughout.

openmc/stats/multivariate.py
```diff
@@ -0,0 +1,543 @@
...
+        ----------
+        elem : xml.etree.ElementTree.Element
+            XML element
+
+        Returns
+        -------
+        openmc.stats.Isotropic
+            Isotropic distribution generated from XML element
+
+        """
```
PullRequestOpen: As it stands, a user can do a nonsensical thing like `iso = Isotropic(elem = "potato")`, and get a valid Isotropic object (since the `elem` isn't used in the method). They could also pass in a PolarAzimuthal "mu-phi" type elem and still get an isotropic. Though you may anticipate the superclass' version of this method will be used most, this method should still check its inputs. Here's a a couple lines to fix that.

```
distribution = get_text(elem, 'type')
if distribution != "isotropic":
    raise Exception("Expected an XML element of type isotropic. Instead got {}".format(distribution))
```
To avoid unforseen/confusing errors (or worse, silently created erroneous objects), something like this can be added to all the `from_xml_element`methods in these two files. If you adopt this approach, then save some typing with a custom Exception class. Something like...

```
class UnexpectedElementError(Exception):
    """
    Raised when trying to create an object with an xml element type that doesn't match the type of the object.
    """
    def __init__(self, expected_type, unexpected_type):
        Exception.__init__(self, "Expected XML element of type {}. Instead got one of type {}".format(expected_type, unexpected_type))
```

openmc/stats/multivariate.py
```diff
@@ -0,0 +1,543 @@
...
+    Attributes
+    ----------
+    mu : openmc.stats.Univariate
+        Distribution of the cosine of the polar angle
+    phi : openmc.stats.Univariate
+        Distribution of the azimuthal angle in radians
+
+    """
+
+    def __init__(self, mu=None, phi=None, reference_uvw=[0., 0., 1.]):
```
PullRequestOpen: Note the danger of using mutable default arguments. It doesn't look like you'll run into trouble as it stands, but check out this [explanation](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) and consider using a stand in value for the default value of `reference_uvw` (or use a tuple as you did in Spatial methods).

openmc/stats/multivariate.py
```diff
@@ -0,0 +1,543 @@
+from abc import ABCMeta, abstractmethod
```
PullRequestOpen: As of python 3.4 you can use `from abc import ABC` and create abstract classes like `class UnitSphere(ABC)`. Just a touch simpler.

openmc/stats/univariate.py
```diff
@@ -0,0 +1,827 @@
...
+        ----------
+        elem : xml.etree.ElementTree.Element
+            XML element
+
+        Returns
+        -------
+        openmc.stats.Discrete
+            Discrete distribution generated from XML element
+
+        """
```
PullRequestOpen: Same suggestion in this file as I had in `multivariate.py` for the `from_xml_element` methods. They'd benefit from Error handling for bad xmls.

openmc/stats/univariate.py
```diff
@@ -0,0 +1,827 @@
...
+    ----------
+    x : Iterable of float
+        Values of the random variable
+    p : Iterable of float
+        Discrete probability for each value
+
+    """
+
+    def __init__(self, x, p):
+        super().__init__()
```
PullRequestOpen: Same as in `multivariate.py`. The superclass' \_\_init\_\_ is just a pass, so there's no need for this line and others like it in this file.

openmc/trigger.py
```diff
@@ -0,0 +1,98 @@
...
+
+    """
+
+    def __init__(self, trigger_type, threshold):
+        self.trigger_type = trigger_type
+        self.threshold = threshold
+        self._scores = []
+
+    def __eq__(self, other):
+        return str(self) == str(other)
```
PullRequestOpen: Generally not a good idea to implement `__eq__` like this. The default `__eq__` method will at least check the properties on a deeper level and not just cast them to strings.

openmc/universe.py
```diff
@@ -0,0 +1,571 @@
...
+
+        """
+
+        if not isinstance(cell, openmc.Cell):
+            msg = 'Unable to remove a Cell from Universe ID="{0}" since "{1}" is ' \
+                  'not a Cell'.format(self._id, cell)
+            raise TypeError(msg)
+
+        # If the Cell is in the Universe's list of Cells, delete it
+        if cell.id in self._cells:
```
PullRequestOpen: These two lines can be replaced with just:

```
self._cells.pop(cell.id, None)
```

openmc/universe.py
```diff
@@ -0,0 +1,571 @@
...
+
+        """
+
+        if not isinstance(cells, Iterable):
+            msg = 'Unable to add Cells to Universe ID="{0}" since "{1}" is not ' \
+                  'iterable'.format(self._id, cells)
+            raise TypeError(msg)
+
+        for cell in cells:
+            self.add_cell(cell)
```
PullRequestOpen: +1 Good job reusing your own code

openmc/volume.py
```diff
@@ -0,0 +1,280 @@
...
+        elif isinstance(domains[0], openmc.Material):
+            self._domain_type = 'material'
+        elif isinstance(domains[0], openmc.Universe):
+            self._domain_type = 'universe'
+        self.ids = [d.id for d in domains]
+
+        self.samples = samples
+
+        if lower_left is not None:
+            if upper_right is None:
```
PullRequestOpen: Should this block be moved to the `else` clause below? The exception says that both coordinates should be specified, and this check will trigger if `lower_left` is specified but `lower_right` is not, which is actually a valid code path according to this class's docstring.

openmc/statepoint.py
```diff
@@ -0,0 +1,689 @@
...
+
+                if not contains_filters:
+                    continue
+
+            # Determine if Tally has the queried Nuclide(s)
+            if nuclides:
+                contains_nuclides = True
+
+                # Iterate over the Nuclides requested by the user
+                for nuclide in nuclides:
```
PullRequestOpen: likewise, can be simply:
```
if !all(nuclide in test_tally.nuclides for nuclide in nuclides):
  continue
```

openmc/statepoint.py
```diff
@@ -0,0 +1,689 @@
...
+                continue
+            if exact_filters and len(filters) != test_tally.num_filters:
+                continue
+
+            # Determine if Tally has the queried score(s)
+            if scores:
+                contains_scores = True
+
+                # Iterate over the scores requested by the user
+                for score in scores:
```
PullRequestOpen: can be simply:
```
if !all(score in test_tally.scores for score in scores):
  continue
```

openmc/summary.py
```diff
@@ -0,0 +1,229 @@
...
+        self._fast_lattices = {}
+
+        self._materials = openmc.Materials()
+        self._nuclides = {}
+        self._macroscopics = []
+
+        self._read_nuclides()
+        self._read_macroscopics()
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", openmc.IDWarning)
```
PullRequestOpen: Why is this ignore done in init? Would recommend moving this inside `_read_geometry` if it's only specific to that method

openmc/summary.py
```diff
@@ -0,0 +1,229 @@
...
+            names = self._f['nuclides/names'][()]
+            awrs = self._f['nuclides/awrs'][()]
+            for name, awr in zip(names, awrs):
+                self._nuclides[name.decode()] = awr
+
+    def _read_macroscopics(self):
+        if 'macroscopics/names' in self._f:
+            names = self._f['macroscopics/names'][()]
+            for name in names:
+                self._macroscopics = name.decode()
```
PullRequestOpen: this just assigns value to `_macroscopics` multiple times? Are you missing a break here?

openmc/summary.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def _finalize_geometry(self, cell_fills):
+
+        # Keep track of universes that are used as fills. That way, we can
+        # determine which universe is NOT used as a fill (and hence is the root
+        # universe)
+        fill_univ_ids = set()
+
+        # Iterate over all Cells and add fill Materials, Universes and Lattices
+        for cell_id, (fill_type, fill_id) in cell_fills.items():
```
PullRequestOpen: The cell_fills you have defined in `_read_cells` method sets value as `(fill_type, fill)`, whereas you read those as `(fill_type, fill_id)`. If they represent same objects, you should use same representation.

openmc/tally_derivative.py
```diff
@@ -0,0 +1,113 @@
...
+        string += '{: <16}=\t{}\n'.format('\tID', self.id)
+        string += '{: <16}=\t{}\n'.format('\tVariable', self.variable)
+
+        if self.variable == 'density':
+            string += '{: <16}=\t{}\n'.format('\tMaterial', self.material)
+        elif self.variable == 'nuclide_density':
+            string += '{: <16}=\t{}\n'.format('\tMaterial', self.material)
+            string += '{: <16}=\t{}\n'.format('\tNuclide', self.nuclide)
+        elif self.variable == 'temperature':
+            string += '{: <16}=\t{}\n'.format('\tMaterial', self.material)
```
PullRequestOpen: you are setting same value in each if condition, it can be simplified by:
```
if self.variable in ['density', 'nuclide_density', 'temperature']:
  string += '{: <16}=\t{}\n'.format('\tMaterial', self.material)
if self.variable == 'nuclide_density':
  string += '{: <16}=\t{}\n'.format('\tNuclide', self.nuclide)
```

openmc/plots.py
```diff
@@ -0,0 +1,850 @@
...
+        return string
+
+    @classmethod
+    def from_geometry(cls, geometry, basis='xy', slice_coord=0.):
+        """Return plot that encompasses a geometry.
+
+        Parameters
+        ----------
+        geometry : openmc.Geometry
+            The geometry the base the plot off of
```
PullRequestOpen: Spelling: I believe you may have meant "The geometry **to** base the plot off of"

openmc/plots.py
```diff
@@ -0,0 +1,850 @@
...
+    def level(self, plot_level):
+        cv.check_type('plot level', plot_level, Integral)
+        cv.check_greater_than('plot level', plot_level, 0, equality=True)
+        self._level = plot_level
+
+    @meshlines.setter
+    def meshlines(self, meshlines):
+        cv.check_type('plot meshlines', meshlines, dict)
+        if 'type' not in meshlines:
+            msg = 'Unable to set on plot the meshlines "{0}" which ' \
```
PullRequestOpen: Question / Possible typo: Did you mean to say "set on plot the meshlines here"? Or did you mean "set the meshlines", "plot the meshlines"?

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+        mgxs = _calculate_mgxs_nuc_macro(this, types, library, orders,
+                                         temperature)
+    elif data_type in ('element', 'material'):
+        mgxs = _calculate_mgxs_elem_mat(this, types, library, orders,
+                                        temperature, ce_cross_sections,
+                                        enrichment)
+    else:
+        raise TypeError("Invalid type")
+
+    # Convert the data to the format needed
```
PullRequestOpen: Optional: Would be more readable to add some color to the final form of `energy_grid` rather than indicating that there is a needed formatting change.

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+
+import openmc.checkvalue as cv
+import openmc.data
+
+# Supported keywords for continuous-energy cross section plotting
+PLOT_TYPES = ['total', 'scatter', 'elastic', 'inelastic', 'fission',
+              'absorption', 'capture', 'nu-fission', 'nu-scatter', 'unity',
+              'slowing-down power', 'damage']
+
+# Supported keywoards for multi-group cross section plotting
```
PullRequestOpen: Spelling: `keywoards` -> `keywords`

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+XI_MT = -2
+
+# MTs to combine to generate associated plot_types
+_INELASTIC = [mt for mt in openmc.data.SUM_RULES[3] if mt != 27]
+PLOT_TYPES_MT = {'total': openmc.data.SUM_RULES[1],
+                 'scatter': [2] + _INELASTIC,
+                 'elastic': [2],
+                 'inelastic': _INELASTIC,
+                 'fission': [18],
+                 'absorption': [27], 'capture': [101],
```
PullRequestOpen: Nit: Would be more readable to keep map construction consistent here by keeping one element per line (by moving `'capture': [101]` to the next line).

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+
+    if data_type == 'nuclide':
+        if isinstance(this, str):
+            nuc = openmc.Nuclide(this)
+        else:
+            nuc = this
+        energy_grid, xs = _calculate_cexs_nuclide(nuc, types, temperature,
+                                                  sab_name, cross_sections)
+        # Convert xs (Iterable of Callable) to a grid of cross section values
+        # calculated on @ the points in energy_grid for consistency with the
```
PullRequestOpen: Question: Did you mean to say "on @ the points" here, or was this meant to be "@ the points" or "on the points"?

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+                 'unity': [UNITY_MT],
+                 'slowing-down power': [2] + _INELASTIC + [XI_MT],
+                 'damage': [444]}
+# Operations to use when combining MTs the first np.add is used in reference
+# to zero
+PLOT_TYPES_OP = {'total': (np.add,),
+                 'scatter': (np.add,) * (len(PLOT_TYPES_MT['scatter']) - 1),
+                 'elastic': (),
+                 'inelastic': (np.add,) * (len(PLOT_TYPES_MT['inelastic']) - 1),
+                 'fission': (), 'absorption': (),
```
PullRequestOpen: Nit: Same as on line 37, be more readable to keep map construction consistent here by keeping one element per line.

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+
+    If the data for the nuclide or macroscopic object in the library is
+    represented as angle-dependent data then this method will return the
+    geometric average cross section over all angles.
+
+    Parameters
+    ----------
+    this : str or openmc.Material
+        Object to source data from
+    data_type : {'nuclide', 'element', material', 'macroscopic'}
```
PullRequestOpen: Forgot a `'` here before `material` (`material'` -> `'material'`)

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+            # Perform the division
+            for line in range(len(types)):
+                data[line, :] /= data_div[line, :]
+                if divisor_types[line] != 'unity':
+                    types[line] += ' / ' + divisor_types[line]
+
+    # Generate the plot
+    if axis is None:
+        fig = plt.figure(**kwargs)
+        ax = fig.add_subplot(111)
```
PullRequestOpen: Question: Any significance to `111` here?  Or is it a preferred scaling factor or something related to the resulting plotting visual?

openmc/plotter.py
```diff
@@ -0,0 +1,910 @@
...
+_PLOT_MGXS_ATTR = {line: line.replace(' ', '_').replace('-', '_')
+                   for line in PLOT_TYPES_MGXS}
+_PLOT_MGXS_ATTR['scatter'] = 'scatter_matrix'
+
+# Special MT values
+UNITY_MT = -1
+XI_MT = -2
+
+# MTs to combine to generate associated plot_types
+_INELASTIC = [mt for mt in openmc.data.SUM_RULES[3] if mt != 27]
```
PullRequestOpen: Optional: Clarifying comment about importance of `27` would increase readability here (is absorption special)?

openmc/openmoc_compatible.py
```diff
@@ -0,0 +1,825 @@
...
+    else:
+        boundary = openmoc.BOUNDARY_NONE
+
+    if openmc_surface.type == 'plane':
+        A = openmc_surface.a
+        B = openmc_surface.b
+        C = openmc_surface.c
+        D = openmc_surface.d
+
+        # OpenMOC uses the opposite sign on D
```
PullRequestOpen: Good comment

openmc/particle_restart.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def __init__(self, filename):
+        self._f = h5py.File(filename, 'r')
+
+        # Ensure filetype and version are correct
+        cv.check_filetype_version(self._f, 'particle restart',
+                                  _VERSION_PARTICLE_RESTART)
+
+    @property
+    def current_batch(self):
```
PullRequestOpen: Is it OK if any of these don't exist in `self._f`? You might want to use `.get()` instead to supply a default case.

openmc/particle_restart.py
```diff
@@ -0,0 +1,92 @@
...
+        Energy of the particle in eV
+    xyz : list of float
+        Position of the particle
+    uvw : list of float
+        Directional cosines of the particle
+
+    """
+
+    def __init__(self, filename):
+        self._f = h5py.File(filename, 'r')
```
PullRequestOpen: Does the `h5py` library handle closing the file? If not, you will need to explicitly close your file handle.

openmc/particle_restart.py
```diff
@@ -0,0 +1,92 @@
...
+    def type(self):
+        return self._f['type'][()]
+
+    @property
+    def n_particles(self):
+        return self._f['n_particles'][()]
+
+    @property
+    def run_mode(self):
+        return self._f['run_mode'][()].decode()
```
PullRequestOpen: Why `decode()`? Why only for this property?

openmc/mesh.py
```diff
@@ -0,0 +1,560 @@
...
+            this cell is a single parallelepiped with boundaries matching
+            the outermost mesh boundary with the boundary conditions from bc
+            applied.
+        cells : iterable of openmc.Cell
+            The list of cells within each lattice position mimicking the mesh
+            geometry.
+
+        """
+
+        cv.check_length('bc', bc, length_min=4, length_max=6)
```
PullRequestOpen: This seems to be the incorrect check. Since only 4 or 6 are valid values, cv.check_length needs to ensure it is 4 or 6, and 5 should fail.

openmc/mesh.py
```diff
@@ -0,0 +1,560 @@
...
+        if upper_right is not None:
+            mesh.upper_right = [float(x) for x in upper_right.split()]
+
+        width = get_text(elem, 'width')
+        if width is not None:
+            mesh.width = [float(x) for x in width.split()]
+
+        return mesh
+
+    def build_cells(self, bc=['reflective'] * 6):
```
PullRequestOpen: The default argument for `bc` is mutable, which is a common pitfall and anti-pattern. I have provided a more detailed explanation in another PR for this issue. [Ref](https://stackoverflow.com/questions/41686829/warning-about-mutable-default-argument-in-pycharm)

Solution is the default should be `None` and in the code check

```python
if bc == None:
    bc = ['reflective'] * 6
```

openmc/mesh.py
```diff
@@ -0,0 +1,560 @@
...
+                      self.dimension[2])
+                lattice.pitch = [dx, dy, dz]
+
+        # Fill Cell with the Lattice
+        root_cell.fill = lattice
+
+        return root_cell, cells
+
+
+def Mesh(*args, **kwargs):
```
PullRequestOpen: This can also be marked as `@Deprecated` with a decorator.
Reference: https://deprecation.readthedocs.io/en/latest/

PR #1429, opened 2019-12-06
Review by PullRequest, Inc. - Part 4

PullRequestOpen:


openmc/deplete/results.py
```diff
@@ -0,0 +1,496 @@
...
+        ----------
+        pos : tuple
+            A three-length tuple containing a stage index, mat index and a nuc
+            index.  All can be integers or slices.  The second two can be
+            strings corresponding to their respective dictionary.
+
+        val : float
+            The value to set data to.
+
+        """
```
PullRequestOpen: We should check the type here with `cv.check_type` since, without it, we could just add anything to our data.

openmc/deplete/results.py
```diff
@@ -0,0 +1,496 @@
...
+            A three-length tuple containing a stage index, mat index and a nuc
+            index.  All can be integers or slices.  The second two can be
+            strings corresponding to their respective dictionary.
+
+        Returns
+        -------
+        float
+            The atoms for stage, mat, nuc
+
+        """
```
PullRequestOpen: We should check the type here

openmc/deplete/results.py
```diff
@@ -0,0 +1,496 @@
...
+            in :attr:`rates`
+
+        Returns
+        -------
+        Results
+            New results object
+        """
+        new = Results()
+        new.volume = {lm: self.volume[lm] for lm in local_materials}
+        new.mat_to_ind = dict(zip(
```
PullRequestOpen: so this is just enumerating, and flipping the indexes. I think this would be much clearer:
`{mat: idx for (idx, mat) in enumerate(local_materials)]}`

This shows that you are 1. enumerating the local_materials 2. constructing a dictionary out of it 
in one line.

openmc/deplete/results.py
```diff
@@ -0,0 +1,496 @@
...
+import copy
+from warnings import warn
+
+import numpy as np
+import h5py
+
+from . import comm, have_mpi, MPI
+from .reaction_rates import ReactionRates
+
+_VERSION_RESULTS = (1, 0)
```
PullRequestOpen: If this is exported for `ResultsList` we shouldnt keep it as a private constant. `VERSION_RESULTS` instead of `_VERSION_RESULTS` would be better

openmc/deplete/results_list.py
```diff
@@ -0,0 +1,167 @@
...
+        Parameters
+        ----------
+        mat : str
+            Material name to evaluate
+        nuc : str
+            Nuclide name to evaluate
+
+        Returns
+        -------
+        time : numpy.ndarray
```
PullRequestOpen: since this is an array of times, instead of just one, maybe `result_times` or atleast `times`

openmc/deplete/results_list.py
```diff
@@ -0,0 +1,167 @@
...
+            Array of times in [s]
+        concentration : numpy.ndarray
+            Total number of atoms for specified nuclide
+
+        """
+        time = np.empty_like(self, dtype=float)
```
PullRequestOpen: `time` is the name of a standard Python library module: https://docs.python.org/3/library/time.html Consider using a different variable name. Also, this value seems to be used like an array, so maybe `times` or `result_times` are appropriate.

openmc/deplete/nuclide.py
```diff
@@ -0,0 +1,570 @@
...
+
+        """
+        elem = ET.Element('nuclide')
+        elem.set('name', self.name)
+
+        if self.half_life is not None:
+            elem.set('half_life', str(self.half_life))
+            elem.set('decay_modes', str(len(self.decay_modes)))
+            elem.set('decay_energy', str(self.decay_energy))
+            for mode, daughter, br in self.decay_modes:
```
PullRequestOpen: Nit: If `mode, daughter, br` is deconstructing a `DecayTuple` type, I believe that would mean that `mode` maps to `DecayTuple.type`.  If this is the case (and intended behavior), I would prefer renaming `mode` to `mode_type` for clarity.  This would also help on line 228.

Note that I definitely do not recommend renaming `mode` to `type`, as `type` is a overloaded term in python.

openmc/deplete/nuclide.py
```diff
@@ -0,0 +1,570 @@
...
+                    0.0 if yield_val is None else yield_val)
+        self.energies = tuple(energies)
+        self.products = tuple(ordered_prod)
+        self.yield_matrix = yield_matrix
+
+    def __len__(self):
+        return len(self.energies)
+
+    def __getitem__(self, energy):
+        if energy not in self.energies:
```
PullRequestOpen: For completeness, you may want to check that the appropriate element exists in `yield_matrix` as well as `energies`.

openmc/deplete/reaction_rates.py
```diff
@@ -0,0 +1,153 @@
...
+    Parameters
+    ----------
+    local_mats : list of str
+        Material IDs
+    nuclides : list of str
+        Depletable nuclides
+    reactions : list of str
+        Transmutation reactions being tracked
+    from_results : boolean
+        If the reaction rates are loaded from results, indexing dictionnaries
```
PullRequestOpen: Nit / Spelling: `dictionnaries` -> `dictionaries`

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+    @classmethod
+    def from_endf(cls, decay_files, fpy_files, neutron_files):
+        """Create a depletion chain from ENDF files.
+
+        Parameters
+        ----------
+        decay_files : list of str
+            List of ENDF decay sub-library files
+        fpy_files : list of str
+            List of ENDF neutron-induced fission product yield sub-library files
```
PullRequestOpen: Thank you for clearly explaining fpy in the comments.  Helps a lot for the new reader.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+            if parent in fpy_data:
+                fpy = fpy_data[parent]
+
+                if fpy.energies is not None:
+                    yield_energies = fpy.energies
+                else:
+                    yield_energies = [0.0]
+
+                yield_data = {}
+                for E, table in zip(yield_energies, fpy.independent):
```
PullRequestOpen: Optional: I would use a different identifier than `table`.  If it is the table of fission products (which I believe it to be by my read below), I would use `fission_products` or `fission_table`.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+        return 'H1'
+
+    # First check if ground state is available
+    if state:
+        product = '{}{}'.format(symbol, A)
+
+    # Find isotope with longest half-life
+    half_life = 0.0
+    for nuclide, data in decay_data.items():
+        m = re.match(r'{}(\d+)(?:_m\d+)?'.format(symbol), nuclide)
```
PullRequestOpen: Optional: If possible, I would anchor the regular expression with "Start of String" (`^`) and End of String (`$`) Anchors.  These tend to produce more robust matches and handle dirty inputs a bit more cleanly.  Note that of course this depends on the inputs into `decay_data`.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+            if len(indexes) == 0:
+                if strict:
+                    raise AttributeError(
+                        "Nuclide {} does not have {} reactions".format(
+                            parent, reaction))
+                missing_reaction.add(parent)
+                continue
+
+            this_sum = sum(sub.values())
+            # sum of branching ratios can be lower than 1 if no ground
```
PullRequestOpen: Thank you for the explanation of the +1 offset.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+            if beta_minus:
+                Z += 1
+            else:
+                Z -= 1
+        product = '{}{}'.format(openmc.data.ATOMIC_SYMBOL[Z], A)
+
+    return product
+
+
+_SECONDARY_PARTICLES = {
```
PullRequestOpen: Optional: Might be more readable if each key/value pair has it's own line.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+                        br = 1.0 - sum(m.branching_ratio.nominal_value
+                                       for m in data.modes[:-1])
+
+                    # Append decay mode
+                    nuclide.decay_modes.append(DecayTuple(type_, target, br))
+
+            if parent in reactions:
+                reactions_available = set(reactions[parent].keys())
+                for name, mts, changes in _REACTIONS:
+                    if mts & reactions_available:
```
PullRequestOpen: Good use of the set intersection operator.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+            rxn_Q = parent.reactions[rxn_index[0]].Q
+
+            # Remove existing reactions
+
+            for ix in reversed(rxn_index):
+                parent.reactions.pop(ix)
+
+            all_meta = True
+
+            for tgt, br in new_ratios.items():
```
PullRequestOpen: Optional: Would clarifying meaning of `tgt` in a comment.

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+                all_meta = all_meta and ("_m" in tgt)
+                parent.reactions.append(ReactionTuple(
+                    reaction, tgt, rxn_Q, br))
+
+            if all_meta and sums[parent_name] != 1.0:
```
PullRequestOpen: Optional: I would add a comment explaining the expected behavior of this block.  By my read, it has to do with expected appended reactions related to when / if unity happens for a sum, however it's not 100% clear from context.  Thanks!

openmc/deplete/chain.py
```diff
@@ -0,0 +1,786 @@
...
+                mass_longest_lived = int(m.group(1))
+                half_life = data.half_life.nominal_value
+
+    # If mass number of longest-lived isotope is less than that of missing
+    # product, assume it undergoes beta-. Otherwise assume beta+.
+    beta_minus = (mass_longest_lived < A)
+
+    # Iterate until we find an existing nuclide
+    while product not in decay_data:
+        if Z > 98:
```
PullRequestOpen: Optional: I would define `98` here as a class / file level constant, or leave a comment here explaining it's significance.

openmc/data/product.py
```diff
@@ -0,0 +1,187 @@
...
+    def to_hdf5(self, group):
+        """Write product to an HDF5 group
+
+        Parameters
+        ----------
+        group : h5py.Group
+            HDF5 group to write to
+
+        """
+        group.attrs['particle'] = np.string_(self.particle)
```
PullRequestOpen: Generally, it can be useful to store these strings in constants rather than use them directly, so that any typos/misspellings would be within the constant name (assuming that the single line where the constant is set is correct) and cause a more obvious failure. eg.:
```python
PARTICLE = 'particle'

...

group.attrs[PARTICLE] = ...
```

openmc/data/product.py
```diff
@@ -0,0 +1,187 @@
...
+from .function import Tabulated1D, Polynomial, Function1D
+
+
+class Product(EqualityMixin):
+    """Secondary particle emitted in a nuclear reaction
+
+    Parameters
+    ----------
+    particle : str, optional
+        What particle the reaction product is. Defaults to 'neutron'.
```
PullRequestOpen: nit: May read better as: `The particle type of the reaction product. Defaults to 'neutron'.`

openmc/data/product.py
```diff
@@ -0,0 +1,187 @@
...
+        yield represents particles from prompt and delayed sources.
+    particle : str
+        What particle the reaction product is.
+    yield_ : openmc.data.Function1D
+        Yield of secondary particle in the reaction.
+
+    """
+
+    def __init__(self, particle='neutron'):
+        self.particle = particle
```
PullRequestOpen: Minor nit: It can be a little easier to read when the order of these match the comment above. Other places in the file, such as the ordering of properties and setters match the comment. Though I'd agree that particle could come first here as it is a constructor parameter.

openmc/data/product.py
```diff
@@ -0,0 +1,187 @@
...
+        self._emission_mode = emission_mode
+
+    @particle.setter
+    def particle(self, particle):
+        cv.check_type('product particle type', particle, str)
+        self._particle = particle
+
+    @yield_.setter
+    def yield_(self, yield_):
+        cv.check_type('product yield', yield_, Function1D)
```
PullRequestOpen: I suspect that I may be misunderstanding this, but the __repr__ method tests for a few different types of yield_. Does this setter accept a smaller subset of those types?

openmc/data/reaction.py
```diff
@@ -0,0 +1,1189 @@
...
+
+        neutron = Product('neutron')
+        neutron.emission_mode = whichnu
+
+        idx = ace.jxs[2]
+        LNU = int(ace.xss[idx])
+        if LNU == 1:
+            # Polynomial function form of nu
+            NC = int(ace.xss[idx+1])
+            coefficients = ace.xss[idx+2 : idx+2+NC].copy()
```
PullRequestOpen: Why copy here? Setting and accessing list items won't modify the backing list since slicing a list like that creates a new list itself.

```
>>> a = [1,2,3,4,5,6]
>>> b = a[2:4]
>>> b
[3, 4]
>>> b[0] = 100  # Edit this
>>> a
[1, 2, 3, 4, 5, 6]  # `a` is unchanged
>>> 


```

openmc/data/reaction.py
```diff
@@ -0,0 +1,1189 @@
...
+                mass_ratio = (A/(A + 1.))**2
+                dist.energy = LevelInelastic(threshold, mass_ratio)
+
+                neutron.distribution.append(dist)
+
+            if (4, mt) in ev.section:
+                for dist in neutron.distribution:
+                    dist.angle = AngleDistribution.from_endf(ev, mt)
+
+            if mt in (18, 19, 20, 21, 38) and (5, mt) in ev.section:
```
PullRequestOpen: This goes for multiple instances in this PR... it might be a good idea to group things into global constants instead of checking against a tuple of ints.

For example:

```
if mt in (18, 19, 20, 21, 38)
```

vs

```
FISSION_REACTION_TYPES = (18, 19, 20, 21, 38)
if mt in FISSION_REACTION_TYPES
```

This allows for better readability as well as maintainability (in the case that you want to add a new type to this).

openmc/data/photon.py
```diff
@@ -0,0 +1,1306 @@
...
+        return data
+
+    def export_to_hdf5(self, path, mode='a', libver='earliest'):
+        """Export incident photon data to an HDF5 file.
+
+        Parameters
+        ----------
+        path : str
+            Path to write HDF5 file to
+        mode : {'r', r+', 'w', 'x', 'a'}
```
PullRequestOpen: Missing an opening quote here

openmc/data/photon.py
```diff
@@ -0,0 +1,1306 @@
...
+        Returns
+        -------
+        openmc.data.IncidentPhoton
+            Photon interaction data
+
+        """
+        if isinstance(group_or_filename, h5py.Group):
+            group = group_or_filename
+        else:
+            h5file = h5py.File(str(group_or_filename), 'r')
```
PullRequestOpen: Need to close this file after you finish reading from it

openmc/data/photon.py
```diff
@@ -0,0 +1,1306 @@
...
+        if 'bremsstrahlung' in group:
+            rgroup = group['bremsstrahlung']
+            data.bremsstrahlung['I'] = rgroup.attrs['I']
+            for key in ('dcs', 'electron_energy', 'ionization_energy',
+                        'num_electrons', 'photon_energy'):
+                data.bremsstrahlung[key] = rgroup[key][()]
+
+        return data
+
+    def export_to_hdf5(self, path, mode='a', libver='earliest'):
```
PullRequestOpen: I've seen this method a few times in this code base, maybe it should be put into a module for better re-use. Also the `mode` should probably not be allowed to be specified by the caller here as you're writing to a file, it wouldn't make sense to open the file in `r` mode.

openmc/data/photon.py
```diff
@@ -0,0 +1,1306 @@
...
+                for i in range(1, 101):
+                    group = f['{:03}'.format(i)]
+                    _BREMSSTRAHLUNG[i] = {
+                        'I': group.attrs['I'],
+                        'num_electrons': group['num_electrons'][()],
+                        'ionization_energy': group['ionization_energy'][()]
+                    }
+
+            filename = os.path.join(os.path.dirname(__file__), 'BREMX.DAT')
+            brem = open(filename, 'r').read().split()
```
PullRequestOpen: This file needs to be closed after reading:

```python
with open(filename) as file_obj:
    brem = file_obj.read().split()
```

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+                                                  e_mid[1:]):
+                    e_left = e - g*_RESONANCE_ENERGY_GRID
+                    energies.append(e_left[e_left > e_lower][::-1])
+                    e_right = e + g*_RESONANCE_ENERGY_GRID[1:]
+                    energies.append(e_right[e_right < e_upper])
+
+                # Concatenate all points
+                energies = np.concatenate(energies)
+
+                # Create 1000 equal log-spaced energies over RRR, combine with
```
PullRequestOpen: Thanks for clearly documenting the importance of 1000.

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+        """
+        if isinstance(ev_or_filename, Evaluation):
+            ev = ev_or_filename
+        else:
+            ev = Evaluation(ev_or_filename)
+
+        atomic_number = ev.target['atomic_number']
+        mass_number = ev.target['mass_number']
+        metastable = ev.target['isomeric_state']
+        atomic_weight_ratio = ev.target['mass']
```
PullRequestOpen: Question: Just want to confirm here that `ev.target['mass']` represents relative atomic mass or another number equivalent to "atomic weight ratio".

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+        """Generate incident neutron continuous-energy data from an ENDF evaluation
+
+        Parameters
+        ----------
+        ev_or_filename : openmc.data.endf.Evaluation or str
+            ENDF evaluation to read from. If given as a string, it is assumed to
+            be the filename for the ENDF file.
+
+        covariance : bool
+            Flag to indicate whether or not covariance data from File 32 should be
```
PullRequestOpen: Optional: I would clarify the significance of `File 32` for readability in this context.

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+            raise NotImplementedError('Cannot export incident neutron data that '
+                                      'originated from an ENDF file.')
+
+        # Open file and write version
+        f = h5py.File(str(path), mode, libver=libver)
+        f.attrs['filetype'] = np.string_('data_neutron')
+        f.attrs['version'] = np.array(HDF5_VERSION)
+
+        # Write basic data
+        g = f.create_group(self.name)
```
PullRequestOpen: Nit: I would use `group` or another identifier rather than `g` here.

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+
+                # Get midpoints between resonances (use min/max energy of
+                # resolved region as absolute lower/upper bound)
+                e_mid = np.concatenate(
+                    ([e_min], (e_peak[1:] + e_peak[:-1])/2, [e_max]))
+
+                # Add grid around each resonance that includes the peak +/- the
+                # width times each value in _RESONANCE_ENERGY_GRID. Values are
+                # constrained so that points around one resonance don't overlap
+                # with points around another. This algorithm is from Fudge.
```
PullRequestOpen: Optional: I would clarify the statement `This algorithm is from Fudge`.  My guess is Fudge is either a paper or a textbook, but it is not entirely clear from context.

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+    reactions : collections.OrderedDict
+        Contains the cross sections, secondary angle and energy distributions,
+        and other associated data for each reaction. The keys are the MT values
+        and the values are Reaction objects.
+    resonances : openmc.data.Resonances or None
+        Resonance parameters
+    resonance_covariance : openmc.data.ResonanceCovariance or None
+        Covariance for resonance parameters
+    temperatures : list of str
+        List of string representations the temperatures of the target nuclide
```
PullRequestOpen: Nit: Did you mean to say "List of string representations **of** the temperatures" rather than "List of string representations the temperatures"?

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+            return [mt] if mt in self else []
+
+    def export_to_hdf5(self, path, mode='a', libver='earliest'):
+        """Export incident neutron data to an HDF5 file.
+
+        Parameters
+        ----------
+        path : str
+            Path to write HDF5 file to
+        mode : {'r', r+', 'w', 'x', 'a'}
```
PullRequestOpen: Spelling / typo: Need a extra `'` to the right of `r+`: `'r+'`

openmc/data/neutron.py
```diff
@@ -0,0 +1,926 @@
...
+        heating_number = ace.xss[i + 4*n_energy : i + 5*n_energy]*EV_PER_MEV
+
+        # Create redundant reactions (total, absorption, and heating)
+        total = Reaction(1)
+        total.xs[strT] = Tabulated1D(energy, total_xs)
+        total.redundant = True
+        data.reactions[1] = total
+
+        if np.count_nonzero(absorption_xs) > 0:
+            absorption = Reaction(101)
```
PullRequestOpen: Optional: I would clarify the significance of `101` and `301` in this circumstance for readability.

openmc/data/multipole.py
```diff
@@ -0,0 +1,535 @@
...
+        if isinstance(group_or_filename, h5py.Group):
+            group = group_or_filename
+        else:
+            h5file = h5py.File(str(group_or_filename), 'r')
+
+            # Make sure version matches
+            if 'version' in h5file.attrs:
+                major, minor = h5file.attrs['version']
+                if major != WMP_VERSION_MAJOR:
+                    raise IOError(
```
PullRequestOpen: I would use a better exception to describe the situation such as a `ModuleNotFoundException` https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError

openmc/data/multipole.py
```diff
@@ -0,0 +1,535 @@
...
+        return fun(E)
+
+    def export_to_hdf5(self, path, mode='a', libver='earliest'):
+        """Export windowed multipole data to an HDF5 file.
+
+        Parameters
+        ----------
+        path : str
+            Path to write HDF5 file to
+        mode : {'r', r+', 'w', 'x', 'a'}
```
PullRequestOpen: This function writes to a file and therefore must provide write access. Would this still be able to write if the caller of this method passed `r` as the mode?

openmc/data/multipole.py
```diff
@@ -0,0 +1,535 @@
...
+            Mode that is used to open the HDF5 file. This is the second argument
+            to the :class:`h5py.File` constructor.
+        libver : {'earliest', 'latest'}
+            Compatibility mode for the HDF5 file. 'latest' will produce files
+            that are less backwards compatible but have performance benefits.
+
+        """
+
+        # Open file and write version.
+        with h5py.File(str(path), mode, libver=libver) as f:
```
PullRequestOpen: This is a better way to interact with Files +1

openmc/data/multipole.py
```diff
@@ -0,0 +1,535 @@
...
+        openmc.data.WindowedMultipole
+            Resonant cross sections represented in the windowed multipole
+            format.
+
+        """
+
+        if isinstance(group_or_filename, h5py.Group):
+            group = group_or_filename
+        else:
+            h5file = h5py.File(str(group_or_filename), 'r')
```
PullRequestOpen: This file needs to be closed when you are finished with it per the docs: http://docs.h5py.org/en/stable/high/file.html#File.close

openmc/data/njoy.py
```diff
@@ -0,0 +1,593 @@
...
+        # Indicate tapes to save for each ACER run
+        tapeout[nace] = fname.format(ace, temperature)
+        tapeout[ndir] = fname.format(xsdir, temperature)
+    commands += 'stop\n'
+    run(commands, tapein, tapeout, **kwargs)
+
+    with open(ace, 'w') as ace_file, open(xsdir, 'w') as xsdir_file:
+        # Concatenate ACE and xsdir files together
+        for temperature in temperatures:
+            text = open(fname.format(ace, temperature), 'r').read()
```
PullRequestOpen: Need to close this file

openmc/data/njoy.py
```diff
@@ -0,0 +1,593 @@
...
+            ace_file.write(text)
+
+            text = open(fname.format(xsdir, temperature), 'r').read()
```
PullRequestOpen: Need to close this file as well

openmc/data/njoy.py
```diff
@@ -0,0 +1,593 @@
...
+                print(line, end='')
+
+        # Check for error
+        if njoy.returncode != 0:
+            raise CalledProcessError(njoy.returncode, njoy_exec,
+                                     ''.join(lines))
+
+        # Copy output files back to original directory
+        for tape_num, filename in tapeout.items():
+            tmpfilename = os.path.join(tmpdir, 'tape{}'.format(tape_num))
```
PullRequestOpen: Python has a builtin library for tempfiles like this if you want to simplify things: https://docs.python.org/3/library/tempfile.html

openmc/data/grid.py
```diff
@@ -0,0 +1,114 @@
+import numpy as np
+
+
+def linearize(x, f, tolerance=0.001):
+    """Return a tabulated representation of a function of one variable.
```
PullRequestOpen: This comment reads slightly awkward to me. How about "Returns a tabulated representation of a one-variable function"

openmc/data/grid.py
```diff
@@ -0,0 +1,114 @@
...
+                i_right = i_left + 1
+                break
+
+        i_right += 1
+
+    for i_remove in range(i_left + 1, i_right - 1):
+        x_out[i_remove] = np.nan
+        y_out[i_remove] = np.nan
+
+    return x_out[np.isfinite(x_out)], y_out[np.isfinite(y_out)]
```
PullRequestOpen: If `x_out` and `y_out` are made a set instead of array, then they can simply be removed from the set, as opposed to setting value to `np.nan` and filtering out in the end. At the end the set can be converted to list.

Alternatively instead of modifying `x_out` directly, create temporary lists for values to be removed and remove them all at once from the output list. 

Both of these will be a cleaner implementation and allow you to remove in-place which not only should be more performant but also potentially get rid of possible bugs due to having illegal values in the list.

openmc/data/grid.py
```diff
@@ -0,0 +1,114 @@
...
+    # Initialize stack
+    x_stack = [x[0]]
+    y_stack = [f(x[0])]
+
+    for i in range(x.shape[0] - 1):
+        x_stack.insert(0, x[i + 1])
+        y_stack.insert(0, f(x[i + 1]))
+
+        while True:
+            x_high, x_low = x_stack[-2:]
```
PullRequestOpen: While it may seem clear to the code author what this function is doing, it'd be great if there's a comment either explaining how the linearize operation is happening, or if a suitable link to the explanation for the calculations is provided.

openmc/data/kalbach_mann.py
```diff
@@ -0,0 +1,402 @@
...
+            # distribution
+            if m == 0:
+                eout_i = eout_continuous
+            elif m == n:
+                eout_i = eout_discrete
+            else:
+                eout_i = Mixture([p_discrete, 1. - p_discrete],
+                                 [eout_discrete, eout_continuous])
+
+            km_r = Tabulated1D(data[0, j:j+n], data[3, j:j+n])
```
PullRequestOpen: It is unclear to me why `km_r` and `km_a` are the same data points except 3 vs 4. Could a comment be added to explain this?

openmc/data/kalbach_mann.py
```diff
@@ -0,0 +1,402 @@
...
+    def precompound(self):
+        return self._precompound
+
+    @property
+    def slope(self):
+        return self._slope
+
+    @breakpoints.setter
+    def breakpoints(self, breakpoints):
+        cv.check_type('Kalbach-Mann breakpoints', breakpoints,
```
PullRequestOpen: Moving `'Kalbach-Mann'` to a variable will allow its reuse and prevent any bugs from typos. Most IDEs will autocomplete a variable.


```
KALBACH_MANN = 'Kalbach-Mann'
cv.check_type(KALBACH_MANN + ' breakpoints', breakpoints, .... )
```


This can be reused in other functions as well

openmc/data/laboratory.py
```diff
@@ -0,0 +1,142 @@
+from collections.abc import Iterable
+from numbers import Real, Integral
+
+import numpy as np
+
+import openmc.checkvalue as cv
+from openmc.stats import Tabular, Univariate, Discrete, Mixture
+from .angle_energy import AngleEnergy
+from .function import INTERPOLATION_SCHEME
```
PullRequestOpen: Unused `INTERPOLATION_SCHEME`

openmc/data/laboratory.py
```diff
@@ -0,0 +1,142 @@
...
+            Laboratory angle-energy distribution
+
+        """
+        params, tab2 = get_tab2_record(file_obj)
+        ne = params[5]
+        energy = np.zeros(ne)
+        mu = []
+        energy_out = []
+        for i in range(ne):
+            params, tab2mu = get_tab2_record(file_obj)
```
PullRequestOpen: `tab2mu` is unused.

Replace with : 

`params, _ = get_tab2_record(file_obj)`

openmc/data/laboratory.py
```diff
@@ -0,0 +1,142 @@
+from collections.abc import Iterable
+from numbers import Real, Integral
+
+import numpy as np
+
+import openmc.checkvalue as cv
+from openmc.stats import Tabular, Univariate, Discrete, Mixture
```
PullRequestOpen: Unused `Mixture`

openmc/data/library.py
```diff
@@ -0,0 +1,173 @@
+import os
+import xml.etree.ElementTree as ET
+import pathlib
+
+import h5py
+
+from openmc.mixin import EqualityMixin
+from openmc._xml import clean_indentation
+from openmc.checkvalue import check_type
```
PullRequestOpen: `check_type` is unused here.

openmc/data/fission_energy.py
```diff
@@ -0,0 +1,380 @@
...
+        if ev.target['isomeric_state'] != incident_neutron.metastable:
+            raise ValueError('The metastable state of the ENDF evaluation '
+                             'does not match the given IncidentNeutron.')
+        if not ev.target['fissionable']:
+            raise ValueError('The ENDF evaluation is not fissionable.')
+
+        if (1, 458) not in ev.section:
+            raise ValueError('ENDF evaluation does not have MF=1, MT=458.')
+
+        file_obj = StringIO(ev.section[1, 458])
```
PullRequestOpen: This file handle should be closed.

openmc/data/energy_distribution.py
```diff
@@ -0,0 +1,1277 @@
...
+
+from .function import Tabulated1D, INTERPOLATION_SCHEME
+from openmc.stats.univariate import Univariate, Tabular, Discrete, Mixture
+import openmc.checkvalue as cv
+from openmc.mixin import EqualityMixin
+from .data import EV_PER_MEV
+from .endf import get_tab1_record, get_tab2_record
+
+
+class EnergyDistribution(EqualityMixin, metaclass=ABCMeta):
```
PullRequestOpen: Well done writing the abstract base class.

openmc/data/endf.c
```diff
@@ -0,0 +1,39 @@
+#include <stdlib.h>
+
```
PullRequestOpen: Given that this code is unusual in that it skips any and all whitespace, and recognizes a `+`/`-` as occurring after an `e` even when the `e` is missing, some documentation up top of the valid inputs may be warranted. Further, some unit tests would be useful to ensure that the code behaves as expected, and to assist any future developers when making changes. As a suggestion, and also taking into account other observations further below:
```c
/* cfloat_endf
 * Parameters:
 *   const char* buffer   - No NULL-check performed
 *   int n                - No bounds checking performed
 * Notes:
 *   This implementation is meant to only be called by a specific Cython function
 *   and as a result deliberately skips certain checks for performance reasons.
 *   This function will insert an 'e' if it is found to be missing. An internal
 *   buffer-overrun is possible with deliberately crafted input, but is not a
 *   concern within the known execution context.
 */
```

openmc/data/endf.c
```diff
@@ -0,0 +1,39 @@
...
+double cfloat_endf(const char* buffer, int n)
+{
+  char arr[12]; // 11 characters plus a null terminator
```
PullRequestOpen: Depending on the function input, this array could be overflowed.

openmc/data/endf.c
```diff
@@ -0,0 +1,39 @@
+#include <stdlib.h>
+
+double cfloat_endf(const char* buffer, int n)
```
PullRequestOpen: A few suggestions here. Given this is only called from Cython I realize that the inputs are constrained, so please disregard if you think that is already sufficient, but I'm also looking out for if this were potentially reused in some other context.
* For `buffer`, check that it isn't NULL.
* For `n`, either change type to `size_t` or `unsigned int`, or check that `n >= 0`.

openmc/data/endf.c
```diff
@@ -0,0 +1,39 @@
...
+{
+  char arr[12]; // 11 characters plus a null terminator
+  int j = 0; // current position in arr
+  int found_significand = 0;
+  int found_exponent = 0;
+  for (int i = 0; i < n; ++i) {
```
PullRequestOpen: Note that there is no bounds checking performed on `n`, and so therefore `i`. Therefore, `arr` could be overflowed. If the input to this function can be customized by a user by using a deliberately crafted input value then at the very least a stack corruption could be triggered and cause the program to crash on return, or a security issue could be potentially exploited.

openmc/data/endf.c
```diff
@@ -0,0 +1,39 @@
...
+    // Skip whitespace characters
```
PullRequestOpen: Tiny nit: this comment might be better moved one line down, ie. immediately connected to the line: `if (c == ' ') continue;`

PR #1430, opened 2019-12-06
Review by PullRequest, Inc. - Part 5

PullRequestOpen:

### DIFF
--- a/include/openmc/distribution_energy.h
+++ b/include/openmc/distribution_energy.h
@@ -1,0 +1,152 @@
+//! \file distribution_energy.h
+//! Energy distributions that depend on incident particle energy
+
+#ifndef OPENMC_DISTRIBUTION_ENERGY_H
+#define OPENMC_DISTRIBUTION_ENERGY_H
+
+#include <vector>
+
+#include "xtensor/xtensor.hpp"
+#include "hdf5.h"
+
+#include "openmc/constants.h"
+#include "openmc/endf.h"
+
+namespace openmc {
+
+//===============================================================================
+//! Abstract class defining an energy distribution that is a function of the
+//! incident energy of a projectile. Each derived type must implement a sample()
+//! function that returns a sampled outgoing energy given an incoming energy
+//===============================================================================
+
+class EnergyDistribution {
+public:
+  virtual double sample(double E) const = 0;
+  virtual ~EnergyDistribution() = default;
+};
+
+//===============================================================================
+//! Discrete photon energy distribution
+//===============================================================================
+
+class DiscretePhoton : public EnergyDistribution {
+public:
+  explicit DiscretePhoton(hid_t group);
+
+  //! Sample energy distribution
+  //! \param[in] E Incident particle energy in [eV]
+  //! \return Sampled energy in [eV]
+  double sample(double E) const;
+private:
+  int primary_flag_; //!< Indicator of whether the photon is a primary or
+                     //!< non-primary photon.
+  double energy_; //!< Photon energy or binding energy
+  double A_; //!< Atomic weight ratio of the target nuclide
+};
+
+//===============================================================================
+//! Level inelastic scattering distribution
+//===============================================================================
+
+class LevelInelastic : public EnergyDistribution {
+public:
+  explicit LevelInelastic(hid_t group);
+
+  //! Sample energy distribution
+  //! \param[in] E Incident particle energy in [eV]
+  //! \return Sampled energy in [eV]
+  double sample(double E) const;
+private:
+  double threshold_; //!< Energy threshold in lab, (A + 1)/A * |Q|
+  double mass_ratio_; //!< (A/(A+1))^2
+};
+
+//===============================================================================
+//! An energy distribution represented as a tabular distribution with histogram
+//! or linear-linear interpolation. This corresponds to ACE law 4, which NJOY
+//! produces for a number of ENDF energy distributions.
+//===============================================================================
+
+class ContinuousTabular : public EnergyDistribution {
+public:
+  explicit ContinuousTabular(hid_t group);
+
+  //! Sample energy distribution
+  //! \param[in] E Incident particle energy in [eV]
+  //! \return Sampled energy in [eV]
+  double sample(double E) const;
+private:
+  //! Outgoing energy for a single incoming energy
+  struct CTTable {
+    Interpolation interpolation; //!< Interpolation law
+    int n_discrete; //!< Number of of discrete energies
+    xt::xtensor<double, 1> e_out; //!< Outgoing energies in [eV]
+    xt::xtensor<double, 1> p; //!< Probability density
+    xt::xtensor<double, 1> c; //!< Cumulative distribution
+  };
+
+  int n_region_; //!< Number of inteprolation regions
+  std::vector<int> breakpoints_; //!< Breakpoints between regions
+  std::vector<Interpolation> interpolation_; //!< Interpolation laws
+  std::vector<double> energy_; //!< Incident energy in [eV]
+  std::vector<CTTable> distribution_; //!< Distributions for each incident energy
+};
+
+//===============================================================================
+//! Evaporation spectrum corresponding to ACE law 9 and ENDF File 5, LF=9.
+//===============================================================================
+
+class Evaporation : public EnergyDistribution {
+public:
+  explicit Evaporation(hid_t group);
+
+  //! Sample energy distribution
+  //! \param[in] E Incident particle energy in [eV]
+  //! \return Sampled energy in [eV]
+  double sample(double E) const;
+private:
+  Tabulated1D theta_; //!< Incoming energy dependent parameter
+  double u_; //!< Restriction energy
+};
+
+//===============================================================================
+//! Energy distribution of neutrons emitted from a Maxwell fission spectrum.
+//! This corresponds to ACE law 7 and ENDF File 5, LF=7.
+//===============================================================================
+
+class MaxwellEnergy : public EnergyDistribution {
+public:
+  explicit MaxwellEnergy(hid_t group);
+
+  //! Sample energy distribution
+  //! \param[in] E Incident particle energy in [eV]
+  //! \return Sampled energy in [eV]
+  double sample(double E) const;
+private:
+  Tabulated1D theta_; //!< Incoming energy dependent parameter
+  double u_; //!< Restriction energy
+};
+
+//===============================================================================
+//! Energy distribution of neutrons emitted from a Watt fission spectrum. This
+//! corresponds to ACE law 11 and ENDF File 5, LF=11.
+//===============================================================================
+
+class WattEnergy : public EnergyDistribution {
+public:
+  explicit WattEnergy(hid_t group);
+
+  //! Sample energy distribution
+  //! \param[in] E Incident particle energy in [eV]
+  //! \return Sampled energy in [eV]
+  double sample(double E) const;
+private:
+  Tabulated1D a_; //!< Energy-dependent 'a' parameter
+  Tabulated1D b_; //!< Energy-dependent 'b' parameter
+  double u_; //!< Restriction energy
+};
+
+} // namespace openmc
+
+#endif // OPENMC_DISTRIBUTION_ENERGY_H
--- a/include/openmc/geometry.h
+++ b/include/openmc/geometry.h
@@ -1,0 +1,80 @@
+#ifndef OPENMC_GEOMETRY_H
+#define OPENMC_GEOMETRY_H
+
+#include <array>
+#include <cmath>
+#include <cstdint>
+#include <vector>
+
+#include "openmc/particle.h"
+
+
+namespace openmc {
+
+//==============================================================================
+// Global variables
+//==============================================================================
+
+namespace model {
+
+extern int root_universe;  //!< Index of root universe
+extern int n_coord_levels; //!< Number of CSG coordinate levels
+
+extern std::vector<int64_t> overlap_check_count;
+
+} // namespace model
+
+//==============================================================================
+// Information about nearest boundary crossing
+//==============================================================================
+
+struct BoundaryInfo {
+  double distance {INFINITY};   //!< distance to nearest boundary
+  int surface_index {0}; //!< if boundary is surface, index in surfaces vector
+  int coord_level;   //!< coordinate level after crossing boundary
+  std::array<int, 3> lattice_translation {}; //!< which way lattice indices will change
+};
+
+//==============================================================================
+//! Check two distances by coincidence tolerance
+//==============================================================================
+
+inline bool coincident(double d1, double d2) {
+  return std::abs(d1 - d2) < FP_COINCIDENT;
+}
+
+//==============================================================================
+//! Check for overlapping cells at a particle's position.
+//==============================================================================
+
+bool check_cell_overlap(Particle* p, bool error=true);
+
+//==============================================================================
+//! Locate a particle in the geometry tree and set its geometry data fields.
+//!
+//! \param p A particle to be located.  This function will populate the
+//!   geometry-dependent data fields of the particle.
+//! \param use_neighbor_lists If true, neighbor lists will be used to accelerate
+//!   the geometry search, but this only works if the cell attribute of the
+//!   particle's lowest coordinate level is valid and meaningful.
+//! \return True if the particle's location could be found and ascribed to a
+//!   valid geometry coordinate stack.
+//==============================================================================
+
+bool find_cell(Particle* p, bool use_neighbor_lists);
+
+//==============================================================================
+//! Move a particle into a new lattice tile.
+//==============================================================================
+
+void cross_lattice(Particle* p, const BoundaryInfo& boundary);
+
+//==============================================================================
+//! Find the next boundary a particle will intersect.
+//==============================================================================
+
+BoundaryInfo distance_to_boundary(Particle* p);
+
+} // namespace openmc
+
+#endif // OPENMC_GEOMETRY_H
--- a/include/openmc/thermal.h
+++ b/include/openmc/thermal.h
@@ -1,0 +1,130 @@
+#ifndef OPENMC_THERMAL_H
+#define OPENMC_THERMAL_H
+
+#include <cstddef>
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "xtensor/xtensor.hpp"
+
+#include "openmc/angle_energy.h"
+#include "openmc/endf.h"
+#include "openmc/hdf5_interface.h"
+#include "openmc/particle.h"
+
+namespace openmc {
+
+//==============================================================================
+// Constants
+//==============================================================================
+
+// Secondary energy mode for S(a,b) inelastic scattering
+// TODO: Convert to enum
+constexpr int SAB_SECONDARY_EQUAL  {0}; // Equally-likely outgoing energy bins
+constexpr int SAB_SECONDARY_SKEWED {1}; // Skewed outgoing energy bins
+constexpr int SAB_SECONDARY_CONT   {2}; // Continuous, linear-linear interpolation
+
+// Elastic mode for S(a,b) elastic scattering
+// TODO: Convert to enum
+constexpr int SAB_ELASTIC_INCOHERENT {3}; // Incoherent elastic scattering
+constexpr int SAB_ELASTIC_COHERENT   {4}; // Coherent elastic scattering (Bragg edges)
+
+//==============================================================================
+// Global variables
+//==============================================================================
+
+class ThermalScattering;
+
+namespace data {
+extern std::vector<std::unique_ptr<ThermalScattering>> thermal_scatt;
+extern std::unordered_map<std::string, int> thermal_scatt_map;
+}
+
+//==============================================================================
+//! Secondary angle-energy data for thermal neutron scattering at a single
+//! temperature
+//==============================================================================
+
+class ThermalData {
+public:
+  ThermalData(hid_t group);
+
+  //! Calculate the cross section
+  //
+  //! \param[in] E Incident neutron energy in [eV]
+  //! \param[out] elastic Elastic scattering cross section in [b]
+  //! \param[out] inelastic Inelastic scattering cross section in [b]
+  void calculate_xs(double E, double* elastic, double* inelastic) const;
+
+  //! Sample an outgoing energy and angle
+  //
+  //! \param[in] micro_xs Microscopic cross sections
+  //! \param[in] E_in Incident neutron energy in [eV]
+  //! \param[out] E_out Outgoing neutron energy in [eV]
+  //! \param[out] mu Outgoing scattering angle cosine
+  void sample(const NuclideMicroXS& micro_xs, double E_in,
+              double* E_out, double* mu);
+private:
+  struct Reaction {
+    // Default constructor
+    Reaction() { }
+
+    // Data members
+    std::unique_ptr<Function1D> xs; //!< Cross section
+    std::unique_ptr<AngleEnergy> distribution; //!< Secondary angle-energy distribution
+  };
+
+  // Inelastic scattering data
+  Reaction elastic_;
+  Reaction inelastic_;
+
+  // ThermalScattering needs access to private data members
+  friend class ThermalScattering;
+};
+
+//==============================================================================
+//! Data for thermal neutron scattering, typically off light isotopes in
+//! moderating materials such as water, graphite, BeO, etc.
+//==============================================================================
+
+class ThermalScattering {
+public:
+  ThermalScattering(hid_t group, const std::vector<double>& temperature);
+
+  //! Determine inelastic/elastic cross section at given energy
+  //!
+  //! \param[in] E incoming energy in [eV]
+  //! \param[in] sqrtkT square-root of temperature multipled by Boltzmann's constant
+  //! \param[out] i_temp corresponding temperature index
+  //! \param[out] elastic Thermal elastic scattering cross section
+  //! \param[out] inelastic Thermal inelastic scattering cross section
+  void calculate_xs(double E, double sqrtkT, int* i_temp, double* elastic,
+                    double* inelastic) const;
+
+  //! Determine whether table applies to a particular nuclide
+  //!
+  //! \param[in] name Name of the nuclide, e.g., "H1"
+  //! \return Whether table applies to the nuclide
+  bool has_nuclide(const char* name) const;
+
+  // Sample an outgoing energy and angle
+  void sample(const NuclideMicroXS& micro_xs, double E_in,
+              double* E_out, double* mu);
+
+  std::string name_; //!< name of table, e.g. "c_H_in_H2O"
+  double awr_;       //!< weight of nucleus in neutron masses
+  double energy_max_; //!< maximum energy for thermal scattering in [eV]
+  std::vector<double> kTs_;  //!< temperatures in [eV] (k*T)
+  std::vector<std::string> nuclides_; //!< Valid nuclides
+
+  //! cross sections and distributions at each temperature
+  std::vector<ThermalData> data_;
+};
+
+void free_memory_thermal();
+
+} // namespace openmc
+
+#endif // OPENMC_THERMAL_H
--- a/src/bank.cpp
+++ b/src/bank.cpp
@@ -1,0 +1,72 @@
+#include "openmc/bank.h"
+
+#include "openmc/capi.h"
+#include "openmc/error.h"
+
+#include <cstdint>
+
+// Explicit template instantiation definition
+template class std::vector<openmc::Particle::Bank>;
+
+namespace openmc {
+
+//==============================================================================
+// Global variables
+//==============================================================================
+
+namespace simulation {
+
+std::vector<Particle::Bank> source_bank;
+std::vector<Particle::Bank> fission_bank;
+std::vector<Particle::Bank> secondary_bank;
+#ifdef _OPENMP
+std::vector<Particle::Bank> master_fission_bank;
+#endif
+
+} // namespace simulation
+
+//==============================================================================
+// Non-member functions
+//==============================================================================
+
+void free_memory_bank()
+{
+  simulation::source_bank.clear();
+  #pragma omp parallel
+  {
+    simulation::fission_bank.clear();
+  }
+#ifdef _OPENMP
+  simulation::master_fission_bank.clear();
+#endif
+}
+
+//==============================================================================
+// C API
+//==============================================================================
+
+extern "C" int openmc_source_bank(void** ptr, int64_t* n)
+{
+  if (simulation::source_bank.size() == 0) {
+    set_errmsg("Source bank has not been allocated.");
+    return OPENMC_E_ALLOCATE;
+  } else {
+    *ptr = simulation::source_bank.data();
+    *n = simulation::source_bank.size();
+    return 0;
+  }
+}
+
+extern "C" int openmc_fission_bank(void** ptr, int64_t* n)
+{
+  if (simulation::fission_bank.size() == 0) {
+    set_errmsg("Fission bank has not been allocated.");
+    return OPENMC_E_ALLOCATE;
+  } else {
+    *ptr = simulation::fission_bank.data();
+    *n = simulation::fission_bank.size();
+    return 0;
+  }
+}
+
+} // namespace openmc
--- a/src/cmfd_solver.cpp
+++ b/src/cmfd_solver.cpp
@@ -1,0 +1,361 @@
+#include "openmc/cmfd_solver.h"
+
+#include <vector>
+#include <cmath>
+
+#include "xtensor/xtensor.hpp"
+
+#include "openmc/error.h"
+#include "openmc/constants.h"
+#include "openmc/capi.h"
+
+namespace openmc {
+
+namespace cmfd {
+
+//==============================================================================
+// Global variables
+//==============================================================================
+
+std::vector<int> indptr;
+
+std::vector<int> indices;
+
+int dim;
+
+double spectral;
+
+int nx, ny, nz, ng;
+
+xt::xtensor<int, 2> indexmap;
+
+} // namespace cmfd
+
+//==============================================================================
+// MATRIX_TO_INDICES converts a matrix index to spatial and group
+// indices
+//==============================================================================
+
+void matrix_to_indices(int irow, int& g, int& i, int& j, int& k)
+{
+  g = irow % cmfd::ng;
+  i = cmfd::indexmap(irow/cmfd::ng, 0);
+  j = cmfd::indexmap(irow/cmfd::ng, 1);
+  k = cmfd::indexmap(irow/cmfd::ng, 2);
+}
+
+//==============================================================================
+// GET_DIAGONAL_INDEX returns the index in CSR index array corresponding to
+// the diagonal element of a specified row
+//==============================================================================
+
+int get_diagonal_index(int row)
+{
+  for (int j = cmfd::indptr[row]; j < cmfd::indptr[row+1]; j++) {
+    if (cmfd::indices[j] == row)
+      return j;
+  }
+
+  // Return -1 if not found
+  return -1;
+}
+
+//==============================================================================
+// SET_INDEXMAP sets the elements of indexmap based on input coremap
+//==============================================================================
+
+void set_indexmap(const int* coremap)
+{
+  for (int z = 0; z < cmfd::nz; z++) {
+    for (int y = 0; y < cmfd::ny; y++) {
+      for (int x = 0; x < cmfd::nx; x++) {
+        if (coremap[(z*cmfd::ny*cmfd::nx) + (y*cmfd::nx) + x] != CMFD_NOACCEL) {
+          int counter = coremap[(z*cmfd::ny*cmfd::nx) + (y*cmfd::nx) + x];
+          cmfd::indexmap(counter, 0) = x;
+          cmfd::indexmap(counter, 1) = y;
+          cmfd::indexmap(counter, 2) = z;
+        }
+      }
+    }
+  }
+}
+
+//==============================================================================
+// CMFD_LINSOLVER_1G solves a one group CMFD linear system
+//==============================================================================
+
+int cmfd_linsolver_1g(const double* A_data, const double* b, double* x,
+                      double tol)
+{
+  // Set overrelaxation parameter
+  double w = 1.0;
+
+  // Perform Gauss-Seidel iterations
+  for (int igs = 1; igs <= 10000; igs++) {
+    double err = 0.0;
+
+    // Copy over x vector
+    std::vector<double> tmpx {x, x+cmfd::dim};
+
+    // Perform red/black Gauss-Seidel iterations
+    for (int irb = 0; irb < 2; irb++) {
+
+      // Loop around matrix rows
+      for (int irow = 0; irow < cmfd::dim; irow++) {
+        int g, i, j, k;
+        matrix_to_indices(irow, g, i, j, k);
+
+        // Filter out black cells
+        if ((i+j+k) % 2 != irb) continue;
+
+        // Get index of diagonal for current row
+        int didx = get_diagonal_index(irow);
+
+        // Perform temporary sums, first do left of diag, then right of diag
+        double tmp1 = 0.0;
+        for (int icol = cmfd::indptr[irow]; icol < didx; icol++)
+          tmp1 += A_data[icol] * x[cmfd::indices[icol]];
+        for (int icol = didx + 1; icol < cmfd::indptr[irow + 1]; icol++)
+          tmp1 += A_data[icol] * x[cmfd::indices[icol]];
+
+        // Solve for new x
+        double x1 = (b[irow] - tmp1) / A_data[didx];
+
+        // Perform overrelaxation
+        x[irow] = (1.0 - w) * x[irow] + w * x1;
+
+        // Compute residual and update error
+        double res = (tmpx[irow] - x[irow]) / tmpx[irow];
+        err += res * res;
+      }
+    }
+
+    // Check convergence
+    err = std::sqrt(err / cmfd::dim);
+    if (err < tol)
+      return igs;
+
+    // Calculate new overrelaxation parameter
+    w = 1.0/(1.0 - 0.25 * cmfd::spectral * w);
+  }
+
+  // Throw error, as max iterations met
+  fatal_error("Maximum Gauss-Seidel iterations encountered.");
+
+  // Return -1 by default, although error thrown before reaching this point
+  return -1;
+}
+
+//==============================================================================
+// CMFD_LINSOLVER_2G solves a two group CMFD linear system
+//==============================================================================
+
+int cmfd_linsolver_2g(const double* A_data, const double* b, double* x,
+                      double tol)
+{
+  // Set overrelaxation parameter
+  double w = 1.0;
+
+  // Perform Gauss-Seidel iterations
+  for (int igs = 1; igs <= 10000; igs++) {
+    double err = 0.0;
+
+    // Copy over x vector
+    std::vector<double> tmpx {x, x+cmfd::dim};
+
+    // Perform red/black Gauss-Seidel iterations
+    for (int irb = 0; irb < 2; irb++) {
+
+      // Loop around matrix rows
+      for (int irow = 0; irow < cmfd::dim; irow+=2) {
+        int g, i, j, k;
+        matrix_to_indices(irow, g, i, j, k);
+
+        // Filter out black cells
+        if ((i+j+k) % 2 != irb) continue;
+
+        // Get index of diagonals for current row and next row
+        int d1idx = get_diagonal_index(irow);
+        int d2idx = get_diagonal_index(irow+1);
+
+        // Get block diagonal
+        double m11 = A_data[d1idx];     // group 1 diagonal
+        double m12 = A_data[d1idx + 1]; // group 1 right of diagonal (sorted by col)
+        double m21 = A_data[d2idx - 1]; // group 2 left of diagonal (sorted by col)
+        double m22 = A_data[d2idx];     // group 2 diagonal
+
+        // Analytically invert the diagonal
+        double dm = m11*m22 - m12*m21;
+        double d11 = m22/dm;
+        double d12 = -m12/dm;
+        double d21 = -m21/dm;
+        double d22 = m11/dm;
+
+        // Perform temporary sums, first do left of diag, then right of diag
+        double tmp1 = 0.0;
+        double tmp2 = 0.0;
+        for (int icol = cmfd::indptr[irow]; icol < d1idx; icol++)
+          tmp1 += A_data[icol] * x[cmfd::indices[icol]];
+        for (int icol = cmfd::indptr[irow+1]; icol < d2idx-1; icol++)
+          tmp2 += A_data[icol] * x[cmfd::indices[icol]];
+        for (int icol = d1idx + 2; icol < cmfd::indptr[irow + 1]; icol++)
+          tmp1 += A_data[icol] * x[cmfd::indices[icol]];
+        for (int icol = d2idx + 1; icol < cmfd::indptr[irow + 2]; icol++)
+          tmp2 += A_data[icol] * x[cmfd::indices[icol]];
+
+        // Adjust with RHS vector
+        tmp1 = b[irow] - tmp1;
+        tmp2 = b[irow + 1] - tmp2;
+
+        // Solve for new x
+        double x1 = d11*tmp1 + d12*tmp2;
+        double x2 = d21*tmp1 + d22*tmp2;
+
+        // Perform overrelaxation
+        x[irow] = (1.0 - w) * x[irow] + w * x1;
+        x[irow + 1] = (1.0 - w) * x[irow + 1] + w * x2;
+
+        // Compute residual and update error
+        double res = (tmpx[irow] - x[irow]) / tmpx[irow];
+        err += res * res;
+      }
+    }
+
+    // Check convergence
+    err = std::sqrt(err / cmfd::dim);
+    if (err < tol)
+      return igs;
+
+    // Calculate new overrelaxation parameter
+    w = 1.0/(1.0 - 0.25 * cmfd::spectral * w);
+  }
+
+  // Throw error, as max iterations met
+  fatal_error("Maximum Gauss-Seidel iterations encountered.");
+
+  // Return -1 by default, although error thrown before reaching this point
+  return -1;
+}
+
+//==============================================================================
+// CMFD_LINSOLVER_NG solves a general CMFD linear system
+//==============================================================================
+
+int cmfd_linsolver_ng(const double* A_data, const double* b, double* x,
+                      double tol)
+{
+  // Set overrelaxation parameter
+  double w = 1.0;
+
+  // Perform Gauss-Seidel iterations
+  for (int igs = 1; igs <= 10000; igs++) {
+    double err = 0.0;
+
+    // Copy over x vector
+    std::vector<double> tmpx {x, x+cmfd::dim};
+
+    // Loop around matrix rows
+    for (int irow = 0; irow < cmfd::dim; irow++) {
+      // Get index of diagonal for current row
+      int didx = get_diagonal_index(irow);
+
+      // Perform temporary sums, first do left of diag, then right of diag
+      double tmp1 = 0.0;
+      for (int icol = cmfd::indptr[irow]; icol < didx; icol++)
+        tmp1 += A_data[icol] * x[cmfd::indices[icol]];
+      for (int icol = didx + 1; icol < cmfd::indptr[irow + 1]; icol++)
+        tmp1 += A_data[icol] * x[cmfd::indices[icol]];
+
+      // Solve for new x
+      double x1 = (b[irow] - tmp1) / A_data[didx];
+
+      // Perform overrelaxation
+      x[irow] = (1.0 - w) * x[irow] + w * x1;
+
+      // Compute residual and update error
+      double res = (tmpx[irow] - x[irow]) / tmpx[irow];
+      err += res * res;
+    }
+
+    // Check convergence
+    err = std::sqrt(err / cmfd::dim);
+    if (err < tol)
+      return igs;
+
+    // Calculate new overrelaxation parameter
+    w = 1.0/(1.0 - 0.25 * cmfd::spectral * w);
+  }
+
+  // Throw error, as max iterations met
+  fatal_error("Maximum Gauss-Seidel iterations encountered.");
+
+  // Return -1 by default, although error thrown before reaching this point
+  return -1;
+}
+
+//==============================================================================
+// OPENMC_INITIALIZE_LINSOLVER sets the fixed variables that are used for the
+// linear solver
+//==============================================================================
+
+extern "C"
+void openmc_initialize_linsolver(const int* indptr, int len_indptr,
+                                 const int* indices, int n_elements, int dim,
+                                 double spectral, const int* cmfd_indices,
+                                 const int* map)
+{
+  // Store elements of indptr
+  for (int i = 0; i < len_indptr; i++)
+    cmfd::indptr.push_back(indptr[i]);
+
+  // Store elements of indices
+  for (int i = 0; i < n_elements; i++)
+    cmfd::indices.push_back(indices[i]);
+
+  // Set dimenion of CMFD problem and specral radius
+  cmfd::dim = dim;
+  cmfd::spectral = spectral;
+
+  // Set number of groups
+  cmfd::ng = cmfd_indices[3];
+
+  // Set problem dimensions and indexmap if 1 or 2 group problem
+  if (cmfd::ng == 1 || cmfd::ng == 2) {
+    cmfd::nx = cmfd_indices[0];
+    cmfd::ny = cmfd_indices[1];
+    cmfd::nz = cmfd_indices[2];
+
+    // Resize indexmap and set its elements
+    cmfd::indexmap.resize({static_cast<size_t>(dim), 3});
+    set_indexmap(map);
+  }
+}
+
+//==============================================================================
+// OPENMC_RUN_LINSOLVER runs a Gauss Seidel linear solver to solve CMFD matrix
+// equations
+//==============================================================================
+
+extern "C"
+int openmc_run_linsolver(const double* A_data, const double* b, double* x,
+                         double tol)
+{
+  switch (cmfd::ng) {
+  case 1:
+    return cmfd_linsolver_1g(A_data, b, x, tol);
+  case 2:
+    return cmfd_linsolver_2g(A_data, b, x, tol);
+  default:
+    return cmfd_linsolver_ng(A_data, b, x, tol);
+  }
+}
+
+void free_memory_cmfd()
+{
+  cmfd::indptr.clear();
+  cmfd::indices.clear();
+  // Resize indexmap to be an empty array
+  cmfd::indexmap.resize({0});
+}
+
+} // namespace openmc
--- a/src/cross_sections.cpp
+++ b/src/cross_sections.cpp
@@ -1,0 +1,433 @@
+#include "openmc/cross_sections.h"
+
+#include "openmc/constants.h"
+#include "openmc/container_util.h"
+#ifdef DAGMC
+#include "openmc/dagmc.h"
+#endif
+#include "openmc/error.h"
+#include "openmc/file_utils.h"
+#include "openmc/hdf5_interface.h"
+#include "openmc/material.h"
+#include "openmc/message_passing.h"
+#include "openmc/mgxs_interface.h"
+#include "openmc/nuclide.h"
+#include "openmc/photon.h"
+#include "openmc/settings.h"
+#include "openmc/simulation.h"
+#include "openmc/string_utils.h"
+#include "openmc/thermal.h"
+#include "openmc/xml_interface.h"
+#include "openmc/wmp.h"
+
+#include "pugixml.hpp"
+
+#include <cstdlib> // for getenv
+#include <unordered_set>
+
+namespace openmc {
+
+//==============================================================================
+// Global variable declarations
+//==============================================================================
+
+namespace data {
+
+std::vector<Library> libraries;
+std::map<LibraryKey, std::size_t> library_map;
+
+}
+
+//==============================================================================
+// Library methods
+//==============================================================================
+
+Library::Library(pugi::xml_node node, const std::string& directory)
+{
+  // Get type of library
+  if (check_for_node(node, "type")) {
+    auto type = get_node_value(node, "type");
+    if (type == "neutron") {
+      type_ = Type::neutron;
+    } else if (type == "thermal") {
+      type_ = Type::thermal;
+    } else if (type == "photon") {
+      type_ = Type::photon;
+    } else if (type == "wmp") {
+      type_ = Type::wmp;
+    } else {
+      fatal_error("Unrecognized library type: " + type);
+    }
+  } else {
+    fatal_error("Missing library type");
+  }
+
+  // Get list of materials
+  if (check_for_node(node, "materials")) {
+    materials_ = get_node_array<std::string>(node, "materials");
+  }
+
+  // determine path of cross section table
+  if (!check_for_node(node, "path")) {
+    fatal_error("Missing library path");
+  }
+  std::string path = get_node_value(node, "path");
+
+  if (starts_with(path, "/")) {
+    path_ = path;
+  } else if (ends_with(directory, "/")) {
+    path_ = directory + path;
+  } else {
+    path_ = directory + "/" + path;
+  }
+
+  if (!file_exists(path_)) {
+    warning("Cross section library " + path_ + " does not exist.");
+  }
+}
+
+//==============================================================================
+// Non-member functions
+//==============================================================================
+
+void read_cross_sections_xml()
+{
+  pugi::xml_document doc;
+  std::string filename = settings::path_input + "materials.xml";
+#ifdef DAGMC
+  std::string s;
+  bool found_uwuw_mats = false;
+  if (settings::dagmc) {
+    found_uwuw_mats = get_uwuw_materials_xml(s);
+  }
+
+  if (found_uwuw_mats) {
+    // if we found uwuw materials, load those
+    doc.load_file(s.c_str());
+  } else {
+#endif
+  // Check if materials.xml exists
+  if (!file_exists(filename)) {
+    fatal_error("Material XML file '" + filename + "' does not exist.");
+  }
+  // Parse materials.xml file
+  doc.load_file(filename.c_str());
+#ifdef DAGMC
+  }
+#endif
+
+  auto root = doc.document_element();
+
+  // Find cross_sections.xml file -- the first place to look is the
+  // materials.xml file. If no file is found there, then we check the
+  // OPENMC_CROSS_SECTIONS environment variable
+  if (!check_for_node(root, "cross_sections")) {
+    // No cross_sections.xml file specified in settings.xml, check
+    // environment variable
+    if (settings::run_CE) {
+      char* envvar = std::getenv("OPENMC_CROSS_SECTIONS");
+      if (!envvar) {
+        fatal_error("No cross_sections.xml file was specified in "
+          "materials.xml or in the OPENMC_CROSS_SECTIONS"
+          " environment variable. OpenMC needs such a file to identify "
+          "where to find data libraries. Please consult the"
+          " user's guide at https://openmc.readthedocs.io for "
+          "information on how to set up data libraries.");
+      }
+      settings::path_cross_sections = envvar;
+    } else {
+      char* envvar = std::getenv("OPENMC_MG_CROSS_SECTIONS");
+      if (!envvar) {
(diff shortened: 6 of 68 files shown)


src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+               MPI_SUM, 0, mpi::intracomm);
+  #endif
+
+  if (mpi::master) {
+    header("cell overlap check summary", 1);
+    std::cout << " Cell ID      No. Overlap Checks\n";
+
+    std::vector<int32_t> sparse_cell_ids;
+    for (int i = 0; i < model::cells.size(); i++) {
+      std::cout << " " << std::setw(8) << model::cells[i]->id_ << std::setw(17)
```
PullRequestOpen: Optional: Looks like you're setting standard widths for the cell_id and check_count_outputs here.  Just wanted to call out that there is no space in the `cout` stream between the cell `id_` and the `overlap_check_count`.

If this output is being fed into another system where precise widths are necessary, this isn't a problem.  However if the `overlap_check_count` winds up being a width of 17, the `id` and the `overlap_check_count` will bump up against one another without a separator.

Would recommend adding a space between `id_` and `overlap_check_count`, similar to elsewhere in the codebase.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+//==============================================================================
+
+void print_generation()
+{
+  // Save state of cout
+  auto f {std::cout.flags()};
+
+  // Determine overall generation and number of active generations
+  int i = overall_generation() - 1;
```
PullRequestOpen: Nit / Optional: Recommend renaming `i` to `generation_index`, and changing "Determine overall generation" to "Determine overall generation index" above.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+  // Restore state of cout
+  std::cout.flags(f);
+}
+
+//==============================================================================
+
+void show_time(const char* label, double secs, int indent_level=0)
+{
+  std::cout << std::string(2*indent_level, ' ');
+  int width = 33 - indent_level*2;
```
PullRequestOpen: Question: Is there a significance of 33 here (is it the sum of a number of other widths)?  Same for line 447.  If so I would make it a class-level constant or otherwise tie it to a more central config.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+      auto alpha = 1 - CONFIDENCE_LEVEL;
+      t_value = t_percentile(1 - alpha*0.5, tally.n_realizations_ - 1);
+    }
+
+    // Write derivative information.
+    if (tally.deriv_ != C_NONE) {
+      const auto& deriv {model::tally_derivs[tally.deriv_]};
+      switch (deriv.variable) {
+      case DIFF_DENSITY:
+        tallies_out << " Density derivative  Material "
```
PullRequestOpen: Nit: Not sure if this was intentional, but there is a extra space between "derivative" and "Material", both here and on lines 667 and 672.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+  // write global tallies
+  const auto& gt = simulation::global_tallies;
+  double mean, stdev;
+  if (n > 1) {
+    if (settings::run_mode == RUN_MODE_EIGENVALUE) {
+      std::tie(mean, stdev) = mean_stdev(&gt(K_COLLISION, 0), n);
+      std::cout << " k-effective (Collision)     = "
+        << mean << " +/- " << t_n1 * stdev << '\n';
+      std::tie(mean, stdev) = mean_stdev(&gt(K_TRACKLENGTH, 0), n);
```
PullRequestOpen: Good use of variable unpacking `std::tie` operator

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+    std::cout << '\n' << out << "\n\n";
+}
+
+//==============================================================================
+
+std::string time_stamp()
+{
+  std::stringstream ts;
+  std::time_t t = std::time(nullptr);   // get time now
+  ts << std::put_time(std::localtime(&t), "%Y-%m-%d %H:%M:%S");
```
PullRequestOpen: Optional: If you're expecting this might be used for debugging, might be useful to add millisecond precision.  Can make a difference when comparing and ordering log outputs.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+  std::cout << "\n";
+}
+
+//==============================================================================
+
+void print_plot()
+{
+  header("PLOTTING SUMMARY", 5);
+  if (settings::verbosity < 5) return;
```
PullRequestOpen: If possible, I would use a constant rather than the magic number `5` here.  Not sure what logging library / levels are being used, but typically constants are provided within the libraries.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+//==============================================================================
+
+void print_generation()
+{
+  // Save state of cout
+  auto f {std::cout.flags()};
+
+  // Determine overall generation and number of active generations
+  int i = overall_generation() - 1;
+  int n = simulation::current_batch > settings::n_inactive ?
```
PullRequestOpen: Nit / Optional: Recommend changing `n` to active_generation_count or `n_active_generations`.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+  auto f {std::cout.flags()};
+
+  // Determine overall generation and number of active generations
+  int i = simulation::current_batch*settings::gen_per_batch - 1;
+  int n = simulation::n_realizations*settings::gen_per_batch;
+
+  // Set format for values
+  std::cout << std::fixed << std::setprecision(5);
+
+  // write out information batch and option independent output
```
PullRequestOpen: Question: Can you clarify "option independent output"?  Did you mean "optional independent output"? Or does "option independent" refer specifically to `gen_per_batch` and the overall generation index in `k_generation `?

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+      "  =========   ========   ====================\n";
+  }
+}
+
+//==============================================================================
+
+void print_generation()
+{
+  // Save state of cout
+  auto f {std::cout.flags()};
```
PullRequestOpen: Nit / Optional: Highly recommend not having `f` as the identifier name in this case.  I normally reserve single-letter variables for either iterators, or specific variable names for mathematical equations.

Both here, and on lines 406, 460, and 538.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+    std::cout << "    r = (" << p->coord_[i].r.x << ", "
+              << p->coord_[i].r.y << ", " << p->coord_[i].r.z << ")\n";
+    std::cout << "    u = (" << p->coord_[i].u.x << ", "
+              << p->coord_[i].u.y << ", " << p->coord_[i].u.z << ")\n";
+  }
+
+  // Display miscellaneous info.
+  if (p->surface_ != 0) {
+    const Surface& surf {*model::surfaces[std::abs(p->surface_)-1]};
```
PullRequestOpen: Optional: I would clarify the significance of the `-1` offset here.

src/output.cpp
```diff
@@ -0,0 +1,736 @@
...
+//==============================================================================
+
+void print_batch_keff()
+{
+  // Save state of cout
+  auto f {std::cout.flags()};
+
+  // Determine overall generation and number of active generations
+  int i = simulation::current_batch*settings::gen_per_batch - 1;
+  int n = simulation::n_realizations*settings::gen_per_batch;
```
PullRequestOpen: Possible bug or clarification: Just calling out here that the calculation (and / or the definition) for overall generation and number of active generations is different than on line 374 and line 375.

Specifically, the definition of "overall generation" may be the same, however, the definition above calls `overall_generation` and does a index lookup into `simulation::k_generation`.  "number of active generations" looks like it sets a floor value based on `settings::n_inactive`.

This difference may be intentional, however if they are intentionally different I would clarify the definitions in the comments.

include/openmc/thermal.h
```diff
@@ -0,0 +1,130 @@
...
+    // Data members
+    std::unique_ptr<Function1D> xs; //!< Cross section
+    std::unique_ptr<AngleEnergy> distribution; //!< Secondary angle-energy distribution
+  };
+
+  // Inelastic scattering data
+  Reaction elastic_;
+  Reaction inelastic_;
+
+  // ThermalScattering needs access to private data members
```
PullRequestOpen: Thank you for the clear explanation for the purpose of a `friend` class.

src/thermal.cpp
```diff
@@ -0,0 +1,289 @@
...
+  for (auto T : temps_to_read) {
+    // Get temperature as a string
+    std::string temp_str = std::to_string(T) + "K";
+
+    // Read exact temperature value
+    double kT;
+    read_dataset(kT_group, temp_str.data(), kT);
+    kTs_.push_back(kT);
+
+    // Open group for temperature i
```
PullRequestOpen: Potential typo: May have meant "temperature T" to indicate the iterator variable?  `i` does not appear to be defined in this scope.

src/thermal.cpp
```diff
@@ -0,0 +1,289 @@
...
+  hid_t kT_group = open_group(group, "kTs");
+
+  // Determine temperatures available
+  auto dset_names = dataset_names(kT_group);
+  auto n = dset_names.size();
+  auto temps_available = xt::empty<double>({n});
+  for (int i = 0; i < dset_names.size(); ++i) {
+    // Read temperature value
+    double T;
+    read_dataset(kT_group, dset_names[i].data(), T);
```
PullRequestOpen: Question: It's slightly unusual to declare a primitive (`double` type here) then pass it to be modified to a function (`read_dataset`) by passing the primitive directly, rather than passing the pointer to the primitive.  I'm sure there's a reason you are doing this here, but I just want to double-check here and make sure that is correct.

src/random_lcg.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+//==============================================================================
+// FUTURE_PRN
+//==============================================================================
+
+extern "C" double
+future_prn(int64_t n)
+{
+  return future_seed(static_cast<uint64_t>(n), prn_seed[stream]) * prn_norm;
```
PullRequestOpen: Thank you for explicitly using `static_cast`

src/reaction.cpp
```diff
@@ -0,0 +1,277 @@
...
+  // <<<<<<<<<<<<<<<<<<<<<<<<<<<< REMOVE THIS <<<<<<<<<<<<<<<<<<<<<<<<<
+}
+
+//==============================================================================
+// Non-member functions
+//==============================================================================
+
+std::string reaction_name(int mt)
+{
+  if (mt == SCORE_FLUX) {
```
PullRequestOpen: consider splitting this if-else conditional into a hash_map lookup for string search cases.

src/reaction.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+namespace openmc {
+
+//==============================================================================
+// Reaction implementation
+//==============================================================================
+
+Reaction::Reaction(hid_t group, const std::vector<int>& temperatures)
+{
+  read_attribute(group, "Q_value", q_value_);
```
PullRequestOpen: attribute field names should be factored out into an enum or something similar.

src/reaction_product.cpp
```diff
@@ -0,0 +1,101 @@
...
+    // Read applicability
+    if (n > 1) {
+      hid_t app = open_dataset(dgroup, "applicability");
+      applicability_.emplace_back(app);
+      close_dataset(app);
+    }
+
+    // Determine distribution type and read data
+    read_attribute(dgroup, "type", temp);
+    if (temp == "uncorrelated") {
```
PullRequestOpen: see comment regarding enums above.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+
+  } else if ((surf->bc_ == BC_REFLECT || surf->bc_ == BC_WHITE) 
+                                    && (settings::run_mode != RUN_MODE_PLOTTING)) {
+    // =======================================================================
+    // PARTICLE REFLECTS FROM SURFACE
+
+    // Do not handle reflective boundary conditions on lower universes
+    if (n_coord_ != 1) {
+      this->mark_as_lost("Cannot reflect particle " + std::to_string(id_) +
+        " off surface in a lower universe.");
```
PullRequestOpen: I would consider using a singular logging method for your application.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+      return;
+    }
+  }
+}
+
+void
+Particle::mark_as_lost(const char* message)
+{
+  // Print warning and write lost particle file
+  warning(message);
```
PullRequestOpen: I see that using message names for "warning" and "printing" are being used. This is not right. Use a C++ logger or capture all write events in a custom class. This will not scale.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+void
+Particle::write_restart() const
+{
+  // Dont write another restart file if in particle restart mode
+  if (settings::run_mode == RUN_MODE_PARTICLE) return;
+
+  // Set up file name
+  std::stringstream filename;
+  filename << settings::path_output << "particle_" << simulation::current_batch
+    << '_' << id_ << ".h5";
```
PullRequestOpen: Move strings into constant variables instead of hard-coding them inlined like this.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+
+  // Create microscopic cross section caches
+  neutron_xs_.resize(data::nuclides.size());
+  photon_xs_.resize(data::elements.size());
+}
+
+void
+Particle::clear()
+{
+  // reset any coordinate levels
```
PullRequestOpen: Follow consistency in comments, either a sentence starts with an upper case everywhere or lower case everywhere.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+    material_ = model::cells[i_cell]->material_[0];
+    sqrtkT_ = model::cells[i_cell]->sqrtkT_[0];
+    return;
+  }
+#endif
+
+  if (find_cell(this, true)) return;
+
+  // ==========================================================================
+  // COULDN'T FIND PARTICLE IN NEIGHBORING CELLS, SEARCH ALL CELLS
```
PullRequestOpen: Why is this comment in upper-case ?

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+void
+Particle::clear()
+{
+  // reset any coordinate levels
+  for (auto& level : coord_) level.reset();
+  n_coord_ = 1;
+}
+
+void
+Particle::create_secondary(Direction u, double E, Type type)
```
PullRequestOpen: function arguments should be constant.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+    write_particle_track(*this);
+    finalize_particle_track(*this);
+  }
+}
+
+void
+Particle::cross_surface()
+{
+  int i_surface = std::abs(surface_);
+  // TODO: off-by-one
```
PullRequestOpen: It it doesn't already exist, I would consider writing a test that fails because of this TODO item, so whoever writes it knows they did it right.

src/particle.cpp
```diff
@@ -0,0 +1,681 @@
...
+  filename << settings::path_output << "particle_" << simulation::current_batch
+    << '_' << id_ << ".h5";
+
+  #pragma omp critical (WriteParticleRestart)
+  {
+    // Create file
+    hid_t file_id = file_open(filename.str(), 'w');
+
+    // Write filetype and version info
+    write_attribute(file_id, "filetype", "particle restart");
```
PullRequestOpen: Sug # 1: These attributes should be part of some class that has constant strings containing all attributes.  Don't type these a string literals inside a function. Use something like,wWrite_Attribute(FILE_ATTRIB.FILETYPE, "particle restart") ...
Do this for all attributes below.

src/particle_restart.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+  // Write meessage
+  write_message("Loading particle restart file " +
+    settings::path_particle_restart + "...", 5);
+
+  // Open file
+  hid_t file_id = file_open(settings::path_particle_restart, 'r');
+
+  // Read data from file
+  read_dataset(file_id, "current_batch", simulation::current_batch);
```
PullRequestOpen: Use the same FILE_ATTRIB.FILETYPE standard attribute fields in reading the dataset, that were used for writing. Do not inline field names such as "generations_per_batch" ...etc.

src/particle_restart.cpp
```diff
@@ -0,0 +1,107 @@
...
+    throw std::runtime_error{"Unexpected run mode: " +
+      std::to_string(previous_run_mode)};
+  }
+  set_particle_seed(particle_seed);
+
+  // Transport neutron
+  p.transport();
+
+  // Write output if particle made it
+  print_particle(&p);
```
PullRequestOpen: Is there a standard logger writing the particle in the print_particle function ?

src/particle_restart.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <array>
+#include <stdexcept>
+#include <string>
+
+namespace openmc {
+
+void read_particle_restart(Particle& p, int& previous_run_mode)
+{
+  // Write meessage
+  write_message("Loading particle restart file " +
```
PullRequestOpen: Is this writing output through a standard logging method ?

src/lattice.cpp
```diff
@@ -0,0 +1,1075 @@
...
+
+ReverseLatticeIter HexLattice::rbegin()
+{return ReverseLatticeIter(*this, universes_.size()-n_rings_);}
+
+//==============================================================================
+
+bool
+HexLattice::are_valid_indices(const int i_xyz[3]) const
+{
+  return ((i_xyz[0] >= 0) && (i_xyz[1] >= 0) && (i_xyz[2] >= 0)
```
PullRequestOpen: Suggest adding some comments here.

src/lattice.cpp
```diff
@@ -0,0 +1,1075 @@
...
+  return out;
+}
+
+//==============================================================================
+
+void
+RectLattice::to_hdf5_inner(hid_t lat_group) const
+{
+  // Write basic lattice information.
+  write_string(lat_group, "type", "rectangular", false);
```
PullRequestOpen: These attributes need to come from a constant string array or enum etc.

src/lattice.cpp
```diff
@@ -0,0 +1,1075 @@
...
+        for (int j = 0; j < nx; j++) {
+          int indx1 = nx*ny*m + nx*k + j;
+          int indx2 = nx*ny*m + nx*(ny-k-1) + j;
+          out[indx2] = model::universes[universes_[indx1]]->id_;
+        }
+      }
+    }
+
+    hsize_t dims[3] {nz, ny, nx};
+    write_int(lat_group, 3, dims, "universes", out.data(), false);
```
PullRequestOpen: Use constants instead of inlining fields/attributes such as "universes"

src/initialize.cpp
```diff
@@ -0,0 +1,291 @@
...
+        // Read and set number of OpenMP threads
+        int n_threads = std::stoi(argv[i]);
+        if (n_threads < 1) {
+          std::string msg {"Number of threads must be positive."};
+          strcpy(openmc_err_msg, msg.c_str());
+          return OPENMC_E_INVALID_ARGUMENT;
+        }
+        omp_set_num_threads(n_threads);
+#else
+        if (mpi::master)
```
PullRequestOpen: The established pattern is to use { } even if there is only one line.

src/initialize.cpp
```diff
@@ -0,0 +1,291 @@
...
+  // Create bank datatype
+  Particle::Bank b;
+  MPI_Aint disp[6];
+  MPI_Get_address(&b.r, &disp[0]);
+  MPI_Get_address(&b.u, &disp[1]);
+  MPI_Get_address(&b.E, &disp[2]);
+  MPI_Get_address(&b.wgt, &disp[3]);
+  MPI_Get_address(&b.delayed_group, &disp[4]);
+  MPI_Get_address(&b.particle, &disp[5]);
+  for (int i = 5; i >= 0; --i) disp[i] -= disp[0];
```
PullRequestOpen: To be consistent with the other commits, consider placing the disp[i] -= disp[0]; on a seperate line from the for loop.

src/geometry_aux.cpp
```diff
@@ -0,0 +1,569 @@
...
+//==============================================================================
+//! Partition some universes with many z-planes for faster find_cell searches.
+
+void
+partition_universes()
+{
+  // Iterate over universes with more than 10 cells.  (Fewer than 10 is likely
+  // not worth partitioning.)
+  for (const auto& univ : model::universes) {
+    if (univ->cells_.size() > 10) {
```
PullRequestOpen: Can this 10 value change in the future ? does it need to be factored out as a variable ?

src/geometry_aux.cpp
```diff
@@ -0,0 +1,569 @@
...
+        }
+      }
+
+      // Partition the universe if there are more than 5 z-planes.  (Fewer than
+      // 5 is likely not worth it.)
+      int n_zplanes = 0;
+      for (auto i_surf : surf_inds) {
+        if (dynamic_cast<const SurfaceZPlane*>(model::surfaces[i_surf].get())) {
+          ++n_zplanes;
+          if (n_zplanes > 5) {
```
PullRequestOpen: Can this value change in the future ? does it need to be factored out as a variable ?

include/openmc/geometry.h
```diff
@@ -0,0 +1,80 @@
...
+//!   valid geometry coordinate stack.
+//==============================================================================
+
+bool find_cell(Particle* p, bool use_neighbor_lists);
+
+//==============================================================================
+//! Move a particle into a new lattice tile.
+//==============================================================================
+
+void cross_lattice(Particle* p, const BoundaryInfo& boundary);
```
PullRequestOpen: I would suggest making `ptr` const. Possible using shared_ptr if possible.

src/geometry.cpp
```diff
@@ -0,0 +1,496 @@
...
+    level_surf_cross = surface_distance.second;
+
+    // Find the distance to the next lattice tile crossing.
+    if (coord.lattice != C_NONE) {
+      auto& lat {*model::lattices[coord.lattice]};
+      std::array<int, 3> i_xyz {coord.lattice_x, coord.lattice_y, coord.lattice_z};
+      //TODO: refactor so both lattice use the same position argument (which
+      //also means the lat.type attribute can be removed)
+      std::pair<double, std::array<int, 3>> lattice_distance;
+      switch (lat.type_) {
```
PullRequestOpen: Might want to log that no switch case was hit in the default case if it is unexpected.

src/geometry.cpp
```diff
@@ -0,0 +1,496 @@
...
+    }
+  }
+
+  return false;
+}
+
+//==============================================================================
+
+bool
+find_cell_inner(Particle* p, const NeighborList* neighbor_list)
```
PullRequestOpen: This function is a bit too big with a lot of branching, can it be factored into a simpler one using helper functions? This will help reduce the cognitive load required for future readers and maintainers of your code.

src/geometry.cpp
```diff
@@ -0,0 +1,496 @@
...
+  llc[2] = bbox.zmin;
+
+  // set upper right corner values
+  urc[0] = bbox.xmax;
+  urc[1] = bbox.ymax;
+  urc[2] = bbox.zmax;
+
+  return 0;
+}
+
```
PullRequestOpen: nit: extra line

src/geometry.cpp
```diff
@@ -0,0 +1,496 @@
...
+          if (error) {
+            std::stringstream err_msg;
+            err_msg << "Overlapping cells detected: " << c.id_ << ", "
+                    << model::cells[p->coord_[j].cell]->id_ << " on universe "
+                    << univ.id_;
+            fatal_error(err_msg);
+          }
+          return true;
+        }
+        ++model::overlap_check_count[index_cell];
```
PullRequestOpen: This `model::overlap_check_count` is being incremented here and this `check_cell_overlap` method is being called with a `#pragma omp for` block but doesn't appear to have any locking or synchronization on it. Should some sort of atomic pragma or similar be used here?

include/openmc/distribution_energy.h
```diff
@@ -0,0 +1,152 @@
...
+
+namespace openmc {
+
+//===============================================================================
+//! Abstract class defining an energy distribution that is a function of the
+//! incident energy of a projectile. Each derived type must implement a sample()
+//! function that returns a sampled outgoing energy given an incoming energy
+//===============================================================================
+
+class EnergyDistribution {
```
PullRequestOpen: Add Abstract or similar keyword to class to denote its abstract.

src/distribution_energy.cpp
```diff
@@ -0,0 +1,356 @@
...
+  }
+}
+
+//==============================================================================
+// LevelInelastic implementation
+//==============================================================================
+
+LevelInelastic::LevelInelastic(hid_t group)
+{
+  read_attribute(group, "threshold", threshold_);
```
PullRequestOpen: You could consider loading attribute names from a constant string attribute list to help improve documentation of all of the available attributes as well as preventing typos in the future.

src/distribution_energy.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+  // Sample outgoing energy based on evaporation spectrum probability
+  // density function
+  double x;
+  while (true) {
+    x = -std::log((1.0 - v*prn())*(1.0 - v*prn()));
+    if (x <= y) break;
+  }
+
+  return x*theta;
```
PullRequestOpen: nit: space before and after '*'

src/distribution_energy.cpp
```diff
@@ -0,0 +1,356 @@
...
+  int n_discrete = distribution_[i].n_discrete;
+  double E_i_1 = distribution_[i].e_out[n_discrete];
+  double E_i_K = distribution_[i].e_out[n_energy_out - 1];
+
+  n_energy_out = distribution_[i+1].e_out.size();
+  n_discrete = distribution_[i+1].n_discrete;
+  double E_i1_1 = distribution_[i+1].e_out[n_discrete];
+  double E_i1_K = distribution_[i+1].e_out[n_energy_out - 1];
+
+  double E_1 = E_i_1 + r*(E_i1_1 - E_i_1);
```
PullRequestOpen: nit: space before and after '*'

src/distribution_energy.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+namespace openmc {
+
+//==============================================================================
+// DiscretePhoton implementation
+//==============================================================================
+
+DiscretePhoton::DiscretePhoton(hid_t group)
+{
+  read_attribute(group, "primary_flag", primary_flag_);
```
PullRequestOpen: Load attribute names from a constant string attribute list.

src/distribution_energy.cpp
```diff
@@ -0,0 +1,356 @@
...
+//==============================================================================
+
+ContinuousTabular::ContinuousTabular(hid_t group)
+{
+  // Open incoming energy dataset
+  hid_t dset = open_dataset(group, "energy");
+
+  // Get interpolation parameters
+  xt::xarray<int> temp;
+  read_attribute(dset, "interpolation", temp);
```
PullRequestOpen: See Sug # 1, same for all use cases below.

src/distribution_energy.cpp
```diff
@@ -0,0 +1,356 @@
...
+DiscretePhoton::DiscretePhoton(hid_t group)
+{
+  read_attribute(group, "primary_flag", primary_flag_);
+  read_attribute(group, "energy", energy_);
+  read_attribute(group, "atomic_weight_ratio", A_);
+}
+
+double DiscretePhoton::sample(double E) const
+{
+  if (primary_flag_ == 2) {
```
PullRequestOpen: Can this 2 change on the future ? You may consider pulling this out into a variable or constant.

src/distribution.cpp
```diff
@@ -0,0 +1,312 @@
...
+  return watt_spectrum(a_, b_);
+}
+
+//==============================================================================
+// Normal implementation
+//==============================================================================
+Normal::Normal(pugi::xml_node node) 
+{
+  auto params = get_node_array<double>(node,"parameters");
+  if (params.size() != 2)
```
PullRequestOpen: Nit: The established pattern in these files is to use { } even if only a single statement.

src/distribution.cpp
```diff
@@ -0,0 +1,312 @@
...
+  } else {
+    return x_[0];
+  }
+}
+
+void Discrete::normalize()
+{
+  // Renormalize density function so that it sums to unity
+  double norm = std::accumulate(p_.begin(), p_.end(), 0.0);
+  for (auto& p_i : p_)
```
PullRequestOpen: Nit: The established pattern in these files is to use { } even if only a single statement.

src/distribution.cpp
```diff
@@ -0,0 +1,312 @@
...
+}
+
+//==============================================================================
+// Uniform implementation
+//==============================================================================
+
+Uniform::Uniform(pugi::xml_node node)
+{
+  auto params = get_node_array<double>(node, "parameters");
+  if (params.size() != 2)
```
PullRequestOpen: Nit: The established pattern in these files is to use { } even if only a single statement.

src/distribution.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+UPtrDist distribution_from_xml(pugi::xml_node node)
+{
+  if (!check_for_node(node, "type"))
+    openmc::fatal_error("Distribution type must be specified.");
+
+  // Determine type of distribution
+  std::string type = get_node_value(node, "type", true, true);
+
+  // Allocate extension of Distribution
```
PullRequestOpen: Instead of a helper function, consider using the factory pattern here.

src/distribution.cpp
```diff
@@ -0,0 +1,312 @@
...
+  return normal_variate(mean_value_, std_dev_);
+}
+
+//==============================================================================
+// Muir implementation
+//==============================================================================
+Muir::Muir(pugi::xml_node node) 
+{
+  auto params = get_node_array<double>(node,"parameters");
+  if (params.size() != 3)
```
PullRequestOpen: Nit: The established pattern in these files is to use { } even if only a single statement.

src/eigenvalue.cpp
```diff
@@ -0,0 +1,654 @@
...
+  } else if ((std::abs(kv[1] - kv[2]) / kv[1] < FP_REL_PRECISION) &&
+             (std::abs(cov(1, 1) - cov(2, 2)) / cov(1, 1) < FP_REL_PRECISION)) {
+    // 1 and 2 match, so only use 0 and 1 in our comparisons
+    i = 0;
+    j = 1;
+
+  } else {
+    // No two estimators match, so set i to -1 and this will be the indicator
+    // to use all three estimators.
+    i = -1;
```
PullRequestOpen: Optional / super nit: It's a small thing, but I actually think having a boolean indicator for this state rather than co-opting `i` would make this cleaner.  I have a small fear that setting `i` to this value could result in silent, hard-to-debug issues if this code is later changed.  In addition, I typically prefer to not co-opt values for these cases unless necessary through hardware, memory, or API constraints.

src/eigenvalue.cpp
```diff
@@ -0,0 +1,654 @@
...
+  double p_sample = static_cast<double>(sites_needed) / total;
+
+  simulation::time_bank_sample.start();
+
+  // ==========================================================================
+  // SAMPLE N_PARTICLES FROM FISSION BANK AND PLACE IN TEMP_SITES
+
+  // Allocate temporary source bank
+  int64_t index_temp = 0;
+  std::vector<Particle::Bank> temp_sites(3*simulation::work_per_rank);
```
PullRequestOpen: Question: Are you using `3` as an initial size multiplier as a "size guess" for the potential size of the expected source bank, or does "3" have other significance?  If 3 has special significance, I would note it here for clarity.

src/eigenvalue.cpp
```diff
@@ -0,0 +1,654 @@
...
+    simulation::keff = simulation::k_generation[i];
+  } else {
+    // Sample mean of keff
+    simulation::k_sum[0] += simulation::k_generation[i];
+    simulation::k_sum[1] += std::pow(simulation::k_generation[i], 2);
+
+    // Determine mean
+    simulation::keff = simulation::k_sum[0] / n;
+
+    if (n > 1) {
```
PullRequestOpen: Optional: It may be worth it here to print a error if `n` is not `> 1`.  Note that this case may be 100% covered by the logic on line `113` (in the "synchronize bank logic") in which case it would be less necessar

src/eigenvalue.cpp
```diff
@@ -0,0 +1,654 @@
...
+
+  // Check to see if two estimators are the same; this is guaranteed to happen
+  // in MG-mode with survival biasing when the collision and absorption
+  // estimators are the same, but can theoretically happen at anytime.
+  // If it does, the standard estimators will produce floating-point
+  // exceptions and an expression specifically derived for the combination of
+  // two estimators (vice three) should be used instead.
+
+  // First we will identify if there are any matching estimators
+  int i, j, k;
```
PullRequestOpen: Optional: Since `k` is not used in the if/else block below, it may be clearer to define `k` closer to usage, maybe right above line 159.

src/eigenvalue.cpp
```diff
@@ -0,0 +1,654 @@
...
+  if (sites_outside) {
+    if (mpi::master) warning("Fission source site(s) outside of entropy box.");
+  }
+
+  // sum values to obtain shannon entropy
+  if (mpi::master) {
+    // Normalize to total weight of bank sites
+    p /= xt::sum(p);
+
+    double H = 0.0;
```
PullRequestOpen: Nit: Is `H` a term related to shannon entropy?  Or the result? Might make it more readable if there was a comment here, or a clarifying comment where `H` gets pushed into the vector on line 564.

src/endf.cpp
```diff
@@ -0,0 +1,261 @@
...
+    for (int j = 0; j < n_regions_; ++j) {
+      if (i < nbt_[j]) {
+        interp = int_[j];
+        break;
+      }
+    }
+  }
+
+  // handle special case of histogram interpolation
+  if (interp == Interpolation::histogram) return y_[i];
```
PullRequestOpen: Optional: Given this is short-circuiting the function, it may make more sense to move this up.  I would recommend including this in the `if` block starting on line `170`, specifically as the first condition.

src/endf.cpp
```diff
@@ -0,0 +1,261 @@
...
+    || mt == N_3HEA || mt == N_3P) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+bool is_inelastic_scatter(int mt)
+{
+  if (mt < 100) {
```
PullRequestOpen: Optional: For `100` here, and for `27` on line 69 and `200` on line 71, might make it more readable to define them as constants (similar to `N_2N0`) or leave a comment clarifying their meaning.

src/endf.cpp
```diff
@@ -0,0 +1,261 @@
...
+  // Read coefficients into a vector
+  read_dataset(dset, coef_);
+}
+
+double Polynomial::operator()(double x) const
+{
+  // Use Horner's rule to evaluate polynomial. Note that coefficients are
+  // ordered in increasing powers of x.
+  double y = 0.0;
+  for (auto c = coef_.crbegin(); c != coef_.crend(); ++c) {
```
PullRequestOpen: Optional: If possible, would recommend making `c` const, similar to the for loop on line `140`.

src/error.cpp
```diff
@@ -0,0 +1,137 @@
...
+#endif
+
+  // Write warning
+  std::cerr << " WARNING: ";
+  output(message, std::cerr, 10);
+
+#ifdef _POSIX_VERSION
+  // Reset color for terminal
+  if (isatty(STDERR_FILENO)) {
+    std::cerr << "\033[0m";
```
PullRequestOpen: Thank you for remembering to reset the color! Too many CLI tools forget to do this and leave terminals in an undesirable state.

src/bank.cpp
```diff
@@ -0,0 +1,72 @@
...
+#endif
+}
+
+//==============================================================================
+// C API
+//==============================================================================
+
+extern "C" int openmc_source_bank(void** ptr, int64_t* n)
+{
+  if (simulation::source_bank.size() == 0) {
```
PullRequestOpen: I'll mention it only here, but are NULL-checks missing deliberately? There are reasons that could _arguably_ mean that they aren't required, but I can't find any references in the source code.

src/bank.cpp
```diff
@@ -0,0 +1,72 @@
...
+//==============================================================================
+
+void free_memory_bank()
+{
+  simulation::source_bank.clear();
+  #pragma omp parallel
+  {
+    simulation::fission_bank.clear();
+  }
+#ifdef _OPENMP
```
PullRequestOpen: Is this also required here?: `simulation::secondary_bank.clear();`

src/cmfd_solver.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+extern "C"
+void openmc_initialize_linsolver(const int* indptr, int len_indptr,
+                                 const int* indices, int n_elements, int dim,
+                                 double spectral, const int* cmfd_indices,
+                                 const int* map)
+{
+  // Store elements of indptr
+  for (int i = 0; i < len_indptr; i++)
+    cmfd::indptr.push_back(indptr[i]);
```
PullRequestOpen: Should these be `clear()`'d first, in case user accidentally calls `openmc_initialize_linsolver` twice? Perhaps calling `free_memory_cmfd()` could make sense.

src/cmfd_solver.cpp
```diff
@@ -0,0 +1,361 @@
...
+//==============================================================================
+// SET_INDEXMAP sets the elements of indexmap based on input coremap
+//==============================================================================
+
+void set_indexmap(const int* coremap)
+{
+  for (int z = 0; z < cmfd::nz; z++) {
+    for (int y = 0; y < cmfd::ny; y++) {
+      for (int x = 0; x < cmfd::nx; x++) {
+        if (coremap[(z*cmfd::ny*cmfd::nx) + (y*cmfd::nx) + x] != CMFD_NOACCEL) {
```
PullRequestOpen: This code looks correct, though these two lines could be a little more readable if the index were precomputed, eg:
```c++
        int idx = (z*cmfd::ny*cmfd::nx) + (y*cmfd::nx) + x;
        if (coremap[idx] != CMFD_NOACCEL) {
          int counter = coremap[idx];
```

src/cross_sections.cpp
```diff
@@ -0,0 +1,433 @@
...
+          already_read.insert(element);
+        }
+      }
+
+      // Read multipole file into the appropriate entry on the nuclides array
+      if (settings::temperature_multipole) read_multipole_data(i_nuclide);
+    }
+  }
+
+  for (auto& mat : model::materials) {
```
PullRequestOpen: A short comment would be good here.

src/cross_sections.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+  std::string directory;
+  if (check_for_node(root, "directory")) {
+    // Copy directory information if present
+    directory = get_node_value(root, "directory");
+  } else {
+    // If no directory is listed in cross_sections.xml, by default select the
+    // directory in which the cross_sections.xml file resides
+    auto pos = filename.rfind("/");
+    if (pos == std::string::npos) {
```
PullRequestOpen: A real platform check would be better here.

src/cross_sections.cpp
```diff
@@ -0,0 +1,433 @@
...
+    // No cross_sections.xml file specified in settings.xml, check
+    // environment variable
+    if (settings::run_CE) {
+      char* envvar = std::getenv("OPENMC_CROSS_SECTIONS");
+      if (!envvar) {
+        fatal_error("No cross_sections.xml file was specified in "
+          "materials.xml or in the OPENMC_CROSS_SECTIONS"
+          " environment variable. OpenMC needs such a file to identify "
+          "where to find data libraries. Please consult the"
+          " user's guide at https://openmc.readthedocs.io for "
```
PullRequestOpen: This URL is valid, though the docs are also available at https://docs.openmc.org/, so I'm wondering if the URL should be changed.

src/tallies/filter_mesh.cpp
```diff
@@ -0,0 +1,132 @@
...
+MeshFilter::to_statepoint(hid_t filter_group) const
+{
+  Filter::to_statepoint(filter_group);
+  write_dataset(filter_group, "bins", model::meshes[mesh_]->id_);
+}
+
+std::string
+MeshFilter::text_label(int bin) const
+{
+  auto& mesh = *model::meshes[mesh_];
```
PullRequestOpen: If `mesh_` is not set appropriately, this dereference operation will cause a crash.

src/tallies/filter_mesh.cpp
```diff
@@ -0,0 +1,132 @@
...
+  auto* filt = dynamic_cast<MeshFilter*>(filt_base);
+
+  // Check the filter type.
+  if (!filt) {
+    set_errmsg("Tried to get mesh on a non-mesh filter.");
+    return OPENMC_E_INVALID_TYPE;
+  }
+
+  // Output the mesh.
+  *index_mesh = filt->mesh();
```
PullRequestOpen: Another potential `nullptr`-dereference.

src/tallies/filter_mesh.cpp
```diff
@@ -0,0 +1,132 @@
...
+    fatal_error(err_msg);
+  }
+}
+
+void
+MeshFilter::get_all_bins(const Particle* p, int estimator, FilterMatch& match)
+const
+{
+  if (estimator != ESTIMATOR_TRACKLENGTH) {
+    auto bin = model::meshes[mesh_]->get_bin(p->r());
```
PullRequestOpen: I'll call it out again here, but generally not much of this code does parameter checks within functions, so for example `p` here could be `nullptr`. Since this function will crash if `p == nullptr`, it means that `p` is not optional, and so a reference may be a better choice than a pointer in order to enforce that and to document to developers that this is the case.

src/tallies/filter_particle.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void
+ParticleFilter::to_statepoint(hid_t filter_group) const
+{
+  Filter::to_statepoint(filter_group);
+  std::vector<std::string> particles;
+  for (auto p : particles_) {
+    switch (p) {
+    case Particle::Type::neutron:
```
PullRequestOpen: I've seen this code, or at least very similar code, repeated several times so far, here and I believe in another review. I'm guessing from the Particle class declaration's use of an enum for Type that subclasses aren't desired. Instead, perhaps the following two methods in Particle would make sense:
```c++
static string type_to_str(Type type) const {
    switch (type) {
        case Particle::Type::neutron:     return "neutron";
        case Particle::Type::photon:      return "photon";
        case Particle::Type::electron:    return "electron";
        case Particle::Type::positron:    return "positron";
        default:                          return "";
    }
}

static Type str_to_type(string typeName) const {
    if      (typeName == "neutron")       return Particle::Type::neutron:
    else if (typeName == "photon")        return Particle::Type::photon:
    else if (typeName == "electron")      return Particle::Type::electron:
    else if (typeName == "positron")      return Particle::Type::positron:
    else                                  return Particle::Type::UNKNOWN;
}
```

src/tallies/filter.cpp
```diff
@@ -0,0 +1,249 @@
...
+  // Allocate according to the filter type
+  auto f = Filter::create(s, filter_id);
+
+  // Read filter data from XML
+  f->from_xml(node);
+  return f;
+}
+
+Filter* Filter::create(const std::string& type, int32_t id)
+{
```
PullRequestOpen: It may be cleaner to use a template or a macro instead of repeating the same code. For example
```
template<T> bool tryFilter(const std::string &name, const std::string &type) {
    if(name == type) {
        model::tally_filters.push_back(std::make_unique<T>());
        return(true);
    }
    return(false);
}
Filter* Filter::create(const std::string& type, int32_t id) {
    bool success = tryFilter<AzimuthalFilter>("azimuthal")
                             || tryFilter<CellFilter>("cell")
                             || ...;
    ...
}
```

src/tallies/filter_sptl_legendre.cpp
```diff
@@ -0,0 +1,209 @@
...
+void
+SpatialLegendreFilter::set_axis(LegendreAxis axis)
+{
+  axis_ = axis;
+}
+
+void
+SpatialLegendreFilter::set_minmax(double min, double max)
+{
+  if (max < min) {
```
PullRequestOpen: This should probably be `max <= min`, because we divide by `max-min`, so if `max==min` we will have an error in our calculations.

src/tallies/trigger.cpp
```diff
@@ -0,0 +1,203 @@
...
+get_tally_uncertainty(int i_tally, int score_index, int filter_index)
+{
+  const auto& tally {model::tallies[i_tally]};
+
+  auto sum = tally->results_(filter_index, score_index, RESULT_SUM);
+  auto sum_sq = tally->results_(filter_index, score_index, RESULT_SUM_SQ);
+
+  int n = tally->n_realizations_;
+  auto mean = sum / n;
+  double std_dev = std::sqrt((sum_sq/n - mean*mean) / (n-1));
```
PullRequestOpen: Do we have a guarantee that n > 1? Should we throw an exception if n < 1

src/tallies/trigger.cpp
```diff
@@ -0,0 +1,203 @@
...
+          auto uncert_pair = get_tally_uncertainty(i_tally, score_index,
+            filter_index);
+          double std_dev = uncert_pair.first;
+          double rel_err = uncert_pair.second;
+
+          // Pick out the relevant uncertainty metric for this trigger.
+          double uncertainty;
+          switch (trigger.metric) {
+            case TriggerMetric::variance:
+              uncertainty = std_dev * std_dev;
```
PullRequestOpen: We taking the square root in other code, and then squaring again. We could change the formula to just calculate std_dev**2 directly.

src/tallies/trigger.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (t.n_realizations_ < 2) continue;
+
+    for (const auto& trigger : t.triggers_) {
+      const auto& results = t.results_;
+      for (auto filter_index = 0; filter_index < results.shape()[0];
+           ++filter_index) {
+        for (auto score_index = 0; score_index < results.shape()[1];
+             ++score_index) {
+          // Compute the tally uncertainty metrics.
+          auto uncert_pair = get_tally_uncertainty(i_tally, score_index,
```
PullRequestOpen: We aren't using this if `trigger.metric == TriggerMetric::not_active`. We should probably re-factor the code to only calculate what we need.

src/tallies/trigger.cpp
```diff
@@ -0,0 +1,203 @@
...
+  return {std_dev, rel_err};
+}
+
+//! Find the limiting limiting tally trigger.
+//
+//! param[out] ratio The uncertainty/threshold ratio for the most limiting
+//!   tally trigger
+//! param[out] tally_id The ID number of the most limiting tally
+//! param[out] score The most limiting tally score bin
+
```
PullRequestOpen: It seems the style elsewhere is to have functions take pointers rather than references. We should probably be consistent.

src/tallies/trigger.cpp
```diff
@@ -0,0 +1,203 @@
...
+              break;
+            case TriggerMetric::relative_error:
+              uncertainty = rel_err;
+              break;
+            case TriggerMetric::not_active:
+              uncertainty = 0.0;
+              break;
+          }
+
+          // Compute the uncertainty / threshold ratio.
```
PullRequestOpen: It may be good to check that the threshold isn't set to 0.

src/tallies/filter_azimuthal.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void AzimuthalFilter::set_bins(gsl::span<double> bins)
+{
+  // Clear existing bins
+  bins_.clear();
+  bins_.reserve(bins.size());
+
+  // Copy bins, ensuring they are valid
+  for (gsl::index i = 0; i < bins.size(); ++i) {
```
PullRequestOpen: nit: The established pattern in this file is to use post increment instead of pre increment.

src/tallies/filter_azimuthal.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+  n_bins_ = bins_.size() - 1;
+}
+
+void
+AzimuthalFilter::get_all_bins(const Particle* p, int estimator,
+                              FilterMatch& match) const
+{
+  double phi;
+  if (estimator == ESTIMATOR_TRACKLENGTH) {
```
PullRequestOpen: If x and y are zero, this can throw an error for std::atan2. We probably want to catch and handle this error.

src/tallies/filter_azimuthal.cpp
```diff
@@ -0,0 +1,85 @@
...
+void
+AzimuthalFilter::to_statepoint(hid_t filter_group) const
+{
+  Filter::to_statepoint(filter_group);
+  write_dataset(filter_group, "bins", bins_);
+}
+
+std::string
+AzimuthalFilter::text_label(int bin) const
+{
```
PullRequestOpen: It might be good to check that bin isn't greater than the total number of bins, and if its throw an exception.

src/tallies/filter_cell.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "openmc/cell.h"
+#include "openmc/error.h"
+#include "openmc/xml_interface.h"
+
+namespace openmc {
+
+void
+CellFilter::from_xml(pugi::xml_node node)
+{
+  // Get cell IDs and convert into indices into the global cells vector
```
PullRequestOpen: nit: Should this be: // Get cell IDs and convert indices into the global cells vector

src/tallies/filter_distribcell.cpp
```diff
@@ -0,0 +1,82 @@
...
+    throw std::runtime_error{err_msg.str()};
+  }
+
+  this->set_cell(search->second);
+}
+
+void
+DistribcellFilter::set_cell(int32_t cell)
+{
+  Expects(cell >= 0);
```
PullRequestOpen: Why is expects only used here and not in other error handling? We may want be more consistent.

src/tallies/filter_energy.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+void
+EnergyFilter::set_bins(gsl::span<const double> bins)
+{
+  // Clear existing bins
+  bins_.clear();
+  bins_.reserve(bins.size());
+
+  // Copy bins, ensuring they are valid
+  for (gsl::index i = 0; i < bins.size(); ++i) {
```
PullRequestOpen: Nit:  The established pattern in this commit is to use post increment instead of pre increment.

src/tallies/filter_energy.cpp
```diff
@@ -0,0 +1,174 @@
...
+void
+EnergyFilter::to_statepoint(hid_t filter_group) const
+{
+  Filter::to_statepoint(filter_group);
+  write_dataset(filter_group, "bins", bins_);
+}
+
+std::string
+EnergyFilter::text_label(int bin) const
+{
```
PullRequestOpen: We should probably have error handling to ensure bin is not greater bin_.size() - 2

src/tallies/filter_energyfunc.cpp
```diff
@@ -0,0 +1,161 @@
...
+  if (energy.size() != y.size()) {
+    fatal_error("Energy grid and y values are not consistent");
+  }
+  energy_.clear();
+  energy_.reserve(energy.size());
+  y_.clear();
+  y_.reserve(y.size());
+
+  // Copy over energy values, ensuring they are valid
+  for (gsl::index i = 0; i < energy.size(); ++i) {
```
PullRequestOpen: nit:  established pattern is to use post increment.

src/tallies/filter_material.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+//==============================================================================
+// C-API functions
+//==============================================================================
+
+extern "C" int
+openmc_material_filter_get_bins(int32_t index, const int32_t** bins, size_t* n)
+{
+  // Make sure this is a valid index to an allocated filter.
```
PullRequestOpen: Lines 82-93 seem to be duplicated throughout the code.  Is it possible to refactor these lines out into a common function?  It might be a good opportunity to reduce the size of the code base.

src/tallies/derivative.cpp
```diff
@@ -0,0 +1,687 @@
...
+      break;
+    }
+  }
+}
+
+void score_collision_derivative(const Particle* p)
+{
+  // A void material cannot be perturbed so it will not affect flux derivatives.
+  if (p->material_ == MATERIAL_VOID) return;
+  const Material& material {*model::materials[p->material_]};
```
PullRequestOpen: Add an empty line before the const line for readability

src/tallies/derivative.cpp
```diff
@@ -0,0 +1,687 @@
...
+  }
+
+  if (id <= 0)
+    fatal_error("<derivative> IDs must be an integer greater than zero");
+
+  std::string variable_str = get_node_value(node, "variable");
+
+  if (variable_str == "density") {
+    variable = DIFF_DENSITY;
+
```
PullRequestOpen: Extra line at end of block is unnecessary.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+  : tally_{tally}
+{
+  // Handle the special case for an iterator that points to the end.
+  if (end) {
+    index_ = -1;
+    return;
+  }
+
+  for (auto i_filt : tally_.filters()) {
+    auto& match {simulation::filter_matches[i_filt]};
```
PullRequestOpen: Good usage of `{}-Initialization` for the most vexing parse.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+      // group bin if it exists
+      if (tally.delayedgroup_filter_ >= 0) {
+
+        // Get the index of the delayed group filter
+        auto i_dg_filt = tally.filters()[tally.delayedgroup_filter_];
+
+        const DelayedGroupFilter& dg_filt {*dynamic_cast<DelayedGroupFilter*>(
+          model::tally_filters[i_dg_filt].get())};
+
+        // Loop over delayed group bins until the corresponding bin is found
```
PullRequestOpen: > Loop over delayed group bins until the corresponding bin is found

Potential Bug: Note that the for loop below will continue to loop even after the corresponding bin is found.  If this is not the intent, I would recommend `break`'ing out of the loop at the end of the `if` block on line 421.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+            // Find the element corresponding to the nuclide
+            auto name = data::nuclides[i_nuclide]->name_;
+            std::string element = to_element(name);
+            int i_element = data::element_map[element];
+            auto& heating {data::elements[i_element].heating_};
+            auto i_grid = p->photon_xs_[i_element].index_grid;
+            auto f = p->photon_xs_[i_element].interp_factor;
+            score = std::exp(heating(i_grid) + f * (heating(i_grid+1) -
+              heating(i_grid))) * atom_density * flux;
+          } else {
```
PullRequestOpen: Optional: You can collapse this `else` and the `if` below into a `else if` statement.  E.g.:

```
} else if(p-material_ != MATERIAL_VOID) {
    ...
}
```

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+               * p->neutron_xs_[p->event_nuclide_].fission * flux
+               / p->neutron_xs_[p->event_nuclide_].absorption;
+      }
+    }
+  } else {
+    if (i_nuclide >= 0) {
+      const Nuclide& nuc {*data::nuclides[i_nuclide]};
+      return get_nuc_fission_q(nuc, p, score_bin) * atom_density * flux
+             * p->neutron_xs_[i_nuclide].fission;
+    } else {
```
PullRequestOpen: Optional: You can collapse this `else` block and the `if` block below into a single `else if` block.   E.g.,:


```
} else if (p-material_ != MATERIAL_VOID) {
  ...
  return score * flux;
}
return 0.0
```

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+          // number of particles that were banked in the fission bank. Since
+          // this was weighted by 1/keff, we multiply by keff to get the proper
+          // score. Loop over the neutrons produced from fission and check which
+          // ones are delayed. If a delayed neutron is encountered, add its
+          // contribution to the fission bank to the score.
+          score = 0.;
+          for (auto i = 0; i < p->n_bank_; ++i) {
+            auto i_bank = simulation::fission_bank.size() - p->n_bank_ + i;
+            const auto& bank = simulation::fission_bank[i_bank];
+            auto g = bank.delayed_group;
```
PullRequestOpen: Optional: I would use another variable name than `g` to make it more readable.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+                    * flux;
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+
+    // Add derivative information on score for differnetial tallies.
```
PullRequestOpen: Spelling: Did you mean "differential" rather than "differnetial"?

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+  // sites, we need to multiply the score by keff to get the true nu-fission
+  // rate. Otherwise, the sum of all nu-fission rates would be ~1.0.
+
+  // loop over number of particles banked
+  for (auto i = 0; i < p->n_bank_; ++i) {
+    auto i_bank = simulation::fission_bank.size() - p->n_bank_ + i;
+    const auto& bank = simulation::fission_bank[i_bank];
+
+    // get the delayed group
+    auto g = bank.delayed_group;
```
PullRequestOpen: optional: Rename variable `delayed_group` or `group` rather than `g`.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+          } else {
+            // If the delayed group filter is not present, compute the score
+            // by multiplying the delayed-nu-fission macro xs by the flux
+            score = p->neutron_xs_[i_nuclide].fission
+              * data::nuclides[i_nuclide]
+              ->nu(E, ReactionProduct::EmissionMode::delayed)
+              * atom_density * flux;
+          }
+        } else {
+          if (tally.delayedgroup_filter_ != C_NONE) {
```
PullRequestOpen: Optional: It's not 100% clear from context what happens when `i_nuclide >= 0`.  Would recommend adding a comment here describing this condition and when it's expected to be executed.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+  }
+  return 0.0;
+}
+
+//! Helper function to obtain the kerma coefficient for a given nuclide
+
+double get_nuclide_neutron_heating(const Particle* p, const Nuclide& nuc,
+    int rxn_index, int i_nuclide)
+{
+  size_t mt = nuc.reaction_index_[rxn_index];
```
PullRequestOpen: Optional: Might make this a bit more readable if the individual conditions were separated by a newline.  E.g.:

```
  size_t mt = nuc.reaction_index_[rxn_index];
  if (mt == C_NONE) return 0.0;

  auto i_temp = p->neutron_xs_[i_nuclide].index_temp;
  if (mt == C_NONE) return 0.0;

  auto i_temp = p->neutron_xs_[i_nuclide].index_temp;
  if (i_temp < 0) return 0.0; // Can be true due to multipole

  const auto& rxn {*nuc.reactions_[mt]};
  const auto& xs {rxn.xs_[i_temp]};
  auto i_grid = p->neutron_xs_[i_nuclide].index_grid;
  if (i_grid < xs.threshold) return 0.0;

  auto f = p->neutron_xs_[i_nuclide].interp_factor;
  return (1.0 - f) * xs.value[i_grid-xs.threshold] + f * xs.value[i_grid-xs.threshold+1];
```

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+          if (i_temp >= 0) { // Can be false due to multipole
+            auto i_grid = p->neutron_xs_[i_nuclide].index_grid;
+            auto f = p->neutron_xs_[i_nuclide].interp_factor;
+            const auto& xs {rxn.xs_[i_temp]};
+            if (i_grid >= xs.threshold) {
+              score = ((1.0 - f) * xs.value[i_grid-xs.threshold]
+                + f * xs.value[i_grid-xs.threshold+1]) * atom_density * flux;
+            }
+          }
+        } else {
```
PullRequestOpen: Optional: This `else` block can be combined with the `if` block below, to slightly simplify the code.

```
} else if (p->material_ != MATERIAL_VOID) {
  ...
}
```

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+  this->compute_index_weight();
+}
+
+FilterBinIter&
+FilterBinIter::operator++()
+{
+  // Find the next valid combination of filter bins.  To do this, we search
+  // backwards through the filters until we find the first filter whose bins
+  // can be incremented.
+  bool done_looping = true;
```
PullRequestOpen: Optional: As an alternative to using the `done_looping` flag, your could declare the `i` index outside the for loop and check that variable in the `if` block below.  This would decrease the amount of state you have to track.  Renaming it would make it more descriptive as well.  E.g.:

```
int tally_index = tally_.filters().size() - 1;
for (; tally_index >= 0; tally_index--) {
   ...
}

// Do this instead of checking `done_looping`
if (tally_index == 0) {
   ...
}
```

If you choose not to do the above, would also recommend choosing a variable name more tied to the business logic than the procedural logic.  Maybe instead of `done_looping`, use `checked_all_tallies` or `visited_every_combination`.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+        } else {
+          // Calculate photon heating cross section on-the-fly
+          if (i_nuclide >= 0) {
+            // Find the element corresponding to the nuclide
+            auto name = data::nuclides[i_nuclide]->name_;
+            std::string element = to_element(name);
+            int i_element = data::element_map[element];
+            auto& heating {data::elements[i_element].heating_};
+            auto i_grid = p->photon_xs_[i_element].index_grid;
+            auto f = p->photon_xs_[i_element].interp_factor;
```
PullRequestOpen: Optional: Recommend better variable than `f`.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+//! Helper function to score fission energy
+//
+//! Pulled out to support both the fission_q scores and energy deposition
+//! score
+
+double score_fission_q(const Particle* p, int score_bin, const Tally& tally,
+  double flux, int i_nuclide, double atom_density)
+{
+  if (tally.estimator_ == ESTIMATOR_ANALOG) {
+    const Nuclide& nuc {*data::nuclides[p->event_nuclide_]};
```
PullRequestOpen: Thank you for dereferencing this, and turning it into a reference const variable.  Makes it significantly more readable.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+  #pragma omp atomic
+  tally.results_(filter_index, score_index, RESULT_VALUE) += score;
+
+  // Reset the original delayed group bin
+  dg_match.bins_[i_bin] = original_bin;
+}
+
+//! Helper function to retrieve fission q value from a nuclide
+
+double get_nuc_fission_q(const Nuclide& nuc, const Particle* p, int score_bin)
```
PullRequestOpen: Optional: Recommend renaming the variable `p` to `particle` for readability, here and throughout.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+      // times the yield as the estimate for the number of neutrons exiting a
+      // reaction with neutrons in the exit channel
+      if (p->event_mt_ == ELASTIC || p->event_mt_ == N_LEVEL ||
+        (p->event_mt_ >= N_N1 && p->event_mt_ <= N_NC)) {
+        // Don't waste time on very common reactions we know have
+        // multiplicities of one.
+        score = p->wgt_last_ * flux;
+      } else {
+        // Get yield and apply to score
+        auto m =
```
PullRequestOpen: Optional: Recommend another variable name other than `m`.  It's hyper local (used on the next line), but it's not 100% clear what the variable has from context, which can open the code up to subtle bugs when refactoring.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+              const Material& material {*model::materials[p->material_]};
+              for (auto i = 0; i < material.nuclide_.size(); ++i) {
+                auto j_nuclide = material.nuclide_[i];
+                auto atom_density = material.atom_density_(i);
+                const auto& nuc {*data::nuclides[j_nuclide]};
+                if (nuc.fissionable_) {
+                  const auto& rxn {*nuc.fission_rx_[0]};
+                  // Tally each delayed group bin individually
+                  for (auto d_bin = 0; d_bin < filt.n_bins(); ++d_bin) {
+                    auto d = filt.groups()[d_bin];
```
PullRequestOpen: Optional: I would use another variable name than `d` to make it more readable.

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+        }
+      } else {
+        if (i_nuclide >= 0) {
+          const auto& nuc {*data::nuclides[i_nuclide]};
+          if (nuc.fissionable_) {
+            const auto& rxn {*nuc.fission_rx_[0]};
+            score = rxn.q_value_ * p->neutron_xs_[i_nuclide].fission
+              * atom_density * flux;
+          }
+        } else {
```
PullRequestOpen: Optional: You could simplify this code slightly by collapsing the `else` and `if` blocks.  E.g.:

```
} else if (p->material_ != MATERIAL_VOID) {
  ...
}
```

src/tallies/tally_scoring.cpp
```diff
@@ -0,0 +1,2347 @@
...
+        double atom_density {material.atom_density_(i)};
+        const Nuclide& nuc {*data::nuclides[j_nuclide]};
+        heating_xs += atom_density * get_nuclide_neutron_heating(p, nuc, rxn_bin, j_nuclide);
+      }
+      if (tally.estimator_ == ESTIMATOR_ANALOG) {
+        heating_xs /= p->macro_xs_.total;
+      }
+    }
+  }
+  score = heating_xs * flux;
```
PullRequestOpen: `heating_xs` can be uninitialized at this point if `i_nuclide` is less than 0 and `p->material_ == MATERIAL_VOID` it could be worth either handling this case or adding an explicit failure case in if this is unexpected to ever happen

src/dagmc.cpp
```diff
@@ -0,0 +1,406 @@
...
+void check_dagmc_file() {
+  std::string filename = settings::path_input + DAGMC_FILENAME;
+  if (!file_exists(filename)) {
+    fatal_error("Geometry DAGMC file '" + filename + "' does not exist!");
+  }
+}
+
+bool get_uwuw_materials_xml(std::string& s) {
+  check_dagmc_file();
+  UWUW uwuw((settings::path_input + DAGMC_FILENAME).c_str());
```
PullRequestOpen: You are checking to see if the file exists on the path listed in line 57, but then you use the path listed on line 65. Consider adding a parameter in the check_dagmc_file procedure that takes the file name you want to check as a parameter.  Otherwise, you could wind up checking one file and then trying to load up another (if someone changes the path on one line and not the other).

src/dagmc.cpp
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  return uwuw_mats_present;
+}
+
+bool read_uwuw_materials(pugi::xml_document& doc) {
+  std::string s;
+  bool found_uwuw_mats = get_uwuw_materials_xml(s);
+  if (found_uwuw_mats) {
+    pugi::xml_parse_result result = doc.load_string(s.c_str());
```
PullRequestOpen: Did you mean to ignore the result on line 89?  If so maybe a comment here could help future readers of the code.

src/dagmc.cpp
```diff
@@ -0,0 +1,406 @@
...
+void load_dagmc_geometry()
+{
+  check_dagmc_file();
+
+  if (!model::DAG) {
+    model::DAG = new moab::DagMC();
+  }
+
+
+  std::string filename = settings::path_input + DAGMC_FILENAME;
```
PullRequestOpen: Same comment as earlier about the file name.

Issue #1443, opened 2020-01-07
duplicate identical surfaces in different regions sometimes lead to max_event errors


eepeterson:
I've noticed that sometimes if you create two different non-connected Regions that share a surface, but you form them with two identical, but distinct Surfaces (with different ids) it can lead to max_event errors. The geometry this happened in was relatively complicated and I have yet to be able to reproduce the same issue in a simpler case, but the problem is resolved if you remove the redundant surface in the geometry.xml file. I think it is some type of infinite recursion issue with redundant surfaces, but I can't tell for sure. 

I think I'd rather have a solution that removes all redundant surfaces when the geometry is exported to xml rather than ensure that I don't create any in the model building process which would put some constraints on the modularity of the model building process. I would think it might also help with performance in terms of keeping only the minimum necessary set of surfaces in the geometry.

My first idea is to write a Geometry.clean method that can be toggled in the export_to_xml method that searches for redundant surfaces and changes all the surface id references to single surface and deletes the others. Any thoughts or suggestions are greatly appreciated.

RossWorrall:
I've had this issue pop up in a geometry with two duplicate z-cones and z-planes. The geometry was converted from CAD-to-CSG and happened to have two z cone surfaces (ZConeOneSided) but on opposite sides of the apex, which OpenMC replaced in the xml with with the same z-cones and z-planes across the apex.

I receive lots of "WARNING: Particle 8750001 underwent maximum number of events." despite all surfaces in the cell being shared with the cell surrounding.

I'm using 0.15.4-dev

To fix automatically I have to use the method geometry.remove_redundant_surfaces().

The below shows the geometry which raises the warning

```xml
<?xml version='1.0' encoding='UTF-8'?>
<geometry>
  <cell id="93" material="void" region="(337 | -338) (-339 -340) (341 | 342) -343 -344 (341 | 342) (345 | 346)" universe="24"/> 
  <cell id="94" material="void" region="347 -348 349 -350 351 -352 ((-337 338) | (339 | 340) | (-341 -342) | 343 | 344 | (-341 -342) | (-345 -346))" universe="24"/>
    <cell id="95" material="void" region="(-347 | 348 | -349 | 350 | -351 | 352) -353" universe="24"/>
  <cell id="96" material="void" region="353" universe="24"/>
  <surface id="337" type="z-cone" coeffs="0.0 0.0 652.49434672 171.40699659829093"/>
[...]

PR #1444, opened 2020-01-08
Removing duplicate surfaces from geometry

eepeterson:
I just wanted to open a PR to let people see what I was thinking about solving issue #1443. By adding a key and hash method to the surface class we can compare surfaces easily for topological equality by comparing, surface type, boundary_type and coefficients and throwing out the name and id number. This means you can take sets of surfaces to get only the unique ones. Then it should be pretty straight-forward to check surfaces in a universe for equality and reduce them to a single surface by replacing the duplicated surfaces in a Regions node list.

I still need to implement the duplicate checking in a universe or geometry, but I wanted to get feedback before going further.

### DIFF
--- a/openmc/geometry.py
+++ b/openmc/geometry.py
@@ -77,15 +77,22 @@ class Geometry(object):
                 if universe.id in volume_calc.volumes:
                     universe.add_volume_information(volume_calc)
 
-    def export_to_xml(self, path='geometry.xml'):
+    def export_to_xml(self, path='geometry.xml', remove_surfs=False):
         """Export geometry to an XML file.
 
         Parameters
         ----------
         path : str
             Path to file to write. Defaults to 'geometry.xml'.
+        remove_surfs : bool
+            Whether or not to remove redundant surfaces from the geometry when
+            exporting
 
         """
+        # Find and remove redundant surfaces from the geometry
+        if remove_surfs:
+            self.remove_redundant_surfaces()
+
         # Create XML representation
         root_element = ET.Element("geometry")
         self.root_universe.create_xml_subelement(root_element, memo=set())
@@ -382,6 +389,26 @@ class Geometry(object):
             surfaces = cell.region.get_surfaces(surfaces)
         return surfaces
 
+    def get_redundant_surfaces(self):
+        """Return all of the topologically redundant surface ids
+
+        Returns
+        -------
+        dict
+            Dictionary whose keys are the ID of a redundant surface and whose
+            values are the topologically equivalent :class:`openmc.Surface`
+            that should replace it.
+
+        """
+        tally = defaultdict(list)
+        for surf in self.get_all_surfaces().values():
+            coeffs = tuple(surf._coefficients[k] for k in surf._coeff_keys)
+            key = (surf._type,) + coeffs
+            tally[key].append(surf)
+        return {replace.id: keep 
+                for keep, *redundant in tally.values()
+                for replace in redundant}
+
     def get_materials_by_name(self, name, case_sensitive=False, matching=False):
         """Return a list of materials with matching names.
 
@@ -578,6 +605,17 @@ class Geometry(object):
                 lattices.add(lattice)
 
         return sorted(lattices, key=lambda x: x.id)
+
+    def remove_redundant_surfaces(self):
+        """Remove redundant surfaces from the geometry"""
+
+        # Get redundant surfaces
+        redundant_surfaces = self.get_redundant_surfaces()
+
+        # Iterate through all cells contained in the geometry 
+        for cell in self.get_all_cells().values():
+            # Recursively remove redundant surfaces from regions
+            cell.region.remove_redundant_surfaces(redundant_surfaces)
 
     def determine_paths(self, instances_only=False):
         """Determine paths through CSG tree for cells and materials.
--- a/openmc/surface.py
+++ b/openmc/surface.py
@@ -2209,6 +2209,21 @@ class Halfspace(Region):
         surfaces[self.surface.id] = self.surface
         return surfaces
 
+    def remove_redundant_surfaces(self, redundant_surfaces):
+        """Recursively remove all redundant surfaces referenced by this region
+
+        Parameters
+        ----------
+        redundant_surfaces : dict
+            Dictionary mapping redundant surface IDs to surface IDs for the 
+            :class:`openmc.Surface` instances that should replace them.
+
+        """
+
+        surf = redundant_surfaces.get(self.surface.id)
+        if surf is not None:
+            self.surface = surf
+
     def clone(self, memo=None):
         """Create a copy of this halfspace, with a cloned surface with a
         unique ID.
--- a/tests/unit_tests/test_geometry.py
+++ b/tests/unit_tests/test_geometry.py
@@ -301,3 +301,50 @@ def test_rotation_matrix():
     assert geom.find((0.0, 0.5, 0.0))[-1] == c3
     assert geom.find((0.0, -0.5, 0.0))[-1] == c1
     assert geom.find((0.0, -1.5, 0.0))[-1] == c2
+
+def test_remove_redundant_surfaces():
+    """Test ability to remove redundant surfaces"""
+
+    m1 = openmc.Material()
+    m1.add_nuclide('U235', 1.0, 'wo')
+    m1.add_nuclide('O16', 2.0, 'wo')
+    m1.set_density('g/cm3', 10.0)
+
+    m2 = openmc.Material()
+    m2.add_element('Zr', 1.0)
+    m2.set_density('g/cm3', 2.0)
+
+    m3 = openmc.Material()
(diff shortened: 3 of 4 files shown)


pshriwise:
I like where you're headed w/ it @eepeterson. What you have here leads quickly to at least a report of duplicate surfaces when calling `Geometry.export_to_xml` with just a few lines in that method. 

I'll think about it a bit more, but nothing major jumps out at me initially.

openmc/surface.py
```diff
@@ -94,6 +94,22 @@ def __repr__(self):
 
         return string
 
+    def __key(self):
+        return (self.type, self.boundary_type) + \
+                tuple(j for i in self.coefficients.items() for j in i)
```
drewejohnson: For this method, we want the order of the coefficients to be consistent, which is not ensured with dictionaries unless care is taken. Thankfully, when Surfaces are constructed, the coefficients dictionary is populated in some deterministic manner ([Sample from `Cone` base class](https://github.com/openmc-dev/openmc/blob/55ca9fe7f81e271e6167aea98677f2d2660eee6b/openmc/surface.py#L1624-L1627)), which can ensure insertion order and iteration as well.  But it may be possible that someone populates this dictionary arbitrarily.
```python
>>> foo = {"x": 2, "y": 4}
>>> bar = {"y": 4, "x": 2}
```
If these dictionaries represented the coefficients for identical surface types, they _should_ hash as identical, but due to the dictionary ordering will be hashed differently. 
```python
>>> list(foo.items())
[("x", 2), ("y", 4)]
>>> list(bar.items())
[("y", 4), ("x", 2)]
```
For robustness, I would recommend using a sorted iteration
```python
>>> list(sorted(bar.items()))
[("x", 2), ("y", 4)]
```

eepeterson: Good point, thanks.

openmc/surface.py
```diff
@@ -94,6 +94,22 @@ def __repr__(self):
...
+
+    def __hash__(self):
+        return hash(self.__key())
+
+    def __eq__(self, other):
+        if isinstance(other, type(self)):
+            return self.__key() == other.__key()
```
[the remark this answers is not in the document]
eepeterson: Thanks for the link and comments @drewejohnson. Would you rather see the key essentially generated in the __hash__ method and the __key method removed entirely? Alternatively I could change __key to _key so that it doesn't invoke the name mangling algorithm.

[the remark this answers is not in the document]
drewejohnson: I think it would be better to have the hash method not rely on `_key`, especially if `__hash__` is the only method that uses `_key`.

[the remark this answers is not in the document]
eepeterson: Sounds good.

eepeterson:
Thanks for the review and the tips @drewejohnson. I'll work on the test case now.

openmc/surface.py
```diff
@@ -94,6 +94,23 @@ def __repr__(self):
 
         return string
 
+    def _key(self):
+        coeffs = list(sorted(self.coefficients.items()))
+        coeff_tup = tuple(j for i in coeffs for j in i)
+        return (self.type, self.boundary_type) + coeff_tup
+
+    def __hash__(self):
+        return hash(self._key())
```
paulromano: This violates an assumption that Python makes about hashable objects, which is that the [hash value is immutable](https://docs.python.org/3/reference/datamodel.html#object.__hash__). Because our surface classes are mutable, that means this key can change, and hence the hash value can change. This can result in very strange behavior, such as the following:
```Python
>>> s1 = openmc.Sphere()
>>> s2 = openmc.Sphere()
>>> collection = {s1, s2}
>>> s1 in collection
True
>>> s1.r = 3
>>> s1 in collection
False
```

eepeterson: Ah, bummer. Would an acceptable solution be to remove the __hash__ method and compare __repr__ for equality with the name and ID fields stripped or go back to a _key variable and compare those keys for equality?

paulromano: Yes, something like that would work. Removing `__hash__` means that you won't be able to use a surface object in a set or as the key in a dictionary. Additionally removing `__eq__` means that you would be able to use them in a set/dict, but the hash value is essentially just based on the memory location of the object, which means that two objects that are the same other than their ID will not go to the same dict key. But, you can probably get by just using the ID as the key in a dict.

eepeterson: @paulromano I've removed the `__hash__` method for the `Surface` class, but that means I needed to remove `__eq__` and  `__ne__` as well since defining those without defining `__hash__` means the `Surface` can't be used as a key in a set which breaks the memoization when translating to xml elements. Unfortunately, this makes it take a lot longer to find all the duplicates, since it requires multiple iterations through all the surfaces, but maybe I'm missing something that could simplify this.

eepeterson: Actually, I could implement `__eq__` as long as `__hash__` is explicitly set to the parent class `__hash__` method - in the case of `Surface` I would assume that would be the `IDManagerMixin` class?

openmc/surface.py
```diff
@@ -94,19 +94,18 @@ def __repr__(self):
...
+        coefficients = '{0: <16}'.format('\tCoefficients') + '\n'
 
-    def __ne__(self, other):
-        return not self == other
+        for coeff, value in list(sorted(self.coefficients.items())):
+            coefficients += '{0: <16}{1}{2}\n'.format(coeff, '=\t', value)
+
+        string += coefficients
+
+        return string
```
drewejohnson: If we make this public, something like `Surface.description`, we can use this as the key in the dictionary from the previous implementation. This gives us the one-pass approach from before, without relying on `__hash__` for Surfaces. My bad for not catching that before...

eepeterson: Ah, good call. I don't know why I didn't think of that.

openmc/geometry.py
```diff
@@ -382,6 +389,24 @@ def get_all_surfaces(self):
...
+        dict
+            Dictionary whose keys are the ID of a redundant surface and whose
+            values are the topologically equivalent :class:`openmc.Surface`
+            that should replace it.
+
+        """
+        tally = defaultdict(list)
+        surfaces = self.get_all_surfaces()
+        for sid, surf in surfaces.items():
+            tally[surf.description].append(sid)
```
paulromano: All we really need for the key here is something that uniquely identifies the surface. To me, this `description` attribute is kind of overkill for what's needed. My suggestion would be to have the key just be a tuple containing the type of the surface and the coefficients. Something like:
```Python
key = (surf._type,) + tuple(surf._coefficients[k] for k in surf._coeff_keys)
```
As an example, for a sphere with radius 3.0, this would produce the tuple `('sphere', 0.0, 0.0, 0.0, 3.0)`. I don't think we really need boundary info because if your geometry has two equivalent surfaces with _different_ boundary conditions, you're really in for trouble already.

openmc/geometry.py
```diff
@@ -382,6 +389,24 @@ def get_all_surfaces(self):
...
+        -------
+        dict
+            Dictionary whose keys are the ID of a redundant surface and whose
+            values are the topologically equivalent :class:`openmc.Surface`
+            that should replace it.
+
+        """
+        tally = defaultdict(list)
+        surfaces = self.get_all_surfaces()
+        for sid, surf in surfaces.items():
```
paulromano: For clarity:
```suggestion
        for surf_id, surf in surfaces.items():
```

openmc/geometry.py
```diff
@@ -382,6 +389,24 @@ def get_all_surfaces(self):
...
+            tally[surf.description].append(sid)
+        duplist = [(ids[0], ids) for ids in tally.values() if len(ids) > 1]
+        return {k : surfaces[sid] for sid, dups in duplist for k in dups[1:]}
```
paulromano: Ok, here's a suggested reworking of this:
```Python
for surf in self.get_all_surfaces().values():
    coeffs = tuple(self._coefficients[k] for k in surf._coeff_keys)
    key = (surf._type,) + coeffs
    tally[key].append(surf)
return {replace.id: keep
    for keep, *redundant in tally.values()
    for replace in redundant}
```

paulromano:
@eepeterson The implementation looks good now :+1: The last thing this PR needs is a unit test.

eepeterson:
@paulromano Sounds good. I added the unit test.

tests/unit_tests/test_geometry.py
```diff
@@ -301,3 +301,54 @@ def test_rotation_matrix():
...
+    assert n_redundant_surfs == 6
+    # Remove redundant surfaces
+    geom.remove_redundant_surfaces()
+    # There should be 0 remaining redundant surfaces
+    n_redundant_surfs = len(geom.get_redundant_surfaces().keys())
+    assert n_redundant_surfs == 0
+
+
+
+
```
drewejohnson: Remove the extra lines

eepeterson: done.

paulromano:
Nice work @eepeterson!

PR #1451, opened 2020-01-14
Preparation for event-based parallelism (Thread private variable removal)

jtramm:
This PR makes two changes for preparing OpenMC for a future PR to introduce event-based parallelism.

## Removal of thread private variables

Threadprivate variables preclude the use of an event-based model in OpenMC. This is because in the event-based model, parallelism is expressed over all particles undergoing a single event, meaning that a given particle history may be processed by any number of different threads. If some of the particle state variables are owned by the threads themselves, this will result in undefined/incorrect results. For things to work correctly in the event-based model, all particle state variables must be owned by the particle's themselves so that they can be processed by different threads.

Most variables that are currently threadprivate are transferred from thread ownership to particle ownership, with the exception of the fission bank. Instead, the fission bank is converted into a regular array with a corresponding length integer, both of which are shared by all OpenMP threads. Over the course of a power iteration as various threads want to append fission sites to the fission bank, they will atomically capture and increment the length integer variable and store their data at the captured index position, e.g.:

```C++
int idx;
#pragma omp atomic capture
idx = simulation::fission_bank_length++;
Particle::Bank* site = simulation::fission_bank + idx;
```

While the fission bank could also have been made into a particle owned vector, looking forward to GPU we will probably want to use a shared fission bank, so we felt this change was best to make now rather than changing it again down the line.

There is also the addition of a sorting step at the end of each iteration, where the shared fission bank is (stable) sorted in order of each fission site's parental ID. This is done to ensure reproducibility regardless of what order the particles contribute to the bank in. In the future, we may want to use C++17 parallel sorting to accomplish this, but current the cost of the serial sort appears to be trivial as it only occures once per iteration.

**This PR is not meant as an improvement for the CPU version of OpenMC -- these changes are very much lateral.** However, removal of thread private variables will allow for OpenMC to support an event-based model, which we believe will be critical for performance on GPU. Once this PR is accepted, a second PR will be made that introduces an alternative option for event-based parallelism on CPU (with GPU support coming down the line). The changes made in this PR will allow for the event-based mode to be introduced by only adding a few additional functions and without any structural changes to be made in OpenMC. While we don't believe event-based parallelism will be very useful for CPU-based architectures, it will be helpful to have it in so as to minimize the code delta between the CPU- and GPU-based versions.

Performance wise, the change from thread to particle ownership of variables should not affect performance in any meaningful way. However, the change to the new shared fission bank method does introduce a significant increase in the number of OpenMP atomic operations, and does introduce some extra overhead due to the stable sorting step. Thus, to ensure we aren't losing any performance from this PR, I plan on doing some empirical testing and will post the results shortly. Early results so far show that the shared fission bank does not appear to cause any significant performance loss even on highly threaded systems.

## Particle history() split up

One small additional change made in this PR is the breaking up of the particle history function into a set of smaller functions, with control flow transferred to simulation.c. These changes are made possible by the elimination of thread private variables. Similar to removal of threadprivate variables, these changes are made to prepare the code for an event-based parallelism option. These changes are included in this PR as I made them concurrently, and found it would be more complicated to unwind them.


jtramm:
I collected some performance data on a dual socket Xeon 8180 CPU (56 cores, 112 threads), when run with 1 MPI rank with OpenMP using all 112 threads. I tested two problems, the standard pincell example (/examples/xml/pincell) and the depleted SMR core from the ECP benchmark suite. For both cases, I ran using 10 inactive + 10 active batches, all with 100k particles per batch. I used gcc 5.5, and ran each configuration 10 times and took the average.

|        | Pincell       |      | SMR | |
| ------------- |:-------------:| -----:|:-------------:|:-------------:|
|        | Inactive           | Active  | Inactive | Active |
| Develop      | 432,921 | 28,166 | 188,693 | 15,250 |
| This PR      | 347,731 | 29,587 | 189,468 | 16,769 |

Overall, it looks like we get a small boost in active cycle performance for both cases, though there is a fairly significant loss in inactive cycle performance for the small pincell case. Not sure what could be causing that, but it doesn't seem to show up for the larger SMR case.

src/particle.cpp
```diff
@@ -129,32 +130,8 @@ Particle::from_source(const Bank* src)
...
-  }
-
-  // Prepare to write out particle track.
-  if (write_track_) add_particle_track();
-
-  // Every particle starts with no accumulated flux derivative.
-  if (!model::active_tallies.empty()) zero_flux_derivs();
-
-  while (true) {
     // Set the random number stream
```
paulromano: The indentation in this function needs to come down two spaces.

src/simulation.cpp
```diff
@@ -570,4 +558,30 @@ void free_memory_simulation()
   simulation::entropy.clear();
 }
 
+void transport_history_based_single_particle(Particle& p)
+{
+  while (true) {
+    p.event_calculate_xs();
+    p.event_advance();
+    if( p.collision_distance_ > p.boundary_.distance )
```
paulromano:
```suggestion
    if (p.collision_distance_ > p.boundary_.distance) {
```

src/simulation.cpp
```diff
@@ -570,4 +558,30 @@ void free_memory_simulation()
...
 }
 
+void transport_history_based_single_particle(Particle& p)
+{
+  while (true) {
+    p.event_calculate_xs();
+    p.event_advance();
+    if( p.collision_distance_ > p.boundary_.distance ) 
+      p.event_cross_surface();
+    else
```
paulromano:
```suggestion
    } else {
```

src/simulation.cpp
```diff
@@ -570,4 +558,30 @@ void free_memory_simulation()
...
+      p.event_collide();
```
paulromano:
```suggestion
      p.event_collide();
    }
```

src/simulation.cpp
```diff
@@ -570,4 +558,30 @@ void free_memory_simulation()
...
+    p.event_revive_from_secondary();
+    if(!p.alive_)
```
paulromano:
```suggestion
    if (!p.alive_)
```

src/tallies/tally_scoring.cpp
```diff
@@ -55,8 +55,8 @@ FilterBinIter::FilterBinIter(const Tally& tally, const Particle* p)
   this->compute_index_weight();
 }
 
-FilterBinIter::FilterBinIter(const Tally& tally, bool end)
-  : tally_{tally}
+FilterBinIter::FilterBinIter(const Tally& tally, bool end, std::vector<FilterMatch>* particle_filter_matches)
```
paulromano: This line is getting a little long. Can you bring that last argument down to a new line?

src/tallies/tally_scoring.cpp
```diff
@@ -144,12 +144,12 @@ FilterBinIter::compute_index_weight()
 //! Helper function used to increment tallies with a delayed group filter.
 
 void
-score_fission_delayed_dg(int i_tally, int d_bin, double score, int score_index)
+score_fission_delayed_dg(int i_tally, int d_bin, double score, int score_index, std::vector<FilterMatch>& filter_matches)
```
paulromano: Same thing here, need a new line

src/particle_restart.cpp
```diff
@@ -100,8 +102,26 @@ void run_particle_restart()
...
+  if (settings::run_CE) {
+    for (auto& micro : p.neutron_xs_) micro.last_E = 0.0;
+  }
+
+  // Prepare to write out particle track.
+  if (p.write_track_) add_particle_track(p);
+
+  // Every particle starts with no accumulated flux derivative.
+  if (!model::active_tallies.empty())
+  {
```
paulromano: Combine with previous line

src/particle_restart.cpp
```diff
@@ -100,8 +102,26 @@ void run_particle_restart()
...
+    p.flux_derivs_.resize(model::tally_derivs.size(), 0.0);
+    std::fill(p.flux_derivs_.begin(), p.flux_derivs_.end(), 0.0);
+  }
+  
+  // Allocate space for tally filter matches
+  p.filter_matches_.resize(model::tally_filters.size());
```
paulromano: The additions here are things that appear exactly as in `initialize_history`. Maybe we should separate them out into a function somewhere?

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+            "in this generation will not be banked.");
+        #pragma omp atomic write
+        simulation::fission_bank_length = simulation::fission_bank_max;
+        skipped++;
+        break;
+      }
+      site = simulation::fission_bank + idx;
+    }
+    else
+    {
```
paulromano: Combine braces on the else line: `} else {`

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
-    site.wgt = 1. / weight;
+  // Determine whether to place fission sites into the shared fission bank
+  // or the secondary particle bank.
+  bool use_fission_bank = false;
+  if (settings::run_mode == RUN_MODE_EIGENVALUE)
+    use_fission_bank = true;
+
+  for (int i = 0; i < nu; ++i) {
+    Particle::Bank* site;
+    if(use_fission_bank)
```
paulromano:
```suggestion
    if (use_fission_bank) {
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+    {
```
paulromano:
```suggestion
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+      int64_t idx;
+      #pragma omp atomic capture
+      idx = simulation::fission_bank_length++;
+      if( idx >= simulation::fission_bank_max )
```
paulromano:
```suggestion
      if (idx >= simulation::fission_bank_max) {
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+      {
```
paulromano:
```suggestion
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
 
-    // Bank source neutrons by copying the particle data
-    site.r = p->r();
-    site.particle = Particle::Type::neutron;
-    site.wgt = 1. / weight;
+  // Determine whether to place fission sites into the shared fission bank
+  // or the secondary particle bank.
+  bool use_fission_bank = false;
+  if (settings::run_mode == RUN_MODE_EIGENVALUE)
+    use_fission_bank = true;
```
paulromano: This could be
```C++
bool use_fission_bank = (settings::run_mode == RUN_MODE_EIGENVALUE)
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
     if (p->delayed_group_ > 0) {
       nu_d[p->delayed_group_-1]++;
     }
+    
+    // Write fission particles to nuBank
+    if(use_fission_bank)
+    {
+      Particle::NuBank* nu_bank_entry;
+      p->nu_bank_.emplace_back();
+      nu_bank_entry = &p->nu_bank_.back();
```
paulromano: Might as well make the declaration here so it can be initialized simultaneously

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
-    p->delayed_group_ = site.delayed_group;
+    p->delayed_group_ = site->delayed_group;
 
     // Increment the number of neutrons born delayed
     if (p->delayed_group_ > 0) {
       nu_d[p->delayed_group_-1]++;
     }
+    
+    // Write fission particles to nuBank
+    if(use_fission_bank)
```
paulromano:
```suggestion
    if (use_fission_bank) {
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+    {
```
paulromano:
```suggestion
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+      nu_bank_entry = &p->nu_bank_.back();
+      nu_bank_entry->wgt             = site->wgt;
+      nu_bank_entry->E               = site->E;
+      nu_bank_entry->delayed_group   = site->delayed_group;
+    }
   }
+  
+  // If shared fission bank was full, and no fissions could be added,
+  // set the particle fission flag to false.
+  if( nu == skipped )
```
paulromano:
```suggestion
  if (nu == skipped) {
```

src/physics.cpp
```diff
@@ -169,28 +168,82 @@ create_fission_sites(Particle* p, int i_nuclide, const Reaction* rx,
...
+  {
```
paulromano:
```suggestion
```

src/physics_mg.cpp
```diff
@@ -114,38 +113,71 @@ create_fission_sites(Particle* p, std::vector<Particle::Bank>& bank)
...
+  if (settings::run_mode == RUN_MODE_EIGENVALUE)
+    use_fission_bank = true;
+
   for (int i = 0; i < nu; ++i) {
-    // Create new bank site and get reference to last element
-    bank.emplace_back();
-    auto& site {bank.back()};
+    Particle::Bank* site;
+    if(use_fission_bank)
+    {
```
paulromano: All the changes that are mentioned for the non-MG version should be made here

src/particle_restart.cpp
jtramm: Yes, that's definitely a good idea. I've split this section into a new helper function so as to avoid the repetition. I couldn't think of a good name for it, so went with "initialize_history_partial()", but maybe there is a better name out there for these tasks -- let me know if something comes to mind.

include/openmc/bank.h
```diff
@@ -21,13 +16,10 @@ namespace openmc {
...
 
 extern std::vector<Particle::Bank> source_bank;
-extern std::vector<Particle::Bank> fission_bank;
-extern std::vector<Particle::Bank> secondary_bank;
-#ifdef _OPENMP
-extern std::vector<Particle::Bank> master_fission_bank;
-#endif
 
-#pragma omp threadprivate(fission_bank, secondary_bank)
+extern Particle::Bank* fission_bank;
```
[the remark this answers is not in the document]
jtramm: That's a good question, I was on the fence about that one. Pros and cons:

Pros of using vector:
- More consistent with the rest of the code
- A little less prone to memory leaks due to having the destructor, and takes something off the to-do list for simulation cleanup.

Cons of using vector:
- For our use case, other developers might be tempted at some point to use vector functions (e.g., push_back()) on it, which would not be allowed. The result of using any of those functions would be undefined behavior which might be hard for a developer to track down the cause of. My thought was having it be just a pointer (might) make a developer more curious as to how it was allocated, and how it is used elsewhere, so therefore might be less error prone. Basically, since we don't want people to use vector functions on this, a good way to explicitly disallow that is to use a pointer rather than a vector.
- This is not a particularly important point, but the fission bank will need to be an allocated pointer on GPU, so doing it this way will minimize code delta. My thought is that if we are making big changes to the fission bank now, might as well bite the bullet and get that done.

Overall, I would probably lean towards just leaving it as an allocated pointer, but I can totally understand if the vector option sounds better. It will definitely work either way, and I can definitely see how the memory safety logistics would be more important for maintainability.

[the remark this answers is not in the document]
paulromano: Ah, right. For the omp atomic update, vector push_back won't work :thinking: Ok, I'm on board then. With respect to memory safety, perhaps the best thing to do would be to use `std::unique_ptr<Particle::Bank[]>` rather than a raw pointer. This would allow us to use `std::make_unique` for creation and then `.reset()` for deletion.

paulromano:
@jtramm I've been doing some performance testing of your branch this morning and things look mostly good. I did a few tests of a pincell problem and a quick run of a full core, and things look fine. For the pincell case, I also starting increasing the number of particles per generation, N, to see what would happen and noticed a slight decrease in the performance with increasing N that I believe is due to `stable_sort`. Interestingly, stable sort has complexity N log^2 N, unless extra memory is available in which case it goes as N log N (see [here](https://devdocs.io/cpp/algorithm/stable_sort)).

To find out what complexity we actually observe in practice, I switched over to running a very simple problem, Jezebel (sphere of Pu), and ran that with N ranging from 1e5 up to 1e8. I added a timer around the `stable_sort` to get some exact measurements and then compared that to N, N log N, and N log^2 N. You can find the full results in [this spreadsheet](https://docs.google.com/spreadsheets/d/1eD0Dpe745Zd5bnXIs84lL0B7Hmf-rB53bXGjz-XPoSg/edit?usp=sharing) (go to the "Sorting cost" tab). It looks like for N between 1 and 10 million, the cost is N log N, but above that it pretty clearly becomes N log^2 N (I'm guessing libstdc++ has some kind of cutoff above which it switches to the memory-optimized version as opposed to the time-optimized version?). As far as I can tell, there is really no way to tell the algorithm "hey, I have memory, please do this in N log N".

Some possible solutions:
1. Write our own stable sort that is guaranteed N log N.
2. Use a third-party library with such a sort
3. Punt on this, with the expectation that in the future we can use a parallel version of stable_sort, which should presumably do much better.
4. Forrest Brown published a transactions paper back in 1992 (vol 65, page 235) about an algorithm for sorting the fission bank in O(N). It entails storing both a parent ID (as you have done in this PR) as well as a "progeny ID". I'll send you the paper and you can tell me what you think.

If you're up for it, option 4 is probably my preferred approach.

include/openmc/bank.h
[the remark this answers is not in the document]
paulromano: Also, where `fission_bank` is declared, it would be good to put a comment explaining why we _can't_ use a vector.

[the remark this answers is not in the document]
jtramm: Ah, that makes sense, I'll go the unique_ptr route and add some comments explaining how it is used.

src/bank.cpp
```diff
@@ -32,13 +41,60 @@ std::vector<Particle::Bank> master_fission_bank;
...
+    int64_t value = simulation::progeny_per_particle[i-1] + tmp;
+    tmp = simulation::progeny_per_particle[i];
+    simulation::progeny_per_particle[i] = value;
+  }
+
+  // TODO: C++17 introduces the exclusive_scan() function which could be
+  // used to replace everything above this point in this function.
+
+  // Create temporary scratch vector for permuting the fission bank
+  std::vector<Particle::Bank> sorted_bank(simulation::fission_bank_length);
```
paulromano: So one thought I had when reading this is that our fission bank is already well over-allocated under normal circumstances. Hypothetically, we could use the upper portion of `fission_bank` to store the sorted sites if the number of banked sites is less than half the size of the fission bank. Not sure how much the memory allocation really matters, but might be worth looking into.

src/bank.cpp
```diff
@@ -32,13 +41,60 @@ std::vector<Particle::Bank> master_fission_bank;
...
+  for (int64_t i = 0; i < simulation::fission_bank_length; i++) {
+    const auto& site = simulation::fission_bank[i];
+    int64_t offset = site.parent_id - 1 - simulation::work_index[mpi::rank];
+    int64_t idx = simulation::progeny_per_particle[offset] + site.progeny_id;
+    sorted_bank[idx] = site;
+  }
+
+  // Copy sorted bank into the fission bank
+  for (int64_t i = 0; i < simulation::fission_bank_length; i++) {
+    simulation::fission_bank[i] = sorted_bank[i];
```
paulromano: This block could be replaced with a `std::copy`

src/bank.cpp
jtramm: That's a good idea! I've added that ability in, so for the typical case it will no longer allocate the extra memory for storing the sorted buffer. It makes the function a little busier, but I think it's worth it to save the memory in most cases.

src/bank.cpp
```diff
@@ -32,13 +41,70 @@ std::vector<Particle::Bank> master_fission_bank;
...
+  
+  // We need a scratch vector to make permutation of the fission bank into
+  // sorted order easy. Under normal usage conditions, the fission bank is
+  // over provisioned, so we can use that as scratch space.
+  Particle::Bank* sorted_bank;
+  std::vector<Particle::Bank> sorted_bank_holder;
+
+  // If there is not enough space, allocate a temporary vector and point to it
+  if (simulation::fission_bank_length > simulation::fission_bank_max / 2) {
+    sorted_bank_holder = std::vector<Particle::Bank>(simulation::fission_bank_length);
```
paulromano:
```suggestion
    sorted_bank_holder.resize(simulation::fission_bank_length);
```

paulromano:
Just wanted to summarize some of my thoughts for posterity. As John has already pointed out, most of the changes from threadprivate to particle-owned variables are inconsequential. The shared fission bank does have some tradeoffs, the downsides being:

- The fact that it is atomically updated from each thread does potentially lead to some shared memory contention.
- In order to maintain reproducibility, it's necessary to sort the fission bank at the end of a generation (before we would serially concatenate threadprivate fission banks into a single master fission bank).

The performance impact is generally only noticeable if the problem you're running is really simple. I've been doing some testing on Jezebel (pu-met-fast-001), and we definitely lose performance due to the shared fission bank. Running this problem on a two socket Xeon Platinum 8260 node with 1e6 particles per generation, I get the following numbers:

- develop branch: 4.18e6 particles/second
- this PR: 2.75e6 particles/second
- this PR (no sorting): 2.97e6 particles/second

For that last data point, I just commented out the fission bank sort (which is only needed for reproducibility). As you can see, most of the performance loss is not due to the sorting but rather just from the shared memory contention on the fission bank. Now, this being said, for a more realistic problem the impact is pretty small because, of course, there's more work for the particle to do in between these performance-impacting events (end of generation sort, atomic update on the fission bank). So overall, I think the benefits of these changes (better modularity, more flexibility for parallel models) outweigh this small cost.

I'll wait until Monday to merge in case any others have feedback.

paulromano:
Thanks once again @jtramm!

PR #1498, opened 2020-02-27
Fix for Issue #1473  Model.run not finding the final statepoint when trigger is being used

alex-lyons:
This should now be fixed, along the lines suggested by Paul Romano for closed PR #1046.
As a convenience, the final statepoint filename is available in model.statepoint after model.run() has completed.
I've written a regression test for this.


simondrichards:
@alex-lyons I notice that the Travis CI builds are failing your regression test on this PR, but the test passes in my clone of your fork. Your fork is a bit behind the openmc-dev repository so it might be worth bringing your fork up to date to see if that changes the results.

paulromano:
@alex-lyons @simondrichards Travis CI will actually test the result of merging the PR branch with `develop` rather than just the PR branch by itself, so merging develop into the PR branch shouldn't actually affect the test results.

simondrichards:
Thanks for the clarification @paulromano. OK, Merging develop into the PR branch should produce the same results as we see in Travis CI. But isn't it possible that the out of date PR branch is currently producing different results, and these may change as a result of merging develop? The  reference results in results_true.dat will have been generated by the PR branch. If we merge develop into the PR branch and then see the same results as we see on Travis CI, then we can update the results in results_true.dat and hopefully Travis CI will then pass.

paulromano:
Yes, that is true. It's certainly worth trying!

alex-lyons:
Paul, I'll take a look at your "more elegant" proposal. I agree in principle, and I felt a bit uneasy about relying on parsing a filename from the output lines as our only means of identifying the last statepoint file. I believe you said earlier that getting the filename this way would also work if the output directory was changed, because it would be included in the filename, but maybe your new proposal using settings.output will do that as well. I'm not sufficiently familiar with where openmc can put its files to know for sure.
I'll pull in the latest openmc-dev:develop before I make another pull request, so hopefully the tests will work.

alex-lyons:
Paul,

I’m looking to implement your latest proposed solution, but I have a couple of questions.


  1.  Is it guaranteed that the output directory, either settings.output[‘path’] or cwd(), is unique to each openmc run?
  2.  Does an openmc run always generate a statepoint file?

If the answer to either of the above is ‘NO’, it would become possible that the “latest” statepoint file might not have been written by this openmc run. It might have been already in the directory (if this run didn’t generate any statepoints), or it may have been written by another concurrent openmc run that happens to use the same output directory.

I can guard against reading possibly old statepoints by testing that its st_mtime is >= the start time of the openmc run, but I can’t think of any way of checking parentage in the case of concurrent runs writing to the same output directory. Maybe it just needs a note in the user guide to say “don’t do that”?

Any thoughts appreciated before I make the changes.

Regards,
Alex Lyons

paulromano:
@alex-lyons I like that you're proactively thinking about how users will break the code in weird ways (because they absolutely will) :smile: The answer to both of those questions is indeed "no". In the first case (two jobs writing concurrently to the same directory), there would likely be _many_ things that go wrong, so this particular issue is probably the least of the worries. For example, both jobs would write a summary.h5 file, one overwriting the other. It's a pretty strong assumption that when you are running OpenMC, your job is the only one writing to its directory. "don't do that" would be the right advice here.

For the second case, this is possible because we do allow users to specify at what batches they want statepoint files written. Your proposed solution is a good one -- we can just raise an exception if no sufficiently recent statepoint file was written, telling the user that their simulation didn't produce a statepoint file.

alex-lyons:
Paul,

I have many years’ experience of breaking code in weird ways 😉

I’ve written the code to determine the last written statepoint to exclude any pre-existing statepoint files. Not sure it’s a good idea to raise an exception if no statepoint is written: after all, the user may have deliberately turned them off. In such a case, it leaves model.statepoint=None and return None for k_effective.

I agree that if a user has concurrent runs writing to the same output directory, they deserve what they get.

Now on to struggling with git…

Alex.

paulromano:
@alex-lyons Sounds good; just ping me for a review once the branch is updated. Thanks!

openmc/model/model.py
```diff
@@ -207,17 +212,34 @@ def run(self, **kwargs):
...
+        for sp in output_dir.glob('statepoint.*.h5'):
+            mtime = sp.stat().st_mtime
+            if mtime >= tstart:  # >= allows for poor clock resolution
+                tstart = mtime
+                self.statepoint = sp
+
+        # Open the last statepoint to get the final k-effective
+        keff = None
+        if self.statepoint:
+            with openmc.StatePoint(self.statepoint) as sp:
```
paulromano: Looks like this is needed to prevent issues on Python 3.5
```suggestion
            with openmc.StatePoint(str(self.statepoint)) as sp:
```

alex-lyons:
I think the API change is OK.
It should be helpful to be able to easily access the last statepoint after the run. This is the minimal way of doing that without breaking the existing API in any way. I can't think of any alternative, e.g. returning a tuple, that doesn't break backwards-compatibility. Having determined the last statepoint, it seems a bit churlish just to extract k-effective from it and then throw it away, which is effectively what the old code did.
Model.run isn't "adding" a new atttribute: it is populating it - the statepoint attribute is None before the run (and after if no statepoint was written), which seems logical.

paulromano:
I agree that it if we absolutely want to preserve backwards compatibility, modifying this attribute is the way to go. However, the downside is that this makes what I would argue is a weird API even less clean. It would mean that some information gets communicated back as a return value and some as an attribute. We really only should be adding attributes on an object where access is required across different methods, which is not the case here. So, I'm suggesting that it's worth breaking backwards compatibility as a way of cleaning this up.

Although we don't make any guarantees about backwards compatibility across releases (given that we are pre 1.0), of course we do think carefully about when it makes sense to break backwards compatibility. My philosophy is that when we identify things like this where there's an opportunity to improve the API and doing so won't cause a lot of user pain, it's worth making these changes sooner rather than being locked in to them for a longer time. In this particular case, the `Model` class is actually not widely advertised, e.g., it doesn't appear in the user's guide at present, so I do feel comfortable changing the behavior for this method.

alex-lyons:
Classification: Public

Paul,

IMHO a proposed change that would break backwards-compatibility of the Model API as you suggest ought to be put in as a separate issue. Strictly, its outside the scope of #1473, which was simply to ensure that the returned k_effective was taken from the last-written statepoint, particularly for the case of triggered runs where the batch number was indeterminate.

However, I do agree with your view that for Model.run() to return the last statepoint Path rather than just the last k_effective would be generally more useful. I chose making it available as an attribute specifically to avoid breaking the existing API, assuming it was published and widely used. As you now state that this is not the case, I'm happy to make that change, particularly as I'm 90% there already. I'll fix any tests that may break as a result.

Regards,
Alex.

paulromano:
Thanks @alex-lyons. One change that will be needed is in our k-eff search function, since that does rely on `Model.run`. The change should be something like:
```diff
diff --git a/openmc/search.py b/openmc/search.py
index 6be8a50ea..0e992a410 100644
--- a/openmc/search.py
+++ b/openmc/search.py
@@ -51,7 +51,9 @@ def _search_keff(guess, target, model_builder, model_args, print_iterations,
     model = model_builder(guess, **model_args)
 
     # Run the model and obtain keff
-    keff = model.run(output=print_output)
+    sp_filename = model.run(output=print_output)
+    with openmc.StatePoint(sp_filename) as sp:
+        keff = sp.k_combined
 
     # Record the history
     guesses.append(guess)
```

alex-lyons:
Third attempt - Model.run now returns the statepoint Path (change to unpublished API).
Also updated StatePoint so it can take a Path as well as a string for the filename.

tests/regression_tests/trigger_statepoint_restart/test.py
```diff
@@ -0,0 +1,132 @@
...
+    cell = openmc.Cell(fill=mat, region=-sph)
+    geometry = openmc.Geometry([cell])
+
+    # Settings
+    settings = openmc.Settings()
+    settings.run_mode = 'eigenvalue'
+    settings.batches = 10
+    settings.inactive = 5
+    settings.particles = 200
+    settings.keff_trigger = {'type': 'std_dev', 'threshold': 0.002}
```
paulromano: I noticed that on Python 3.5, this test fails due to the inputs not matching. The underlying reason is that dictionaries in Python 3.5 are not guaranteed to retain insertion order. As a remedy, what we can do is sort the (key, value) pairs coming out of the dictionary when we write them to the settings.xml file:
```diff
diff --git a/openmc/settings.py b/openmc/settings.py
index 794c97b61..56125365e 100644
--- a/openmc/settings.py
+++ b/openmc/settings.py
@@ -809,9 +809,9 @@ class Settings(object):
         if self._keff_trigger is not None:
             element = ET.SubElement(root, "keff_trigger")
 
-            for key in self._keff_trigger:
+            for key, value in sorted(self._keff_trigger.items()):
                 subelement = ET.SubElement(element, key)
-                subelement.text = str(self._keff_trigger[key]).lower()
+                subelement.text = str(value).lower()
 
     def _create_energy_mode_subelement(self, root):
         if self._energy_mode is not None:
```

For Python 3.6 and above, my guess is that the tests are failing because of the threshold here. With this threshold, we end up running 47 batches. The more batches we run, the higher the likelihood of two simulations diverging due to very small differences in, e.g., implementations of basic mathematical functions. Making this threshold higher will help to reduce the number of batches. For example, picking a value of 0.0045 would result in only 12 total batches.

tests/regression_tests/trigger_statepoint_restart/test.py
```diff
@@ -0,0 +1,132 @@
...
+    settings.trigger_max_batches = 1000
+    settings.trigger_batch_interval = 1
+    settings.trigger_active = True
+    settings.verbosity = 1 # to test that this works even with no output
+ 
+    # Tallies
+    tallies = openmc.Tallies()
+    t = openmc.Tally()
+    t.scores = ['flux']
+    tallies.append(t)
```
paulromano:
```suggestion
    t = openmc.Tally()
    t.scores = ['flux']
    tallies = openmc.Tallies([t])
```

tests/regression_tests/trigger_statepoint_restart/test.py
```diff
@@ -0,0 +1,132 @@
...
+    plots = openmc.Plots()
+
+    # Put it all together
+    model = openmc.model.Model(materials=materials,
+                               geometry=geometry,
+                               settings=settings,
+                               tallies=tallies,
+                               plots=plots)
+    return model
+
```
paulromano: Per PEP8, two blank lines between top-level functions or classes
```suggestion


```

tests/regression_tests/trigger_statepoint_restart/test.py
```diff
@@ -0,0 +1,132 @@
...
+            assert str(k_combined_1) == str(k_combined_2), \
+                'Different final k_combined after restart'
+            self._write_inputs(self._get_inputs())
+            self._compare_inputs()
+            self._test_output_created()
+            self._write_results(self._get_results())
+            self._compare_results()
+        finally:
+            self._cleanup()
+
```
paulromano:
```suggestion


```

tests/regression_tests/trigger_statepoint_restart/test.py
```diff
@@ -0,0 +1,132 @@
...
+ 
+    # Plots (none)
+    plots = openmc.Plots()
+
+    # Put it all together
+    model = openmc.model.Model(materials=materials,
+                               geometry=geometry,
+                               settings=settings,
+                               tallies=tallies,
+                               plots=plots)
```
paulromano: If plots are empty, you don't need to specify it:
```suggestion
    # Put it all together
    model = openmc.model.Model(materials=materials,
                               geometry=geometry,
                               settings=settings,
                               tallies=tallies)
```

tests/regression_tests/trigger_statepoint_restart/test.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture
+def model():
+
+    # Materials
+    materials = openmc.Materials()
+    mat = openmc.Material()
+    mat.set_density('g/cm3', 4.5)
+    mat.add_nuclide('U235', 1.0)
+    materials.append(mat)
```
paulromano:
```suggestion
    mat = openmc.Material()
    mat.set_density('g/cm3', 4.5)
    mat.add_nuclide('U235', 1.0)
    materials = openmc.Materials([mat])
```

tests/regression_tests/trigger_statepoint_restart/test.py
alex-lyons: OK, although I note that in the equivalent _keff_trigger_from_xml_element method, the keys (there are only two) are hard-coded. I shall do the same for _create_output_subelement which also loops over unsorted keys.

alex-lyons:
Not sure what's going on here... the new test trigger_statepoint_restart works fine on my machine but fails consistently in travis, all versions. I've done the pytest --update and the updated output is being used in the tests, but the travis-calculated keff is just different. Could there be some file that I've not checked-in? I'll double check.

paulromano:
Hmm, that is strange. So far I haven't been able to reproduce the failures on Travis, even using the same OS (Ubuntu 16.04 on a docker image). I'll keep digging and let you know if I come up with anything. Sorry this has dragged on!

paulromano:
Ok, I think I have an idea of what's going on. The statepoint name that is given for the test is `statepoint.0*.h5`. Then, in the `execute_test` method, it calls `self._get_results()` which uses the statepoint name to glob for matching files. The problem is that there is no guarantee about what order you'll get statepoint files back in. The test ends up writing statepoints at batches 10 and 13, so on Travis it appears that it's picking up the results from `statepoint.0010.h5`, not `statepoint.0013.h5`. When you and I run locally, we just happen to get lucky and get `statepoint.0013.h5`. Does that make sense @alex-lyons?

alex-lyons:
I've sorted the glob of statepoint files, so ordering should now be consistent and correct. Fortunately the 'real' glob was in _test_output_created which I'd already overridden, so I didn't have to override any more of the standard testing harness.
We'll see what happens...

paulromano:
@alex-lyons It looks like your last commit did fix the failing test :tada: However, it also rolled back a change in the xtl git submodule. Can you undo that change? The following should work:
```bash
git reset HEAD~
git submodule update
git add tests/regression_tests/trigger_statepoint_restart/test.py
git commit ...
git push -f ...
```

alex-lyons:
Paul, I've done the "git reset HEAD~", etc,  you suggested, but I'm not sure if it's made any difference. I admit I'm finding git on github a bit occult. I've no idea why XTL should have been "rolled back" by anything I've done. Also not sure how I seemed to suddenly generate 4 review requests!
I'll look in again next week to see what has happened, but I can't really spend much more time on this.

paulromano:
@alex-lyons No problem -- I've just fixed up your branch and submitted a PR to replace this one. Your name is still listed on all the commits.

Regarding the review requests, we have a [CODEOWNERS](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) that lists "owners" for specific files. If you change one of those files in a pull request, the person will automatically get requested for a review. One of your merge commit resulted in other files changing which is why github suddenly requested reviews from others.

paulromano:
This PR was superseded by #1536, which has now been merged.

PR #1502, opened 2020-02-28
Fix for enumeration values not explicitly handled in switch statements

simondrichards:
This is a fairly trivial PR to fix examples of switch statements which don't explicitly handle all enumeration values. These throw compiler warnings when I build on MacOS. This PR allows the MacOS build to complete with no warnings.


nelsonag:
Thanks @simondrichards!

Issue #1512, opened 2020-03-05
MGXS mode doesn't work with interpolation temperature method


dryuri92:
> ERROR: MGXS Library does not contain cross sections for fuel at temperatures
        that bound 1027.000000

kTs list of data contains temperatures: 300, 600, 900, 1200, 1800, 2100 K
The problem doesn't work with mgxs library when temperature method is interpolation. This mistake take place for both cases temperature of cell is equal one from the "kts" list of mgxs.h5 or doesn't.

nelsonag:
Thanks again for this @dryuri92!

PR #1514, opened 2020-03-06
Fix the bug on MGXS mode with Interpolation method

dryuri92:
This PR closed the Issue #1512. The problem related with incorrect initialization `xt::xarray<double>available_temps(num_temps)`  which have been lead to `available_temps` array elements to have the same address. I fixed it according to https://xtensor.readthedocs.io/en/latest/container.html with `shape` vector. In addition I found interpolation procedure themselves didn't work right and I repaired it too.


src/mgxs.cpp
```diff
@@ -380,16 +383,17 @@ Mgxs::Mgxs(const std::string& in_name, const std::vector<double>& mat_kTs,
     // If we are doing nearest temperature interpolation, then we don't need
     // to do the 2nd temperature
     int num_interp_points = 2;
+    std::vector<double> interp(micros.size());
```
nelsonag: Can you move this ``std::vector`` initialization and the next to below the ``if (settings::temperature_method == TemperatureMethod::Nearest)...`` line? The comment block above is describing the two lines that set ``num_interp_points`` so its better to have them together and next to that comment.

dryuri92: ok

dryuri92:
> We currently don't have a test of MGXS temperature interpolation. Would you be willing to add a unit test of the nearest and linear interpolants for an MGXS case?
@nelsonag Thank you for review!
Yes I've the same idea about of testing nesserity of this feature testing! But I don't know how do it better way. I looked at the unit testing directory and didn't found out existing test related to multi-group mode. At the regression directory there are a number of mg tests. And I don't how what is better to do:

- to add temperature setting to existing tests at e.g. `mg_basic`, `mg_convert`, etc;

- to create new regression test/tests with check up criteria is `keff`;

- to create new unit test as you mentioned above but I don't know where is criteria for comparison
at unit test case.
?

nelsonag:
The way I would do it is to create a series of regression tests that use an
infinite-homogeneous model and a 2-group MGXS provided at >=2 temperatures.
Then, that model and data would be run as follows to ensure the code is
exericsed
- Microscopic cross sections:
  - Nearest temperature interpolation:
      1. Run with a material T == a T on the temperature grid point
      2. Run with a material T closer to a lower temperature grid point*
      3. Run with a material T closer to a higher temperature grid point
  - Linear interpolation
      1. Run with a material T == a T on the temperature grid point
      2. Run with a material T somewhere between temperature grid points*
- Macroscopic cross sections: [Same as for microscopic]
  - Nearest temperature interpolation:
      1. Run with a material T == a T on the temperature grid point
      2. Run with a material T closer to a lower temperature grid point*
      3. Run with a material T closer to a higher temperature grid point
  - Linear interpolation
      1. Run with a material T == a T on the temperature grid point
      2. Run with a material T somewhere between temperature grid points*

*The items with asterisks are the minimum necessary, but the rest would be
what is necessary for a full verification.

The comparison criteria then can simply be the analytic keff you would
expect based on the hand-calculated interpolated cross sections. If the
cross sections are different enough with temperature, then the change in
keff should be significant enough to be observed.

One final note, I recognize there is an argument to be made for using
1-group instead of 2-group xs for simplicity, but I'd prefer 2-group over
1-group as that is slightly more robust as it opens up more ways for group
indexing to fail.

Let me know if this is something you would like to tackle, if not we can
separate it out from this PR.

Thanks!
Adam

dryuri92:
@nelsonag  Sorry for a long response I've almost finished  a test but faced with a trouble with case 

> Linear interpolation
      1. Run with a material T == a T on the temperature grid point

i.e. in such case when interpolation coefficients is (1.0, 0.0). I researched code and found that in xsdata.cpp in function `XsData::combine` we accumulate cross-sections according whit that coefficients. But scatter matrix calulated in scattdata.cpp are not accumulated on the contrary it taking a last interpolation value i.e.  0.0 and it lead to fully incorrect calculation. I've been made some changes in last commit  but not sure in this corrections and ask you for review it.

src/scattdata.cpp
```diff
@@ -65,7 +65,10 @@ ScattData::base_combine(size_t max_order,
   xt::xtensor<double, 3> this_matrix({groups, groups, max_order}, 0.);
   xt::xtensor<double, 2> mult_numer({groups, groups}, 0.);
   xt::xtensor<double, 2> mult_denom({groups, groups}, 0.);
-
+  // TODO: Need to review this:
+  if (this->scattxs.size() > 0) {
+    this_matrix = this->get_matrix(max_order);
+  }
```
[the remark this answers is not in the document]
nelsonag: Wouldn't this initialization, with the for loop at lines 97 to 99 double count the first matrix (i.e., the one from those_scatts[0])?

src/scattdata.cpp
```diff
@@ -110,7 +113,19 @@ ScattData::base_combine(size_t max_order,
 
   // Combine mult_numer and mult_denom into the combined multiplicity matrix
   xt::xtensor<double, 2> this_mult({groups, groups}, 1.);
-  this_mult = xt::nan_to_num(mult_numer / mult_denom);
```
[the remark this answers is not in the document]
nelsonag: What problem did you observe with the xt::nan_to_num(mult_numer / mult_denom) approach that the explicit loop is intended to fix?

src/scattdata.cpp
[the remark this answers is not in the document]
dryuri92: As I understand the logic of the code .. it can be written as follows:
`this->scatter_matrix = amin * scattermatrix[temperature_min] ;
 this->scatter_matrix = this->scatter_matrix  + amax*scatter_matrix[temperature_max]` 
At the first iteration of a cycle in xsdata.cpp function `combine` 
At lines 97-99 `scalars` consist from element `{amin}` with `that_scatter` or `those_scatts[0]` is pointed to `scattermatrix[temperature_min]` 
at the second one `scalars` consist from element `{amax}` with `that_scatter` or `those_scatts[0]` is pointed to `scattermatrix[temperature_max]`. And the result we get this_matrix equal to this part `amax*scatter_matrix[temperature_max]`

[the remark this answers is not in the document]
dryuri92: Shortly we didn't save a value of scatter_matrix after first cycle

src/scattdata.cpp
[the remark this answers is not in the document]
dryuri92: It is more complicated issue but the reasaon is the same:

> we didn't save a value of scatter_matrix after first cycle

and it relates `this_mult` array too. Because when we at cross point our last coefficient is 0 and we should divide on 
`this_mult = xt::nan_to_num(mult_numer / mult_denom);` null vector. Function `xt::nan_to_num` converts all NaN to zero leading to when throug simmulation we sample a particle at code line 788 this source
  ```
// Update the weight to reflect neutron multiplicity
  wgt *= mult[gin][i_gout];
```
we always get a "0" is a weight.

**Comment by dryuri92** [comment]
@nelsonag I've just uploaded series of regression tests called `mg_temperature` in the order how you described here:

> - Microscopic cross sections:
>   - Nearest temperature interpolation:
>       1. Run with a material T == a T on the temperature grid point
>       2. Run with a material T closer to a lower temperature grid point*
>       3. Run with a material T closer to a higher temperature grid point
>   - Linear interpolation
>       1. Run with a material T == a T on the temperature grid point
>       2. Run with a material T somewhere between temperature grid points*
> - Macroscopic cross sections: [Same as for microscopic]
>   - Nearest temperature interpolation:
>       1. Run with a material T == a T on the temperature grid point
>       2. Run with a material T closer to a lower temperature grid point*
>       3. Run with a material T closer to a higher temperature grid point
>   - Linear interpolation
>       1. Run with a material T == a T on the temperature grid point
>       2. Run with a material T somewhere between temperature grid points*

By now I temporary save as result both openmc and analytical eigenvalues for analysis and comparison ability.
I hope these tests be helpful for the issue analyze.

File: tests/regression_tests/mg_temperature/build_2g.py [hunk 4]
```diff
@@ -0,0 +1,291 @@
...
+    mg_cross_sections_file_macro.export_to_hdf5('macro_2g.h5')
+    # Returning the macro_xs dict is needed for analytical solution
+    return macro_xs
+
+
+def analytical_solution_2g_therm(xsmin, xsmax=None, wgt=1.0):
+    """ Calculate eigenvalue based on analytical solution for eq Lf = (1/k)Qf
+        in two group for infinity dilution media in assumption of group
+        boundary in thernmal spectra < 1.e+3 Ev
+        Parametres:
```
nelsonag: typo: "Parametres" -> "Parameters"

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+    mg_cross_sections_file_micro.export_to_hdf5('micro_2g.h5')
+    mg_cross_sections_file_macro.export_to_hdf5('macro_2g.h5')
+    # Returning the macro_xs dict is needed for analytical solution
+    return macro_xs
+
+
+def analytical_solution_2g_therm(xsmin, xsmax=None, wgt=1.0):
+    """ Calculate eigenvalue based on analytical solution for eq Lf = (1/k)Qf
+        in two group for infinity dilution media in assumption of group
+        boundary in thernmal spectra < 1.e+3 Ev
```
nelsonag: "thernmal" -> "thermal"

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+    """Create a dictionary with two group cross-section"""
+    xs_macro = {}
+    for t, d1 in xs_micro.items():
+        xs_macro[t] = {}
+        for r, d2 in d1.items():
+            temp = []
+            xs_macro[t][r] = {}
+            for n, v in d2.items():
+                temp.append(d2[n])
+            xs_macro[t][r]['macro'] = sum(temp)
```
nelsonag: Perhaps a comment should be added explaining why the 'macro' key needed if this is already the xs_macro variable. I *think* its so you can reuse build_openmc_xs_lib for micro and macro, but clarity would be good.

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+def analytical_solution_2g_therm(xsmin, xsmax=None, wgt=1.0):
+    """ Calculate eigenvalue based on analytical solution for eq Lf = (1/k)Qf
+        in two group for infinity dilution media in assumption of group
+        boundary in thernmal spectra < 1.e+3 Ev
+        Parametres:
+        ----------
+        xsmin : dict
+            - macro cross-sections dictonary with minimum range temperature
```
nelsonag: The rest of the OpenMC python docstrings dont have the leading dash before the description of the parameter. For consistency, could you delete these dashes?

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+        xsmax : dict
+            - macro cross-sections dictonary with maximum range temperature
+              by default: None not used for standalone temperature
+        wgt : double
+            - weight for interpolation by default 1.0
+        Returns:
```
nelsonag: Add an extra line before "Returns" and make sure the "-----------" has the same number of dashes as Returns has spaces (7).

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+        in two group for infinity dilution media in assumption of group
+        boundary in thernmal spectra < 1.e+3 Ev
+        Parametres:
+        ----------
+        xsmin : dict
+            - macro cross-sections dictonary with minimum range temperature
+        xsmax : dict
+            - macro cross-sections dictonary with maximum range temperature
+              by default: None not used for standalone temperature
+        wgt : double
```
nelsonag: in python these "double"s are called "float" (Confusing... I know...)

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+            - weight for interpolation by default 1.0
+        Returns:
+        ---------
+        keff : np.double - analytical eigenvalue of critical eq matrix
```
nelsonag: and this np.double would be a np.float64

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+    # Exporting library to hdf5 files
+    mg_cross_sections_file_micro.export_to_hdf5('micro_2g.h5')
+    mg_cross_sections_file_macro.export_to_hdf5('macro_2g.h5')
+    # Returning the macro_xs dict is needed for analytical solution
+    return macro_xs
+
+
+def analytical_solution_2g_therm(xsmin, xsmax=None, wgt=1.0):
+    """ Calculate eigenvalue based on analytical solution for eq Lf = (1/k)Qf
+        in two group for infinity dilution media in assumption of group
```
nelsonag: The function docstrings should be indented to the same level as the triple quotes.

tests/regression_tests/mg_temperature/build_2g.py
nelsonag: "dictonary" -> "dictionary"

nelsonag: This is in a few locations, fix all

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+        sa = xsmin['absorption']['macro'] * wgt + \
+             xsmax['absorption']['macro'] * (1 - wgt)
+        ss12 = xsmin['scatter']['macro'][0][1][0] * wgt + \
+               xsmax['scatter']['macro'][0][1][0] * (1 - wgt)
+        nsf = xsmin['nu-fission']['macro'] * wgt + \
+              xsmax['nu-fission']['macro'] * (1 - wgt)
+    L = np.array([sa[0] + ss12, 0.0, -ss12, sa[1]]).reshape(2, 2)
+    Q = np.array([nsf[0], nsf[1], 0.0, 0.0]).reshape(2, 2)
+    arr = np.linalg.inv(L).dot(Q)
+    return np.linalg.eigvals(arr)[1]
```
nelsonag: Since the eigenvalues produced from np.linalg.eigvals are not necessarily ordered, would it be better to return np.amax(np.linalg....) instead of indexing with [1]?

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,291 @@
...
+    """ Building an infinite medium for openmc multi-group testing
+        Parametres:
+        ----------
+        xsnames : list of str()
+            - list with xs names
+        xslibname:
+            - name of hdf5 file with cross-section library
+        temperature : float
+            - value of a current temperature in K
+        tempmethod : str {'nearest', 'interpolstion'} by default 'nearest'
```
nelsonag: 'interpolstion' -> 'interpolation'

nelsonag: No need to state the type (str), just leave the options in the curly braces. Also move the default to the next line with a description.

tests/regression_tests/mg_temperature/macro/interpolation/case1/geometry.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='utf-8'?>
```
nelsonag: These xml files dont need to be stored in the git repository.

tests/regression_tests/mg_temperature/macro/nearest/case3/results_true.dat
```diff
@@ -0,0 +1,4 @@
+k-combined:
+1.411726E+00 2.151179E-03
```
nelsonag: This doesn't seem to match the analytic solution. Do we understand why?

tests/regression_tests/mg_temperature/micro/interpolation/case1/results_true.dat
```diff
@@ -0,0 +1,4 @@
+k-combined:
+1.422739E+00 1.661420E-03
```
nelsonag: ditto

tests/regression_tests/mg_temperature/micro/nearest/case3/results_true.dat
```diff
@@ -0,0 +1,4 @@
+k-combined:
+1.411726E+00 2.151179E-03
+k-analytical:
+1.407830E+00
```
nelsonag: ditto

tests/regression_tests/mg_temperature/test.py
```diff
@@ -0,0 +1,106 @@
...
+from tests.regression_tests.mg_temperature.build_2g import *
+from tests.testing_harness import *
+
+
+class MgTemperatureTestHarness(TestHarness):
+
+    def execute_test(self):
+        """Run OpenMC with the appropriate arguments and check the outputs."""
+        base_dir = os.getcwd()
+        print("Base dir is {}".format(base_dir))
```
nelsonag: Looks like a debug statement, remove in the final commit

tests/regression_tests/mg_temperature/test.py
```diff
@@ -0,0 +1,106 @@
+import os
+from tests.regression_tests.mg_temperature.build_2g import *
+from tests.testing_harness import *
+
+
+class MgTemperatureTestHarness(TestHarness):
+
+    def execute_test(self):
```
nelsonag: I would prefer if the results_*.dat files were all aggregated together in a single file in tests/regression_tests/mg_temperature/ instead of located in the case subdirectories. Then, all of those case subdirectories can be completely removed after the test executes.

dryuri92:
> Thank you very much @dryuri92. These additions are great. The requested changes are in the following categories:
> 
> * editorial issues with docstrings
> * accuracy of test results
> * test directory structure.
>   Take a look and let me know if you have any questions. Thanks again!

I just want help to this very ambigious project! Yes, I will clean code in accordance of your point upper and will work on tests structure. As for differences there are two reasons for difference between analytical solution and openmc one:

1. (less) A low-diagonal element is in a scatter matrix for MC simulation. In analytical solution i didn't consider that
2. (more) Insufficance of statistics when I increased a number of batches in 10 times I got a good accordance up to 0.001-0.002  in a value of difference. 

Hence I have a question what preferable to do keep the results the same or increase a statistics paramtres but at the same time increase a time for test?

nelsonag:
For the two items on your list:
1. Ah that makes sense. Since we don't need to test that up-scattering is captured as well (there is no reason to think that if down-scattering is interpolated that up-scattering wont be), can you just remove the up-scatter component, thus improving ability to match the analytic solution.
2. The runtime due to these tests is not very large as-is, right? If so, I'm ok with the extra factor of 10 since these infinite 2g problems should be pretty darn quick.

nelsonag:
Regarding the update item, I am in general in support of that. However, the correct place for that would either be the https://github.com/openmc-dev/data repository. I personally dont have the time to write a MGXS conversion script for those codes, but would welcome any pull requests which do that (and show that the data is processed correctly with adequate V&V)

dryuri92:
> Regarding the update item, I am in general in support of that. However, the correct place for that would either be the https://github.com/openmc-dev/data repository. I personally dont have the time to write a MGXS conversion script for those codes, but would welcome any pull requests which do that (and show that the data is processed correctly with adequate V&V)

I'm currently work with a number of multigroup libraries in applying it in openmc. If result will be good and have enough time I want to share their with community. 
P.S. I checked code and uploaded it in last commits.

tests/regression_tests/mg_temperature/build_2g.py
```diff
@@ -0,0 +1,297 @@
...
+    for t, d1 in xs_micro.items():
+        xs_macro[t] = {}
+        for r, d2 in d1.items():
+            temp = []
+            xs_macro[t][r] = {}
+            for n, v in d2.items():
+                temp.append(d2[n])
+    # The name 'macro' is needed to store data at the same level
+    # of a xs_macro dictionary as for xs_micro and use it in
+    # function build_openmc_xs_lib
```
nelsonag: Let's make this comment indented to the same level as the for loop it resides in

tests/regression_tests/mg_temperature/test.py
```diff
@@ -0,0 +1,118 @@
...
+import shutil
+
+
+class MgTemperatureTestHarness(TestHarness):
+
+    def execute_test(self):
+        """Run OpenMC with the appropriate arguments and check the outputs."""
+        base_dir = os.getcwd()
+        overall_results = []
+        print("Base dir is {}".format(base_dir))
```
nelsonag: Can you remove the print statement here?

tests/regression_tests/mg_temperature/test.py
```diff
@@ -0,0 +1,118 @@
...
+            self._cleanup()
+            for f in ['micro_2g.h5', 'macro_2g.h5']:
+                if os.path.exists(f):
+                    os.remove(f)
+
+    def update_results(self):
+        """Update the results_true using the current version of OpenMC."""
+        base_dir = os.getcwd()
+        overall_results = []
+        print("Base dir is {}".format(base_dir))
```
nelsonag: Same with this print statement: remove it

nelsonag:
Thanks for this great PR @dryuri92! I will merge once the tests pass

Issue #1517, opened 2020-03-10
Initial Implementation of IncidentProton


SamPUG:
This issue is added in response to a requirement from @Shimwell and @ukaea.

Initial IncidentProton functionality has been requested so that proton data from ACE or ENDF libraries can be read and plotted. 

I plan on implementing an initial IncidentProton class which reads only cross section data, ignoring any distribution data. This will include the implementation of the following methods:
- `from_endf`
- `from_ace`
- `export_to_hdf5`
- `from_hdf5`

Development of the IncidentProton class will be based on the existing `data.IncidentPhoton` and `data.IncidentNeutron` classes.

I also plan to make the necessary developments to `plotter.py` to allow the proton data to be plotted.

paulromano:
Out of curiosity, what is the intended use of incident proton data?

Shimwell:
> Out of curiosity, what is the intended use of incident proton data?

One of the reasons is that there are proton induced tritium producing reactions that might just help with a bit of tritium production for those spherical reactors with small breeder blankets. The ability of OpenMC to conveniently make materials and extract the material cross section is a useful feature for examining potential proton induced rections and allows us to make use of alot of existing infrastructure. Another reason is if OpenMC is going to consider supporting the transport of protons for machines with big magnets like accelerators (and perhaps edge cases in fusion magnet heating) then a good place to start is reading the cross section data. Perhaps a final reason is that it was seen as easy enough to implement with in the contract timeframe and allows us to grow the community. Identify accessable tasks 1 year before they would be done that the existing community is not already considering is tricky and this was perhaps the worst task of all the 13. So in general the task all have a combination of: fusion relevance, unlikely to be done otherwise, easy enough to gaurentee it can be done in the timeframe, not so difficult that it deterred potential applications. I hope to be ambitious with the tasks in the future if there is a similar activity. In addition to the actual 13 tasks it is the positive experience of the process and contributors experience that helps us demonstrate the future applicability of this wonderful code for fusion with ability to implement features being something that OpenMC leads the market in 😀

paulromano:
Well put; thanks @Shimwell!

simondrichards:
Now it's my turn to be curious: Are there any plans for OpenMC to support transport of charged particles in magnetic fields?

Shimwell:
> Now it's my turn to be curious: Are there any plans for OpenMC to support transport of charged particles in magnetic fields?

Looking forward to the answer to this one,

[...]

PR #1520, opened 2020-03-13
Added options for scaling axes in plotter.py

SamPUG:
While testing developments for #1517 I found it helpful to change the plotting axes. Currently the axis limits are hard coded.

I've added some optional arguments to `plot_xs` to adjust the axis scaling and energy/y limits.


SamPUG:
Thanks @paulromano. I didn't think about getting a reference to the axis back out of the returned figure!

PR #1530, opened 2020-03-23
adding elements to a material from formula

Shimwell:
This PR adds the ability to quickly make a material from a formula. This is useful when making a few fusion relevant materials like lithium ceramics (e.g. Li4SiO4), molten salts (e.g. Li2BeF4), neutron multipliers (e.g. Be12Ti), superconductors (e.g. Nb3Ti). The formula can also accept brackets like Al2(SO4)3 

The most simple use case is below
```
m = openmc.Material()
m.add_elements_from_formula('Li4SiO4')
```

Arguments like enrichment are also allowed
```
m = openmc.Material()
m.add_elements_from_formula('Li4SiO4', enrichment=60, enrichment_target='Li6')
```

There is some checking of the formula in the method to ensure that the material is created correctly. However some of the checking is also carried out by add_element which is called. 

Unit tests have also been added to check the various value errors and check the results are as expected for a few representative cases.

### DIFF
--- a/openmc/material.py
+++ b/openmc/material.py
@@ -1,9 +1,10 @@
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, Counter
 from collections.abc import Iterable
 from copy import deepcopy
 from numbers import Real, Integral
 from pathlib import Path
 import warnings
+import re
 from xml.etree import ElementTree as ET
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_material.py
```diff
@@ -58,6 +58,61 @@ def test_elements_by_name():
...
+    nuc_dens = m.get_nuclide_atom_densities()
+    for nuclide in ref_dens:
+        assert nuc_dens[nuclide][1] == pytest.approx(ref_dens[nuclide], 1e-2)
+
+    # testing the correct nuclides are added to the Material when enriched
+    m = openmc.Material()
+    m.add_elements_from_formula('Li4SiO4',
+                                enrichment=60.,
+                                enrichment_target='Li6')
+    ref_dens = {'Li6': 0.2666, 'Li7': 0.1777,
```
drewejohnson: Would it make sense to check the elemental fractions before diving in to the atom densities? This way we can easily verify that the method is assigning four parts lithium to one part silicon to four parts oxygen, independent of the enrichments and abundances?

tests/unit_tests/test_material.py
```diff
@@ -58,6 +58,61 @@ def test_elements_by_name():
...
+                'Si28': 0.102477, 'Si29': 0.0052035, 'Si30': 0.0034301,
+                'O16': 0.443386, 'O17': 0.000168}
+    nuc_dens = m.get_nuclide_atom_densities()
+    for nuclide in ref_dens:
+        assert nuc_dens[nuclide][1] == pytest.approx(ref_dens[nuclide], 1e-2)
+
+    # testing the use of brackets
+    m = openmc.Material()
+    m.add_elements_from_formula('Mg2(NO3)2')
+    ref_dens = {'Mg24': 0.157902, 'Mg25': 0.02004, 'Mg26': 0.022058,
```
drewejohnson: Same comment as above, but with 2 parts manganese, 2 parts nitrogen, and 6 parts oxygen

openmc/material.py
```diff
@@ -588,6 +589,97 @@ def add_element(self, element, percent, percent_type='ao', enrichment=None,
...
+
+        """
+        cv.check_type('formula', formula, str)
+
+        # Tokenizes the formula and check validity of tokens
+        tokens = re.findall(r"([A-Z][a-z]*)(\d*)|(\()|(\))(\d*)", formula)
+        for row in tokens:
+            for token in row:
+                if token.isalpha():
+                    if token not in list(openmc.data.ATOMIC_NUMBER.keys())[1:]:
```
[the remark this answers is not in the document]
Shimwell: Good spot

openmc/material.py
```diff
@@ -588,6 +589,97 @@ def add_element(self, element, percent, percent_type='ao', enrichment=None,
...
+    def add_elements_from_formula(self, formula, percent_type='ao', enrichment=None,
+                                  enrichment_target=None, enrichment_type=None):
+        """Add a elements from a chemical formula to the material.
+
+        Parameters
+        ----------
+        formula : str
+            Formula to add, e.g., 'C2O', 'C6H12O6', or (NH4)2SO4.
+            Note this is case sensitive, elements must start with an uppercase
+            character. Any numbers will be converted to integers (rounded down).
```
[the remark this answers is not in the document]
Shimwell: Yep, I can do this and I shall add a test for such a case.

openmc/material.py
```diff
@@ -588,6 +589,97 @@ def add_element(self, element, percent, percent_type='ao', enrichment=None,
...
+        mat_stack = [Counter()]
+        for symbol, multi1, opening_bracket, closing_bracket, multi2 in tokens:
+            if symbol:
+                mat_stack[-1][symbol] += int(multi1 or 1)
+            if opening_bracket:
+                mat_stack.append(Counter())
+            if closing_bracket:
+                stack_top = mat_stack.pop()
+                for i in stack_top:
+                    mat_stack[-1][i] += int(multi2 or 1) * stack_top[i]
```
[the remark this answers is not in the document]
Shimwell: Nice, that is clearer

openmc/material.py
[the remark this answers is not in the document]
Shimwell: I have now added a test to check decimal values result in an error being raised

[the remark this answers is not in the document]
Shimwell: I have now added a check to the add_elements_from_material method to check for the existence of a '.' in the formula which then gives more helpful instructions to the user. Decimal places were previously raising a not value error but they were grouped in with other forbidden formula entry chars. This new error is more specific to the problem

tests/unit_tests/test_material.py
Shimwell: This would be nice and I am sure I can do it. 

I shall see if I can find an elegant way to do it using the existing methods. The .get_nuclide_atom_densities() method makes it easy to check the nuclides but if I have understood correctly there is not an equivalent method for element checking.

Perhaps in the future I can add a few more methods to the material class like get_elements(), get_element_densities(), get_element_atom_densities()

drewejohnson: As an interim step, you could do something like
```python
elem = collections.defaultdict(float)
for nuclide, adens in m.get_nuclide_atom_densities().values():
    if nuclide.startswith("Li"):
        elem["Li"] += adens
    # repeat for other elements
elem_percents = {key: value / sum(elem.values) for key, value in elem.items())
assert elem_percents["Li"] == pytest.approx(4)
# repeat for other elements
```
Granted, this works only if you know the elements that should exist. Not too much a burden for this test, and doesn't require adding (and testing) new ``Material`` methods

Shimwell: That looks great, thanks for the help. I have implement this test

tests/unit_tests/test_material.py
Shimwell: I have also implement this test

drewejohnson:
It looks like the [build has passed](https://travis-ci.org/github/openmc-dev/openmc/builds/666335597), but isn't being reported back to GitHub. It may be related to [some outages yesterday](https://www.traviscistatus.com/incidents/ntxbkxt8lrx9)

Pinging @paulromano for advice. Do we need to restart a build?

tests/unit_tests/test_material.py
```diff
@@ -3,7 +3,7 @@
 import openmc.stats
 import openmc.examples
 import pytest
-
+from collections import defaultdict
```
paulromano: PEP8: Order of imports should be 1) standard library, 2) thick-party, 3) local, each separated by a blank line. Can you move this above the openmc/pytest imports?

tests/unit_tests/test_material.py
```diff
@@ -58,6 +58,96 @@ def test_elements_by_name():
     assert a._nuclides == b._nuclides
     assert b._nuclides == c._nuclides
 
+def test_adding_elements_by_formula():
```
paulromano: PEP8: two blank lines between functions at top level
```suggestion

def test_add_elements_by_formula():
```

tests/unit_tests/test_material.py
Shimwell: Moved the import as suggested and imports now follow the specified order

PR #1557, opened 2020-05-07
Hot fix: mgxs library Cell domain filtering

nelsonag:
This PR fixes 2 minor bug:

1. ``data::temperature_min`` and ``data::temperature_max`` are to keep track of the min and max temperatures read, respectively, finally being written to screen after all data is read.
a. However, these min/max values were not the global min/max, but were instead the min/max values read for a particular isotope.
b. For example, if the temperature min/max range is 294K and 600K and a nuclide is encountered who we only needed to read 900K data from, the temperature_min and temperature_max values would both become 900.
c. This was fixed with a simple change: for min, set it with infinity, and just update the min value to be the lesser of the current min or the global min. Opposite for max.

2. The ``openmc.mgxs.Library`` class's ``domains`` setter was not allowing you to assign cell domains if they were filled with universes. This was simply because we were obtaining the list of all possible cells by using  `geometry.get_all_material_cells()` instead of `geometry.get_all_cells()`. Since there is no reason to disallow a MGXS Library domain to be a universe-filled cell, I am now using `geometry.get_all_cells()`. Distribcell types still use `get_all_material_cells()` as there you are looking for the actual distributed material.

3. EDIT (added bug fix in 2nd commit): When creating an `openmc.MGXSLibrary` instance from an `openmc.mgxs.Library` class with MGXS tallied by nuclide, the nuclide name ended up being appended twice. First in `openmc.mgxs.Library.create_mg_library(...)`, and second in `openmc.mgxs.get_xsdata(...)`. I therefore removed the `openmc.mgxs.Library.create_mg_library(...)` appending o the nuclide name and let the lower level method handle it.

### DIFF
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -1265,6 +1265,7 @@ class Library:
                     i += 1
 
         else:
+            # import pdb; pdb.set_trace()
             # Create the xsdata object and add it to the mgxs_file
             for i, domain in enumerate(self.domains):
                 if self.by_nuclide:
@@ -1277,8 +1278,6 @@ class Library:
                         xsdata_name = 'set' + str(i + 1)
                     else:
                         xsdata_name = xsdata_names[i]
-                    if nuclide != 'total':
-                        xsdata_name += '_' + nuclide
 
                     xsdata = self.get_xsdata(domain, xsdata_name,
                                              nuclide=nuclide, xs_type=xs_type)
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -1265,7 +1265,6 @@ class Library:
                     i += 1
 
         else:
-            # import pdb; pdb.set_trace()
             # Create the xsdata object and add it to the mgxs_file
             for i, domain in enumerate(self.domains):
                 if self.by_nuclide:
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -192,7 +192,9 @@ class Library:
         if self._domains == 'all':
             if self.domain_type == 'material':
                 return list(self.geometry.get_all_materials().values())
-            elif self.domain_type in ['cell', 'distribcell']:
+            elif self.domain_type == 'cell':
+                return list(self.geometry.get_all_cells().values())
+            elif self.domain_type in 'distribcell':
                 return list(self.geometry.get_all_material_cells().values())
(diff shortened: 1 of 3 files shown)


paulromano:
So in practice, the only thing that the min/max temperature are used for is a check in `Cell::set_temperature` when temperature interpolation is being used. It may seem weird, but it's actually intentional that the global minimum temperature is the maximum of the minimum temperatures of individual nuclides. This ensures that when a user/coupling code sets a temperature of a cell, it is guaranteed to have cross sections available at that temperature. If you want to guarantee that cross sections are available over a range of temperatures, we have a way to do that:
```Python
settings = openmc.Settings
settings.temperature = {'range': (300.0, 1500.0)}
```

nelsonag:
@paulromano Hmm, I see. The problem I was trying to fix is the misleading output after loading xs:

> Minimum neutron data temperature: 900.000000 K
> Maximum neutron data temperature: 900.000000 K
> 
 when the model had 600 and 900K data. This would lead to user concern of "wait, is the code grabbing my temperatures correctly?"

So, after undoing my 'fix', the options then are to:

1. Find some alternate scheme that also prints that data so the user can't misinterpret.
2. Delete that output.
3. Make that output more useful somehow (print specific xs temperatures loaded for each iso).
4. Do nothing.

3 is the best, but could get quite verbose. So doing it right would require a bit more than the scope of this hotfix (probably include in statepoint as that is our 'log file'). 
2 is reasonable since the value-add of this output doesnt seem to be very much, IMHO. 1 is also probably outside the scope. 4 is not unreasonable, but now I can't unsee the oddity so my personal preference is not this.

Thoughts? My preference is 2.

paulromano:
The other potential option here is to change how we decide on what temperatures to load. One reasonable approach would be to first scan all materials/cells for temperatures to get the overall range of temperatures in the problem, and then use that range when loading cross section data for all nuclides. That would result in more data being loaded, but I think it's probably closer to what one might expect the code to do. The `settings.temperature['range']` option would still give a way to override this if desired.

nelsonag:
Hmm, I'd prefer not to increase the data memory usage if possible.

It's sounding like I should revert that portion of this PR and create an
issue so we can progress with the rest of the PR. Ok with that?

paulromano:
@nelsonag yes, that sounds good.

paulromano:
Thanks @nelsonag!

PR #1568, opened 2020-05-20
Fixed electron energy grid in thick target bremsstrahlung approximation

NuclearEngideer:
Suggested:

The BREMX.DAT file has a coarse energy grid for electron energies from 1keV to 10GeV. In the current photon data parsing script, the grid is interpolated for 200 points between 1keV and 1GeV. I am suggesting using the entire range of the available data.


NuclearEngideer:
I didn't think to check what

NuclearEngideer:
I didn't think to check what happens in the physics when electrons have energy greater than 1GeV. This "suggestion" wouldn't mean anything if something else is used above 1GeV.

PR #1573, opened 2020-05-27
More surface constructors (great for externally linked mini-apps!)

gridley:
Hi all,

I recently was looking to play around with a modified particles-in-a-box eigenvalue algorithm, and was looking to save some work on writing the plane intersection and reflection routines. The constructors I've added here allow surfaces to be created in externally linked apps, e.g.:

```c++
// Create the box surfaces to simplify reflection stuff
openmc::SurfaceXPlane s_x1(box_lower_left.x);
openmc::SurfaceXPlane s_x2(box_upper_right.x);
openmc::SurfaceYPlane s_y1(box_lower_left.y);
openmc::SurfaceYPlane s_y2(box_upper_right.y);
openmc::SurfaceZPlane s_z1(box_lower_left.z);
openmc::SurfaceZPlane s_z2(box_upper_right.z);

std::vector<openmc::Surface*> surfaces {
  &s_x1,
  &s_x2,
  &s_y1,
  &s_y2,
  &s_z1,
  &s_z2};

...

  for (auto& part: particles) {

    // Skip dead particles
    if (part.wgt == 0) continue;

    // Get distances to boundaries
    int i_surf = 0;
    part.distance_to_bdry = openmc::INFTY;
    for (openmc::Surface* surf: surfaces) {
      double dist = surf->distance(part.r, part.u, part.surf_id==i_surf);
      if (dist < part.distance_to_bdry) {
        part.distance_to_bdry = dist;
        part.bdry_surf = i_surf;
      }
      i_surf++;
    }

```

Of course, I would completely understand it if the maintainers would prefer to not add code which isn't explicitly used in the main codebase. This is part of an overarching dream to generalize OpenMC's classes to work well as a Monte Carlo library that aids in research rather than simply being a monolithic program.

### DIFF
--- a/include/openmc/surface.h
+++ b/include/openmc/surface.h
@@ -196,7 +196,7 @@ public:
   explicit PeriodicSurface(pugi::xml_node surf_node);
 
   // Default constructor has no periodic surface associated
-  PeriodicSurface();
+  PeriodicSurface() : i_periodic_(C_NONE) {}
 
   //! Translate a particle onto this surface from a periodic partner surface.
   //! \param other A pointer to the partner surface in this periodic BC.
--- a/include/openmc/surface.h
+++ b/include/openmc/surface.h
@@ -195,6 +195,9 @@ public:
 
   explicit PeriodicSurface(pugi::xml_node surf_node);
 
+  // Default constructor has no periodic surface associated
+  PeriodicSurface();
+
   //! Translate a particle onto this surface from a periodic partner surface.
   //! \param other A pointer to the partner surface in this periodic BC.
   //! \param r A point on the partner surface that will be translated onto
@@ -218,6 +221,7 @@ class SurfaceXPlane : public PeriodicSurface
 {
 public:
   explicit SurfaceXPlane(pugi::xml_node surf_node);
+  explicit SurfaceXPlane(double x) : x0_(x) {}
   double evaluate(Position r) const;
   double distance(Position r, Direction u, bool coincident) const;
   Direction normal(Position r) const;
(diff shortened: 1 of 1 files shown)


gridley:
Actually, I'm just going to close this since it's unlikely anyone is going to go using this.

cjwyett:
Apologies for digging up a dead thread but I was hoping to see a response to this. The code may be useful. 

>  I would completely understand it if the maintainers would prefer to not add code which isn't explicitly used in the main codebase. This is part of an overarching dream to generalise OpenMC's classes to work well as a Monte Carlo library that aids in research rather than simply being a monolithic program.

Interesting comments, OpenMC could make a nice do-it-all MC library. On the other hand there are implementation and maintenance costs, and also the risk of the codebase becoming too convoluted.

pshriwise:
Thanks for reviving this discussion, @cjwyett. I meant to respond, but never got around to it. I’m personally for adding some constructors like these. We’re seeing more and more use of our C++ API externally and I like the capability this set of changes adds to that interface.

Regarding @gridley’s comment on adding code that isn’t used in the main codebase, as we encourage more use of our C++ API directly I think we’re starting to see a need for more direct testing of these interfaces to verify their behavior. It certainly does add some maintenance cost, but overall I think it’s worthwhile as long as the burden continues to be shared by out development community. It might be worth also having a discussion about what that testing framework might look like considering that we currently test nearly everything via our Python API and C-API interfaces.

pshriwise:
Along those lines, we’ve looked into automating the generation of a Python interface to our C++ interface using pybind. A student got started on that work last fall but no one has had time to continue it since then.

PR #1606, opened 2020-07-10
Added classes to mgxs.py

mkreher13:
Sam Shaner wrote a significant amount of code to implement a transient capability into OpenMC. However, since his code was never merged, it fell severely out of date. I have restored compatibility, and I am merging that code in incrementally. In this pull request, I have included SurfaceMGXS and Current classes in mgxs.py. 

The current requires a mesh domain and uses a MeshSurfaceFilter. I have tested the Current class on simple and complex geometries, and have also shown results of my code to match Sam Shaner's. Several regression tests (mgxs_library_condense for example) test all available MGXS types, and all passed after my addition. 

I would like @nelsonag to verify that _divide_by_density should be set to True in SurfaceMGXS since this is a new variable he just added to all the classes, and I wasn't 100% sure.


mkreher13:
My system didn't show any of these multigroup tests failing, as they are showing here, otherwise I would have tried to fix that before the pull request. I could use some help getting them running smoothly, since this my first time contributing to OpenMC. 

First of all, in the following tests, all the MGXS types are tested with a domain of 'material', however, the Current MGXS type can only be used on a 'mesh' domain: 
-mgxs_library_condense
-mgxs_library_distribcell (here the domain is distribcell, not material)
-mgxs_library_hdf5
-mgxs_library_no_nuclides
-mgxs_library_nuclides

Now the test mgxs_library_mesh also failed, even though the current tally should work with a mesh domain. So this tells me I may need to update the expected results of the test based on the new code. 

Let me know what you think.

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+        # tally data is stored in order of increasing energies
+        if order_groups == 'increasing':
+            xs = xs[..., ::-1, :, :]
+
+        if squeeze:
+            # We want to squeeze out everything but the polar, azimuthal,
+            # and energy group data.
+            xs = self._squeeze_xs(xs)
+
+        return xs
```
GiudGiud, line 6261: Instead of having this huge chunk of code repeating the get_xs in the MGXS class, just call the inherited one (with super) and reshape the returned xs to take into account the surfaces.

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+        if self.domain_type == 'mesh' and xs_type == 'micro':
+            msg = 'Unable to get micro xs for mesh domain since the mesh ' \
+                  'cells do not know the nuclide densities in each mesh cell.'
+            raise ValueError(msg)
+
+        filters = []
+        filter_bins = []
+
+        # Construct a collection of the domain filter bins
+        if not isinstance(subdomains, string_types):
```
paulromano, line 6175:
```suggestion
        if not isinstance(subdomains, str):
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
                              ' for xs_type other than "macro"')
+
+class SurfaceMGXS(MGXS,metaclass=ABCMeta):
+    """An abstract multi-group cross section for some energy group structure
+    on the surfaces of a mesh domain.
+    This class can be used for both OpenMC input generation and tally data
+    post-processing to compute surface- and energy-integrated multi-group cross
+    section for multi-group neutronics calculations.
+    NOTE: Users should instantiate the subclasses of this abstract class.
+    Parameters
```
paulromano, line 5908:
```suggestion
    NOTE: Users should instantiate the subclasses of this abstract class.

    Parameters
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+    domain_type : {'mesh'}
+        The domain type for spatial homogenization
+    energy_groups : openmc.mgxs.EnergyGroups
+        The energy group structure for energy condensation
+    by_nuclide : bool
+        If true, computes cross sections for each nuclide in domain
+    name : str, optional
+        Name of the multi-group cross section. Used as a label to identify
+        tallies in OpenMC 'tallies.xml' file.
+    Attributes
```
paulromano, line 5921:
```suggestion
        tallies in OpenMC 'tallies.xml' file.

    Attributes
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+    @property
+    def filters(self):
+        group_edges = self.energy_groups.group_edges
+        energy_filter = openmc.EnergyFilter(group_edges)
+        mesh = _DOMAIN_TO_FILTER[self.domain_type](self.domain).mesh
+        meshsurface_filter = openmc.MeshSurfaceFilter(mesh)
+        filters = [[meshsurface_filter, energy_filter]]
+
+        return self._add_angle_filters(filters)
+
```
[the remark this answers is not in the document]
paulromano:
```suggestion
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+    necessary data to compute multi-group cross sections from a
+    :class:`openmc.StatePoint` instance. The derived multi-group cross section
+    can then be obtained from the :attr:`TotalXS.xs_tally` property.
+    For a spatial domain :math:`V` and energy group :math:`[E_g,E_{g-1}]`, the
+    total cross section is calculated as:
+    .. math::
+       \frac{\int_{r \in V} dr \int_{4\pi} d\Omega \int_{E_g}^{E_{g-1}} dE \;
+       \sigma_t (r, E) \psi (r, E, \Omega)}{\int_{r \in V} dr \int_{4\pi}
+       d\Omega \int_{E_g}^{E_{g-1}} dE \; \psi (r, E, \Omega)}.
+    Parameters
```
paulromano, line 6284:
```suggestion
       d\Omega \int_{E_g}^{E_{g-1}} dE \; \psi (r, E, \Omega)}.

    Parameters
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+    domain_type : {'material', 'cell', 'distribcell', 'universe', 'mesh'}
+        The domain type for spatial homogenization
+    groups : openmc.mgxs.EnergyGroups
+        The energy group structure for energy condensation
+    by_nuclide : bool
+        If true, computes cross sections for each nuclide in domain
+    name : str, optional
+        Name of the multi-group cross section. Used as a label to identify
+        tallies in OpenMC 'tallies.xml' file.
+    Attributes
```
paulromano, line 6297:
```suggestion
        tallies in OpenMC 'tallies.xml' file.

    Attributes
```

openmc/mgxs/mgxs.py
```diff
@@ -6,6 +7,7 @@
 
 import h5py
 import numpy as np
+from six import string_types
```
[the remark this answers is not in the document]
mkreher13: You're right, it's not needed for Current. I'll remove it for now. I'll need it later when I merge the DiffusionCoefficient class.

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+    sparse : bool
+        Whether or not the MGXS' tallies use SciPy's LIL sparse matrix format
+        for compressed data storage
+    loaded_sp : bool
+        Whether or not a statepoint file has been loaded with tally data
+    derived : bool
+        Whether or not the MGXS is merged from one or more other MGXS
+    hdf5_key : str
+        The key used to index multi-group cross sections in an HDF5 data store
+    """
```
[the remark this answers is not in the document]
mkreher13: All the attributes of the new classes are present in the other existing classes as well. Are there specific ones that are no longer needed?

openmc/mgxs/mgxs.py
mkreher13: I feel this could go either way. There are 3 other classes that redefine get_xs() to change the function in some ways, and not others. So to follow the same format, I would be inclined to keep this function since it is not identical. 
In this case 1/3 of the function is different, so I could justify leaving it as is.

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+    Parameters
+    ----------
+    domain : openmc.RegularMesh
+        The domain for spatial homogenization
+    domain_type : {'mesh'}
+        The domain type for spatial homogenization
+    energy_groups : openmc.mgxs.EnergyGroups
+        The energy group structure for energy condensation
+    by_nuclide : bool
+        If true, computes cross sections for each nuclide in domain
```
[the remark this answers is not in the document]
mkreher13: You're right, in fact an error occurs if by_nuclide = True, since the domain is a mesh, and a mesh doesn't have nuclides. 
I will change it so that by_nuclide is inherited from MGXS, but raises an explicit error if it's True. Then I can eliminate it as a parameter, attribute, and in all the if statements of the SurfaceMGXS class and Current class.

openmc/mgxs/mgxs.py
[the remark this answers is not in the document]
GiudGiud: Nope you're right. I missed them somehow.

openmc/mgxs/mgxs.py
GiudGiud: It could. If you remove the code for the per_nuclide and micro options then I'm fine with it, it wont be as much repeated code

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,448 @@ def get_units(self, xs_type='macro'):
...
+
+    # Store whether or not the number density should be removed for microscopic
+    # values of this data
+    _divide_by_density = False
+
+    def __init__(self, domain=None, domain_type=None, energy_groups=None,
+                 by_nuclide=False, name=''):
+        super(SurfaceMGXS, self).__init__(domain, domain_type, energy_groups,
+                                          by_nuclide, name)
+
```
GiudGiud, line 5988:
```suggestion
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,448 @@ def get_units(self, xs_type='macro'):
...
+        # Use tally "slicing" to ensure that tallies correspond to our domain
+        # NOTE: This is important if tally merging was used
+        if self.domain_type == 'mesh':
+            filters = []
+            filter_bins = []
+        elif self.domain_type != 'distribcell':
+            filters = [_DOMAIN_TO_FILTER[self.domain_type]]
+            filter_bins = [(self.domain.id,)]
+        # Distribcell filters only accept single cell - neglect it when slicing
+        else:
```
GiudGiud, line 6067: this case is explicitly forbidden by the previous loop

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,448 @@ def get_units(self, xs_type='macro'):
...
+
+        # Clear any tallies previously loaded from a statepoint
+        if self.loaded_sp:
+            self._tallies = None
+            self._xs_tally = None
+            self._rxn_rate_tally = None
+            self._loaded_sp = False
+
+        # Make a list of the surface bins
+        #surface_bins = list(range(1, 4 * self.domain.n_dimension + 1))
```
GiudGiud, line 6079: delete if not needed

openmc/mgxs/mgxs.py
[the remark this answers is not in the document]
GiudGiud: Good point.
I don't actually know whether an attribute that is both unused and should not be used should be present in the docstrings. @paulromano or @nelsonag  do you have an opinion on this ?

nelsonag:
Hi @mkreher13! Thanks for the changes, I have been lax in getting to this PR so far. I will try and take a deep dive soon. In the meantime I have a high-level question to start off with: As @GiudGiud pointed out, a current is fundamentally different than a multigroup cross section; I fear this could lead to confusion in using the interface. That is, "why do I need a current score and a current MGXS class?" or "why is there not a Flux MGXS type"? Can you put this into context of other changes coming? Is this as part of a step to generate nodal discontinuity factors and the diffusion coefficient adjustments for CMFD?

mkreher13:
I see what you mean. The context is getting all the prerequisites ready for the python kinetics code. The kinetics code runs Monte Carlo & CMFD together to implement the frequency transform method for running transients. In order to get all the right multi-group cross sections, currents, and diffusion coefficients, I need to:
1) include Current, as a SurfaceMGXS
2) include DiffusionCoefficient, as a TransportMGXS
3) include the kinetics code
4) include examples of transient input files 
From there, my plan is to use OpenMC to further develop the kinetics code to include thermal feedback. 

That's the whole picture. Thoughts?

GiudGiud:
I suppose the current class is used to compute the nonlinear/correction diffusion coefficients that are present in the CMFD. These diffusion coefficients feel like a good (better?) instance of a surfaceMGXS. They could be tallied directly. However, the implementation could be a little odd & difficult within the MGXS framework since they require fluxes from cells on both sides of the surface, geometrical information about the size of the cell and so on.

mkreher13:
While there might be better ways to organize the Current and Diffusion Coefficient calculations here, altering the way they are calculated will upend how Sam's kinetics code works. He wrote an impressive suite for transient capabilities, and if the documentation is clear about the differences between classic MGXS and these new ones, I think they can be used effectively to add a whole new functionality to OpenMC. I suggest they can be improved upon and re-organized later, but until they get initially merged, we're going to be working with transient code that is constantly becoming deprecated.

openmc/mgxs/mgxs.py
mkreher13: Can you please elaborate? The distribcell domain are allowed in the previous loop.

GiudGiud: My bad, didn't see.

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,444 @@ def get_units(self, xs_type='macro'):
         else:
             raise ValueError('Unable to return the units of InverseVelocity'
                              ' for xs_type other than "macro"')
+
+
+class SurfaceMGXS(MGXS):
```
nelsonag, line 5900: Thoughts on renaming to `MeshSurfaceMGXS` to be consistent with the nomenclature we use elsewhere for these surface tallies?

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,444 @@ def get_units(self, xs_type='macro'):
...
+        if self.loaded_sp:
+            self._tallies = None
+            self._xs_tally = None
+            self._rxn_rate_tally = None
+            self._loaded_sp = False
+
+        # Find, slice and store Tallies from StatePoint
+        # The tally slicing is needed if tally merging was used
+        for tally_type, tally in self.tallies.items():
+            #surface_filter = openmc.SurfaceFilter(surface_bins)
```
nelsonag, line 6080: why commented out? if not needed, remove (same on next line)

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,444 @@ def get_units(self, xs_type='macro'):
...
+            xs = self._squeeze_xs(xs)
+
+        return xs
+
+
+class Current(SurfaceMGXS):
+    r"""A current multi-group cross section.
+    This class can be used for both OpenMC input generation and tally data
+    post-processing to compute spatially-homogenized and energy-integrated
+    multi-group total cross sections for multi-group neutronics calculations. At
```
nelsonag, line 6242: Looks like this docstring was taken from TotalXS and not updated for Current. Fix.

openmc/mgxs/mgxs.py
```diff
@@ -507,7 +508,8 @@ def tallies(self):
                 self._tallies[key] = openmc.Tally(name=self.name)
                 self._tallies[key].scores = [score]
                 self._tallies[key].estimator = estimator
-                self._tallies[key].filters = [domain_filter]
+                if score != 'current':
+                    self._tallies[key].filters = [domain_filter]
```
nelsonag, line 512: I'm confused by this if block, where do you set the filter for the current tally?

mkreher13: In this block, every tally gets a domain-specific filter. Then later in the class that corresponds to that tally, the non-domain specific filters get added. 
In the case of the current, the domain is a mesh. However, the current does not operate with a MeshFilter, but a MeshSurfaceFilter. For that reason, I had to exclude it from adding the domain-specific filter (MeshFilter). I add the non-domain specific filter (MeshSurfaceFilter) in the SurfaceMGXS class at line 5994.
The other classes behave similarly. For example, MultiplicityMatrixXS adds its non-domain specific filters on line 4850, ProbabilityMatrix on line 5023, Chi on line 5380, etc.

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5895,444 @@ def get_units(self, xs_type='macro'):
...
+        An (optional) tally precision trigger given to each tally used to
+        compute the cross section
+    scores : list of str
+        The scores in each tally used to compute the multi-group cross section
+    filters : list of openmc.Filter
+        The filters in each tally used to compute the multi-group cross section
+    tally_keys : list of str
+        The keys into the tallies dictionary for each tally used to compute
+        the multi-group cross section
+    estimator : {'analog'}
```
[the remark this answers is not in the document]
mkreher13: Ok, I think I understand what you're saying. It was originally both options {'track length', 'analog'} but @GiudGiud requested it be changed to only analog. Should we change it back then since both are allowed?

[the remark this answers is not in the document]
GiudGiud: I think it would be confusing to say it's a tracklength estimator when it's clearly analog. 1. is the way to go.

GiudGiud:
Agreed on the capability being very important and worth merging, and having had to deal with old unmerged code for my PhD as well, I can feel the pain. The only concern is that it breaks the Liskov substitution principle in that a few of the attributes and functions of a MGXS are irrelevant for a SurfaceMGXS. 

This calls for a later rework of MGXS. Probably a split between VolumeMGXS and SurfaceMGXS with a MGXS main class. 
For now, as soon as this is tested and documented, I'd approve it.

paulromano:
@mkreher13 We are going to be making our next release imminently. Are you ok with this being merged after the release is made? That will give more time for testing and expanding the transient-related capabilities for a future release.

mkreher13:
@paulromano No problem. Does the status of this PR?

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5897,466 @@ def get_units(self, xs_type='macro'):
...
+        super(SurfaceMGXS, self).__init__(domain, domain_type, energy_groups,
+                                          by_nuclide, name)
+
+
+    @property
+    def scores(self):
+        return [self.rxn_type]
+
+    @property
+    def domain(self):
```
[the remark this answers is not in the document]
mkreher13: I had to put all the domain properties back in so that I could do a check_type on domain_type, make sure it's a mesh and raise a proper error.

mkreher13:
Alright, I got the testing up to date, so all the checks pass now. I think I got to all the requested changes, so just let me know if this is ready to be merged.

openmc/mgxs/mgxs.py
mkreher13: I think this is the last unresolved conversation. Do you think we should rename it or leave that for a later change ?

GiudGiud: I'd agree with the renaming. If I end up putting transport GET / discontinuity factors (which are defined on fuel outer surfaces) in mgxs I'll either rename it back or create another class of SurfaceMGXS.

mkreher13:
Is everything in order here with regard to the code? "Changes requested" is still red, so let me know what remains to be resolved. 
I can't figure out where to update documentation, so please point me in the right direction for that.

GiudGiud:
Have a look at
[mgxs.rst](https://github.com/openmc-dev/openmc/blob/213a33f42e6ac93fcd7db4ac7195004beae92696/docs/source/pythonapi/mgxs.rst#multi-group-cross-sections)

I think you need to explicitly add the current mgxs there. I'm not sure, but I think the mgxs module only shows up in the Python API tab and the examples in the documentation. So that would be all. Once you'll have merged the whole capability, a jupyter notebook example would be great. For now it's too early imo.

openmc/mgxs/mgxs.py
```diff
@@ -5959,13 +5959,9 @@ class SurfaceMGXS(MGXS):
         two to account for both the incoming and outgoing current from the
         mesh cell surfaces.
     num_nuclides : int
-        The number of nuclides for which the multi-group cross section is
-        being tracked. This is unity if the by_nuclide attribute is False.
+        Unused n SurfaceMGXS
```
GiudGiud: in*
Also it'd be better to be consistent with the previous lines "for" or "in"

tests/regression_tests/mgxs_library_distribcell/test.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, *args, **kwargs):
         self.mgxs_lib.by_nuclide = False
 
         # Test all MGXS types
-        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES + \
+        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES[:-1] + \
```
GiudGiud, line 26: Would prefer a loop that excludes the current mgxs explicitly. If someone adds a mgxs after, they'd have to update the -1 to -2.

tests/regression_tests/mgxs_library_mesh/results_true.dat
```diff
@@ -180,38 +180,38 @@
...
    mesh 1               group in nuclide    mean std. dev.
         x  y       surf                                   
-0       1  1  x-min out        1   total  0.0000  0.000000
-1       1  1   x-min in        1   total  0.0000  0.000000
-2       1  1  x-max out        1   total  0.2738  0.093735
 3       1  1   x-max in        1   total  0.1892  0.011302
-4       1  1  y-min out        1   total  0.0000  0.000000
-5       1  1   y-min in        1   total  0.0000  0.000000
-6       1  1  y-max out        1   total  0.2358  0.041204
+2       1  1  x-max out        1   total  0.2738  0.093735
```
GiudGiud, line 184: any reason the order of the bins is flipped ?

openmc/mgxs/mgxs.py
```diff
@@ -6239,14 +6239,17 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
                 df = df[df['group out'].isin(groups)]
 
         mesh_str = 'mesh {0}'.format(self.domain.id)
+        surfaces = df[(mesh_str,'surf')]
+        df.drop(columns=[(mesh_str,'surf')],inplace=True)
+        df.insert(len(self.domain.dimension),(mesh_str,'surf'),surfaces)
```
GiudGiud, line 6244: Use [pandas' shift](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.shift.html)

openmc/mgxs/mgxs.py
```diff
@@ -6239,14 +6239,17 @@ def get_pandas_dataframe(self, groups='all', nuclides='all',
                 df = df[df['group out'].isin(groups)]
 
         mesh_str = 'mesh {0}'.format(self.domain.id)
+        surfaces = df[(mesh_str,'surf')]
```
GiudGiud, line 6242:
```suggestion
        surfaces = df[(mesh_str, 'surf')]
```

tests/regression_tests/mgxs_library_condense/test.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, *args, **kwargs):
         self.mgxs_lib.by_nuclide = False
 
         # Test all MGXS types
-        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES + \
+        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES[:-1] + \
```
GiudGiud, line 22: Condensing a multigroup current to a coarser group structure is theoretically possible so it should be tested.

tests/regression_tests/mgxs_library_no_nuclides/test.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, *args, **kwargs):
         self.mgxs_lib.by_nuclide = False
 
         # Test all MGXS types
-        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES + \
+        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES[:-1] + \
```
GiudGiud, line 23: You could leave currents in that one too ?
Or is there an issue with the output format ? (dimensions of array?)

tests/regression_tests/mgxs_library_hdf5/test.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, *args, **kwargs):
         self.mgxs_lib.by_nuclide = False
 
         # Test all MGXS types
-        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES + \
+        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES[:-1] + \
```
GiudGiud, line 26: Would leave currents in here too.

GiudGiud:
The "get_xs" routine seems to not be tested, see [coveralls](https://coveralls.io/builds/32233056/source?filename=openmc%2Fmgxs%2Fmgxs.py#L6023).
I think it was covered by the regression test for the other classes.

tests/regression_tests/mgxs_library_hdf5/test.py
mkreher13: The problem here is that the domain type imposed is 'material', but current can only work on a 'mesh' domain.

tests/regression_tests/mgxs_library_no_nuclides/test.py
mkreher13: The problem here is that the domain type imposed is 'material', but current can only work on a 'mesh' domain.

tests/regression_tests/mgxs_library_condense/test.py
mkreher13: The problem here is that the domain type imposed is 'material', but current can only work on a 'mesh' domain.

tests/regression_tests/mgxs_library_mesh/results_true.dat
mkreher13: Can you please explain what you mean? What is flipped? It all seems to be in proper order to me.

tests/regression_tests/mgxs_library_no_nuclides/test.py
GiudGiud: Ok it's fine to drop that one.

tests/regression_tests/mgxs_library_hdf5/test.py
GiudGiud: If adapting the current unit test is not possible, I think you need to create another.
btw, _hdf5_key does not seem to be set for "current" (also "total" and "transport"). Can you please check what it defaults to in a hdf5 file?

tests/regression_tests/mgxs_library_condense/test.py
GiudGiud: Ok I'm fine with dropping that one. I can't think of a use case, CMFD current tallies should be done in the desired group structure from the get-go.

tests/regression_tests/mgxs_library_mesh/results_true.dat
GiudGiud: It goes 3,2,1,0,7,6,5,4. It also goes x-max before x-min and in before out.
The previous order seems to have been x-min before x-max and out before in.

mkreher13: In my previous commits, the order was arbitrary and in fact was inconsistent across different python versions in the test results. I was able to fix the order using sort_values(by= ... 'surf' ...), so the order is the one that will stick. It doesn't really matter if min is before max, as long as it is consistent across systems. 
The numbering on the left-hand-side doesn't seem to matter. What matters is imposing the order in the mesh, first by x, then y, then z (then 'surf' in the case of current). If you notice the other results also have the numbers listed in an unordered way. That's because the order is imposed in sort_values() in mgxs.

tests/regression_tests/mgxs_library_no_nuclides/test.py
mkreher13: get_xs() isn't tested anywhere for any of the MGXS types. All the tests rely on get_mgxs() from the library class. So two options:
1) I don't test get_xs() and remain consistent with the other classes
2) I add a test for get_xs() in every test since it was missing

openmc/mgxs/mgxs.py
mkreher13: Panda's shift requires knowing where the column is first, and then moving it where you want it. 
The issue here is that pandas creates the order arbitrarily. So I don't know where it shows up, and therefore I don't know how much to shift it or in what direction. 
So for the test results to match & pass, I had to impose that it's at the end. The best solution I found was to drop & insert it where I want. That way, I don't need to know where it was.

tests/regression_tests/mgxs_library_hdf5/test.py
mkreher13: It defaults to "current". Do I need to add that information anywhere? 

I think I will expand mgxs_library_mesh to include all the MGXS function tests. It's the only test that handles the mesh domain, so this way all the classes can be tested for how they handle a mesh domain in every function. This seems to be missing anyway.

mkreher13: I just extended mgxs_library_condense and mgxs_library_hdf5 to include current by modifying their domain from 'material' to 'mesh'. So now both those functionalities are tested for current.

tests/regression_tests/mgxs_library_no_nuclides/test.py
GiudGiud: get_xs isnt tested explicitly, but it was being called by one of the tests for every other MGXS types as you can see on coveralls. 
Now that you have changed the condense and hdf5 tests to allow the current MGXS, it is also tested for MeshSurfaceMGXS
Option 2, a unit test for get_xs() would be better. As you want, what you did is good for this PR

openmc/mgxs/mgxs.py
GiudGiud: Good point. Can you specify the column placement directly then ?
like df = df[[mesh_str, 'surf', other_column_name_1, other_column_name_2]]

tests/regression_tests/mgxs_library_mesh/results_true.dat
GiudGiud: Ok agreed sort_values() is the way to go.

openmc/mgxs/mgxs.py
GiudGiud: Approving the PR after that. @nelsonag @paulromano do you have more comments

mkreher13: It's a little more complicated than that because the column mesh_str is a dataframe itself which contains columns x, y, z, surf. 
The sub-columns of that mesh column can't be set easily 'inplace' (that is, without a copy, delete, replace) since it is embedded in df. It might be possible, but I spent some hours on it without success. There wasn't much literature on embedded data frames.

mkreher13: Is this piece of code going to be acceptable?

tests/regression_tests/mgxs_library_no_nuclides/test.py
```diff
@@ -19,8 +19,12 @@ def __init__(self, *args, **kwargs):
         self.mgxs_lib = openmc.mgxs.Library(self._model.geometry)
         self.mgxs_lib.by_nuclide = False
 
-        # Test all MGXS types
-        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES + \
+        # Test all relevant MGXS types
+        relevant_MGXS_TYPES = []
+        for item in openmc.mgxs.MGXS_TYPES:
+            if item != 'current':
+                relevant_MGXS_TYPES = relevant_MGXS_TYPES + [item]
```
paulromano, line 26:
```suggestion
        relevant_MGXS_TYPES = [item for item in openmc.mgxs.MGXS_TYPES
                               if item != 'current']
```

tests/regression_tests/mgxs_library_nuclides/test.py
```diff
@@ -17,8 +17,13 @@ def __init__(self, *args, **kwargs):
...
         self.mgxs_lib = openmc.mgxs.Library(self._model.geometry)
         self.mgxs_lib.by_nuclide = True
-        # Test all MGXS types
-        self.mgxs_lib.mgxs_types = openmc.mgxs.MGXS_TYPES
+
+        # Test relevant all MGXS types
+        relevant_MGXS_TYPES = []
+        for item in openmc.mgxs.MGXS_TYPES:
+            if item != 'current':
+                relevant_MGXS_TYPES = relevant_MGXS_TYPES + [item]
```
paulromano, line 25:
```suggestion
        relevant_MGXS_TYPES = [item for item in openmc.mgxs.MGXS_TYPES
                               if item != 'current']
```

openmc/mgxs/mgxs.py
```diff
@@ -5891,3 +5898,458 @@ def get_units(self, xs_type='macro'):
...
+        """
+
+        if not isinstance(groups, str):
+            cv.check_iterable_type('groups', groups, Integral)
+        cv.check_value('xs_type', xs_type, ['macro'])
+
+        df = self.xs_tally.get_pandas_dataframe(paths=paths)
+
+        # Remove the score column since it is homogeneous and redundant
+        if self.domain_type == 'mesh':
```
paulromano, line 6228: Is there ever a case where the domain type is not mesh?

tests/regression_tests/mgxs_library_hdf5/test.py
```diff
@@ -54,8 +62,8 @@ def _get_results(self, hash_output=False):
             for domain in self.mgxs_lib.domains:
                 for mgxs_type in self.mgxs_lib.mgxs_types:
                     outstr += 'domain={0} type={1}\n'.format(domain.id, mgxs_type)
-                    avg_key = 'material/{0}/{1}/average'.format(domain.id, mgxs_type)
-                    std_key = 'material/{0}/{1}/std. dev.'.format(domain.id, mgxs_type)
+                    avg_key = 'mesh/{0}/{1}/average'.format(domain.id, mgxs_type)
+                    std_key = 'mesh/{0}/{1}/std. dev.'.format(domain.id, mgxs_type)
```
paulromano, line 66:
```suggestion
                    avg_key = 'mesh/{}/{}/average'.format(domain.id, mgxs_type)
                    std_key = 'mesh/{}/{}/std. dev.'.format(domain.id, mgxs_type)
```

mkreher13:
The same system keeps timing out during the build/testing, without any other errors. Not sure how to fix or work around.

paulromano:
Thanks for persevering with this one @mkreher13. Sorry it took us so long to get it in!

PR #1612, opened 2020-07-16
Assign prev_results to correct member in TrasnportOperator's constructor

kingyue737:
a little more fix to keep user-supplied prev_results on operator #1610


kingyue737:
Thank you, @drewejohnson !

PR #1619, opened 2020-07-27
CMake MPI detection updates

RonRahaman:
In this PR, CMake will test for MPI by using the cache variable `CMAKE_CXX_COMPILER`, rather than the environment variable `CXX`.  

This will not change the behavior for anyone accustomed to using the env variable.  This is because CMake searches for usable compilers in this order and then sets the cache var:
1.  Already-set cache variables: (`CMAKE_CXX_COMPILER` from command line or GUI)
2. Environment variables (`$CXX`)
3. Common C++ compiler names in `$PATH`

This should fix a few minor problems.  For example, it will fix the behavior of `find(HDF5)` when re-running CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,8 +30,8 @@ option(dagmc    "Enable support for DAGMC (CAD) geometry"        OFF)
 #===============================================================================
 
 set(MPI_ENABLED FALSE)
-if($ENV{CXX} MATCHES "(mpi[^/]*|CC)$")
-  message(STATUS "Detected MPI wrapper: $ENV{CXX}")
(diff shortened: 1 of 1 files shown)


RonRahaman:
I also updated this to use `FindMPI` for CMake versions 3.10+.  In 3.10, the FindMPI module was greatly improved and is actually usable :) .  See the release notes:  https://cmake.org/cmake/help/v3.10/release/3.10.html#modules.

RonRahaman:
After some more testing, I see that the `FindMPI` has a few corner cases.  So I will put that on the back burner and propose just the first commit.

PR #1646, opened 2020-08-25
Fixing existing typos and errors in MGXS routines

mkreher13:
Over the course of my summer internship, I used a lot of new-to-me mgxs routines and identified several errors in the existing code. 
1) create_mg_mode didn't account for cells filled with a lattice instead of a material
2) get_xsdata referenced 'beta' but then pointed to 'nu-fission' - obviously a typo. Once fixed, worked just like the others. This routine is tested whenever create_mg_library is called, but not with delayed mgxs_types. 
3) XSdata class had beta, decay_rate, inverse_velocity listed in attributes, but missing from property definitions. I'm sure this class is tested somewhere, but obviously not covering those three attributes. 
4) Warning message for setting Legendre order greater than 0 was erroneous.

### DIFF
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -412,7 +412,7 @@ class Library:
             if self.correction == 'P0' and legendre_order > 0:
                 msg = 'The P0 correction will be ignored since the ' \
                       'scattering order {} is greater than '\
-                      'zero'.format(self.legendre_order)
+                      'zero'.format(legendre_order)
                 warn(msg, RuntimeWarning)
                 self.correction = None
         elif self.scatter_format == 'histogram':
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -1402,12 +1402,12 @@ class Library:
                 if self.domain_type == 'material':
                     # Fill all appropriate Cells with new Material
                     for cell in all_cells:
-                        if cell.fill.id == domain.id:
+                        if isinstance(cell.fill, openmc.Material) and cell.fill.id == domain.id:
                             cell.fill = material
 
                 elif self.domain_type == 'cell':
                     for cell in all_cells:
-                        if cell.id == domain.id:
+                        if isinstance(cell.fill, openmc.Material) and cell.id == domain.id:
                             cell.fill = material
 
         return mgxs_file, materials, geometry
--- a/openmc/mgxs/library.py
+++ b/openmc/mgxs/library.py
@@ -1407,7 +1407,11 @@ class Library:
 
                 elif self.domain_type == 'cell':
                     for cell in all_cells:
-                        if isinstance(cell.fill, openmc.Material) and cell.id == domain.id:
+                        if not isinstance(cell.fill, openmc.Material):
+                            warn('If the library domain includes a lattice or universe cell '
+                            'in conjunction with a consituent cell of that lattice/universe, '
+                            'the multi-group simulation will fail') 
+                        if cell.id == domain.id:
                             cell.fill = material
 
         return mgxs_file, materials, geometry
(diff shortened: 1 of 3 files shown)


openmc/mgxs/library.py
```diff
@@ -1402,12 +1402,12 @@ def create_mg_mode(self, xsdata_names=None, bc=['reflective'] * 6):
...
                     # Fill all appropriate Cells with new Material
                     for cell in all_cells:
-                        if cell.fill.id == domain.id:
+                        if isinstance(cell.fill, openmc.Material) and cell.fill.id == domain.id:
                             cell.fill = material
 
                 elif self.domain_type == 'cell':
                     for cell in all_cells:
-                        if cell.id == domain.id:
+                        if isinstance(cell.fill, openmc.Material) and cell.id == domain.id:
```
nelsonag, line 1410: This change makes it so only a cell that was filled with a material in the upstream model will receive the MGXS material. That doesnt make much sense to me.

If the user requested MGXS over a cell, then they will receive a single MGXS representing the homogenized region of whatever was in that cell, regardless of if it was a lattice, universe, material, etc. Therefore we shouldn't care what the upstream fill type was when assigning this material. Right?

mkreher13: When using create_mg_mode with a cell or material domain, the user is getting mgxs for every cell/material in the lattice, so why and how would they use the cross section for the lattice that contains multiple other cells/materials? If the user wanted a mgxs for a lattice, they would use a mesh domain where the lattice fit into a single mesh cell (i.e. 'smearing' the lattice materials into a single multi-group material).

Furthermore, using create_mg_mode with a lattice cell creates errors at runtime. If I run a problem that:
1) has a lattice cell
2) builds an mgxs library with domain_type = 'cell' or 'material'
3) uses create_mg_mode()
4) updates the settings to 'multi-group' and runs openmc
I get errors where OpenMC can't find the cell or material specified. 

I suggest you try running the OpenMC example "Multi-Group Mode Part II" with create_mg_mode, and let me know what you think.

nelsonag: You're correct that the materials are assigned incorrectly when running create_mg_mode in Multi-Group Mode Part II. The fix you propose on lines 1404/5 handles that correctly and I am great with merging that in.

However, I am talking about the problem where the tally domain is the cell which contains the lattice or universe. This can be recreated in the Multi-Group Mode Part II notebook by setting ``mgxs_lib.domains = [root_cell]`` with a ``cell`` domain type. I just verified that doing this worked correctly in the old code but is broken in the new.

To your first point, this is a frequent application and should be handled. For example, a user would want MGXS for the cell that contains a lattice if they are creating MGXS homogenized over an assembly.

mkreher13: Ok, I see! So we need a fix where it allows the user to set domain = get_all_cells().values (which currently doesn't work if one of the cells is a lattice/universe cell), but also works if the domain is expressly supposed to be a lattice/universe cell.

mkreher13: The issue arises if your domain includes a lattice cell and a cell that is part of that lattice. Example: 
domain = [root_cell] works
domain = [fuel_cell] works
domain = [root_cell, fuel_cell] doesn't work. Mixing lattice & constituent cells together causes the simulation to fail. Understandably so, since it creates ambiguity about which cross section to use. 

Using geometry.get_all_cells().values can easily put the user in this situation. But this is technically user error, I guess? In this case, a better warning/error message might be a sufficient fix.

nelsonag: I think that's sufficient. Thanks!

nelsonag:
Thanks again @mkreher13, looks great to me!

PR #1650, opened 2020-09-03
Examine if the region exists before removing redundant surfaces

kingyue737:
It's not necessary for cell to have a region (for example an assembly cell nested in a core lattice). This PR prevents errors when removing redundant surfaces for geometry with cells whose region is None.


paulromano (approved):
Thanks for the fix @kingyue737!

PR #1651, opened 2020-09-03
Fix plotting of individual universe levels.

pshriwise:
Attempting to plot a specific universe level is risky right now after the changes in #1201. If the universe level is higher than the number of coordinates for the particle at a given pixel location in the image, it's possible that an invalid access in the particle coordinates can occur, resulting in a segfault. The relevant chunk of code is here:

https://github.com/openmc-dev/openmc/blob/828e0aa7dfc2e9774d8ff7532e5269723e3b6e55/src/plot.cpp#L63-L71

This PR corrects that by setting any pixel color to `NOT_FOUND` when the coordinate level of the particle is lower than the requested universe level. For density and temperature plots, the particle's coordinate level is always used.

This also exposes the `model::n_coord_level` variable in the CAPI and the `openmc.lib` module to allow our [plotting application](https://github.com/openmc-dev/plotter) to correctly populate the number of available universe levels.

### DIFF
--- a/include/openmc/geometry.h
+++ b/include/openmc/geometry.h
@@ -18,7 +18,7 @@ namespace openmc {
 namespace model {
 
 extern int root_universe;  //!< Index of root universe
-extern int n_coord_levels; //!< Number of CSG coordinate levels
+extern "C" int n_coord_levels; //!< Number of CSG coordinate levels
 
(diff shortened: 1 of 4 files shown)


paulromano (approved):
Thanks @pshriwise!

PR #1654, opened 2020-09-07
Criticality search: return the minimum critical value

cjwyett:
I've had a few experiences using `search_for_keff` where one of the guesses is closer to 1 than the final guess. This PR adjusts zero_value to correspond to the keff that is: 
a) critical (i.e. above 1) 
b) closest to sub-critical.

A better approach might be to return the value closest to the target.


paulromano:
Thanks for the proposed change @cjwyett. I think rather than making an ad hoc change in the existing implementation, we ought to look into replacing it with a better algorithm. I've seen a number of papers that propose better algorithms than what we currently do. If you're interested in taking a shot at this, let me know and we can create a new issue to discuss it.

cjwyett:
Cheers @paulromano. I'd be delighted to work on this (after an impending deadline). Would be great to discuss and have a bit of a think, so please go ahead with the issue.

cjwyett:
#1656

cjwyett:
I have a few months free to really focus on this now. Will start coding shortly.

cjwyett:
Apologies for not prioritising this. 

Here's a slightly modified version of the Morrow (2007) ABA. Still a WIP.  

I would appreciate some further instruction on how to proceed. As far as I can tell there are two choices:

**1.  Improve the adaptive batching algorithm**

TODO
- Use restart files to save time when n_a is changed.
- Add a verbosity argument for debugging
- Change n_a by the amount prescribed by morrow (instead of arbitrarily doubling it)
- Ways to handle weird reactivity coefficients
-  Create tests; will need to prepare for different use cases (control rods, molten salts, critical enrichment, critical radius, etc)
- Update documentation.

**2. Find a better algorithm** 
The RMC algos look very promising and deserve investigation. 

My personal recommendation would be to not merge this. The algorithm itself is highly unstable, especially around the target keff. Sources of error include inappropriate eigenvalue parameters (too easy to do) and non-linear reactivity coefficients. Let's have a go at the [perturbation](https://www.sciencedirect.com/science/article/pii/S0306454914005404) method.

paulromano (commented):
@cjwyett Thanks for starting some work here to improve the criticality search. I agree with you that we probably don't want to proceed with the method from the 2007 Morrow paper. I would actually recommend that we implement the regula falsi method described in the 2013 Gill paper (if you don't have a copy, shoot me a message at promano@anl.gov and I can send it to you). The perturbation methods are nice, but they require that you're able to calculate the derivative with respect to the parameter of interest directly in a tally. While we do have the ability to do tally derivatives, they are limited in scope and wouldn't have the same level of generality as the `search_for_keff` function currently does.

In the case where we want to increase the number of batches (as required for both the Morrow and Gill methods), the most efficient way to do this would be to use the `openmc.lib` bindings to the C/C++ API, which are lightly documented [here](https://docs.openmc.org/en/latest/pythonapi/capi.html). These bindings allow you to run a calculation for a certain number of batches, perform some logic, and then continue running more batches, e.g.
```Python
openmc.lib.init()
openmc.lib.simulation_init()

# Run N batches
for _ in range(N):
    openmc.lib.next_batch()

# Do something interesting, like getting k-effective and then figuring out how many more batches to run
k = openmc.lib.keff()
...
more_batches = ...

# Run more batches
for _ in range(more_batches):
    openmc.lib.next_batch()

...

# finish up
openmc.lib.finalize()
```
Note that this also solves the issue of using a restart file since you never actually have to restart. You're just putting some logic in the middle of the transport simulation. As a bonus, you also save on the time having to reload cross sections at initialization.

cjwyett:
Cheers @paulromano. `openmc.lib` is definitely a better way to go about this. I sent you an email regarding the regula falsi paper, hopefully this will be a more successful venture.

nelsonag:
Hello @cjwyett, I am working on the path laid out in #1859, and the next step would be upgrading the critical search capability. Is the work discussed herein still active and something I should plan on using or should I pick it up?
Thanks!

cjwyett:
Hi @nelsonag, sincere apologies for the delay, I've been really busy testing industry and then coming back for my PhD. 

I would really like to finish this. When will you need it by? If it's urgent I can try and knock something together this week. Otherwise I can take a really good long look at it w/c 18th

nelsonag:
That’s perfectly acceptable; just wanted to sync up and avoid double work/taking something you were interested in! No need for the urgency of this week.

cjwyett:
Apologies again, I'll definitely have some time to look at this over Chrimbo

nelsonag:
Sounds good and thanks for taking a look, Cyrus!

PR #1655, opened 2020-09-08
Implementation of WMP library generation from ENDF

liangjg:
This PR tries to implement WMP library generation workflow in OpenMC, using Vector Fitting approach.

You can now generate a WMP library from ENDF file with the new `openmc.data.WindowedMultipole.from_endf` method.
``` python
endf_file = '/opt/nucdata/endf-b-vii.1/neutrons/n-001_H_001.endf'

# generate a library with specified windowing size and curve fitting order
nuc = openmc.data.WindowedMultipole.from_endf(endf_file, wmp_options={"n_win": 400, "n_cf":3})

# Or, let OpenMC search from a range of window size and CF order
# and select the optimal WMP library (using built-in performance metric)
nuc_opt = openmc.data.WindowedMultipole.from_endf(endf_file, wmp_options={"search": True})
```

Through this approach, any nuclide with 0K point-wise data can be converted to WMP format, no resonance parameters are required. 
To achieve this, several steps are integrated:
- run NJOY on the ENDF file to get 0K ACE and extract relevant point-wise data (up to RRR boundary or the lowest threshold)
- perform Vector Fitting on point-wise data to get multipole parameters. This relies on an independent package `vectfit` (https://github.com/mit-crpg/vectfit).
- do windowing on the vector-fitted poles to get the WMP representation. As shown above, an automatic optimization procedure is implemented to get the best performance library by searching among different windowing parameters.

It should be noted that for complex resonant nuclides like U238, straightforward vector fitting is not doable because of both the time cost and memory constraints with the huge matrix operations. An energy domain decomposition approach is adopted in the workflow, i.e. carrying out the Vector Fitting and windowing on partitioned energy pieces. This makes the approach practical. I just managed to process a whole WMP library from ENDFB/VIII.0 with all 556 nuclides. The processing time for different nuclides varies widely, from seconds to tens of hours, depending on the complexity of the point-wise data.

For this PR, in addition to the new functions in `openmc.data.WindowedMultipole`, I was thinking if it is necessary to add some test cases on these functions. A little trouble here is I may have to install [`vectfit` package](https://github.com/mit-crpg/vectfit) on Travis, which relies on `conda`. Note currently 'import vectfit' is done when the relevant functions are called so 'import openmc' doesn't rely on `vectfit` package by default.

### DIFF
--- a/openmc/data/multipole.py
+++ b/openmc/data/multipole.py
@@ -175,7 +175,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
 
     """
 
-    # import vectfit package: https://github.com/mit-crpg/vectfit
+    # import vectfit package: https://github.com/liangjg/vectfit
     import vectfit as vf
 
     ne = energy.size
@@ -568,7 +568,7 @@ def _windowing(mp_data, rtol=1e-3, atol=1e-5, n_win=None, n_cf=None,
 
     """
 
-    # import vectfit package: https://github.com/mit-crpg/vectfit
+    # import vectfit package: https://github.com/liangjg/vectfit
     import vectfit as vf
 
     # unpack multipole data
--- a/openmc/data/multipole.py
+++ b/openmc/data/multipole.py
@@ -448,23 +448,19 @@ def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
     E_min, E_max = energy[0], energy[-1]
     n_points = energy.size
     total_xs = nuc_ce[1].xs['0K'](energy)
-    if 2 in nuc_ce:
+    try:
         elastic_xs = nuc_ce[2].xs['0K'](energy)
-    else:
+    except:
         elastic_xs = np.zeros_like(total_xs)
-    if 27 in nuc_ce:
+    try:
         absorption_xs = nuc_ce[27].xs['0K'](energy)
-    else:
-        if 101 in nuc_ce:
-            absorption_xs = nuc_ce[101].xs['0K'](energy)
-            if 18 in nuc_ce:
-                absorption_xs += nuc_ce[18].xs['0K'](energy)
-        else:
-            absorption_xs = np.zeros_like(total_xs)
+    except:
+        absorption_xs = np.zeros_like(total_xs)
     fissionable = False
-    if 18 in nuc_ce:
+    try:
         fission_xs = nuc_ce[18].xs['0K'](energy)
         fissionable = True
+    except:pass
 
     # make vectors
     if fissionable:
--- a/openmc/data/multipole.py
+++ b/openmc/data/multipole.py
@@ -203,7 +203,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
     test_s = np.sqrt(test_energy)
     weight = 1.0/f
 
-    # very small cross sections can lead to huge weights, which will harm the 
+    # very small cross sections can lead to huge weights, which will harm the
     # fitting accuracy
     MIN_CROSS_SECTION = 1e-7
     for i in range(nmt):
@@ -631,7 +631,7 @@ def _windowing(mp_data, rtol=1e-3, atol=1e-5, n_win=None, n_cf=None,
         inbegin = sqrt(E_min) + spacing * iw
         inend = inbegin + spacing
         incenter = (inbegin + inend) / 2.0
-        # extend window energy range for Doppler broadening 
+        # extend window energy range for Doppler broadening
         if iw == 0 or sqrt(alpha)*inbegin < 4.0:
             e_start = inbegin**2
         else:
@@ -1024,7 +1024,7 @@ class WindowedMultipole(EqualityMixin):
             Dictionary of keyword arguments, e.g. {'rtol':0.001, 'log':True},
             passed to :func:`openmc.data.multipole._vectfit_nuclide`
         wmp_options : dict
-            Dictionary of keyword arguments, e.g. {'search': True, 'log': True},
+            Dictionary of keyword arguments, e.g. {'search':True, 'log': True},
             passed to :func:`openmc.data.WindowedMultipole.from_multipole`
 
         Returns
--- a/openmc/data/multipole.py
+++ b/openmc/data/multipole.py
@@ -32,7 +32,7 @@ _FIT_S = 0       # Scattering
 _FIT_A = 1       # Absorption
 _FIT_F = 2       # Fission
 
-# Upper temperature limit
+# Upper temperature limit (K)
 TEMPERATURE_LIMIT = 3000
 
 # Logging control
@@ -190,7 +190,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
     ne_test = (ne - 1)*n_finer + 1
     test_energy = np.interp(np.arange(ne_test),
                             np.arange(ne_test, step=n_finer), energy)
-    test_energy[[0, -1]] = energy[[0, -1]] # avoid numerical issue
+    test_energy[[0, -1]] = energy[[0, -1]]  # avoid numerical issue
     test_xs_ref = np.zeros((nmt, ne_test))
     for i in range(nmt):
         test_xs_ref[i] = np.interp(test_energy, energy, ce_xs[i])
@@ -235,7 +235,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
 
     # perform VF with increasing orders
     found_ideal = False
-    n_discarded = 0 # for accelation, number of discarded searches
+    n_discarded = 0  # for accelation, number of discarded searches
     best_quality = best_ratio = -np.inf
     for i, order in enumerate(orders):
         if log:
@@ -275,7 +275,8 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
             # assess the result on test grid
             test_xs = vf.evaluate(test_s, new_poles, residues) / test_energy
             abserr = np.abs(test_xs - test_xs_ref)
-            relerr = abserr / test_xs_ref
+            with np.errstate(invalid='ignore'):
+                relerr = abserr / test_xs_ref
             if np.any(np.isnan(abserr)):
                 maxre, ratio, ratio2 = np.inf, -np.inf, -np.inf
             elif np.all(abserr <= atol):
@@ -302,7 +303,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
 
             if quality > best_quality:
                 if log >= DETAILED_LOGGING:
-                    print("  Best by far!")
+                    print("  Best so far!")
                 found_better = True
                 best_quality, best_ratio = quality, ratio
                 best_poles, best_residues = new_poles, residues
@@ -385,7 +386,8 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
 
     return (mp_poles, mp_residues)
 
-def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
+
+def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_pieces=None,
                      log=False, path_out=None, mp_filename=None, **kwargs):
     r"""Generate multipole data for a nuclide from ENDF.
 
@@ -395,8 +397,6 @@ def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
         Path to ENDF evaluation
     njoy_error : float, optional
         Fractional error tolerance for processing point-wise data with NJOY
-    vf_error : float, optional
-        Fractional error tolerance for data fitting
     vf_pieces : integer, optional
         Number of equal-in-momentum spaced energy pieces for data fitting
     log : bool or int, optional
@@ -515,8 +515,8 @@ def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
         e_end_idx = np.searchsorted(energy, e_end, side='left') + 1
         e_idx = range(e_start_idx, min(e_end_idx + 1, n_points))
 
-        p, r = _vectfit_xs(energy[e_idx], ce_xs[:, e_idx], mts, rtol=vf_error,
-                           log=log, path_out=path_out, **kwargs)
+        p, r = _vectfit_xs(energy[e_idx], ce_xs[:, e_idx], mts, log=log,
+                           path_out=path_out, **kwargs)
 
         poles.append(p)
         residues.append(r)
@@ -542,6 +542,7 @@ def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
             print("Dumped multipole data to file: {}".format(mp_filename))
 
     return mp_data
+
 
 def _windowing(mp_data, n_cf, rtol=1e-3, atol=1e-5, n_win=None, spacing=None,
                log=False):
@@ -595,7 +596,6 @@ def _windowing(mp_data, n_cf, rtol=1e-3, atol=1e-5, n_win=None, spacing=None,
             n_win = int((sqrt(E_max) - sqrt(E_min)) / spacing)
             E_max = (sqrt(E_min) + n_win*spacing)**2
         else:
-            warnings.warn("Number of windows or window size not specified.")
             n_win = 1000
     # inner window size
     spacing = (sqrt(E_max) - sqrt(E_min)) / n_win
@@ -671,7 +671,8 @@ def _windowing(mp_data, n_cf, rtol=1e-3, atol=1e-5, n_win=None, spacing=None,
 
             # assess the result
             abserr = np.abs(xs_fit + xs_wp - xs_ref)
-            relerr = abserr / xs_ref
+            with np.errstate(invalid='ignore'):
+                relerr = abserr / xs_ref
             if not np.any(np.isnan(abserr)):
                 re = relerr[abserr > atol]
                 if re.size == 0 or np.all(re <= rtol) or \
@@ -702,7 +703,7 @@ def _windowing(mp_data, n_cf, rtol=1e-3, atol=1e-5, n_win=None, spacing=None,
         poles_unused[i_piece][lp:rp] = 0
 
     # flatten and shrink by removing unused poles
-    data = [] # used poles and residues
+    data = []  # used poles and residues
     for ip in range(n_pieces):
         used = (poles_unused[ip] == 0)
         # stack poles and residues for library format
@@ -736,6 +737,7 @@ def _windowing(mp_data, n_cf, rtol=1e-3, atol=1e-5, n_win=None, spacing=None,
     wmp.broaden_poly = np.ones((n_win,), dtype=bool)
 
     return wmp
+
 
 class WindowedMultipole(EqualityMixin):
     """Resonant cross sections represented in the windowed multipole format.
@@ -1010,18 +1012,20 @@ class WindowedMultipole(EqualityMixin):
         return out
 
     @classmethod
-    def from_endf(cls, endf_file, vf_options=None, wmp_options=None):
+    def from_endf(cls, endf_file, log=False, vf_options=None, wmp_options=None):
         """Generate windowed multipole neutron data from an ENDF evaluation.
 
         Parameters
         ----------
         endf_file : str
             Path to ENDF evaluation
-        vf_options : dict
-            Dictionary of keyword arguments, e.g. {'rtol': 0.001, 'log': True},
+        log : bool or int, optional
+            Whether to print running logs (use int for verbosity control)
+        vf_options : dict, optional
+            Dictionary of keyword arguments, e.g. {'njoy_error': 0.001},
             passed to :func:`openmc.data.multipole._vectfit_nuclide`
-        wmp_options : dict
-            Dictionary of keyword arguments, e.g. {'search': True, 'log': True},
+        wmp_options : dict, optional
+            Dictionary of keyword arguments, e.g. {'search': True, 'rtol': 0.01},
             passed to :func:`openmc.data.WindowedMultipole.from_multipole`
 
         Returns
@@ -1037,6 +1041,10 @@ class WindowedMultipole(EqualityMixin):
 
         if wmp_options is None:
             wmp_options = {}
+
+        if log:
+            vf_options.update(log=log)
+            wmp_options.update(log=log)
 
         # generate multipole data from EDNF
         mp_data = _vectfit_nuclide(endf_file, **vf_options)
@@ -1106,7 +1114,7 @@ class WindowedMultipole(EqualityMixin):
                            wmp.n_windows * 0.01)
                 if best_wmp is None or metric > best_metric:
                     if log:
-                        print("Best library by far.")
+                        print("Best library so far.")
                     best_wmp = deepcopy(wmp)
                     best_metric = metric
 
--- a/openmc/data/multipole.py
+++ b/openmc/data/multipole.py
@@ -5,6 +5,7 @@ import os
 import h5py
 import numpy as np
 from scipy.signal import find_peaks
+import numpy.polynomial.polynomial as poly
 
 import openmc.checkvalue as cv
 from ..exceptions import DataError
@@ -86,11 +87,11 @@ def _broaden_wmp_polynomials(E, dopp, n):
 
     Parameters
     ----------
-    E : Real
+    E : float
         Energy to evaluate at.
-    dopp : Real
+    dopp : float
         sqrt(atomic weight ratio / kT) in units of eV.
-    n : Integral
+    n : int
         Number of components to the polynomial.
 
     Returns
@@ -137,7 +138,7 @@ def _broaden_wmp_polynomials(E, dopp, n):
 
 def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
                 n_vf_iter=30, log=False, path_out=None, **kwargs):
-    r"""Generate multipole data from point-wise cross sections.
+    r"""Convert point-wise cross section to multipole data via Vector Fitting.
 
     Parameters
     ----------
@@ -145,13 +146,13 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
         Energy array
     ce_xs : np.ndarray
         Point-wise cross sections to be fitted
-    mts : Iterable of Integral
+    mts : Iterable of int
         Reaction list
-    rtol : Real, optional
+    rtol : float, optional
         Relative error tolerance
-    atol : Real, optional
+    atol : float, optional
         Absolute error tolerance
-    orders : Iterable of Integral, optional
+    orders : Iterable of int, optional
         A list of orders (number of poles) to be searched
     n_vf_iter : Integral, optional
         Number of maximum VF iterations
@@ -204,9 +205,9 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
             weight[i, ce_xs[i]<=MIN_CROSS_SECTION] = \
                max(weight[i, ce_xs[i]>MIN_CROSS_SECTION])
 
+    # detect peaks (resonances) and determine VF order search range
     peaks, _ = find_peaks(ce_xs[0]+ce_xs[1])
     n_peaks = peaks.size
-    # order search
     if orders is not None:
         # make sure orders are even integers
         orders = list(set([int(i/2)*2 for i in orders if i>=2]))
@@ -219,6 +220,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
         print("Found {} peaks".format(n_peaks))
         print("Fitting orders from {} to {}".format(orders[0], orders[-1]))
 
+    # perform VF with increasing orders
     found_ideal = False
     n_discarded = 0 # for accelation, number of discarded searches
     best_quality = best_ratio = -np.inf
@@ -371,7 +373,7 @@ def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
 
 def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
                      log=False, path_out=None, mp_filename=None, **kwargs):
-    r"""Convert point-wise cross section to multipole data via Vector Fitting.
+    r"""Generate multipole data for a nuclide from ENDF.
 
     Parameters
     ----------
@@ -481,8 +483,7 @@ def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
 
     alpha = nuc_ce.atomic_weight_ratio/(K_BOLTZMANN*TEMPERATURE_LIMIT)
 
-    poles, residues = None, None
-    piece_idxs = [0]
+    poles, residues = [], []
     # VF piece by piece
     for i_piece in range(vf_pieces):
         if log:
@@ -504,13 +505,8 @@ def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
         p, r = _vectfit_xs(energy[e_idx], ce_xs[:, e_idx], mts, rtol=vf_error,
                        log=log, path_out=path_out, **kwargs)
 
-        if poles is None:
-            poles, residues = p, r
-        else:
-            poles = np.hstack(poles, p)
-            residues = np.hstack(residues, r)
-
-        piece_idxs.append(piece_idxs[-1] + p.size)
+        poles.append(p)
+        residues.append(r)
 
     # gather multipole data into a dictionary
     mp_data = {"name": nuc_ce.name,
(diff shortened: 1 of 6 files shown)


liangjg:
Update: added 2 unit test cases generating the wmp library. `vectfit` package is selectively installed in Travis CI with the dependencies built from the source.

paulromano:
Thanks for the PR @liangjg! Sorry I haven't had a chance to look at this yet. I'll try to get around to reviewing it later this week or early next week.

liangjg:
@paulromano, no worries, no rush for this to be merged.
Let me cc Jiankai @rockfool here in case he is interested in taking a look.

paulromano (changes requested):
@liangjg Thanks once again for this PR, and sorry it's taken me so long to give you some feedback. Most of my comments are pretty minor and overall the PR is in pretty good shape. I've only looked through the code so far and will continue my review by actually playing around with the feature. Also, I'm going to think about the testing situation -- it might make sense for us to be using conda in our CI runs as opposed to installing all Python packages through pip, so that might simply matters if we can pull in the xtensor-related dependencies from conda-forge.

tests/unit_tests/test_data_multipole.py
```diff
@@ -5,6 +5,14 @@
 import pytest
 import openmc.data
 
+no_vectfit = False
+try:
+    import vectfit as vf
+except:
+    no_vectfit = True
+
+vf_only = pytest.mark.skipif(no_vectfit, reason="vectfit package not installed")
```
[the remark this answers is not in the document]
liangjg: Seems [pytest.importorskip](https://docs.pytest.org/en/latest/reference.html#pytest-importorskip-ref) is suitable for skipping tests on a missing import at module level. But I don't want to skip other existing `multipole` tests when `vectfit` is missing.

tools/ci/travis-install-vectfit.sh
```diff
@@ -0,0 +1,45 @@
...
+XTL_BRANCH='0.6.9'
+XTL_REPO='https://github.com/xtensor-stack/xtl'
+
+XTENSOR_BRANCH='0.21.2'
+XTENSOR_REPO='https://github.com/xtensor-stack/xtensor'
+
+XTENSOR_PYTHON_BRANCH='0.24.1'
+XTENSOR_PYTHON_REPO='https://github.com/xtensor-stack/xtensor-python'
+
+XTENSOR_BLAS_BRANCH='0.17.1'
```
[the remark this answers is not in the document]
liangjg: Any version later than 0.15 should be ok. I'm going to update vectfit repo readme accordingly.

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+def _vectfit_xs(energy, ce_xs, mts, rtol=1e-3, atol=1e-5, orders=None,
+                n_vf_iter=30, log=False, path_out=None, **kwargs):
+    r"""Convert point-wise cross section to multipole data via Vector Fitting.
+
+    Parameters
+    ----------
+    energy : np.ndarray
+        Energy array
+    ce_xs : np.ndarray
+        Point-wise cross sections to be fitted
```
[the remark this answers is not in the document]
liangjg: Done

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    # import vectfit package: https://github.com/liangjg/vectfit
+    import vectfit as vf
+
+    ne = energy.size
+    nmt = len(mts)
+    if ce_xs.shape != (nmt, ne):
+        raise ValueError('Inconsistent cross section data.')
+
+    # construct test data: interpolate xs with finer grids
+    N_FINER = 10
```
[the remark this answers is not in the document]
liangjg: ok

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    atol : float, optional
+        Absolute error tolerance
+    orders : Iterable of int, optional
+        A list of orders (number of poles) to be searched
+    n_vf_iter : int, optional
+        Number of maximum VF iterations
+    log : bool or int, optional
+        Whether to print running logs
+    path_out : str, optional
+        Path to save the figures
```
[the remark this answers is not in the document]
liangjg: updated

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    **kwargs
+        Additional keyword arguments
```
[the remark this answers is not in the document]
liangjg: removed

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    test_energy[[0, -1]] = energy[[0, -1]] # avoid numerical issue
+    test_xs_ref = np.zeros((nmt, ne_test))
+    for i in range(nmt):
+        test_xs_ref[i] = np.interp(test_energy, energy, ce_xs[i])
+
+    if log:
+        print("Energy: {:.3e} to {:.3e} eV ({} points)".format(
+              energy[0], energy[-1], ne))
+
+    # inputs
```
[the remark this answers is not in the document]
liangjg: comments updated

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    f = ce_xs * energy # sigma*E
+    s = np.sqrt(energy) # sqrt(E)
+    test_s = np.sqrt(test_energy)
+    weight = 1.0/f
+
+    # very small cross sections can lead to huge weights, which will harm the
+    # fitting accuracy
+    MIN_CROSS_SECTION = 1e-7
```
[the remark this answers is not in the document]
liangjg: done

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+        found_better = False
+        # fitting iteration
+        for i_vf in range(n_vf_iter):
+            if log >= DETAILED_LOGGING:
+                print("VF iteration {}/{}".format(i_vf+1, n_vf_iter))
+
+            # call vf
+            try:
+                poles, residues, cf, f_fit, rms = vf.vectfit(f, s, poles, weight)
+            except:
```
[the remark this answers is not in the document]
liangjg: It is very rare but sometimes `vectfit` breaks when `f` is big, likely RAM exhaustion? I just removed `try/except` here。

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+            if np.any(np.isnan(abserr)):
+                maxre, ratio, ratio2 = np.inf, -np.inf, -np.inf
+            elif np.all(abserr <= atol):
+                maxre, ratio, ratio2 = 0., 1., 1.
+            else:
+                maxre = np.max(relerr[abserr > atol])
+                ratio = np.sum((relerr<rtol) | (abserr<atol)) / relerr.size
+                ratio2 = np.sum((relerr<10*rtol) | (abserr<atol)) / relerr.size
+
+            quality = ratio + ratio2 - min(0.1*maxre, 1) - 0.001*new_poles.size
```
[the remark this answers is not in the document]
liangjg: This is a less rigorous metric for choosing the optimal fitting results. It lacks a clear mathematical basis and is is expected to be improved. The idea is: it is preferred to have more points within accuracy tolerance, smaller maximum deviation, and fewer poles (e,g. 10 more poles can only be accepted if 1% more points are satisfying the tolerance). It can be improved  Comments were added.

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    mp_poles = best_poles[real_idx+conj_idx]
+    mp_residues = np.concatenate((best_residues[:, real_idx],
+                                  best_residues[:, conj_idx]*2), axis=1)/1j
+    if log:
+        print("Final number of poles: {}".format(mp_poles.size))
+
+    if path_out:
+        import matplotlib
+        matplotlib.use("agg")
+        import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
liangjg: ok

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+        elastic_xs = np.zeros_like(total_xs)
+    try:
+        absorption_xs = nuc_ce[27].xs['0K'](energy)
+    except:
+        absorption_xs = np.zeros_like(total_xs)
+    fissionable = False
+    try:
+        fission_xs = nuc_ce[18].xs['0K'](energy)
+        fissionable = True
+    except:pass
```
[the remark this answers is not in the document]
liangjg: Done

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    njoy_error : float, optional
+        Fractional error tolerance for processing point-wise data with NJOY
+    vf_error : float, optional
+        Fractional error tolerance for data fitting
+    vf_pieces : integer, optional
+        Number of equal-in-mementum spaced energy pieces for data fitting
+    log : bool or int, optional
+        Whether to print running logs
+    path_out : str, optional
+        Path to write out data files
```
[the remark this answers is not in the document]
liangjg: done

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    mp_data = {"name": nuc_ce.name,
+               "AWR": nuc_ce.atomic_weight_ratio,
+               "E_min": E_min,
+               "E_max": E_max,
+               "poles": poles,
+               "residues": residues}
+
+    # dump multipole data to file
+    if path_out:
+        import pickle
```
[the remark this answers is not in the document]
liangjg: ok

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    rtol : float, optional
+        Relative error tolerance
+    atol : float, optional
+        Absolute error tolerance
+    orders : Iterable of int, optional
+        A list of orders (number of poles) to be searched
+    n_vf_iter : int, optional
+        Number of maximum VF iterations
+    log : bool or int, optional
+        Whether to print running logs
```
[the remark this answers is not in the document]
liangjg: comments updated

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    # inner window size
+    spacing = (sqrt(E_max) - sqrt(E_min)) / n_win
+    # make sure inner window size is smaller than energy piece size
+    if spacing > piece_width:
+        raise ValueError('Window spacing cannot be larger than piece spacing.')
+
+    # determine curve fit order
+    if n_cf is None:
+        warnings.warn("Curvefit order not specified.")
+        n_cf = 5
```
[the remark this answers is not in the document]
liangjg: agree. updated

openmc/data/multipole.py
```diff
@@ -128,6 +141,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+        poles, residues = mp_poles[i_piece], mp_residues[i_piece]
+        n_poles = poles.size
+
+        # generate energy points for fitting: equally spaced in momentum
+        n_points = min(max(100, int((e_end - e_start)*4)), 10000)
+        energy_sqrt = np.linspace(np.sqrt(e_start), np.sqrt(e_end), n_points)
+        energy = energy_sqrt**2
+
+        # reference xs from multipole form
+        xs_ref = vf.evaluate(energy_sqrt, poles, residues*1j) / energy
```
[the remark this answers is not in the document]
liangjg: the 1j comes from a slight difference between the vector fitting formula and multipole formula.

wmp 0k

vf&space;\approx&space;\sum_{j=1}^{N}&space;\frac{\mathbf{r}_j}{s-p_j}&space;)

openmc/data/multipole.py
```diff
@@ -391,6 +1008,110 @@ def from_hdf5(cls, group_or_filename):
 
         return out
 
+    @classmethod
+    def from_endf(cls, endf_file, vf_options={}, wmp_options={}):
```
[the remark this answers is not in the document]
liangjg: ok

openmc/data/multipole.py
```diff
@@ -391,6 +1008,110 @@ def from_hdf5(cls, group_or_filename):
...
+
+                # update arguments dictionary
+                kwargs.update(n_win=n_w, n_cf=n_cf)
+
+                # windowing
+                try:
+                    wmp = _windowing(mp_data, log=log, **kwargs)
+                except Exception as e:
+                    if log:
+                        print('Failed: ' + str(e))
```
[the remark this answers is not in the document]
liangjg: it is inside a loop for searching windowing configurations, I expect searching continues when it breaks for one case?

liangjg:
@paulromano thanks for the comments. Most of them should have been addressed.

tests/unit_tests/test_data_multipole.py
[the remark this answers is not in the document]
paulromano: You can use it for individual functions, e.g.
```Python
def test_from_endf():
    pytest.importerskip('vectfit')
    ...
```
will result in the test being skipped if `vectfit` can't be imported

openmc/data/multipole.py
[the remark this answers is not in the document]
paulromano: Got it; thanks!

openmc/data/multipole.py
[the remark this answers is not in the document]
paulromano: Ah, ok. I think a short comment mentioning this would be useful

openmc/data/multipole.py
```diff
@@ -1,14 +1,24 @@
...
+from copy import deepcopy
+import warnings
 
+import os
 import h5py
+import pickle
 import numpy as np
+from scipy.signal import find_peaks
+import matplotlib
+matplotlib.use("agg")
```
paulromano, line 12: This seems intrusive to me -- anyone who imports openmc.data would have their matplotlib backend potentially changed. Why does it need to be set manually here?

openmc/data/multipole.py
```diff
@@ -391,6 +1010,115 @@ def from_hdf5(cls, group_or_filename):
...
+        """
+
+        if isinstance(mp_data, str):
+            # load multipole data from file
+            with open(mp_data, 'rb') as f:
+                mp_data = pickle.load(f)
+
+        # windowing with specific options
+        if not search:
+            return _windowing(mp_data, log=log, **kwargs)
```
paulromano, line 1077: Now that `n_cf` is a required argument in `_windowing`, it needs to be passed here.

openmc/data/multipole.py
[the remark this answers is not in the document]
liangjg: ok. comments updated.

tests/unit_tests/test_data_multipole.py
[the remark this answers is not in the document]
liangjg: Great. Thanks!

openmc/data/multipole.py
liangjg: Sorry, this was incorrectly copied from a notebook I guess. Just removed.

paulromano (changes requested):

openmc/data/multipole.py
```diff
@@ -391,6 +1009,115 @@ def from_hdf5(cls, group_or_filename):
...
+                    break
+
+                # select wmp library with metric:
+                # - performance: average # used poles per window and CF order
+                # - memory: # windows
+                metric = -(wmp.poles_per_window * 10. + wmp.fit_order * 1. +
+                           wmp.n_windows * 0.01)
+                if best_wmp is None or metric > best_metric:
+                    if log:
+                        print("Best library by far.")
```
paulromano, line 1109: Should this be "Best library so far"?

openmc/data/multipole.py
```diff
@@ -129,6 +143,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+
+            if log >= DETAILED_LOGGING:
+                print("  # poles: {}".format(new_poles.size))
+                print("  Max relative error: {:.3f}%".format(maxre*100))
+                print("  Satisfaction: {:.1f}%, {:.1f}%".format(ratio*100, ratio2*100))
+                print("  Quality: {:.2f}".format(quality))
+
+            if quality > best_quality:
+                if log >= DETAILED_LOGGING:
+                    print("  Best by far!")
```
paulromano, line 305: Same thing here --- "Best so far"?

paulromano:
@liangjg Could you say a little bit about how you generally run `WindowedMultipole.from_endf` when you are generating a library? I assume you use the `search=True` option during the windowing stage? Do you leave all other parameters as the defaults?

openmc/data/multipole.py
```diff
@@ -129,6 +143,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+    alpha = awr / (K_BOLTZMANN*TEMPERATURE_LIMIT)
+
+    # determine window size
+    if n_win is None:
+        if spacing is not None:
+            # Ensure the windows are within the multipole energy range
+            n_win = int((sqrt(E_max) - sqrt(E_min)) / spacing)
+            E_max = (sqrt(E_min) + n_win*spacing)**2
+        else:
+            warnings.warn("Number of windows or window size not specified.")
```
paulromano, line 598: Again, it seems weird to me to give a warning for not specifying an optional parameter

openmc/data/multipole.py
```diff
@@ -129,6 +143,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+            if n_real_poles > 0:
+                if log >= DETAILED_LOGGING:
+                    print("  # real poles: {}".format(n_real_poles))
+                new_poles, residues, cf, f_fit, rms = \
+                      vf.vectfit(f, s, new_poles, weight, skip_pole=True)
+
+            # assess the result on test grid
+            test_xs = vf.evaluate(test_s, new_poles, residues) / test_energy
+            abserr = np.abs(test_xs - test_xs_ref)
+            relerr = abserr / test_xs_ref
```
paulromano, line 278: Since it looks we're expecting the possibility of nans being produced, probably makes sense to silence the warnings about them
```suggestion
            with np.errstate(invalid='ignore'):
                relerr = abserr / test_xs_ref
```

openmc/data/multipole.py
```diff
@@ -129,6 +143,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+
+    # transform xs (sigma) and energy (E) to f (sigma*E) and s (sqrt(E)) to be
+    # compatible with the multipole representation
+    f = ce_xs * energy
+    s = np.sqrt(energy)
+    test_s = np.sqrt(test_energy)
+
+    # inverse weighting is used for minimizing the relative deviation instead of
+    # absolute deviation in vector fitting
+    weight = 1.0/f
```
paulromano, line 210: It looks like in some cases, we get divide by zeros (for example, He4 from ENDF/B-VIII.0). Will this cause any problems?

openmc/data/multipole.py
```diff
@@ -129,6 +143,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+            else:
+                xs_wp = np.zeros_like(xs_ref)
+
+            # do least square curve fit on the remains
+            coefs = np.linalg.lstsq(matrix, (xs_ref - xs_wp).T, rcond=None)[0]
+            xs_fit = (matrix @ coefs).T
+
+            # assess the result
+            abserr = np.abs(xs_fit + xs_wp - xs_ref)
+            relerr = abserr / xs_ref
```
paulromano, line 674:
```suggestion
            with np.errstate(invalid='ignore'):
                relerr = abserr / xs_ref
```

openmc/data/multipole.py
```diff
@@ -391,6 +1009,115 @@ def from_hdf5(cls, group_or_filename):
 
         return out
 
+    @classmethod
+    def from_endf(cls, endf_file, vf_options=None, wmp_options=None):
```
paulromano, line 1013: It might be convenient to have a `log` argument here that gets passed to the underlying functions so that it doesn't have to be specified for both `vf_options` and `wmp_options`. It's hard for me to imagine a case where you'd want it to be True for one set of options and False for the other.

openmc/data/multipole.py
paulromano: After more testing, it seems like this is especially common for fissionable nuclides

openmc/data/multipole.py
```diff
@@ -129,6 +143,600 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+        print("Final number of poles: {}".format(mp_poles.size))
+
+    if path_out:
+        if not os.path.exists(path_out):
+            os.makedirs(path_out)
+        for i, mt in enumerate(mts):
+            fig, ax1 = plt.subplots()
+            lns1 = ax1.loglog(test_energy, test_xs_ref[i], 'g', label="ACE xs")
+            lns2 = ax1.loglog(test_energy, best_test_xs[i], 'b', label="VF xs")
+            ax2 = ax1.twinx()
```
paulromano, line 365: For a bunch of nuclides, I get the following at this line:
```
openmc/data/multipole.py:366: UserWarning: Data has no positive values, and therefore cannot be log-scaled.
```

openmc/data/multipole.py
liangjg: You're right. For the cross section that is too small (less than 1e-7, including zeros), the weight will be adjusted specifically, as shown in [the following lines](https://github.com/openmc-dev/openmc/blob/b842ae78eed362af216c23bfb25ae86620fec7d0/openmc/data/multipole.py#L212-L219).

openmc/data/multipole.py
liangjg: Yeah. I removed this warning.

openmc/data/multipole.py
liangjg: done

openmc/data/multipole.py
liangjg: I guess `n_cf` is packed in `kwargs` and will be passed to `_windowing` automatically?
I tested `h1 = openmc.data.WindowedMultipole.from_multipole("./H1/H1_mp.pickle", n_cf=2, n_win=400)` is working while
`h1 = openmc.data.WindowedMultipole.from_multipole("./H1/H1_mp.pickle", n_win=400)` will give a error
`TypeError: _windowing() missing 1 required positional argument: 'n_cf'`

openmc/data/multipole.py
liangjg: done

openmc/data/multipole.py
liangjg: ok

openmc/data/multipole.py
liangjg: Ok. `log` argument added. If specified, it will overwrite the logs in `vf_options` and `wmp_options`, otherwise, one can still specify them individually.

openmc/data/multipole.py
paulromano: In that case, we probably ought to ignore divide by zeros:
```suggestion
    with np.errstate(divide='ignore'):
        weight = 1.0/f
```

openmc/data/multipole.py
paulromano: Yeah, the use case where I was running into this was:
```Python
openmc.data.WindowedMultipole.from_endf(filename)
```
In which case `n_cf` is not specified anywhere. We need some default value of `n_cf` for this case.

paulromano:
I did some testing of ENDF/B-VIII.0 and ran into trouble on a few nuclides. Namely, for the following nuclides, I wasn't able to successfully generate a `WindowedMultipole` object:
- S36, Ar36, Ar38, Co58m, Am244

The `_vectfit_nuclide` function runs successfully, but at the windowing stage, I end up with the following exception:
```Python
~/openmc/openmc/data/multipole.py in from_endf(cls, endf_file, vf_options, wmp_options)
   1044 
   1045         # windowing
-> 1046         return cls.from_multipole(mp_data, **wmp_options)
   1047 
   1048     @classmethod

~/openmc/openmc/data/multipole.py in from_multipole(cls, mp_data, search, log, **kwargs)
   1115         if log:
   1116             print("Final library: {} poles, {} windows, {:.2g} poles per window, "
-> 1117                   "{} CF order".format(best_wmp.n_poles, best_wmp.n_windows,
   1118                    best_wmp.poles_per_window, best_wmp.fit_order))
   1119 

AttributeError: 'NoneType' object has no attribute 'n_poles'
```
Note that this is with `search=True`. There are a bunch of nuclides that only run successfully if `search=True`. In the default case where `search=False`, for these nuclides it ends up complaining about the pure curvefit failing for the first window. This includes:
- Be7, Na22, P31, K40, Ca46, Ru105, Cd113, Xe123, Pm144, Sm149, Gd153, Gd155, Gd157, Tb158, Ta180, Hg204, Pa230, Pu237

I'm wondering if perhaps we should make `search=True` the default since these failures with `search=False` don't seem too rare. What do you think @liangjg? Also, for the five nuclides above, were you able to get WMP data processed successfully?

openmc/data/multipole.py
```diff
@@ -129,6 +143,604 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+                                    energy[0], energy[-1], mt))
+            plt.savefig(fig_file)
+            plt.close()
+            if log:
+                print("Saved figure: {}".format(fig_file))
+
+    return (mp_poles, mp_residues)
+
+
+def _vectfit_nuclide(endf_file, njoy_error=5e-4, vf_error=1e-3, vf_pieces=None,
```
paulromano, line 390: I think perhaps we should make this function public (no leading underscore). I can imagine a processing workflow where you run this function on all nuclides and then afterward do the windowing for all.

openmc/data/multipole.py
liangjg: ok

liangjg:
@paulromano Thanks for your comments:

1. `search=True` or `search=False`? It is always recommended to use `search=True` unless you know exactly how to set the window size and curvefit order. I myself always use `search=True` to generate a library. I agree with you to make the search=True the default, which is consistent with the most common user cases. I'll update the code.

2. Like what you found, there are indeed cases that `search=True` cannot give a WMP representation successfully, this is mostly because the point-wise data itself is too bad, i.e. not smooth enough to get a correct fitting within accuracy tolerance even with the finest window and highest curve fit order, for nuclides like `Ar38`, 'Xe123' and so on. If you check the 0K xs, you probably see the abnormal segmentally curved xs, especially for absorption in lower energy (I suspect the interpolation parameters are stored mistakenly in ENDF). In these cases, I have to relax the accuracy tolerance, from `1e-3` to `5e-3` or even `1e-2` to get a full library. For ENDFBVIII.0, there are around several dozens of nuclides with larger tolerances. I didn't evaluate how these nuclides will impact the accuracy of transport simulation but I guess they are not too bad in comparison to the true uncertainty of the original point-wise cross section.

3. Here is the script I used to generate the full library of ENDFBVIII.0: https://gist.github.com/liangjg/dba56938c13ebc663dc99ccfebcc50d0 . It looks messy because of the use of trial and error method with a lot of `try/except`s, to relax the accuracy tolerance step by step as explained above. But I did generate a full WMP library from ENDFBVIII.0 (it was slow though). I ran some preliminary validation by comparing it with 0K ACE libraries using the scripts in https://github.com/mit-crpg/WMP_Library.

openmc/data/multipole.py
liangjg: Yeah, I did separate the fitting and windowing step in the workflow. Updated.

openmc/data/multipole.py
liangjg: Can you give me an example nuclide which showed the warning so I can repeat? I probably encountered the same warning before but forgot the nuclides.

paulromano: Looks like H3 with search turned off should do it. Here's a reproducer:
```Python
openmc.data.WindowedMultipole.from_endf(
    '/opt/data/endf/endf-b-viii.0/neutrons/n-001_H_003.endf',
    log=True,
    vf_options={'path_out': '.'},
    wmp_options={'search': False}
)
```

tests/unit_tests/test_data_multipole.py
```diff
@@ -46,3 +46,19 @@ def test_export_to_hdf5(tmpdir, u235):
...
     u235.export_to_hdf5(filename)
     assert os.path.exists(filename)
+
+
+def test_from_endf():
+    pytest.importorskip('vectfit')
+    endf_data = os.environ['OPENMC_ENDF_DATA']
+    endf_file = os.path.join(endf_data, 'neutrons', 'n-001_H_001.endf')
+    return openmc.data.WindowedMultipole.from_endf(
+            endf_file, wmp_options={"n_win": 400, "n_cf": 3})
```
paulromano, line 56: Now that search is turned on by default, we need to disable it for this test:
```suggestion
            endf_file, wmp_options={"search": False, "n_win": 400, "n_cf": 3})
```

liangjg: My bad. `{"n_win": 400, "n_cf": 3}` somehow indicates we don't want `search`, so I added a check in the code to set `search=False` automatically in this case.

openmc/data/multipole.py
liangjg: Yeah, checks added. I also changed `with np.errstate(invalid='ignore'):` to `with np.errstate(invalid='ignore', divide='ignore'):` so the running on `H3` should be clean now.

openmc/data/multipole.py
```diff
@@ -129,6 +143,608 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+
+    # parse energy and cross sections
+    energy = nuc_ce.energy['0K'][:E_max_idx + 1]
+    E_min, E_max = energy[0], energy[-1]
+    n_points = energy.size
+    total_xs = nuc_ce[1].xs['0K'](energy)
+    try:
+        elastic_xs = nuc_ce[2].xs['0K'](energy)
+    except KeyError:
+        elastic_xs = np.zeros_like(total_xs)
```
paulromano, line 461: Every nuclide should have elastic scattering, so if it doesn't appear something is seriously wrong
```suggestion
    elastic_xs = nuc_ce[2].xs['0K'](energy)
```

openmc/data/multipole.py
```diff
@@ -129,6 +143,608 @@ def _broaden_wmp_polynomials(E, dopp, n):
...
+            with np.errstate(invalid='ignore', divide='ignore'):
+                relerr = abserr / test_xs_ref
+                if np.any(np.isnan(abserr)):
+                    maxre, ratio, ratio2 = np.inf, -np.inf, -np.inf
+                elif np.all(abserr <= atol):
+                    maxre, ratio, ratio2 = 0., 1., 1.
+                else:
+                    maxre = np.max(relerr[abserr > atol])
+                    ratio = np.sum((relerr < rtol) | (abserr < atol)) / relerr.size
+                    ratio2 = np.sum((relerr < 10*rtol) | (abserr < atol)) / relerr.size
```
paulromano, line 289: Should this block really be under `with np.errstate`? I thought only the `relerr = abserr / test_xs_ref` line would cause warnings.

openmc/data/multipole.py
```diff
@@ -391,6 +1017,130 @@ def from_hdf5(cls, group_or_filename):
...
+    @classmethod
+    def from_multipole(cls, mp_data, search=None, log=False, **kwargs):
+        """Generate windowed multipole neutron data from multipole data.
+
+        Parameters
+        ----------
+        mp_data : dictionary or str
+            Dictionary or Path to the multipole data stored in a pickle file
+        search : bool, optional
+            Whether to search for optimal window size and curvefit order
```
paulromano, line 1070: Since the default is now hidden in the code, the description here should describe what the default behavior is

tests/unit_tests/test_data_multipole.py
paulromano, line 56: To make sure that all the lines of code for `if log` get run during tests, I'd suggest changing to
```suggestion
            endf_file, log=True, wmp_options={"n_win": 400, "n_cf": 3})
```

tests/unit_tests/test_data_multipole.py
```diff
@@ -46,3 +46,19 @@ def test_export_to_hdf5(tmpdir, u235):
...
+
+
+def test_from_endf_search():
+    pytest.importorskip('vectfit')
+    endf_data = os.environ['OPENMC_ENDF_DATA']
+    endf_file = os.path.join(endf_data, 'neutrons', 'n-095_Am_244.endf')
+    return openmc.data.WindowedMultipole.from_endf(
+            endf_file, wmp_options={"search": True, 'rtol':1e-2})
```
paulromano, line 64:
```suggestion
            endf_file, log=True, wmp_options={"search": True, 'rtol':1e-2})
```

paulromano (changes requested):
One final round of comments here and then I think this is ready to be merged. Thanks @liangjg!

openmc/data/multipole.py
liangjg: `relerr` contains `invalid value` (`nan`) so operations on `relerr` in the following lines, like `(relerr < rtol)` will cause warnings as well. 
The following piece of code will give two times of `RuntimeWarning: invalid value encountered`:
```
import numpy as np
a = b = np.array([[1,2,3.],[0., 0., 0.]])
c = b/a
c < 0.001
```

openmc/data/multipole.py
liangjg: comments added

tests/unit_tests/test_data_multipole.py
liangjg: ok

tests/unit_tests/test_data_multipole.py
liangjg: ok

openmc/data/multipole.py
liangjg: Agree. I've never seen a nuclide without `elastic xs`. Updated.

paulromano (approved):

paulromano:
Thanks for this contribution @liangjg!

PR #1659, opened 2020-09-14
fix typo in energy unit in dose coefficients

liangjg:
The energies should be multiplied by `1E6` when converted from `MeV` to `eV`.


paulromano (approved):
Thanks for the fix @liangjg!

PR #1667, opened 2020-09-23
Added DiffusionCoefficient class to mgxs.py

mkreher13:
To continue in the effort of merging Sam Shaner's transient capability,
this next PR adds the class DiffusionCoefficient to mgxs.py. It uses
the TransportXS class, but has some unique features in get_condensed_xs()
requiring its own function. There won't be any other MGXS classes added
after this.

I updated the docstrings and the appropriate regression tests.
Since my system outputs different formatting for the
mgxs_library_distribcell test, it is possible this test will continue
to fail. However, I did my best to modify it by hand so that it will pass.


openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+
+    """
+
+    def __init__(self, domain=None, domain_type=None, groups=None, nu=False,
+                 by_nuclide=False, name='', num_polar=1, num_azimuthal=1):
+        super(DiffusionCoefficient, self).__init__(domain, domain_type, groups,
+                                                   nu, by_nuclide, name,
+                                                   num_polar, num_azimuthal)
+        self._rxn_type = 'diffusion-coefficient'
+        cv.check_value('nu for diffusion-coefficient', nu, [False])
```
nelsonag, line 2970: Why is `True` not allowed as a possible value for `nu`? 

Seems like it would be necessary if the downstream applications used the scattering production matrix and a reduced absorption xs.

openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+
+            # Compute total cross section
+            total_xs = self.tallies['total'] / self.tallies['flux (tracklength)']
+
+            # Compute transport correction term
+            trans_corr = self.tallies['scatter-1'] / self.tallies['flux (analog)']
+
+            # Compute the diffusion coefficient
+            transport = total_xs - trans_corr
+            dif_coef = transport**(-1) / 3.0
```
nelsonag, line 2995: Why the `**(-1)` and not `1.0 / (3.0 * transport)`? That is, since we already incur the cost of division might as well trade the cost of the exponentiation for the cost of multiplication

openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+
+    @property
+    def xs_tally(self):
+        if self._xs_tally is None:
+            if self.tallies is None:
+                msg = 'Unable to get xs_tally since tallies have ' \
+                      'not been loaded from a statepoint'
+                raise ValueError(msg)
+
+            self._xs_tally = self.rxn_rate_tally / self.tallies['flux (tracklength)']
```
nelsonag, line 3010: Nothing wrong with your code here that I can think of, but I'm trying to remember why we didnt just use `self.rxn_rate_tally` in `TransportXS.xs_tally(...)` like you did here.... Did you stumble on that thought too?

openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+                # Slice Legendre expansion filter and change name of score
+                p1_tally = p1_tally.get_slice(filters=[openmc.LegendreFilter],
+                        filter_bins=[('P1',)],
+                        squeeze=True)
+                p1_tally._scores = ['scatter-1']
+
+            total = self.tallies['total'] / self.tallies['flux (tracklength)']
+            trans_corr = p1_tally / self.tallies['flux (analog)']
+            transport = (total - trans_corr)
+            dif_coef = transport**(-1) / 3.0
```
nelsonag, line 3055: ditto

openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+                p1_tally.filters[-2] = new_filt
+
+                # Slice Legendre expansion filter and change name of score
+                p1_tally = p1_tally.get_slice(filters=[openmc.LegendreFilter],
+                        filter_bins=[('P1',)],
+                        squeeze=True)
+                p1_tally._scores = ['scatter-1']
+
+            total = self.tallies['total'] / self.tallies['flux (tracklength)']
+            trans_corr = p1_tally / self.tallies['flux (analog)']
```
nelsonag, line 3053: If self._rxn_rate_tally is not None, then where would p1_tally come from?

openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+                           self.num_groups, equality=True)
+        cv.check_value('upper coarse energy', coarse_groups.group_edges[-1],
+                       [self.energy_groups.group_edges[-1]])
+        cv.check_value('lower coarse energy', coarse_groups.group_edges[0],
+                       [self.energy_groups.group_edges[0]])
+
+        # Clone this MGXS to initialize the condensed version
+        condensed_xs = copy.deepcopy(self)
+
+        if condense_dif_coef:
```
nelsonag, line 3038: Not sure I understand the need for this

openmc/mgxs/mgxs.py
```diff
@@ -2829,6 +2832,294 @@ def nu(self, nu):
...
+            self._xs_tally = self.rxn_rate_tally / self.tallies['flux (tracklength)']
+            self._compute_xs()
+
+        return self._xs_tally
+
+    def get_condensed_xs(self, coarse_groups, condense_dif_coef=True):
+        """Construct an energy-condensed version of this cross section.
+        Parameters
+        ----------
+        coarse_groups : openmc.mgxs.EnergyGroups
```
nelsonag, line 3019: add condense_dif_coef to docstring

nelsonag (changes requested):
Thanks for this @mkreher13! Only a few comments within!

nelsonag:
We should also discuss why your system's results and the Travis test results differ so that it requires manual modifications of the test results

openmc/mgxs/mgxs.py
mkreher13: Good point.  The checkvalue() implies that nu must be False, but as I think about it, I can't see why that is necessary. I guess anywhere the transport-xs is calculated, there is a need to allow for scatter multiplication. 

If I allow nu=True, do I have to add a 'nu-diffusion-coefficient' rxn_type?

Also, everywhere where nu is used, the docstrings say "defaults to True". But in all the __init__ functions, I see that nu=False by default. Is that an inconsistency I should fix up everywhere?

openmc/mgxs/mgxs.py
mkreher13: The idea of cloning the original MGXS is to initialize the condensed_xs with all the existing attributes. This is done also in the main MGXS class. 

As for the "if condense_dif_coef" I'm not sure either. Who would call get_condensed_xs for the DiffusionCoefficient class and set condense_dif_coef to False? I am comfortable removing that if-statement.

openmc/mgxs/mgxs.py
mkreher13: I get a TypeError if I use 1.0 / (3.0 * transport) : unsupported operant type(s) for /: 'float' and 'Tally' which doesn't show up if using the exponentiation.

openmc/mgxs/mgxs.py
mkreher13: I get a TypeError if I use 1.0 / (3.0 * transport) : unsupported operant type(s) for /: 'float' and 'Tally' which doesn't show up if using the exponentiation.

openmc/mgxs/mgxs.py
mkreher13: I noticed the difference and wondered if that would pose a problem. 
The only difference I see is that in TransportXS.xs_tally(...) the p1_tally is divided by ['flux (analog)'] while the total tally by ['flux (tracklength)']. So if we can just divide both by the tracklength flux, then we can use my expression in both classes. If not, I can change mine to match TransportXS.

mkreher13:
About the distribcell test: if you look at the Checks, you'll see they pass on all the systems listed. Well, that's not 100% true. If you look into the first system, 5081.1, you'll see that the distribcell test didn't pass, it Xfail'ed. That's because although the results are identical, the format is different. So I'm apparently running on that sort of system. When I use my default output, 5081.1 passes, and all the rest fail. So I need to always commit formatting that fits all the other systems and Xfail's on my own. 

I can guarantee it's just a formatting issue, and not anything to worry about concerning the results themselves.

openmc/mgxs/mgxs.py
mkreher13: Good point. I could just remove the statement "if self._rxn_rate_tally is None".

mkreher13: Actually, can you explain this question to me some more? I see the same structure in TransportXS and am wondering how it is resolved in that class.

openmc/mgxs/mgxs.py
nelsonag: > If I allow nu=True, do I have to add a 'nu-diffusion-coefficient' rxn_type?
Yeah, I think that would be the way to go.

> Also, everywhere where nu is used, the docstrings say "defaults to True". But in all the init functions, I see that nu=False by default. Is that an inconsistency I should fix up everywhere?
Yeah, but change the docstring itself as opposed to changing the default.

openmc/mgxs/mgxs.py
nelsonag: I'd agree that we should just remove the ``if condense_dif_coef`` block. Seems like there is no need for that because of course one wants that.

openmc/mgxs/mgxs.py
nelsonag: No need with the above change.

openmc/mgxs/mgxs.py
nelsonag: That structure is in ``TransportXS.rxn_rate_tally`` (I think that's what you're referring to), but in that case it acts only as a means to not re-calculate something that has already been computed as it can be relatively costly.

Here, I'm not sure what the purpose of the ``if self._rxn_rate_tally is None`` block is. If you compare this block with what is in ``def rxn_rate_tally`` for the Diffusion Coefficient, it looks like you can replace
```
            if self._rxn_rate_tally is None:

                p1_tally = self.tallies['scatter-1']
                old_filt = p1_tally.filters[-2]
                new_filt = openmc.EnergyFilter(old_filt.values)
                p1_tally.filters[-2] = new_filt

                # Slice Legendre expansion filter and change name of score
                p1_tally = p1_tally.get_slice(filters=[openmc.LegendreFilter],
                        filter_bins=[('P1',)],
                        squeeze=True)
                p1_tally._scores = ['scatter-1']

            total = self.tallies['total'] / self.tallies['flux (tracklength)']
            trans_corr = p1_tally / self.tallies['flux (analog)']
            transport = (total - trans_corr)
            dif_coef = transport**(-1) / 3.0
            dif_coef *= self.tallies['flux (tracklength)']
```

with one of the two options:

```
            if self._rxn_rate_tally is None:

                p1_tally = self.tallies['scatter-1']
                old_filt = p1_tally.filters[-2]
                new_filt = openmc.EnergyFilter(old_filt.values)
                p1_tally.filters[-2] = new_filt

                # Slice Legendre expansion filter and change name of score
                p1_tally = p1_tally.get_slice(filters=[openmc.LegendreFilter],
                        filter_bins=[('P1',)],
                        squeeze=True)
                p1_tally._scores = ['scatter-1']

                total = self.tallies['total'] / self.tallies['flux (tracklength)']
                trans_corr = p1_tally / self.tallies['flux (analog)']
                transport = (total - trans_corr)
                dif_coef = transport**(-1) / 3.0
                dif_coef *= self.tallies['flux (tracklength)']
```

OR

```
            dif_coef = self.rxn_rate_tally
            <insert code to convet dif_coef from sparse to dense as it looks like the rest of the function expects dense>
```

nelsonag: Real quick on this one: you added in the if/else block here; if ``dif_coef = self.rxn_rate_tally`` works fine, then cant you just drop the whole if block and only use ``dif_coef = self.rxn_rate_tally``? 

(the ``rxn_rate_tally`` property includes the ``if self._rxn_rate_tally is None`` block already.

nelsonag:
@mkreher13 I'm fine with everything save for that one comment above. When we resolve it I'll go ahead and merge.

openmc/mgxs/mgxs.py
mkreher13: I did try that, but the results were different and resulted in mgxs_library_condense failing. 

When I only use dif_coef = self.rxn_rate_tally, it doesn't slice out the Legendre expansion. A series of print-statements showed me that it definitely calls the property self.rxn_rate_tally(self) which has all the same code. But it didn't do the slice for some reason.

When I used the if/else block, the slicing occurred as expected and the test passed.

nelsonag (approved):

nelsonag:
Ok, I'm all set then @mkreher13.  Thanks for putting this together!

PR #1677, opened 2020-10-03
Update to how volume calculation results are verified in `load_results`.

pshriwise:
A [user recently reported](https://openmc.discourse.group/t/error-when-load-volume-calculation-results-of-multiple-materials/769) that if domain IDs aren't in the same order for a `VolumeCalculation` object and the set of results being loaded, a failure is reported. 

This is a quick fix for that problem by doing the ID comparison using sets in `VolumeCalculation.load_results`.


paulromano (approved):
Thanks @pshriwise!

PR #1685, opened 2020-10-08
Remove redundant error check that the Position is in the mesh

aprilnovak:
The `get_bin` method performs a check that the `Position` is within the bounds of the `RegularMesh`. However, the `get_indices` method that is called just after this error check already performs a within-bounds check (albeit with a different approach, not by comparing coordinate values to `lower_left` and `upper_right`, but instead comparing the computed x, y, z indices in the mesh against the range [1, `shape_[i]`] for each `i`.

So I think this can safely be deleted!

### DIFF
--- a/src/mesh.cpp
+++ b/src/mesh.cpp
@@ -193,16 +193,6 @@ RegularMesh::RegularMesh(pugi::xml_node node)
 
 int RegularMesh::get_bin(Position r) const
 {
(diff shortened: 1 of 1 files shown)


pshriwise (approved):
Did a couple of spot-checks on some models I had lying around. No notable changes in performance or behavior. Nice to get rid of this unnecessary block! Thanks, @aprilnovak.

PR #1686, opened 2020-10-09
(tiny) Fix depletion restart example

gridley:
It seems that the example `delpetion_restart.py` currently fails on develop because the `ResultsList` class was at some point changed from taking the name of an HDF5 restart file in its `__init__` to having this file name passing through `from_hdf5`. It seems this example was never adjusted, which I found to be failing. I was thinking my code broke something, but luckily it turns out to be this tiny fix. :smile: 

Also, I have snuck in a small grammar fix in the contributing docs.

### DIFF
--- a/docs/source/devguide/workflow.rst
+++ b/docs/source/devguide/workflow.rst
@@ -50,7 +50,7 @@ Contributing
 ------------
 
 Now that you understand the basic development workflow, let's discuss how an
-individual to contribute to development. Note that this would apply to both new
+individual can contribute to development. Note that this would apply to both new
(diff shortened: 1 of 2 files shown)


nelsonag (approved):
Thanks @gridley, looks good to me.

nelsonag:
I restarted the failing Travis run (network connection issue), when that is done I will merge this in. Due to the simplicity of this one no need to wait the whole review period.

nelsonag:
Nailed it! Thanks again @gridley for taking the effort to put this one in! Always appreciated!

PR #1694, opened 2020-10-15
Improvement to microscopic MG mode calculations

nelsonag:
This PR fixes issues found when using OpenMC's MG mode on a problem with microscopic cross sections defined. Most of these small changes are superficial at best (clarifying error messages about invalid Nuclide names, or minor changes to code comments, for example). 

The larger changes are related to the following:
- Simplification to the process used to interpolate on temperatures while creating macroscopic xs from microscopic
- Simplification to the process used to compute a macroscopic scattering matrix from microscopic data
- Correction to the process used to create a "macroscopic" chi matrix from "microscopic" Chi matrices. The code now creating a "macroscopic" chi using fission source-weighting. This is consistent with the approach that DIF3D uses. This will be a source of error for users. In the spirit of MG-mode, I'd like to implement some other options so the user can evaluate the cost of this commonly made approximation. I am not addressing that in this PR to keep it small and to limit any changes to tests. Expect those options in a future PR, however.


paulromano (approved):
Thanks @nelsonag. The changes themselves look good to me. Just to make sure -- we don't expect any test results to change as a result of the changes here? (i.e., is there something we're not covering in the tests?)

nelsonag:
I was slightly surprised that mg_nuclides didn’t change, but I justified it to myself that we run so few particles in these tests that there was, in effect, only one fissioning isotope in the resultant MGXS library.  With 1 fissioning isotope, the previous chi effect was perfectly fine.

So, I think we’re just missing more particles.

src/mgxs.cpp
```diff
@@ -377,25 +378,43 @@ Mgxs::Mgxs(const std::string& in_name, const std::vector<double>& mat_kTs,
...
-    for (int interp_point = 0; interp_point < num_interp_points; interp_point++) {
-      for (int m = 0; m < micros.size(); m++) {
-        interp[m] = (1. - micro_t_interp[m]) * atom_densities[m];
-        temp_indices[m] = micro_t[m] + interp_point;
-        micro_t_interp[m] = 1. - micro_t_interp[m];
+    // Now combine the microscopic data at each relevant temperature
+    // We will do this by treating the multiple temperatures of a nuclide as
+    // a different nuclide. Mathematically this just means the temperature
+    // interpolant is included in the number density.
+    // These interpolants are contained within interp.
```
GiudGiud, line 385:
```suggestion
    // These interpolants are contained within interpolant.
```

src/scattdata.cpp
```diff
@@ -36,6 +37,13 @@ ScattData::base_init(int order, const xt::xtensor<int, 1>& in_gmin,
     energy[gin] = in_energy[gin];
     mult[gin] = in_mult[gin];
 
+    // Make sure the multiplicity does not have 0s
+    for (int go = 0; go < mult[gin].size(); go++) {
+      if (mult[gin][go] == 0.) {
```
GiudGiud, line 42: What do you think about adding a warning there? It's uncommon having a zero multiplicity could almost only be deliberate

GiudGiud (approved):
Just two soft suggestions

nelsonag:
Thanks @GiudGiud I agree with your changes and have implemented both!

PR #1708, opened 2020-10-28
Add a materials.xml printing function at each step of depletion

rockfool:
This PR is to add a materials.xml printing function at each step of depletion. I added it because I wanted to run neutron transport based on the depleted materials from a previously completed depletion simulation. Now this PR is motivated by the discussion [here](https://openmc.discourse.group/t/create-materials-xml-from-depletion-results/823/6). This option is set to be false by default and turned on at users' request. Only the nuclides with the number density bigger than 1e-50 (to be discussed) and covered by cross_sections.xml (with neutron cross sections) could be printed into "materials_n#.xml" file. Any comment or idea is much appreciated.

### DIFF
--- a/scripts/openmc-track-to-vtk
+++ b/scripts/openmc-track-to-vtk
@@ -46,7 +46,7 @@ def main():
         n_coords = track.attrs['n_coords']
         coords = []
         for i in range(n_particles):
-            coords.append(track['coordinates_' + str(i + 1)].value)
+            coords.append(track['coordinates_' + str(i + 1)][()])
             for j in range(n_coords[i]):
                 points.InsertNextPoint(coords[i][j,:])
 
--- a/tests/regression_tests/plot/test.py
+++ b/tests/regression_tests/plot/test.py
@@ -42,10 +42,10 @@ class PlotTestHarness(TestHarness):
                 # Add voxel data to results
                 with h5py.File(fname, 'r') as fh:
                     outstr += fh.attrs['filetype']
-                    outstr += fh.attrs['num_voxels'].tostring()
-                    outstr += fh.attrs['lower_left'].tostring()
-                    outstr += fh.attrs['voxel_width'].tostring()
-                    outstr += fh['data'].value.tostring()
+                    outstr += fh.attrs['num_voxels'].tobytes()
+                    outstr += fh.attrs['lower_left'].tobytes()
+                    outstr += fh.attrs['voxel_width'].tobytes()
+                    outstr += fh['data'][()].tobytes()
 
         # Hash the information and return.
         sha512 = hashlib.sha512()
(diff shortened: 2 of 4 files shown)


drewejohnson (changes requested):
Thanks for pushing these changes @rockfool. I think @gridley might have some insight as he originated the discussion in question. For the short term, there are some merge conflicts with the current main branch.

I will admit I haven't done a full review yet. But from the skim, I'm curious why this approach was chosen over having this be something a `Result` object can perform, e.g. 
```python
r = ResultsList.from_hdf(<depletion result file>)
r[j].export_materials_to_xml()  # export step j
```

As they're both written, it is difficult to do something like "start at day 125" or "start at burnup 12.5 MWd/kgHM", as both require knowledge of the depletion step for the point of interest. I think if we can provide such a feature, this will become a very useful feature. Even for our normal restart methods now that I think about it :thinking:

gridley:
Hey Drew, sure thing. The reason to _not_ make this a method of the results class is that you need to do this in order to produce a reasonable XML file:

```                             if nuc_name in self.nuclides_with_data: ```

So, you are inherently tied to whatever nuclear data library you're using in this case. Hm. So, if not this, then we would maybe have to pass a depletion operator to the `export_materials_to_xml` or perhaps just the name of the neutron XS library you plan to use, somehow. Or maybe we could just automatically pull it from the operator? But you're not guaranteed to construct a depletion operator in an arbitrary script where `export_materials_to_xml` would be called. Curious to hear you're thinking as to what would be most convenient.

My preference would be to just check `OPENMC_CROSS_SECTIONS` and only output stuff to `materials.xml` if it can be found there. It's a little black-boxy, maybe, but would be most convenient. And of course providing a means to override the chosen `cross_sections.xml` file used would be included.

drewejohnson:
Thanks for the insight @gridley. Having some information into what nuclides are in the data library will be crucial, as you've pointed out and provided through this PR. My biggest hesitations with this automatically writing all the material xs files are 1) finding the current step index to match a point of interest after running the simulation and 2) the excess of files that could be created. For a high fidelity depletion simulation, you may have several hundreds of depletion steps. and if you're simulating a problem with hundreds to thousands of burnable materials (full core, TRISO, et.c), writing a material file at every step can take a noticeable amount of time. Likely not as much as the full transport sweep, but this is also a serial step for what it's worth.

If my understanding of the use cases this feature would support are correct, people want this information not to do a depletion run. Otherwise passing the results list would be sufficient. Working with that understanding, I don't think we'll want to put users through the work of having to construct a depletion operator just to have access to the cross sections.

I think it's reasonable that the `OPENMC_CROSS_SECTIONS` environment variable could be used to identify the nuclides to be written. Or explicitly indicate the cross section file. Or a subset of nuclides to write (e.g. don't write fission products). To a comment made by @rockfool, we could also allow the threshold be configurable by the users

PR #1710, opened 2020-10-29
Remove MOAB Static Library Build

pshriwise:
Our CI builds with DAGMC usually take around 48 minutes, but they will occasionally go over the time limit. This PR removes static library builds which we don't rely on for CI anyway and shortens the DAGMC CI job by ~4 min. This should greatly reduce the frequency of these job timeouts.


paulromano (approved):
Simple change here that should hopefully avoid the intermittent test failures. Thanks @pshriwise!

PR #1717, opened 2020-11-03
 Enable access to the MOAB interface pointer from the Unstructured Mesh class through the C API

helen-brooks:
- Promotes MOAB interface inside Unstructured mesh class to a shared pointer
- External access to MOAB interface via model namespace through a map from Mesh ID -> MOAB pointer (allows multiple concurrent unstructured meshes)
- Additional attributes for Unstructured mesh (create, load) to allow the user to call the MOAB Interface constructor and load_file methods themselves through the C API
- New test: external_mesh (tests that the external C API works with no change in results compared to internal load, and that inconsistent settings exit correctly)


pshriwise:
Thanks for the PR @helen-brooks! Can you provide a little context as to the use case for these changes?

helen-brooks:
> Thanks for the PR @helen-brooks! Can you provide a little context as to the use case for these changes?
Sure. The context / use case is for embedding OpenMC via the C API in an external application. By both allowing access to the underlying MOAB interface for the unstructured mesh, and adding controls that block the automatic creation and loading of a mesh into MOAB inside OpenMC,  this provides maximal flexibility where additional operations on that mesh may be required. In my case that is in fact the creation of a mesh that is already in memory in another format, so MOAB's load_file method is never called. Even if a given user does not want to access the mesh prior to its use for tallying in OpenMC (in which case they can set create=true and load=true) the mere ability to access the MOAB pointer allows for a more detailed querying of the mesh data that could be useful in post-processing.

helen-brooks:
@pshriwise  I also noticed that the Travis build appears to have failed. I'm not entirely sure why that might be. I don't have any issues on my own system, and didn't change anything to do with the build scripts. Perhaps it's possible I missed an #ifdef DAGMC somewhere. But if you know what's going on and it's something I can look at while the review is ongoing then please let me know :-)

makeclean:
I think its because only unstructured mesh should have any idea about moabptrs, and therefore maybe is in the wrong part of the file?

helen-brooks:
> I think its because only unstructured mesh should have any idea about moabptrs, and therefore maybe is in the wrong part of the file?

Yes, I just realised I did indeed miss an #ifdef DAGMC  statement around the model::moabPtrs in mesh.cpp (and didn't remember to test compilation with dagmc=off).  Is it possible to just ammend this PR, or do I need to submit a new one?

makeclean:
Make your changes in your branch, commit and then push and then CI will launch again

helen-brooks:
Ah excellent. Wasn't' sure it would automatically update. All done

pshriwise:
Thanks for clarifying @helen-brooks! I agree that it would be a nice capability to have, particularly for multiphysics simulations.

After giving this some thought, I'm wondering if what you want to accomplish might be possible entirely on the C/C++ side without modification of the Python API.

The addition of an `UnstructuredMesh::(moab::Interface*)` or `UnstructuredMesh::(std::shared_ptr<moab::Interface>` constructor could be used to create the object directly in C++. This object can then be added to the `openmc::model::meshes` vector and used in a `MeshFilter`.

As for the C side:

Right now we have some capability in place for allocating space for new mesh objects via the `openmc_extend_meshes` CAPI function -- though you would need more than that I think. In a branch I'm working on that supports unstructured mesh for libMesh, I have a simple function for creating new unstructured meshes from a filename in the CAPI, [`openmc_add_unstructured_mesh`](https://github.com/pshriwise/openmc/blob/40d5e359625fa2a356bbfcdbe79cb8c6f5c0102e/src/mesh.cpp#L1272). It sounds like what you'd like to do is very similar, but passing a `MOAB::Interface*` rather than a filename. This would also require the additional `UnstructuredMesh` constructor described above.

Would some combination of these options work equally well for your purposes?

makeclean:
always a big fan of leaving the python alone :+1:

makeclean:
@helen-brooks could you rebase this branch and push again? @pshriwise any chance of a reminder where we are on this one?

helen-brooks:
@makeclean The outcome I believe @pshriwise and I agreed upon was to not merge this PR as is, due to incoming changes in the  mesh class (something to do with a libmesh interface, I think). We thought we'd go down a slightly different route of adding in an alternative constructor as Patrick describes in his previous comment. I intend to close this PR when I've seen the new Mesh changes come in, then make a new changeset and new PR. I don't know what the status of these changes is (haven't been following closely...) Could @aprilnovak comment on this?

pshriwise:
Looking at the goals of this PR in the description, I think the changes added during the addition of support for libMesh unstructured mesh in #1735 partially cover what's needed here. A new CAPI function is available for adding an `UnstructuredMesh` to OpenMC that takes a library type (`moab`/`libmesh`) and a filename.

https://github.com/openmc-dev/openmc/blob/4c4b7a82343fea6082978fd0b899eb50917737d7/src/mesh.cpp#L1382-L1384

What remains to be supported is access to the MOAB pointer of the `MOABMesh` class, which should be a pretty quick PR adding an accessor method. Right now it is stored on the class as an `openmc::unique_ptr`. If we're only accessing it on the `MOABMesh` class, I'd expect that would be fine. From the description of the scenario above

>  In my case that is in fact the creation of a mesh that is already in memory in another format...

we may want to consider making the MOAB instance attribute an `openmc::shared_ptr` and add a constructor for `MOABMesh` which takes a MOAB instance instead of a filename.

Also, if the mesh is changing between OpenMC runs in Aurora we may need to provide methods for updating the data structures on the `MOABMesh` class (kDTree, barycentric coords, etc.) as well?

helen-brooks:
This PR is now superseded by #1851 so I am now closing this PR.

PR #1718, opened 2020-11-04
Fixed several bugs related to decay-rate

mkreher13:
1) In mdgxs.py, DecayRate, get_xs(). There is no need to reverse the order
since there isn't an energy group filter anymore. Also, the new_shape
addition formats the cross section like all the others. I updated
mgxs_library_hdf5 test results accordingly.

2) In material.cpp. If you use a macroscopic density other than 1.0,
the code still pushed back '1.0'. My change fixes that. Using a macro
density other than 1.0 wasn't tested anywhere, so I modifed mg_convert
to take a density of 1.1. The default macro density of 1.0 is used in
many other tests.

3) tally_scoring.cpp. Notice that the score is accumulated, then sent
to the function score_fission_delayed_dg where the tally is accumulated.
This amounts to "double accumulation". It makes sense to accumulate the score
over the delayed groups if there is no delayed group filter, but in this
case, when delayed group filter is not None, the score needs to reset
each time.

In other parts of SCORE_DECAY_RATE, I think the distinction between
accumulating and resetting the score is correct. The only place I wasn't
sure was line 1057 (and its equivalent in multi-group, line 1962).

It's too bad the decay-rate isn't tested very thoroughly. Although I
don't have time to do it myself, I think adding a DelayedGroups filter
to the mg_tallies test would be a good start.


mkreher13:
@nelsonag Let me know if there are any clarifications I can make to help this pull request along

nelsonag (commented):
Oof, thanks for the reminder @mkreher13!!

No comment on the code, and once again thank you very much for basically having to go through this pain by virtue of being the first user on the kinetics parameters stuff since... well I guess at least before the C++ conversion!

No question on the code, but this:
- Can you summarize for me what did you do to verify the correctness of the mg_convert test?

mkreher13:
Thanks! 

In my branch transient-learn, I am running a full transient of the C5G7 benchmark in multi-group mode. The transient is a moderator density drop. Before making the change in materials.cpp, the transient wasn’t progressing (because the density kept getting pushed back as 1.0). After the change in materials.cpp, the moderator density dropped as expected and the results were good. 

For the mg_convert test: on one hand, all the k-eff results are different, as expected. On the other hand, the k-eff values are not consistently all larger or smaller - I would have expected a density change to produce a consistent response. So while I didn’t verify that the mg_convert test is producing physical results, I do have evidence from C5G7 that the changes in materials.cpp produce physical results. 

If that’s insufficient, it might be safer to keep this test untouched and unfortunately not have a test for macro densities other than 1.

nelsonag (approved):
@mkreher13 Thanks for the explanation. What you explained is roughly equivalent to what is done for many of these tests which can only be tested by simulating histories, and so that is sufficient.

I approve and will merge.

PR #1719, opened 2020-11-05
Fix the way that depletion operator distributes burnable materials

kingyue737:
As reported in [discourse](https://openmc.discourse.group/t/modification-of-differentiate-burnable-mats-in-depletion/841), volume of a burnable material will be divided by mistake multiple times, if different cells are filled by it. Furthermore, the cells that appear only once in geometry will not be assigned a new cloned burnable material even if this material has multiple instances.

This PR fixes this issue by distributing each depletable material in all cell instances filled by it.


drewejohnson (approved):
Good find @kingyue737!

PR #1726, opened 2020-11-18
Export of depletion results to openmc.Materials

gridley:
Builds off #1714 

This PR just ties in a few pieces of feedback from the PR mentioned there. Now, OPENMC_CROSS_SECTIONS or the cross_sections element of a materials.xml file is used to determine which nuclides should be added to depleted materials. I could have made a PR to @rockfool's fork, but didn't so you don't have to worry about this Jiankai. And then we could get feedback from Drew and Paul faster on this.

Anyways, this also includes the recommendation of returning a `Materials` instance, and does not export to XML by default. Rather, if a certain kwarg is provided for a new XML file name, the new materials XML file will be written. The new `Materials` instance is always returned, though.

Non-burnable materials are also included in the result.

Was there anything else you guys (@paulromano and @drewejohnson) think would be useful for this?


drewejohnson:
Apologies in the delay getting to this. Been a wild few weeks on my end.

To the matter at hand. I'm confused as to what PR should be reviewed for this feature: this or #1714. Is @rockfool is still working on their implementation? I'm not sure how I feel about having two PRs up for the same feature, especially with one being a fork of the other.

I do like the fact that this pulls in information from a `materials.xml` file, but I'm not sure if that should be the responsibility of this method. I'm still partial to the idea of just extracting the burnable materials, and then updating outside this method - https://github.com/openmc-dev/openmc/pull/1714#issuecomment-721824650 A workflow like
```python
r = openmc.deplete.ResultsList.from_hdf(rfile)
burnable = r.export_materials()
all_mats = openmc.Materials.from_xml(mfile)
# or from Materials built by a script and not exported yet
all_mats.update(burnable)
all_mats.export_to_xml("materials.xml")
```
is more verbose than the current PRs, but I think it yields a lot more control to the user / developer. For the case where new materials are added through burnup (refueling, reprocessing operations? maybe not) these may not exist in a file that has been currently exported. But this also could be some premature optimization / YAGNI

rockfool:
I am also sorry for this delayed follow-up. I think we can proceed with this in parallel. And you (@drewejohnson ) can approve the one you like.

gridley:
Yeah, even in the refueling/reprocessing case, if it's anything like serpent, new materials are never added over the duration of the calculation. So, yeah, I mainly just put in this separate PR since @rockfool probably has lots of other stuff to do aside from address comments on that PR. Guess it should have gone in as a PR to your PR Jiankai, but since there are so few changes, I figured it would be easiest to just put in another.

Anyways, yeah, IMO it's a bit YAGNI to add in that level of generality. If someone comes up on the user list asking for this feature, we'll know what to do--and it won't be very much work to rewrite this to accommodate.

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+
+    def export_to_materials_xml(self, burnup_index, nuc_with_data=None, new_xml_name=None):
+        """Export Materials.xml file from the depletion results
+
+        Parameters
+        ----------
+        burn_index : int
+            Index of burnup step to evaluate
+        nuc_with_data : Iterable of str, optional
+            nuclides list to evaluate with neutron data. If not provided,
```
drewejohnson, line 205: The first sentence could be improved because we don't really need it to be a list, just some iterable of strings. Maybe something like `Nuclides with cross section data" and then a small note on why this is necessary (some isotopes tracked through depletion do not have xs)

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+        mat_file : Materials
+            A modified Materials instance containing depleted material data,
+            plus original isotopic compositions of non-depletable materials
+        """
+        result = self[burnup_index]
+
+        # Only materials found in the original materials.xml file will be
+        # updated. If for some reason you have modified OpenMC to produce
+        # new materials as depletion takes place, this method will not
+        # work as expected and leave out that material.
```
drewejohnson, line 224: I think this deserves to be in the docstring. It doesn't really help a developer, and we don't tell the user that we are pulling from `materials.xml` which should exist in the working directory

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+            else:
+                this_library = DataLibrary.from_xml()
+
+            # Find neutron libraries we have access to
+            available_cross_sections = []
+            for lib in this_library.libraries:
+                if lib['type'] == 'neutron':
+                    available_cross_sections.extend(lib['materials'])
+            if not available_cross_sections:
+                raise Exception('No neutron libraries found in cross_sections.xml')
```
drewejohnson, line 247: Not a huge fan of throwing a generic `Exception` but I also can't think of a great replacement. Maybe something like https://github.com/openmc-dev/openmc/blob/2115cae40aa242ec0e2df31c8a43b92f2ba3babb/openmc/exceptions.py#L21

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+        # results, and save them to the new depleted xml file.
+        for mat in mat_file:
+            if str(mat.id) in result.mat_to_ind.keys():
+                mat.volume = result.volume[str(mat.id)]
+                for nuc in result.nuc_to_ind.keys():
+                    atoms = result[0, str(mat.id), nuc]
+                    if atoms > 0.0:
+                        atoms_per_barn_cm = 1e-24 * atoms / mat.volume
+                        if nuc in available_cross_sections:
+                            mat.add_nuclide(nuc, atoms_per_barn_cm)
```
drewejohnson, line 259: I think we can use the `mat_to_ind` and `nuc_to_ind` dicts to slice directly into `Results.data` 
https://github.com/openmc-dev/openmc/blob/2115cae40aa242ec0e2df31c8a43b92f2ba3babb/openmc/deplete/results.py#L87-L93
 by iterating with `result.nuc_to_ind.items()`. This would allow us to pull out the material data with `results[0, str(mat.id), :]` and then index directly into the data. Then we don't have to call `Results.__getitem__` for every nuclide.

drewejohnson, line 259: I'd prefer to place this section earlier in the loop, and then continuing if the nuclide does not have xs. Skip out on some indexing and minor math operations on the assumption that more nuclides will be present in the depletion chain than have cross sections. Might not hold for reduced chains like CASL though

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+
+        if new_xml_name:
+            mat_file.export_to_xml(path=new_xml_name)
```
drewejohnson, line 262: Is it really a burden to do something like `results.export_materials().export_to_xml(new_xml_name)`?

drewejohnson (changes requested):
@rockfool please see the comments here if you wish to incorporate them into #1714 

I've made some recommendations for how I believe this PR could be improved. One big thing that both PRs are missing is testing. 

I will bring up the modular approach from before one more time, since it would be my preference. That would allow easier testing in my opinion: can we get the burnable materials from the result file. There is some overlap, and the functionality would not be fully complete w/o the `Materials.update` feature. Which I'm happy to take on

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+
+        return mat_file
```
paulromano, line 264: I agree with @drewejohnson here. I'd much prefer the method just return a `Materials` instance, and leave the XML exporting as a separate concern.

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+        # work as expected and leave out that material.
+        mat_file = openmc.Materials.from_xml("materials.xml")
+
+        # Only nuclides with valid transport data will be written to
+        # the new materials XML file. The precedence of nuclides to select
+        # is first ones provided as a kwarg here, then ones specified
+        # in the materials.xml file if provided, then finally from
+        # the environment variable OPENMC_CROSS_SECTIONS.
+        if nuc_with_data:
+            available_cross_sections = nuc_with_data
```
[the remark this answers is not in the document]
gridley: Alright, sounds good. I think it's best to avoid set conversion though. If the user passes in a set, great. This code is in no way performance-critical or slow from the get-go, so I see no point in converting to a set. It just adds extra code. May as well keep it minimal. What if the user passes in some iterable which isn't convertible to a set? Unlikely, yes, but I'd rather not convert for those two reasons.

openmc/deplete/results_list.py
```diff
@@ -192,3 +193,72 @@ def get_depletion_time(self):
...
+            if mat_file.cross_sections:
+                this_library = DataLibrary.from_xml(path=mat_file.cross_sections)
+            else:
+                this_library = DataLibrary.from_xml()
+
+            # Find neutron libraries we have access to
+            available_cross_sections = []
+            for lib in this_library.libraries:
+                if lib['type'] == 'neutron':
+                    available_cross_sections.extend(lib['materials'])
```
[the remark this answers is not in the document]
gridley: Here, I definitely agree if we're constructing it. I've incorporated that change.

openmc/deplete/results_list.py
gridley: Done!

openmc/deplete/results_list.py
gridley: Agreed, done.

openmc/deplete/results_list.py
gridley: Why add this extra code when it would convey precisely the same message? I find it unlikely here other code would wrap this in sufficiently complicated exception handling blocks to where throwing a plain Exception would be inconvenient.

openmc/deplete/results_list.py
gridley: Why do this? Yes, that is another way of writing this code, and if you're after high performance, sure it would be great to implement this, but this is simply not a segment of code where performance is going to matter. I would really rather not rewrite this considering it does what we expect at the moment.

gridley: Fair enough. I think this is a very reasonable change. I've added it.

openmc/deplete/results_list.py
gridley: Nope, definitely agree it's best to just return the Materials object and have the user call export_to_xml. I've made that change.

openmc/deplete/results_list.py
gridley: Agreed, done.

gridley:
Could I get some suggestions on how to test this? Not sure if it should be in regression or units. Given the dependence on an initial materials.xml and a depletion statepoint file, maybe this would fit nicely into an existing test. Not familiar enough to say where though.

drewejohnson:
I think our depletion regression test would be the most complete place to test this method. We don't have a `materials.xml` file there but it's the only place we have a full depletion file. You can probably use some of the test fixtures to grab the model or operator to print the material file.

To explain some of my comments, or the motivation therein, I recently was burned by a depletion simulation where a non-performance critical part ended up taking an inordinate amount of the simulation time. So that was fresh in my mind when writing the review. I'll address the comments directly, but I wanted to put that forward first

openmc/deplete/results_list.py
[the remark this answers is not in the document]
drewejohnson: Looking at some timing comparisons, finding something in a set is about 10x faster than finding something in a list. But we also take a noticeable hit in the list (or other iterable) to set conversion. So here it doesn't make as much sense as below when we're making the set by hand. (which you pointed out below too)

openmc/deplete/results_list.py
drewejohnson: The net effect to someone who isn't wrapping this in try / except blocks doesn't change, absolutely. But this also helps distinguish between exceptions that may be raised through this routine however small it may be.

The reason I would prefer raising a `DataError` is that the type of exception raised more closely relates to why the exception was raised. And the extra code would just be an import at the top of the file and switching this line

openmc/deplete/results_list.py
drewejohnson: Completely fair. If we find that this method gets too burdensome we can revisit this

I was just dealing with how slow some non-critical depletion reporting routines can eat a lot of simulation time for large problems (full core, thousands of burnable regions) and found myself trying to prematurely optimize :slightly_smiling_face:

gridley:
Ah, that's very interesting Drew, I wouldn't have guessed. Which piece was slowing you down a lot? This piece may be good for optimization if anyone starts doing multiple statepoint calculations at different depletion states. I'll have to test performance on a larger problem at some point.

Anyways, I'll come back to this PR this weekend! Thanks for the prompt responses.

openmc/deplete/results_list.py
gridley: Ah, didn't realize the data error exception was already present in OpenMC. I was kind of tired while responding to feedback earlier. Will definitely add this on the next iteration.

drewejohnson:
To clarify, it wasn't OpenMC. I would be very interested in timing breakdowns of how OpenMC performs when depleting large problems but haven't had the chance to tackle that yet. If you find something interesting, please pass it along

And just ping me when this is ready for another review :slightly_smiling_face:

openmc/deplete/results_list.py
gridley: Alright, got this in.

gridley:
@drewejohnson alright, the more specific exceptions are in now, as well as a test for the ResultsList->openmc.Materials conversion. The test just makes a materials.xml for the initial step, then creates a materials.xml for the final step, and ensures that the SHA512 of each file matches. I verified the XML files look as expected, too.

openmc/deplete/results_list.py
```diff
@@ -192,3 +194,78 @@ def get_depletion_time(self):
...
         return times
+
+    def export_to_materials(self, burnup_index, nuc_with_data=None):
+        """Converts a depletion time point composition into an openmc.Materials
+        instance, useful for running transport calculations on individual
+        depletion time points
+
+        This requires a materials.xml file to be present in the current
+        directory. This should correspond to the materials initially
+        present in the depletion calculation.
```
drewejohnson, line 205: [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) indicates the first line of a function / method docstring should be a complete sentence, e.g.,
```suggestion
        """Create an openmc.Materials with compositions from a depletion point.
        
        This is useful for running transport calculations on individual
        depletion time points. This requires a materials.xml file to be present in 
        the current directory. This should correspond to the material initially
        present in the depletion calculation.
```

openmc/deplete/results_list.py
```diff
@@ -192,3 +194,78 @@ def get_depletion_time(self):
...
+            Index of burnup step to evaluate
+        nuc_with_data : Iterable of str, optional
+            nuclides iterable to evaluate with neutron data
+
+            This must be specified because not all nuclides appearing in
+            depletion results have associated neutron cross sections, and
+            as such cannot be used in subsequent transport calculations.
+            If not provided, nuclides from the cross_sections element of
+            materials.xml will be used. If that element is not present,
+            nuclides from OPENMC_CROSS_SECTIONS will be used.
```
drewejohnson, line 224: > iterable of str, optional

and

> This must be specified

seem to disagree

drewejohnson: Also, can you format this block as a single paragraph?

openmc/deplete/results_list.py
```diff
@@ -192,3 +194,78 @@ def get_depletion_time(self):
...
+        Returns
```
drewejohnson, line 225:
```suggestion

        Returns
```

openmc/deplete/results_list.py
```diff
@@ -192,3 +194,78 @@ def get_depletion_time(self):
...
+        mat_file = openmc.Materials.from_xml("materials.xml")
+
+        # Only nuclides with valid transport data will be written to
+        # the new materials XML file. The precedence of nuclides to select
+        # is first ones provided as a kwarg here, then ones specified
+        # in the materials.xml file if provided, then finally from
+        # the environment variable OPENMC_CROSS_SECTIONS.
+        if nuc_with_data:
+            if not all(isinstance(nuclide_name, str) for nuclide_name in nuc_with_data):
+                raise InvalidArgumentError("Expected an iterable of strings as acceptable nuclide data.")
```
drewejohnson, line 241: Non-blocking, but we have `openmc.checkvalue.check_iterable_type https://github.com/openmc-dev/openmc/blob/18af12925a99866b6d07e6769ac2346b4c5085ba/openmc/checkvalue.py#L63
that can handle this for you

tests/regression_tests/deplete/test.py
```diff
@@ -130,3 +131,44 @@ def test_full(run_in_tmpdir, problem, multiproc):
 
     # Check that no additional tallies are loaded from the files
     assert np.all(n_tallies == 0)
+
+# Checks openmc.Materials objects can be created from depletion results
+def test_depletion_results_to_material(run_in_tmpdir, problem):
+
```
drewejohnson, line 137: Gives us two spaces between functions per pep8 format spec
```suggestion


def test_depletion_results_to_material(run_in_tmpdir, problem):
    """Checks openmc.Materials objects can be created from depletion results"""
```

tests/regression_tests/deplete/test.py
```diff
@@ -130,3 +131,44 @@ def test_full(run_in_tmpdir, problem, multiproc):
...
+    last_step_materials = res_ref.export_to_materials(-1)
+
+    # Check that the conversion of the final step depleted materials XML
+    # file hashes to what we expect.
+    output_xml_file = 'last_step_materials.xml'
+    last_step_materials.export_to_xml(path=output_xml_file)
+    with open(output_xml_file, 'rb') as result_file:
+        result_file_hash = hashlib.sha512()
+        for line in result_file:
+            result_file_hash.update(line)
```
drewejohnson, line 160: The hash based testing is nice because we don't have to track the potentially large xml file generated by this test. But it can hide the reason why this test is failing (#1338). I'd prefer checking if the strings are identical directly.  And (optionally) using [`difflib.unified_diff`](https://docs.python.org/3/library/difflib.html#difflib.unified_diff) to print the difference if two files differ

drewejohnson (changes requested):
Thanks for the updates @gridley! I've got some comments on formatting of docstrings and testing, primarily moving from hash-based tests.

Also, can you merge in the latest develop to pick up the CI changes?

tests/regression_tests/deplete/test.py
gridley: Personally, I'd prefer not to make this change. Because this is a very minor piece of OpenMC, I would like to minimize the overall LOC we add to the repo in implementing this functionality.

openmc/deplete/results_list.py
gridley: The first thing only required changing the first instance of the word "not" to "if"

also did the single paragraph formatting

gridley:
@drewejohnson Do you definitely want me to rewrite the test to not use hashing? I really doubt this is going to fail in the future, and wouldn't even be terribly hard to debug if it does fail, which is why I initially elected to use hashing rather than a direct diff.

drewejohnson (commented):
@gridley I brought up the hash based testing because I know it has caused grief in the past. Enough that an issue was created to move away from it.

I understand your sentiment about not wanting to track in the xml file. Just not sure how to square the inclusion of a hash-based test with a desire from other developers to move away from them. But we also haven't made any movement away from them yet...

I know this PR has been up for a bit with the testing in but that was over the holiday break. I'd like to give anyone with some stronger opinions on #1338 and it's implications to this PR another day to chime in 

cc @openmc-dev/technical-committee

gridley:
Oh sheesh. I seem to have not kept up with the trend of moving away from hash-based testing. I have been frustrated deciphering hash-based tests in the past as well. If that is a general movement with OpenMC, I am very happy to follow along.

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
...
                 time, time_units, atol, rtol)
         )
+
+    def export_to_materials(self, burnup_index, nuc_with_data=None):
+        """Return openmc.Materials created from these depletion results.
+
+        Parameters
+        ----------
+        burn_index : int
+            Index of burnup step to evaluate
```
paulromano, line 308: It would be good to have a "See Also" section that mentions the `get_step_where` method

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
...
+        nuc_with_data : Iterable of str, optional
+            Nuclides iterable to evaluate with neutron data.
```
paulromano, line 310: This sounds awkward -- maybe something like "Nuclides to include in resulting materials".

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
...
+            This must be specified if not all nuclides appearing in
```
paulromano, line 311: "can be specified" would be more accurate since the default behavior is probably the common case

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
...
+            as such cannot be used in subsequent transport calculations.
+            If not provided, nuclides from the cross_sections element of
+            materials.xml will be used. If that element is not present,
+            nuclides from OPENMC_CROSS_SECTIONS will be used.
+
+        Returns
+        -------
+        mat_file : Materials
+            A modified Materials instance containing depleted material data,
+            plus original isotopic compositions of non-depletable materials
```
paulromano, line 322:
```suggestion
            A modified Materials instance containing depleted material data
            and original isotopic compositions of non-depletable materials
```

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
...
+        mat_file = Materials.from_xml("materials.xml")
+
+        # Only nuclides with valid transport data will be written to
+        # the new materials XML file. The precedence of nuclides to select
+        # is first ones provided as a kwarg here, then ones specified
+        # in the materials.xml file if provided, then finally from
+        # the environment variable OPENMC_CROSS_SECTIONS.
+        if nuc_with_data:
+            if not all(isinstance(nuclide_name, str) for nuclide_name in nuc_with_data):
+                raise InvalidArgumentError("Expected an iterable of strings as acceptable nuclide data.")
```
paulromano, line 339: This line exceeds 100 characters, which goes over my loose limit of 90ish.

paulromano: Also, I agree with @drewejohnson's original suggestion to use `check_iterable_type`

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
...
+
+        # Overwrite material definitions, if they can be found in the depletion
+        # results, and save them to the new depleted xml file.
+        for mat in mat_file:
+            if str(mat.id) in result.mat_to_ind.keys():
+                mat.volume = result.volume[str(mat.id)]
+                for nuc in result.nuc_to_ind.keys():
+                    if nuc not in available_cross_sections:
+                        continue
+                    atoms = result[0, str(mat.id), nuc]
```
paulromano, line 364:
```suggestion
            mat_id = str(mat.id)
            if mat_id in result.mat_to_ind:
                mat.volume = result.volume[mat_id]
                for nuc in result.nuc_to_ind:
                    if nuc not in available_cross_sections:
                        continue
                    atoms = result[0, mat_id, nuc]
```

tests/regression_tests/deplete/test.py
paulromano: I'd prefer we not add more hashed results if at all possible. How big is the raw data we're talking about here?

paulromano (changes requested):
Thanks for taking this on @gridley! This is a really useful addition for people working on depletion simulations. A few minor comments I have:

tests/regression_tests/deplete/test.py
gridley: Not big, I am working on changing it to plaintext this morning.

openmc/deplete/results_list.py
```diff
@@ -296,3 +298,72 @@ def get_step_where(
             "relative tolerances {} and {}.".format(
                 time, time_units, atol, rtol)
         )
+
+    def export_to_materials(self, burnup_index, nuc_with_data=None):
+        """Return openmc.Materials created from these depletion results.
```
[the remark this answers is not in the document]
gridley: Nice clarity improvement!

openmc/deplete/results_list.py
gridley: Good suggestion here. Was unaware iterating over dictionary keys was the default behavior.

openmc/deplete/results_list.py
gridley: Good idea! I've now added this.

openmc/deplete/results_list.py
gridley: Sounds good!

openmc/deplete/results_list.py
gridley: Yeah, OK, I give, that way is a lot cleaner. Change is in now.

gridley:
Alright, so to test out the file diff stuff, i just screwed up the reference file to see how this looks:

```
```
```

```        # Check all lines match, printing along the way
        success = True
        for line in diff_vs_expected:
            if line:
                success = False
                print(line)
>       assert success
E       assert False

/home/gavin/Code/openmc/tests/regression_tests/deplete/test.py:180: AssertionError
------------------------------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------------------------------
---

+++

@@ -12,7 +12,7 @@
     <nuclide ao="0.046295399999999994" name="O16" />
     <nuclide ao="0.00011276839999999999" name="O17" />
     <nuclide ao="4.611985879920533e-07" name="Cs135" />
-    <nuclide ao="9.843562041057666e-09" name="Gd156" />
+    <nuclide ao="7.843562041057666e-09" name="Gd156" />
     <nuclide ao="5.092224423114367e-10" name="Gd157" />
     <nuclide ao="1.1458035030465554e-08" name="I135" />
     <nuclide ao="1.0044107689339677e-05" name="U234" />
```

Any further thoughts?

tests/regression_tests/deplete/last_step_reference_materials.xml
```diff
@@ -0,0 +1,1537 @@
+<?xml version='1.0' encoding='utf-8'?>
+<materials>
+  <material depletable="true" id="1" temperature="293.6" volume="0.06674732823364922">
+    <density units="sum" />
+    <nuclide ao="1e-14" name="Xe135" />
```
gridley, line 5: OK. looks like there are redundant isotopes in the file as it stands. Hm. Going to have to look into this. This PR never ends...

gridley:
OK, after a little bit more fiddling, I believe this is fully ready to go.

paulromano (approved):

paulromano:
@drewejohnson Are you satisfied with this PR?

drewejohnson (approved):
Thanks @gridley for working through this PR! And thanks to @rockfool for kicking it off.

Looks like GH is expecting some additional CI checks. Maybe related to #1743 ? Might have to merge in changes from the develop branch

pshriwise:
Shauna Coxsey*

Patrick Shriwise
608-446-8173

gridley:
OK, rebased, let's see if CI works now...

Issue #1731, opened 2020-11-25
Potential CI Service Transition


pshriwise:
@paulromano and I were discussing Travis' plans for changing its CI billing model. 

 https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

The new model doesn't require payment, seeing as this is still a public project, but it does make our build time much more restrictive (~10k minutes per month). Given that a conservative estimate for a single CI build for us is ~240 minutes, I don't think that's going to cut it. :grimacing:

I'm creating this issue to hopefully start a discussion on how to address this going forward. In the likely scenario where we move a new service, some of the alternatives brought up were:

  - Circle CI
  - GitHub actions
  - Azure pipelines

I'm all ears for any thoughts or additions to the list!

pshriwise:
The only one of these I have experience with is Circle CI, which makes for some really quick builds as it can cache images for specific steps in the CI build and really cut down on the run times. It can be a bit of a pain to set up though, and I'm not sure how well it will handle updating images for changes in our dependencies -- something I rather like about our current CI system. I know very little about GitHub actions (I believe @Shimwell has some experience there), but the idea of having CI in the same platform as the repo is attractive to me.

makeclean:
When @shimwell see's this he'll be giddy :) There are a slew of benefits, one of them is automatic production of containers after merge, that was always true in Travis and Circle too, but I think it's easier with GH actions
________________________________

Shimwell:
I have been very impressed with how easy Github Actions is to setup and how it has added features. The reviews comparing Circle and Github Actions I've read sometimes have updates as Github Actions has been improved even during the review. I would expect Github Actions to continue to add features and to have the benefit of having a single platform but the disadvantage of lock in to a particular platform. 

This table offers straight forward comparison https://knapsackpro.com/ci_comparisons/circle-ci/vs/github-actions

I have some a minimal OpenMC test repositories that might be of interest for you[ https://github.com/Shimwell/example-git-actions-docker-openmc]( https://github.com/Shimwell/example-git-actions-docker-openmc)

We are actually trying to decided between the Circle and Github Actions over on the [Paramak](https://github.com/ukaea/paramak) repo and are currently using both, we still can't decide even after using them both on the same repo:
- Github Actions to build the dockerfile and then run pytest 
- Github Actions to build push the docker image to dockerhub.
- Circle CI to pull a static docker image from dockerhub and test.
- Circle CI to build the docker image and then run pytest

[...]

PR #1735, opened 2020-12-09
Support for libMesh unstructured mesh filters 

pshriwise:
This adds support for LibMESH unstructured mesh filters. The current `UnstructuredMesh` class is changed to an abstract class containing a general, interface for unstructured mesh representations. The `openmc::LibMesh` and `openmc::MOABMesh` classes implement those interfaces as needed for each library. This relies on the latest libMesh release (1.6.0) which includes some changes from the following PRs for initialization and build compatibility:

  - https://github.com/libMesh/libmesh/pull/2682
  - https://github.com/libMesh/libmesh/pull/2548

### libMesh Initialization
This was a little bit tricky and involves the management of a `libmesh::LibMeshInit` object, which is required for the creation of new mesh instances and file loading. The primary issue here is whether or not OpenMC is the primary application. OpenMC will initialize the libMesh library for standalone simulations, passing on an MPI communicator and the number of threads to use. If libMesh is already initialized, OpenMC expects that the `libMesh::LibMeshInit` object is already set by the external application driving OpenMC.

### `libMesh::PointLocatorBase` Objects
These objects access the Octree build around the libMesh elements. Because calls to these objects are not automatically threadsafe, a threadprivate `PointLocatorBase` has been added to the `LibMesh` class. One of these objects gets initialized per thread during construction of the `LibMesh` object.

While we typically are trying to avoid thread private variables in favor of state on the `Particle` class, the state associated with the `libMesh::PointLocatorBase` object is tied explicitly to the specific `libMesh::Mesh` instance in the `openmc::LibMesh` class. The associated state on the object is an optimization cache which should not change the behavior of point location queries, so there's no need to keep it on the Particle class.

### Estimators

No ray tracing capabilities currently exist in libMesh, but I'm hopeful that they will in the future. As a result, this filter can only be used with a collision estimator. The appropriate checks have been added during initialization to ensure any tallies using one of these meshes as a filter are not using a tracklength estimator.

### IO Changes
A minor addition to the XML formats is the specification of a library to use (either MOAB or libMESH) when using this filter.

This addition is supported in the associated Python classes for exporting XML files and importing from HDF5. 

### CAPI Changes
The `openmc_add_unstructured_mesh` function is added here which allows an unstructured mesh to be created externally.

### Testing

The same test approach from #1527 is used here. Because both MOAB and libMesh can read exodus files, I've updated the unstructured meshes in the test suite to use that format and updated the MOAB build in CI to enable compilation with `netcdf`. 

The libMesh tests are currently implemented in our Travis system, but they may need to be updated as the libMesh build takes the majority of our allotted time in CI and we'll likely be moving away from that system in the near future (#1731).

Update: 

As @aprilnovak pointed out below, this implementation works for any 3D element supported by LibMesh's `PointLocator`.


gridley:
This is awesome. Just checking: have you seen Derek Gaston and Logan Harbour’s ray tracing module in MOOSE? It fills that gap in libMesh. There should be a PR up on the MOOSE repo for it now.

pshriwise:
@gridley I did! Very excited about it :) I'm hoping that some of that gets pushed into libMesh itself at some point. :crossed_fingers:

src/mesh.cpp
```diff
@@ -102,6 +121,70 @@ StructuredMesh::bin_label(int bin) const {
...
+    // write volume of each tet
+    std::vector<double> tet_vols;
+    xt::xtensor<double, 2> centroids({(size_t)this->n_bins(), 3});
+    for (int i = 0; i < this->n_bins(); i++) {
+      tet_vols.emplace_back(this->volume(i));
+      auto c = this->centroid(i);
+      xt::view(centroids, i, xt::all()) = xt::xarray<double>({c.x, c.y, c.z});
+    }
+
+    write_dataset(mesh_group, "volumes", tet_vols);
```
aprilnovak, line 183: I know this was around before, but would you remind me the motivation for writing the volumes and centroid for the UM? It doesn't look like a similar action is performed for the `RectilinearMesh`, for instance. Just wondering if those structured meshes should perhaps be made consistent with the UM case.

src/mesh.cpp
```diff
@@ -1852,6 +1960,208 @@ UnstructuredMesh::write(std::string base_filename) const {
...
+
+LibMesh::LibMesh(const std::string& filename)
+{
+  filename_ = filename;
+  initialize();
+}
+
+void LibMesh::initialize()
+{
+  m_ = std::make_unique<libMesh::Mesh>(settings::LMI->comm(), 3);
```
aprilnovak, line 1980: For readability, I'd change `3` to `n_dimension_`

src/tallies/tally.cpp
```diff
@@ -282,6 +282,21 @@ Tally::Tally(pugi::xml_node node)
...
   }
+
+#ifdef LIBMESH
+  // ensure a tracklength tally isn't used with a libMesh filter
+  for (auto i : this->filters_) {
+    auto df = dynamic_cast<MeshFilter*>(model::tally_filters[i].get());
+    if (df) {
+      auto lm = dynamic_cast<LibMesh*>(model::meshes[df->mesh()].get());
+      if (lm && estimator_ == TallyEstimator::TRACKLENGTH) {
+        fatal_error("A tracklength estimator cannot be used with "
```
aprilnovak, line 293: It looks like there are a few different places that you check whether the estimator is TRACKLENGTH. Do we need this one here in addition to the check in `bins_crossed`?

tests/regression_tests/unstructured_mesh/inputs_true1.dat
```diff
@@ -57,35 +57,34 @@
...
-  <mesh id="3">
+  <mesh id="1">
     <dimension>10 10 10</dimension>
     <lower_left>-10.0 -10.0 -10.0</lower_left>
     <upper_right>10.0 10.0 10.0</upper_right>
   </mesh>
-  <mesh id="4" type="unstructured">
-    <filename>test_mesh_tets.h5m</filename>
+  <mesh id="2" library="libmesh" type="unstructured">
+    <filename>test_mesh_tets.exo</filename>
```
aprilnovak, line 73: In `LibMesh::write(...)`, the message suggests you're writing a file with `.e` extension, but here you are supplying a file with `.exo` extension. In my experience these two extensions are interchangeable (can just rename `cyl.exo` to `cyl.e` and things work fine), but I think it'd be nice for consistency purposes to just stick with one or the other from a documentation POV.

Same comment applies for several other files in this PR.

src/mesh.cpp
```diff
@@ -102,6 +121,70 @@ StructuredMesh::bin_label(int bin) const {
   }
 }
 
+//==============================================================================
+// Unstructured Mesh implementation
+//==============================================================================
+
+UnstructuredMesh::UnstructuredMesh(pugi::xml_node node) : Mesh(node) {
+  n_dimension_ = 3;
```
aprilnovak, line 129: Should there be any additional checks either here, or in the derived classes, that the mesh is indeed 3D? For instance, the constructor for `libMesh::Mesh` that is used later takes a dimension; from the libMesh documentation:

> Constructor. Takes dim, the dimension of the mesh. The mesh dimension can be changed (and may automatically be changed by mesh generation/loading) later.

Does anything break if we pass 3 as the mesh dimension into `libMesh` mesh, when in fact the mesh is not 3D?

src/initialize.cpp
```diff
@@ -54,6 +59,34 @@ int openmc_init(int argc, char* argv[], const void* intracomm)
...
+  // pass command line args, empty MPI communicator, and number of threads
+  settings::LMI = std::make_unique<libMesh::LibMeshInit>(argc, argv, 0, n_threads);
+#endif
+}
+
+if (!settings::LMI) {
+  fatal_error("libMesh::LibMeshInit object isn't set.");
+}
+#endif
+
```
aprilnovak, line 89: This is fine as-is, but if I had to be really nit-picky, I would maybe move this after `read_input_xml()`? Because we don't need to do these steps unless the user is actually using an unstructured mesh, right? But for simplicity's sake I'm totally fine with always doing this.

src/mesh.cpp
```diff
@@ -102,6 +121,70 @@ StructuredMesh::bin_label(int bin) const {
...
+
+  // check the mesh type
+  if (check_for_node(node, "type")) {
```
aprilnovak, line 132: Does this check need to be here? In `read_meshes(pugi::xml_node)`, we only build a `libMesh` or `moabMesh` if the node type is ``unstructured``:

```
 #ifdef DAGMC
     } else if (mesh_type == "unstructured" && mesh_lib == "moab") {
       model::meshes.push_back(std::make_unique<MOABMesh>(node));
 #endif
 #ifdef LIBMESH
     } else if (mesh_type == "unstructured" && mesh_lib == "libmesh") {
       model::meshes.push_back(std::make_unique<LibMesh>(node));
 #endif
```

aprilnovak (changes requested):
This looks great @pshriwise! These are all the comments I have now just from looking at the source. I'm going to spend some time testing this out and may have additional comments from the user POV.

aprilnovak:
One question - it looks like `MoabMesh` is based on tetrahedral meshes, but it doesn't look like there's any such notation for `LibMesh`. Are there any limitations on what elements can be used in the `LibMesh` mesh? nekRS/Nek5000 are limited to hexahedral elements, which is why I'm asking.

include/openmc/settings.h
```diff
@@ -6,6 +6,7 @@
 
 #include <array>
 #include <cstdint>
+#include <memory>
```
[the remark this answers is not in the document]
pshriwise: Ah yeah I had a variable declared in there that required it and moved it elsewhere to reduce changes. Missed this include though. Thanks!

src/mesh.cpp
pshriwise: This provides a way to perform at least a rudimentary visualization of the data that we can support in the Python API without explicit storage of the element vertices, faces, connectivity, etc. 

The second unstructured mesh example demonstrates this (https://docs.openmc.org/en/stable/examples/unstructured-mesh-part-ii.html). It's not ideal, but it's something relatively simple that we can support.

pshriwise: Now that I'm thinking about this... given that we have an option for writing a "marked up" version of the mesh. We could write only the mesh in a VTK format and markup the mesh using the python vtk module :thinking:

src/tallies/tally.cpp
pshriwise: Good question. In my mind, they do serve different purposes. This check helps us by failing early when reading tallies from XML input. The other in `bins_crossed` provides some helpful output if a user is setting up tallies using the CAPI/C++API and manages to define a tally using a LibMesh filter and TL estimator -- in which case this check would be bypassed.

tests/regression_tests/unstructured_mesh/inputs_true1.dat
pshriwise: Good point. Best to be consistent about that. I'll opt for `.e` and update for consistency.

src/tallies/tally.cpp
aprilnovak: gotcha, sounds good!

src/mesh.cpp
pshriwise: Ah yeah we should definitely check that. I'll add that in.

pshriwise: I suppose a check that the input geometry is 3D in the appropriate location would also be good.

src/initialize.cpp
pshriwise: Totally valid point. We should only need that object to be set if we're actually going to be using the capability. I'll move that check into the `LibMesh::initialize` method just before we use it to construct a mesh.

src/mesh.cpp
aprilnovak: That might be a nice feature - the "rudimentary visualization" you're talking about is the point cloud?

src/mesh.cpp
aprilnovak: Yes, both of those would be good. I think you'll need to put the check on the libMesh mesh's dimension (i.e. what is in the exodus file) after calling `initialize()` based on the doc string from libMesh above.

libMesh also has two different ideas of "dimension" - for instance, a 2-D plane that isn't aligned with a Cartesian axis could be considered either 2-D or 3-D (because one coordinate x, y, or z isn't fixed for all points). I think we probably want the 2-D check.

src/initialize.cpp
aprilnovak: I like that solution better than what I proposed. Would you be able to access `argv` and `argc` though?

src/mesh.cpp
pshriwise: As the code stands, this is true. Long-term though, I don't mind having an extra check in the constructor. Having the check in the XML constructor ties the object creation to the XML spec we provide and insulates it from any exterior logic, which I like. Open to other thoughts about this though!

pshriwise:
> One question - it looks like MoabMesh is based on tetrahedral meshes, but it doesn't look like there's any such notation for LibMesh. Are there any limitations on what elements can be used in the LibMesh mesh? nekRS/Nek5000 are limited to hexahedral elements, which is why I'm asking.

There are no limitations on the type of 3D elements used in the LibMesh mesh. I forgot to mention that in the PR description. So you can use any type you like, but no TL estimators for now.

src/initialize.cpp
pshriwise: Oh I just meant we'd move the check that the object is set.

```
if (!settings::LMI) {
    fatal_error("Attempting to use an unstructured mesh without a libMesh::LibMeshInit object.");
}
```

src/mesh.cpp
pshriwise: I was just looking at that as well. I think I used the wrong one in my latest commit. I'll make that change shortly.

src/mesh.cpp
pshriwise: Yeah.

src/initialize.cpp
aprilnovak: oh gotcha! yes that sounds good to me

aprilnovak:
Ok, I'm trying this out now. Regarding documentation, a few comments (I'll just update this comment if I find other places)

- [ ] Add CMake option to `docs/source/userguide/install.rst`

src/mesh.cpp
aprilnovak: That's fine with me! Just checking

aprilnovak:
I think the `openmc-plot-mesh-tally` script needs to be updated to handle the unstructured mesh option? I got this error when trying it:

```
AttributeError: 'UnstructuredMesh' object has no attribute 'dimension
```

aprilnovak:
Works nicely, @pshriwise! Here's a sphere mesh I put at the center of your cube problem in `regresstion_tests/unstructured_mesh` (radius 3 - so, the entire mesh is in the problem):

Screen Shot 2020-12-11 at 10 32 40 AM

And then here's a sphere that is bigger than the mesh - the tallies are indeed zero in the regions that extend past the cube:

Screen Shot 2020-12-11 at 10 42 29 AM

Very cool! No question here, just admiring :)

src/mesh.cpp
aprilnovak: Can confirm that this check works for a 2-D plane not parallel to a Cartesian axis.

aprilnovak: Oh actually, but it looks like if you error at this stage, some memory isn't handled correctly:

```
 Reading tallies XML file...
 ERROR: Mesh file mesh.e specified for use in an unstructured mesh is not a 3D
        mesh.
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=-1
:
system msg for write_line failure : Bad file descriptor
Memory leak detected!
Compile in DEBUG mode with --enable-reference-counting
for more information
```

Maybe this isn't important since you're failing anyways, but if it could be fixed that'd be preferable for me.

src/mesh.cpp
```diff
@@ -102,6 +121,70 @@ StructuredMesh::bin_label(int bin) const {
...
+    auto temp = get_node_value(node, "type", true, true);
+    if (temp != "unstructured") {
+      fatal_error(fmt::format("Invalid mesh type: {}", temp));
+    }
+  }
+
+  // get the filename of the unstructured mesh to load
+  if (check_for_node(node, "filename")) {
+    filename_ = get_node_value(node, "filename");
+  }
```
aprilnovak, line 142:
```suggestion
    if (!file_exists(filename_)) {
      fatal_error("Mesh file '" + filename_ + "' does not exist!");
    }
  }
```

src/mesh.cpp
pshriwise: Yeah same. We should handle that differently. Thanks for checking that! I'll look into that shortly.

src/mesh.cpp
aprilnovak: Not sure how to do a similar commit suggestion for a line that hasn't changed... anyways, if you like this, you'd need to `#include "openmc/file_utils.h"` as well.

This check isn't strictly necessary since it still fails, but without it, you get a big dump of a stack trace that is kind of messy and just says "Error reading ExodusII file", not necessarily the more specific message that it's missing.

pshriwise: Mmm yeah this is much better than an Exodus error. I'll include it and the necessary header.

src/mesh.cpp
pshriwise: Welp, I was hoping this would be simple but, apparently, all existing `libMesh::Mesh` instances need to be cleaned up before the `LibMeshInit` object we maintain gets destroyed or we'll see this output. This makes it trick as we could potentially have several mesh instances loaded and it seems messy to have to go find all of those in our vector of meshes -- particularly because all of this is happening inside an object constructor. I'll have to think on this a bit more.

src/state_point.cpp
```diff
@@ -54,6 +54,11 @@ openmc_statepoint_write(const char* filename, bool* write_source)
   // Write message
   write_message("Creating state point " + filename_ + "...", 5);
 
+#if defined(LIBMESH) || defined(DAGMC)
+      // write unstructured mesh tallies to VTK
+      write_unstructured_mesh_results();
```
paulromano, line 59:
```suggestion
  // write unstructured mesh tallies to VTK
  write_unstructured_mesh_results();
```

src/state_point.cpp
```diff
@@ -730,6 +737,20 @@ void write_unstructured_mesh_results() {
 
       int n_realizations = tally->n_realizations_;
 
+      for (int i_score = 0; i_score < tally->scores_.size(); i_score++) {
+        for (int i_nuc = 0; i_nuc < tally->nuclides_.size(); i_nuc++) {
```
paulromano, line 741: Looks like this could be a range-for loop?

src/tallies/filter_mesh.cpp
```diff
@@ -43,6 +43,8 @@ const
   } else {
     model::meshes[mesh_]->bins_crossed(p, match.bins_, match.weights_);
   }
+
+  double total = std::accumulate(match.weights_.begin(), match.weights_.end(), 0.0);
```
paulromano, line 47: Remove?

tools/ci/gha-install-libmesh.sh
```diff
@@ -0,0 +1,24 @@
...
+# make -j4 install
+# export LIBMESH_PC=$HOME/LIBMESH/lib/pkgconfig/
+# rm -rf $HOME/LIBMESH/build
+
+cd $HOME
+git clone https://github.com/idaholab/moose MOOSE
+cd MOOSE
+
+./scripts/update_and_rebuild_petsc.sh CXX=mpicxx CC=mpicc FC=mpifort
+./scripts/update_and_rebuild_libmesh.sh CXX=mpicxx CC=mpicc FC=mpifort
```
paulromano, line 22: Is this going to be replaced with a plain install of libmesh?

paulromano (changes requested):
Hey @pshriwise -- here's an initial cut at a review (I know you're still working on this but thought I'd give some feedback along the way)

.github/workflows/ci.yml
```diff
@@ -60,17 +65,22 @@ jobs:
...
         name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-
+      -
+        name: Set up CMake
+        uses: jwlawson/actions-setup-cmake@v1.5
+        with:
+          cmake-version: 3.12.x
```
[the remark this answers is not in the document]
pshriwise: Older versions of CMake's PkgConfig module don't seem to play nicely with libMesh's configure files, unfortunately. I'm looking into the exact version we'll need to build with libMesh so we can return an error or warning if `libmesh` is enabled and the CMake version isn't sufficient.

include/openmc/mesh.h
```diff
@@ -22,12 +22,26 @@
...
+#include "libmesh/point.h"
+#endif
+
 namespace openmc {
 
 //==============================================================================
 // Global variables
 //==============================================================================
 
+extern "C" const bool libmesh_enabled;
```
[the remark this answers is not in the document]
pshriwise: Ah yeah, thanks. I'll update the `dagmc_enabled` variable while I'm at it.

include/openmc/mesh.h
```diff
@@ -410,6 +475,66 @@ class UnstructuredMesh : public Mesh {
...
+
+  //! Translate an element pointer to a bin index
+  int get_bin_from_element(const libMesh::Elem* elem) const;
+
+  // Data members
+  std::unique_ptr<libMesh::Mesh> m_; //!< pointer to the libMesh mesh instance
+  std::vector<std::unique_ptr<libMesh::PointLocatorBase>> PL_; //!< per-thread point locators
+  std::unique_ptr<libMesh::EquationSystems> equation_systems_; //!< pointer to the equation systems of the mesh
+  std::string eq_system_name_; //!< name of the equation system holding OpenMC results
+  std::map<std::string, unsigned int> variable_map_; //!< mapping of variable names (tally scores) to libMesh variable numbers
```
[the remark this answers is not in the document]
pshriwise: I don't expect this will ever be a very large map, but an unordered map should be fine so we might as well use one.

src/mesh.cpp
```diff
@@ -1177,6 +1263,46 @@ openmc_extend_meshes(int32_t n, int32_t* index_start, int32_t* index_end)
...
+                           "by this build of OpenMC", lib_name));
+    return OPENMC_E_INVALID_ARGUMENT;
+  }
+
+  int mesh_id = 0;
+  for (const auto& m : model::meshes) { mesh_id = std::max(m->id_, mesh_id); }
+  mesh_id += 1;
+
+  model::meshes.back()->id_ = mesh_id + 1;
+  *id = mesh_id;
```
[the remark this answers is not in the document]
pshriwise: Mmm good point. Thanks for catching that. After implementing `Mesh::set_id` I'm wondering if we could generalize this for all of our classes with IDs... I'll opt to leave that as a separate task though :smiley:

src/mesh.cpp
```diff
@@ -1667,18 +1772,21 @@ UnstructuredMesh::get_bin_from_ent_handle(moab::EntityHandle eh) const {
...
 
 moab::EntityHandle
-UnstructuredMesh::get_ent_handle_from_bin(int bin) const {
+MOABMesh::get_ent_handle_from_bin(int bin) const
+{
   if (bin >= n_bins()) {
     fatal_error(fmt::format("Invalid bin index: ", bin));
   }
-  return ehs_[bin];
+  return ehs_[0] + bin;
```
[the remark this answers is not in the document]
pshriwise: That is a performance optimization to avoid a search in the `MOAB::Range` data structure. Because we've just loaded the file and haven't modified the mesh, the set of EntityHandles in the MOAB instance is guaranteed to be contiguous, so we can rely on arithmetic here instead of indexing the `Range`.

src/mesh.cpp
```diff
@@ -1852,6 +1963,225 @@ UnstructuredMesh::write(std::string base_filename) const {
...
+int LibMesh::n_surface_bins() const
+{
+  int n_bins = 0;
+  for (int i = 0; i < this->n_bins(); i++) {
+    const libMesh::Elem& e = get_element_from_bin(i);
+    n_bins += e.n_faces();
+    // if this is a boundary element, sure to count boundary faces
+    // twice
+    for (int j = 0; j < e.n_sides(); j++) {
+      if (e.neighbor_ptr(j) == NULL) { n_bins++; }
```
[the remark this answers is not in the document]
pshriwise: Sure would. Thanks!

src/mesh.cpp
```diff
@@ -1852,6 +1963,225 @@ UnstructuredMesh::write(std::string base_filename) const {
...
+    auto var_num = eqn_sys.add_variable(std_dev_name, libMesh::CONSTANT, libMesh::MONOMIAL);
+    variable_map_[std_dev_name] = var_num;
+  }
+}
+
+void
+LibMesh::remove_score(const std::string& var_name)
+{
+  return;
+}
```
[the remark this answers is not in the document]
pshriwise: Yep. I need to rethink how the mesh writing is done. It's very different between MOAB and libMesh, so I think the best thing to do might be to contain all of the writing in a single method instead of the many methods that are used now.

src/mesh.cpp
```diff
@@ -1852,6 +1963,225 @@ UnstructuredMesh::write(std::string base_filename) const {
...
+
+  #ifdef _OPENMP
+  int thread_num = omp_get_thread_num();
+  #else
+  int thread_num = 0;
+  #endif
+
+  const auto& point_locator = PL_.at(thread_num);
+
+  auto e = (*point_locator)(p);
```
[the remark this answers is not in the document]
pshriwise: `e` is a libMesh element pointer. I'll update the name to `elem_ptr` for clarity. It's the only type that can be returned from the `libmesh::PointLocator`. Based on the `libMesh` interface only returning pointers or references and never copies of these objects I think it's probably best to avoid copying this object.

tools/ci/gha-install-libmesh.sh
pshriwise: Yeah, it's on my list.

include/openmc/mesh.h
```diff
@@ -243,72 +268,115 @@ class RectilinearMesh : public StructuredMesh
...
+  // Constructors
+  UnstructuredMesh() {};
+  UnstructuredMesh(pugi::xml_node node);
+  UnstructuredMesh(const std::string& filename);
 
-  void bins_crossed(const Particle& p,
-                    std::vector<int>& bins,
-                    std::vector<double>& lengths) const override;
+  // Methods
+public:
```
aprilnovak, line 281: Can we remove this second usage of `public` and just nest the non-constructor methods with the constructors? I don't see other places where we use the `public` access two times for methods

include/openmc/mesh.h
```diff
@@ -399,17 +467,77 @@ class UnstructuredMesh : public Mesh {
...
+  int get_bin_from_element(const libMesh::Elem* elem) const;
+
+  // Data members
+  std::unique_ptr<libMesh::Mesh> m_; //!< pointer to the libMesh mesh instance
+  std::vector<std::unique_ptr<libMesh::PointLocatorBase>> pl_; //!< per-thread point locators
+  std::unique_ptr<libMesh::EquationSystems> equation_systems_; //!< pointer to the equation systems of the mesh
+  std::string eq_system_name_; //!< name of the equation system holding OpenMC results
+  std::unordered_map<std::string, unsigned int> variable_map_; //!< mapping of variable names (tally scores) to libMesh variable numbers
+  libMesh::BoundingBox bbox_; //!< bounding box of the mesh
+  libMesh::Elem* first_element_; //!< pointer to the first element in the mesh (maybe should be a key?)
```
aprilnovak, line 537: Does the "maybe should be a key?" note indicate that anything remains to be addressed here? If not, I'd perhaps omit this or change the wording to explain what that means

src/mesh.cpp
```diff
@@ -102,6 +154,73 @@ StructuredMesh::bin_label(int bin) const {
...
+
+void
+UnstructuredMesh::to_hdf5(hid_t group) const
+{
+    hid_t mesh_group = create_group(group, fmt::format("mesh {}", id_));
+
+    write_dataset(mesh_group, "type", "unstructured");
+    write_dataset(mesh_group, "filename", filename_);
+    write_dataset(mesh_group, "library", this->library());
+    // write volume of each tet
```
aprilnovak, line 210: This is really minor, but since this supports any element type in libMesh, I'd remove the names having to do with tets here, and maybe just replace "tet" by "elem". Could be less confusing for someone reading this

src/mesh.cpp
```diff
@@ -102,6 +154,73 @@ StructuredMesh::bin_label(int bin) const {
...
+    std::vector<double> tet_vols;
+    xt::xtensor<double, 2> centroids({(size_t)this->n_bins(), 3});
+    for (int i = 0; i < this->n_bins(); i++) {
+      tet_vols.emplace_back(this->volume(i));
```
aprilnovak, line 214: I don't have a question here, just a comment - it would be great if we could leverage this volume calculation in Cardinal so we don't have to explicitly tell OpenMC what the volumes are for normalization!

src/mesh.cpp
```diff
@@ -1852,6 +1989,221 @@ UnstructuredMesh::write(std::string base_filename) const {
...
+  return m_->n_elem();
+}
+
+int LibMesh::n_surface_bins() const
+{
+  int n_bins = 0;
+  for (int i = 0; i < this->n_bins(); i++) {
+    const libMesh::Elem& e = get_element_from_bin(i);
+    n_bins += e.n_faces();
+    // if this is a boundary element, sure to count boundary faces
```
aprilnovak, line 2068: Can you elaborate more on this? Only faces on the boundary are counted twice?

src/state_point.cpp
```diff
@@ -748,9 +748,12 @@ void read_source_bank(hid_t group_id, std::vector<Particle::Bank>& sites, bool d
   H5Tclose(banktype);
 }
 
-#ifdef DAGMC
 void write_unstructured_mesh_results() {
 
+#if !defined(LIBMESH) && !defined(DAGMC)
```
aprilnovak, line 753: Does this check need to be here, since you only call `write_unstructured_mesh_results` if either LIBMESH or DAGMC is defined?

aprilnovak:
@pshriwise this looks really good! I just have a few small comments above, in addition to:

- In order to build with libMesh, you need to pass `DCMAKE_PREFIX_PATH=/path/to/libmesh/build` to `cmake`. This might be intuitive for lots of users, but I feel like it wouldn't hurt to mention it in the `userguide/install.rst`. 
- The `openmc-plot-mesh-tally` script doesn't work with unstructured meshes: it errors with `AttributeError: 'UnstructuredMesh' object has no attribute 'dimension'`. I think we should either fix this or at least put an error in saying that the script doesn't support unstructured mesh tallies, to be clear that the error isn't caused by the user.

src/state_point.cpp
pshriwise: Nope. Safe to remove this. Thanks!

src/mesh.cpp
pshriwise: I'll update the comment to be more clear here. 

All element faces should account for two bins -- a bin for particles that cross the face in each direction. In looping over the mesh elements this happens naturally for interior faces, but faces on the boundary will only be counted once. The inner for loop here increments the number of surface bins for any boundary faces of the element to account for this.

aprilnovak: Ah, gotcha, that makes sense! The original comment had me thinking that only the boundary faces are counted twice - thanks for clarifying

pshriwise: Glad to have revisited this. The `libMesh::Elem::neighbor_ptr_range()` method makes this loop a bit more readable.

src/mesh.cpp
pshriwise: If we're using the unstructured mesh tally, we currently do use the element volumes to normalize the tally values. The volumes in the input file are used for the cell-based tally only -- though I do think we could clean streamline that a bit.

src/mesh.cpp
pshriwise: Ah, yeah. Great point. I'll update that.

src/mesh.cpp
aprilnovak: gotcha, thanks!

include/openmc/mesh.h
pshriwise: Thanks for pointing this out! There was still some work to be done here. In the end, what I wanted was the element ID according to libMesh.

Seeing as all I do with this pointer is get the element's ID later in `LibMesh::get_bin_from_element`, I'm just going to store that `dof_id_type` rather than the raw pointer.

include/openmc/mesh.h
pshriwise: Yeah totally. Accidentally left that in there.

pshriwise:
> In order to build with libMesh, you need to pass DCMAKE_PREFIX_PATH=/path/to/libmesh/build to cmake. This might be intuitive for lots of users, but I feel like it wouldn't hurt to mention it in the userguide/install.rst.

Good point. This is true for DAGMC also, so I went ahead and added a similar comment and example for it as well.

> The openmc-plot-mesh-tally script doesn't work with unstructured meshes: it errors with AttributeError: 'UnstructuredMesh' object has no attribute 'dimension'. I think we should either fix this or at least put an error in saying that the script doesn't support unstructured mesh tallies, to be clear that the error isn't caused by the user.

Agreed. Not a very helpful error message if someone attempted to use this script with unstructured meshe. Adding support for visualizing UM in that script is non-trivial and would be better tackled as a separate endeavor. For now, I added a check for unstructured meshes in the statepoint file. The script will present a warning message if they are found and tallies with `MeshFilter`'s using unstructured meshes will not appear in the combo box.

.github/workflows/ci.yml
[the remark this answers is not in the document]
pshriwise: Ok, I finally got to the bottom of this. The `FindPkgConfig` module shipped with CMake pre 3.12.0 has an odd quirk. If _any_ library paths are specified in the `pkgconfig` file, only those paths would be searched. So if system library locations are not included explicitly, they would not be searched when finding libraries. Here is the relevant change in CMake:

https://gitlab.kitware.com/cmake/cmake/-/commit/ac5731a7e380349f19dc319e6c31e189b5faba93

This is a problem when configuring for our libMesh build because `zlib` is present in the `libmesh.pc` file. It doesn't seem like it's standard for these files to contain system library locations, so while it's not ideal I think the best solution for now is to  require CMake >=3.12.x when libmesh is enabled. I'll adjust CI so that CMake 3.12 is only used for builds with libmesh enabled so we're still ensuring compatibility with older versions of CMake in other builds.

aprilnovak (approved):

.github/workflows/ci.yml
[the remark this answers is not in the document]
pshriwise: Ok I was conflating issues here. I think I originally added that CMake specification to figure out what version of CMake would cause this configuration problem and isolate that problem so I could also sort out a libmesh configuration problem I was running into (which is now also resolved).

CMake versions < 3.12 when building with libmesh are still going to be problematic though, so I've added a check in the `FindLIBMESH.cmake` module to protect against this.

pshriwise:
Made a couple of updates here. The main one being the return of the `settings::libmesh_comm` pointer. Without this, our libMesh mesh instance creation requires that we own a `libmesh::LibMeshInit` object, which won't necessarily be the case for applications where OpenMC is being used as a library.

include/openmc/mesh.h
```diff
@@ -243,72 +265,115 @@ class RectilinearMesh : public StructuredMesh
...
+  // Methods
+private:
 
-  std::pair<std::vector<double>, std::vector<double>>
-  plot(Position plot_ll, Position plot_ur) const override;
+  //! Setup method for the mesh. Builds data structures,
+  //! sets up element mapping, creates bounding boxes, etc.
+  virtual void initialize() = 0;
+
+public:
```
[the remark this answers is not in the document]
paulromano: This looks like it hasn't been changed. I'd prefer to see:
```C++
class UnstructuredMesh : public Mesh {
public:
   ...
private:
   ...
}

File: src/mesh.cpp (hunk 21 above)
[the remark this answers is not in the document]
Comment (paulromano) [reply src/mesh.cpp]: My point was that if you do `auto e`, you _will_ get a copy, although if it's returning a pointer then you're effectively getting a copy of the pointer. `auto&` would probably be safer

File: src/tallies/tally.cpp [hunk 29]
```diff
@@ -662,6 +677,18 @@ Tally::score_name(int score_idx) const {
   return reaction_name(scores_[score_idx]);
 }
 
+std::string
+Tally::nuclide_name(int nuclide_idx) const {
+  if (nuclide_idx != -1) {
+    return "total";
```
paulromano, line 683: Unless I'm misunderstanding something, I think this check should be on the value of `nuclide` below, not on `nuclide_idx`. The default value of `nuclides_` in a tally is a vector with a single element storing -1.

tools/ci/gha-install-libmesh.sh
```diff
@@ -0,0 +1,24 @@
...
+# libMESH install
+cd $HOME
+mkdir LIBMESH && cd LIBMESH
+git clone https://github.com/libmesh/libmesh -b v1.6.0 --recurse-submodules
+mkdir build && cd build
+export METHODS="opt"
+
+if [[ $MPI == 'y' ]]; then
+../libmesh/configure --prefix=$HOME/LIBMESH CXX=mpicxx CC=mpicc FC=mpifort F77=mpif77 \
+--enable-exodus --disable-netcdf-4 --disable-eigen --disable-lapack
```
paulromano, line 16:
```suggestion
    ../libmesh/configure --prefix=$HOME/LIBMESH CXX=mpicxx CC=mpicc FC=mpifort F77=mpif77 \
        --enable-exodus --disable-netcdf-4 --disable-eigen --disable-lapack
```

tools/ci/gha-install-libmesh.sh
```diff
@@ -0,0 +1,24 @@
...
+else
+../libmesh/configure --prefix=$HOME/LIBMESH --enable-exodus --disable-netcdf-4 --disable-eigen --disable-lapack --disable-mpi
```
paulromano, line 18:
```suggestion
    ../libmesh/configure --prefix=$HOME/LIBMESH --enable-exodus --disable-netcdf-4 --disable-eigen --disable-lapack --disable-mpi
```

tools/ci/gha-install-libmesh.sh
```diff
@@ -0,0 +1,24 @@
+
+#!/bin/bash
+set -ex
+
+CURRENT_DIR=$(pwd)
+
+# libMESH install
+cd $HOME
```
paulromano, line 8:
```suggestion
# libMESH install
pushd $HOME
```
See `popd` suggestion below

tools/ci/gha-install-libmesh.sh
```diff
@@ -0,0 +1,24 @@
...
+../libmesh/configure --prefix=$HOME/LIBMESH CXX=mpicxx CC=mpicc FC=mpifort F77=mpif77 \
+--enable-exodus --disable-netcdf-4 --disable-eigen --disable-lapack
+else
+../libmesh/configure --prefix=$HOME/LIBMESH --enable-exodus --disable-netcdf-4 --disable-eigen --disable-lapack --disable-mpi
+fi
+make -j4 install
+export LIBMESH_PC=$HOME/LIBMESH/lib/pkgconfig/
+rm -rf $HOME/LIBMESH/build
+
+cd $CURRENT_DIR
```
paulromano, line 24:
```suggestion
popd
```

paulromano (changes requested):
This is looking really good @pshriwise. A few more minor suggestions below

openmc/mesh.py
```diff
@@ -622,12 +629,13 @@ class UnstructuredMesh(MeshBase):
         An iterable of element centroid coordinates, e.g. [(0.0, 0.0, 0.0),
         (1.0, 1.0, 1.0), ...]
     """
-
-    def __init__(self, filename, mesh_id=None, name=''):
+    def __init__(self, filename, library, mesh_id=None, name=''):
```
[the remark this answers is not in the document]
pshriwise: MOAB can be built with support for exodus files, so I think it's best to ask the user to be explicit.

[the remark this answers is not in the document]
pshriwise: I remember now that I was able to take advantage of this in the regression tests to reduce the number of binary files in the repo. :smile:

src/tallies/tally.cpp
pshriwise: Ah, yeah totally. Good catch. Thanks!

tools/ci/gha-install-libmesh.sh
pshriwise: Cool! I wasn't aware of this. Much cleaner.

paulromano (approved):

PR #1741, opened 2021-01-05
:whale: adding badges for docker builds

Shimwell:
This tiny little PR just adds a couple of lines to the README.md

- A new badges for the docker build status.
- A new badges for the docker build status with DAGMC.
- A small change to the GitHub Actions CI badge so that the link points to the CI actions workflow instead of the general Actions URL. Now that we have quite a few GitHub Actions perhaps it is better to link to a specific Action

To be consistent with the existing badges these new badges point to the development branches


pshriwise (approved):
Looks good! Thanks for these updates @Shimwell.

PR #1745, opened 2021-01-14
Quick fix related to the shape of decay-rate.

mkreher13:
@nelsonag You may recall a PR on September 17th where I found that the shape of the decay-rate mgxs was incompatible between the C++ code and the python code. I fixed that by allowing the shape of decay-rate to only be [DG], as opposed to also allowing [DG][G]. So if there are 6 delayed groups, the shape of decay-rate is now (6,). 

That fixed quite a few bugs, but yesterday I noticed a file where the shape [DG][G] was still accepted. This PR eliminates that option in mgxs_library.py. 

I also found a small bug: in the DecayRate class of mdgxs.py I has previously kept a line that added an arbitrary extra dimension to the decay-rate, making it (6,1) instead of (6,). This came up as I was trying to plot some cross sections, and the shape checkvalue was failing. If I remove the extra dimension, the shape remains (6,) and the checkvalue passes and the plotting routine is functional. 

The only reason I had originally kept that line was to make the formatting in the mgxs_library_hdf5 test output prettier, but it occurred to me that it was just wrong to add an extra dimension like that, and obviously it messed up the shape in other parts of the code.


nelsonag (approved):
Looks good to me, thanks @mkreher13!

PR #1752, opened 2021-02-01
Refactor read_input_xml

helen-brooks:
This PR is a small refactor of the method `read_input_xml` from initialization.cpp. The purpose of the refactor is to make all code therein contained within public methods, such that should a developer of an external code wish to perform a partial re-initialisation e.g. in the context of a tightly-coupled multiphysics application, this is made more feasible (i.e. the developer may call some subset of those methods).

The method `finalize_geometry` from geometry_aux.cpp  no longer takes arguments, as the function `get_temperatures` which utilises those arguments has been moved into a new method `finalize_cross_sections` that has been added to cross_sections.cpp.

There are no new tests as there should be no behaviour changes.

### DIFF
--- a/include/openmc/cross_sections.h
+++ b/include/openmc/cross_sections.h
@@ -70,6 +70,9 @@ void read_ce_cross_sections(const std::vector<std::vector<double>>& nuc_temps,
 //! Read cross_sections.xml and populate data libraries
 void read_ce_cross_sections_xml();
 
+//! Load nuclide and thermal scattering data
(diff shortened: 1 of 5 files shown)


aprilnovak:
Looks good to me! I like this from a conceptual perspective, too - it's clearer that the temperature initialization really has to do with the data.

aprilnovak (approved):

helen-brooks:
Thanks April!

PR #1753, opened 2021-02-03
Adiabatic

mkreher13:
This PR contains the python code needed to run an adiabatic transient in a high-order/low-order scheme. The high-order method is time-dependent Monte Carlo with the adiabatic approximation (time derivatives=0) and the low-order method is a time-dependent CMFD. I have attached @samuelshaner's thesis for the theoretical background. All of his published results have been reproduced with this code. 

All the code is contained in openmc/openmc/kinetics. It is all Python and involves zero changes to the source code. The idea is that if you like the adiabatic code, we can discuss merging the improvement to the adiabatic approximation, which is the frequency transform method. That does involve quite a few changes to the source code. You can see the fully functional frequency transform method in my branch transient-learn. 

I have included 2 examples in this PR, one in continuous-energy and one in multi-group. They are simple with low neutron counts. I have NOT included any unit testing or regression testing because I want to get feedback on the code first before I put the work into creating all those tests. 

[ShanerThesis.pdf](https://github.com/openmc-dev/openmc/files/5915214/ShanerThesis.pdf)


examples/transient_adiabatic_ce/pin-td.py
```diff
@@ -0,0 +1,128 @@
...
+solver.seed                         = 1
+solver.min_outer_iters              = 1
+solver.use_pcmfd                    = False
+solver.use_agd                      = False
+solver.condense_dif_coef            = True
+solver.chi_delayed_by_delayed_group = True
+solver.chi_delayed_by_mesh          = False
+solver.use_pregenerated_sps         = False
+solver.run_on_cluster               = False
+solver.log_file_name                = 'log_file.h5'
```
gridley, line 126: It seems this is using the default value, maybe we should remove the lines here which are just using the default value from the solver interface.

examples/transient_adiabatic_mg/c5g7-td-2d-mg.py
```diff
@@ -0,0 +1,142 @@
+import numpy as np
+import os
+import openmc
+import openmc.mgxs
+import openmc.kinetics as kinetics
+
+from geometry_2d_mg import materials, surfaces, universes, cells, lattices, geometry, mgxs_lib_file
+from mgxs_lib import mgxs_data
+materials_file = openmc.Materials(geometry.get_all_materials().values())
+os.environ['OPENMC_MG_CROSS_SECTIONS'] = 'mgxs.h5'
```
gridley, line 10: Does the environment variable definitely have to be set? That seems like maybe something the user should be able to set. For instance, if someone set this to a different value, they wouldn't want it changed. Why not take the approach used in `pincell_multigroup/build_xml.py`?

examples/transient_adiabatic_mg/c5g7-td-2d-mg.py
```diff
@@ -0,0 +1,142 @@
...
+full_assembly_mesh.lower_left  = [-32.13, -32.13, -64.26]
+full_assembly_mesh.width = [ 21.42,  21.42,  128.52]
+
+# Instantiate an EnergyGroups object for the diffuion coefficients
+fine_groups = openmc.mgxs.EnergyGroups()
+fine_groups.group_edges = [0., 0.13, 0.63, 4.1, 55.6, 9.2e3, 1.36e6, 1.0e7]
+
+# Instantiate an EnergyGroups object for the transient solve
+energy_groups = openmc.mgxs.EnergyGroups()
+energy_groups.group_edges = [0., 0.13, 0.63, 4.1, 55.6, 9.2e3, 1.36e6, 1.0e7]
```
gridley, line 63: `energy_groups.group_edges = fine_groups.group_edges`

examples/transient_adiabatic_mg/c5g7-td-2d-mg.py
```diff
@@ -0,0 +1,142 @@
...
+# Instantiate an EnergyGroups object for one group data
+one_group = openmc.mgxs.EnergyGroups()
+one_group.group_edges = [fine_groups.group_edges[0], fine_groups.group_edges[-1]]
+
+# Instantiate a clock object
+t_outer = np.arange(0., 1.0, 5.e-1)
+clock = openmc.kinetics.Clock(start=0., end=0.5, dt_inner=1.e-2, t_outer=t_outer)
+
+# Prescribe the transient as a dictionary of densities and temperatures 
+Transient = {}
```
gridley, line 74: Same as the other comment

openmc/kinetics/__init__.py
```diff
@@ -0,0 +1,4 @@
+from openmc.kinetics.clock import *
```
gridley, line 1: Avoid wildcard imports, according to PEP8 (these style changes are always annoying, sorry!)

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn't known in advance).

openmc/kinetics/aux_functions.py
```diff
@@ -0,0 +1,218 @@
+import scipy.sparse as sps
+from scipy.sparse.linalg import spsolve
+
+import numpy as np
+from numpy.lib.stride_tricks import as_strided as ast
+
+
+def gcd(a, b):
```
gridley, line 8: `from math import gcd`

openmc/kinetics/aux_functions.py
```diff
@@ -0,0 +1,218 @@
...
+    # Compute the initial source
+    old_source = M * flux
+    norm = old_source.mean()
+    old_source  = old_source / norm
+    flux  = flux / norm
+    k_eff = 1.0
+
+    print('initial k {}'.format((M*flux).sum() / (A*flux).sum()))
+
+    for i in range(10000):
```
gridley, line 186: Let's name that constant somewhere, like `max_eigenvalue_iterations` or something. In fact, it should probably be a kwarg to this function with that default value.

openmc/kinetics/aux_functions.py
```diff
@@ -0,0 +1,218 @@
...
+        # Compute new source
+        new_source = M * flux
+
+        # Compute and set k-eff
+        k_eff = new_source.mean()
+
+        # Scale the new source by 1 / k-eff
+        new_source  = new_source / k_eff
+
+        # Compute the residual
```
gridley, line 200: This is not actually the residual vector (Ax-Bx/k), so maybe the comment should more accurately reflect that. Something like `fractional_solution_change`, maybe?

openmc/kinetics/aux_functions.py
```diff
@@ -0,0 +1,218 @@
...
+        residual_array = (new_source - old_source) / new_source
+        residual_array[residual_array == -np.inf] = 0.
```
gridley, line 202: Use the `nan_inf_to_zero` function you had written above instead. Also, it may be good to just set this up so that infinities and nans never show up in the first place. You could instead do:

```
nonzero = new_source != 0
residual_array = (new_source[nonzero]-old_source[nonzero])/new_source[nonzero]
```

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+        self._amplitude_mesh = openmc.RegularMesh()
+        self._amplitude_mesh.type = mesh.type
+        self._amplitude_mesh.dimension = mesh.dimension
+        self._amplitude_mesh.lower_left = mesh.lower_left
+
+        unity_mesh = openmc.RegularMesh()
+        unity_mesh.type = mesh.type
+        unity_mesh.dimension = [1,1,1]
+        unity_mesh.lower_left  = mesh.lower_left
+        if mesh.width is not None:
```
gridley, line 306: I believe you can also do simply `if mesh.width` here if you want to make it more concise

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+            self._amplitude_mesh.width = mesh.width
+            unity_mesh.width = [i*j for i,j in zip(mesh.dimension, mesh.width)]
+        else:
+            self._amplitude_mesh.width = [i-j for i,j in zip(mesh.upper_right, mesh.lower_left)]
+            unity_mesh.width = [i-j for i,j in zip(mesh.upper_right, mesh.lower_left)]
+        self._unity_mesh = unity_mesh
+
+        # Set the power mesh to the shape mesh if it has not be set
+        if self.shape_mesh is None:
```
gridley, line 315: same on these

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+        new_source = '{}/source.h5'.format(self.directory)
+
+        # Run OpenMC
+        if not self.use_pregenerated_sps:
+            if self.run_on_cluster:
+
+                # Copy job file to run directory
+                cmd_str = 'cp ' + self.job_file  + ' ' + self.directory
+                subprocess.Popen(cmd_str, shell=True)
+                time.sleep(5)
```
gridley, line 559: We should definitely find a more robust way to ensure this file operation works correctly. I think an alternative may be to open that cp operation process, then have some utility function that checks the file is written. This function could work by running in a loop trying to open the file and exiting once an open succeeds. Of course you would then add a small sleep call in that loop, otherwise it would be nasty to try opening something millions of times in a second.

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+
+                 # Launch job
+                cmd_str = 'qsub -P moose ' + self.job_file
```
gridley, line 562: I get that moose is popular, but we're not all under their funding :smile: 

Maybe there should be some variable for additional arguments to `qsub` that this line of code could use.

Also, if someone is using slurm or another job scheduler that has different commands (for instance, slurm would be `sbatch` here instead), this won't work. Maybe there's a library out there which can abstract away these interfaces to various job schedulers.

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+                 # Get the job number
+                job_name = proc.stdout.readlines()[0]
+                job_number = job_name.split('.')[0]
+
+                 # Pause and wait for run to finish
+                cmd_str = 'qstat | grep ' + str(job_number)
+                is_file_running = 'initially'
+                elapsed_time = 0
+                while (is_file_running is not ''):
+                    time.sleep(10)
```
gridley, line 574: Similarly, this sleep thing could be better

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+                 # Launch job
+                cmd_str = 'qsub -P moose ' + self.job_file
+                proc = subprocess.Popen(cmd_str, cwd=self.directory, stdout=subprocess.PIPE, shell=True)
+
+                 # Get the job number
+                job_name = proc.stdout.readlines()[0]
+                job_number = job_name.split('.')[0]
+
+                 # Pause and wait for run to finish
+                cmd_str = 'qstat | grep ' + str(job_number)
```
gridley, line 570: likewise, `slurm` would use `squeue` here. And I'm sure there are lots of other schedulers with different commands.

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+        settings_file = copy.deepcopy(self.settings_file)
+
+        # Create a new random seed for the xml file
+        if self.constant_seed:
+            settings_file.seed = self.seed
+        else:
+            settings_file.seed = np.random.randint(1, 1e6, 1)[0]
+
+        if self.mgxs_lib_file:
+            self.materials_file.cross_sections = './mgxs.h5'
```
gridley, line 851: Should this be here? I get the feeling it should just use whatever MGXS cross section file the user has set

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,886 @@
...
+            self.mgxs_lib_file.export_to_hdf5(self.directory + '/mgxs.h5')
```
gridley, line 852: same, why always mgxs.h5?

gridley (changes requested):
Wow, this is awesome. Can't wait to see it in action. I get the feeling the meshed precursor approach is going to be a quite accurate approximation. I haven't tested the code, but do have a few initial notes on the code that I think should be addressed before merging.

OK, so, initial notes:

- `transient_adiabatic_ce` should probably just be `pincell_transient` for a few reasons. Firstly, for all the other OpenMC examples, the continuous energy cases are not marked as "ce", because that's kind of the assumed default operation mode. Secondly, it fits with the other example case names better like "pincell_depletion" and "pincell_multigroup". Thirdly, to someone just getting used to an MC code where examples are important, "ce" and "mg" may not be acronyms they'd be familiar with. And lastly, I think if other transient methodologies are added in the future, it would make sense to put them in this same directory and just have the python script change things automatically to run the problem but using a different technique like alpha mode or frequency transform.

- Another thing is that @ilhamv is working on adding alpha mode to OpenMC, or at least has but the code seems tough to merge for the same reasons as your frequency transform code, Miriam. In particular, it seems both need to add some additional cached cross sections. Of course that's kind of to be expected, considering the close relationship between alpha mode and frequency transform. So I think there will be a way to get both of these into the main codebase in a clean way, but will require a little bit of preparation ahead of time. Particularly, we need to come up with a data structure for cross section caches that only allocates the memory it needs (e.g., not allocate depletion reaction cache data during transient runs, not allocate alpha mode data during frequency transform runs, etc.) As the paid code reviewer says "the current approach will not scale". (Especially on GPU where we need to never waste memory and always maintain contiguity of relevant data to allow coalesced memory accesses)

- Lastly, we need to have more of the C5G7-TD suite fully ran before merging this, maybe minus the 3D tests. In particular, I'd really like to also see that this capability can indeed handle a prompt supercritical transient before merging as well, which is not included in C5G7-TD. That could be done on an IHM problem, for instance.

Amazing work, again!

openmc/kinetics/solver.py
gridley: Yep, there is a library for that:

https://pypi.org/project/jobrunner/

I know adding additional required dependencies is frowned upon, but maybe it would make sense to make this an optional dependence if a user wants to run this type of distributed transient calculation. Alternatively, I'm sure the calculation flow could be slightly revised so that all these changes are done using mpi4py rather than actually scheduling separate jobs.

mkreher13: I'm actually thinking about removing the whole "run_on_cluster" clause. I thought it could be useful to keep, but it does seem too specific. Each cluster is going to have its own job submission process.

examples/transient_adiabatic_ce/pin-td.py
```diff
@@ -0,0 +1,128 @@
...
+
+# Instantiate a clock object
+t_outer = np.arange(0., 1.0, 5.e-1)
+clock = openmc.kinetics.Clock(start=0., end=0.5, dt_inner=1.e-2, t_outer=t_outer)
+
+# Prescribe the transient as a dictionary of densities and temperatures 
+Transient = {}
+
+# Create entries for each material 
+for material in materials_file:
```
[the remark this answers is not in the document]
mkreher13: Thanks for this! I felt like the dictionary was way too clunky, and I needed some new ideas. I'll play around with what you suggested!

examples/transient_adiabatic_ce/pin-td.py
mkreher13: It's just an example, it's not slowing anything down. I think it's good to have the example file show all the options the user has.

openmc/kinetics/__init__.py
mkreher13: Sure, it wouldn't be hard for me to specify each class or function since there aren't too many. But wildcard imports are done all over OpenMC, so I'm not sure we are following that standard. See openmc/openmc/__init__.py for example.

examples/transient_adiabatic_ce/pin-td.py
gridley: Very reasonable point!

examples/transient_adiabatic_mg/c5g7-td-2d-mg.py
mkreher13: I think this could go either way. It's an example input, so it doesn't really make a difference if you write out the full array again or not. I tend to like laying out the whole thing for the user so they don't get confused and think that energy_groups has to equal fine_groups.

openmc/kinetics/aux_functions.py
mkreher13: That worked great, so included it! Thanks.

gridley (commented):
Changes are looking good! I'm going to look into those potential changes to the microscopic cross section caching tonight!

mkreher13:
@gridley Awesome, thanks! But no worries if you can't get to it. Since this PR _doesn't_ require changes to the source code, I want to decouple it from the frequency transform method. While there are quite a few barriers to getting the FTM merged, I think this adiabatic method is a good middle ground to getting some form of transient capability into OpenMC while remaining minimally-invasive.

gridley:
Yeah, agreed, this is a great starting point to get a framework for transients rolling.

paulromano:
Thanks for submitting this PR @mkreher13! I'll try to provide some feedback by the end of the week.

paulromano:
@mkreher13 Sorry this is taking me a bit longer to review than I had hoped. Just wanted to let you know I haven't forgotten it and it's still on my radar.

examples/transient_adiabatic_ce/pin-td.py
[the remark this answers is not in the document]
mkreher13: I've been looking into this more. Sam originally had made changes to cell.py, material.py, and geometry.py to add a time variable. Then, cell.translation and material.density could be input as arrays along with a time array. Then, when defining your material density, instead of specifying a float, you could give an array. This sounds sort of similar to what you suggested. 
However, all this is for a prescribed transient, which is not really a long-term solution for transients in general. We want to work toward coupled transients eventually. So I'm hesitant to make any time-dependent change in the OpenMC material and cell code. In the end, I think I prefer the input-file-only method of specifying the transient.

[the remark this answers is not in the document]
gridley: Makes sense to me!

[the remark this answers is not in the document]
mkreher13: I still want to make it less clunky. I'm thinking about including this dictionary, or a different structure into the kinetics code so that the user can just update parts of it instead of having to generate & fill the whole "transient" for each problem. A work in progress.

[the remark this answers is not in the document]
gridley: Nice :D going to be exciting to see it all come together!

mkreher13:
I've been brainstorming improvements to the definition of the transient. Currently, it's a clunky dictionary that the user has to create that contains the changes for material density and temperature at every time point. I also want to add control rod transients which would just require a cell translation. I see three options:

**Option 1** (currently in place): manually update the xml files in the kinetics code based on arrays of materials/cell changes. 
_Pros:_ code is already written this way
_Cons:_ very clunky input; changing the cell translation of an existing geometry file isn't as easily done as the material changes

**Option 2**: use the CAPI to make updates instead of the xml files. 
_Pros:_ potentially smoother in the kinetics code
_Cons:_ the CAPI does not seem to work between simulations, only during them which is not compatible this time-stepping scheme; doesn't solve the clunky input problem since the user would still have to pass arrays with all the changes; cell.set_translation isn't currently an option in the CAPI

**Option 3**: the user pre-generates xml files for every state of the transient, numbers them, and the code just updates which xml files to read at which time step
_Pros:_ clears up the input file (no clunky arrays to pass); makes the code much more amenable to external coupling
_Cons:_ a bit much if the only change is a single density change

I'm leaning toward option 3. The user can just copy/paste a lot of code and generate folders of xml files that correspond to each reactor state. It seems like a lot if it's a simple transient, but it works for both complex and simple transients.

gridley:
On option 2: I think it would be possible to simply add some different control flow to the C++ side to wait on some flag to be set by the python side in case calculations like transients are being done. This would be useful for more than just transients, because this would come with the benefit of using the fission bank already in memory to kick off the next simulation. This would be useful in stuff like criticality searches where 1) you want to kick off simulations on the last fission bank and 2) reloading cross sections each time is redundant work.

I am guessing the current transient implementation reloads cross sections at each step.

To implement this, we'd just add some static flag variable marked as volatile in simulation.h to be controlled by python. When the calculation finishes, openmc_finalize could be called as usual, but the program would proceed to a waiting phase where it periodically checks if the coupling flag has been set to "go" by the C API. Once the desired changes are made through the C API, the python code could set that flag, and the next simulation would proceed from there.

Of course this would require some different naming convention for the statepoint files, maybe appending an integer index after them to indicate which run it is. I think this approach would look pretty similar to the multiphysics coupling interface they have in Serpent, but would be a lot more powerful because the C API can do _a lot_.

mkreher13:
@paulromano is there anything I can do to make this PR more digestible?

paulromano:
@mkreher13 So sorry I dropped the ball on this one! Mostly I need to familiarize myself with the methods so I can understand what your code is doing :) I should have some time to dedicate to this next week. If you haven't heard anything from me by mid-next week, please ping me again.

openmc/kinetics/__init__.py
gridley: Fair enough!

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+
+        self.states[time_point].load_mgxs()
+
+    def create_state(self, time_point):
+
+        if time_point in ['START', 'END', 'FORWARD_OUTER', 'PREVIOUS_OUTER']:
+            state = openmc.kinetics.OuterState(self.states)
+            state.chi_delayed_by_delayed_group = self.chi_delayed_by_delayed_group
+            state.chi_delayed_by_mesh = self.chi_delayed_by_mesh
+            if time_point in ['FORWARD_OUTER', 'PREVIOUS_OUTER']:
```
paulromano, line 557:
```suggestion
            if time_point in ('FORWARD_OUTER', 'PREVIOUS_OUTER'):
```

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+        state.fine_groups = self.fine_groups
+        state.one_group = self.one_group
+        state.num_delayed_groups = self.num_delayed_groups
+        state.use_agd = self.use_agd
+        state.use_pcmfd = self.use_pcmfd
+        state.time_point = time_point
+        state.clock = self.clock
+        state.k_crit = self.k_crit
+        state.core_volume = self.core_volume
+        state.log_file = self.log_file
```
paulromano, line 578: Why not just have `state` keep a reference to the solver and get all these attributes from the solver when needed?

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+    @property
+    def energy_groups(self):
+        return self._energy_groups
+
+    @property
+    def fine_groups(self):
+        return self._fine_groups
+
+    @property
+    def tally_groups(self):
```
paulromano, line 196: This property is not described in the class docstring

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+        statepoint_file = openmc.StatePoint(sp_new_name, False)
+        statepoint_file.link_with_summary(summary_file)
+
+        # Load mgxs library
+        for mgxs in self.states[time_point].mgxs_lib.values():
+            mgxs.load_from_statepoint(statepoint_file)
+
+        self.states[time_point].load_mgxs()
+
+    def create_state(self, time_point):
```
paulromano, line 551: If this method is not meant to be called by a user (I assume it is not), the convention is to make it private by prepending an underscore, i.e. `def _create_state`

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+        # Export the tallies file to xml
+        tallies_file.export_to_xml(self.directory + '/tallies.xml')
+
+    def solve(self):
+
+        # Create run directory
+        if not os.path.exists(self.directory):
+            os.makedirs(self.directory)
+
+        # Create states and run initial OpenMC on initial state
```
paulromano, line 774: To me it doesn't look like `create_state` runs OpenMC. does this comment need to be updated?

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+        state.log_file = self.log_file
+
+        if self.tally_groups is not None:
+            state.tally_groups = self.tally_groups
+        else:
+            state.tally_groups = self.energy_groups
+
+        self.states[time_point] = state
+
+    def compute_initial_flux(self):
```
paulromano, line 587: Same comment here about methods being "private". Can you go through each method and if it's not intended to be used by a user, add a leading underscore?

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+        structure.
+    initial_power : float
+        The initial core power (in MWth).
+    k_crit : float
+        The initial eigenvalue.
+    mpi_args : list of str, optional
+        MPI execute command and any additional MPI arguments to pass,
+        e.g. ['mpiexec', '-n', '8'].
+    threads : int
+        The number of OpenMP threads to use.
```
paulromano, line 71: I would replace these two options with a general `run_kwargs`, which would be a dictionary of keyword arguments that are passed through to `openmc.run`. This helps it be future-proof in that if any keyword arguments of `openmc.run` change, we don't need to add more options here. At the point `openmc.run` is called, you would just have `openmc.run(..., **run_kwargs)`

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+    one_group : openmc.mgxs.groups.EnergyGroups
+        EnergyGroups which specifies the a one-energy-group structure.
+    energy_groups : openmc.mgxs.groups.EnergyGroups
+        EnergyGroups which specifies the energy groups structure.
+    fine_groups : openmc.mgxs.groups.EnergyGroups
+        EnergyGroups used to tally the transport cross section that will be
+        condensed to get the diffusion coefficients in the coarse group
+        structure.
+    initial_power : float
+        The initial core power (in MWth).
```
paulromano, line 64: In the depletion module, we specify power in [W], so I would stick with the same convention here for consistency

examples/c5g7_transient/c5g7-td-2d-mg.py
```diff
@@ -0,0 +1,139 @@
...
+
+from geometry_2d_mg import materials, surfaces, universes, cells, lattices, geometry, mgxs_lib_file
+from mgxs_lib import mgxs_data
+materials_file = openmc.Materials(geometry.get_all_materials().values())
+os.environ['OPENMC_MG_CROSS_SECTIONS'] = 'mgxs.h5'
+
+###############################################################################
+# Define problem settings
+
+settings_file = openmc.Settings()
```
paulromano, line 15: Don't we need `settings_file.energy_mode = "multi-group"`?

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+    @property
+    def log_file_name(self):
+        return self._log_file_name
+
+    @property
+    def job_file(self):
+        return self._job_file
+
+    @property
+    def multi_group(self):
```
paulromano, line 264: Doesn't look like this attribute is used anywhere

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+        # Create a new random seed for the xml file
+        if self.constant_seed:
+            settings_file.seed = self.seed
+        else:
+            settings_file.seed = np.random.randint(1, 1e6, 1)[0]
+
+        if self.mgxs_lib_file:
+            self.materials_file.cross_sections = './mgxs.h5'
+            self.mgxs_lib_file.export_to_hdf5(self.directory + '/mgxs.h5')
+            settings_file.energy_mode = 'multi-group'
```
paulromano, line 798: It should be the responsibility of the user to make sure their `Settings` object has this set properly, rather than having this class selectively set it

openmc/kinetics/state.py
```diff
@@ -0,0 +1,1503 @@
...
+    Attributes
+    ----------
+    amplitude_mesh : openmc.mesh.Mesh
+        Mesh by which shape is computed on.
+    shape_mesh : openmc.mesh.Mesh
+        Mesh by which power is reconstructed on.
+    unity_mesh : openmc.mesh.Mesh
+        Mesh with one cell convering the entire geometry..
+    tally_mesh : openmc.mesh.Mesh
+        Mesh to tally currents
```
paulromano, line 34: Assuming these should all specify `openmc.RegularMesh`

paulromano (changes requested):
@mkreher13 Thanks for starting this! Overall, I don't see any major issues toward getting this merged. Generally, if code is living on the Python side and is relatively independent (as this is), it's not very intrusive. Obviously as this is expanded to the full frequency transform method, that will be more intrusive since it will involve changes on the C++ side. I think perhaps it would be good for us to set up a meeting to discuss the proper steps toward that.

I haven't gone through your code in its entirety yet but wanted to at least get you some initial feedback. In general:
- Classes and methods need to have docstrings that describe their purpose, arguments, and return values
- In the code itself, what would be really helpful for review purposes (and for posterity) is to add comments that relate the code to specific equations from Shaner's thesis

Once those are added, I can go back and review in full. Thanks once again for moving us toward having a real transient capability, and so sorry that it's taken me this long to get you some real feedback!

openmc/kinetics/clock.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+
+TIME_POINTS = ['START',
+               'PREVIOUS_OUTER',
+               'FORWARD_OUTER',
+               'PREVIOUS_INNER',
+               'FORWARD_INNER',
+               'END']
```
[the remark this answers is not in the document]
mkreher13: I don’t think enums works here. There is never a number assigned directly to the names, and they refer to different times on the clock depending which time step you are in. These names are purely strings used in various dictionaries across the code to keep track of previous & forward states.

mkreher13:
Thank you Paul! I added docstrings, and I addressed your preliminary comments. So hopefully it is all ready for a full review now.

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,831 @@
...
+    def outer_tolerance(self, tolerance):
+        self._outer_tolerance = tolerance
+
+    @method.setter
+    def method(self, method):
+        self._method = method
+
+    @min_outer_iters.setter
+    def min_outer_iters(self, iters):
+        self._min_outer_iters = iters
```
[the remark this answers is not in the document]
paulromano: It looks like several of the setters (`initial_power`, `k_crit`, `run_kwargs`, `states`) still don't do anything more than setting an underlying attribute.

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,869 @@
...
+        condensed to get the diffusion coefficients in the coarse group
+        structure.
+    tally_groups : openmc.mgxs.groups.EnergyGroups
+        EnergyGroups used in all tallies except the diffusion coefficients
+        (see fine_groups).
+    initial_power : float
+        The initial core power (in W).
+    k_crit : float
+        The initial eigenvalue.
+    run_kwargs : dictionary of keyword arguments passed to openmc.run.
```
paulromano, line 68:
```suggestion
    run_kwargs : dict
        Keyword arguments passed to :func:`openmc.run`.
```

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,869 @@
...
+        EnergyGroups which specifies the energy groups structure.
+    fine_groups : openmc.mgxs.groups.EnergyGroups
+        EnergyGroups used to tally the transport cross section that will be
+        condensed to get the diffusion coefficients in the coarse group
+        structure.
+    tally_groups : openmc.mgxs.groups.EnergyGroups
+        EnergyGroups used in all tallies except the diffusion coefficients
+        (see fine_groups).
+    initial_power : float
+        The initial core power (in W).
```
paulromano, line 65: For consistency with the rest of our documentation:
```suggestion
        The initial core power in [W].
```

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,869 @@
...
+    use_pcmfd : bool
+        Whether to use p-CMFD.
+    num_delayed_groups : int
+        The number of delayed neutron precursor groups.
+    states : OrderedDict of openmc.kinetics.State
+        States of the problem.
+    use_pregenerated_sps : bool
+        Whether to use pregenerated statepoint files.
+    core_volume : float
+        The core volume used to normalize the initial power.
```
paulromano, line 82: Need to mention units

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,869 @@
...
+    log_file_name : str
+        Log file name (excluding directory prefix).
+    min_outer_iters : int
+        Minimum number of outer iterations to take.
+
```
paulromano, line 87: There are several attributes not currently described in the docstring: `use_agd`, `inner_tolerance`, `outer_tolerance`, `method`

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,869 @@
...
+        cv.check_type('directory', directory, str)
+        self._directory = directory
+
+    @amplitude_mesh.setter
+    def amplitude_mesh(self, mesh):
+
+        cv.check_type('amplitude mesh', mesh, openmc.RegularMesh)
+        self._amplitude_mesh = mesh
+        unity_mesh = openmc.RegularMesh()
+        unity_mesh.dimension = [1,1,1]
```
paulromano, line 260:
```suggestion
        unity_mesh.dimension = (1, 1, 1)
```

openmc/kinetics/clock.py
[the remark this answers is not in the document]
paulromano: Enums are usually used when you _don't_ care about/need an underlying value. They work great if you need a set of unique, symbolic names, which seems to match the intent of these strings. The Python documentation gives some good examples where the [underlying value is omitted altogether](https://docs.python.org/3/library/enum.html#omitting-values). Probably the best thing to do in this case is to [use a descriptive string](https://docs.python.org/3/library/enum.html#using-a-descriptive-string).

openmc/kinetics/state.py
```diff
@@ -0,0 +1,1572 @@
...
+import h5py
+import numpy as np
+import scipy.sparse as sps
+
+import openmc
+import openmc.kinetics
+import openmc.mgxs
+
+
+class State(ABC):
```
paulromano, line 15: My earlier comment stands -- although this class is intended to be abstract (via subclassing `ABC`), no methods are marked as abstract, which means any class could subclass `State` without having to implement specific methods. What is it that the subclasses (`OuterState`, `InnerState`) need to implement?

openmc/kinetics/solver.py
```diff
@@ -0,0 +1,869 @@
...
+        state.fine_groups = self.fine_groups
+        state.one_group = self.one_group
+        state.num_delayed_groups = self.num_delayed_groups
+        state.use_agd = self.use_agd
+        state.use_pcmfd = self.use_pcmfd
+        state.time_point = time_point
+        state.clock = self.clock
+        state.k_crit = self.k_crit
+        state.core_volume = self.core_volume
+        state.log_file = self.log_file
```
paulromano, line 569: Earlier comment remains -- why not have `State` keep a reference to the `Solver` object and retrieve attributes from it as needed?

paulromano (changes requested):
Thanks for the changes @mkreher13 and sorry once again it has taken me some time to get feedback to you. The major thing missing now are tests -- because all of this new code is not tested anywhere, the PR fails the status checks by virtue of our coverage dropping significantly. Adding some tests will help move this closer to being merged.

openmc/kinetics/state.py
```diff
@@ -0,0 +1,1503 @@
...
+import openmc.kinetics
+import openmc.mgxs
+
+
+if sys.version_info[0] >= 3:
+    basestring = str
+
+
+@add_metaclass(ABCMeta)
+class State(object):
```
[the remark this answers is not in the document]
mkreher13: Some aspects of this code are legacy from what Sam originally wrote. I'm not experienced with abstract classes, except what I've seen in OpenMC mgxs code. Is it sufficient to make it not abstract anymore? That's what I've done in this update.

openmc/kinetics/solver.py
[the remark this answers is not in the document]
mkreher13: Do you recommend removing the setters, or the properties, or both? I have not worked with setters before, so I am mostly trying to follow suit from what I see in OpenMC's other Python files.

openmc/kinetics/clock.py
```diff
@@ -0,0 +1,73 @@
...
+    def dt_inner(self):
+        return self._dt_inner
+
+    @dt_inner.setter
+    def dt_inner(self, dt_inner):
+        self._dt_inner = float(dt_inner)
+
+    @t_outer.setter
+    def t_outer(self, t_outer):
+        self._t_outer = np.float64(t_outer)
```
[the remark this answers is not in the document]
mkreher13: I moves the times to the t_outer setter. As for float() I have to use np.float() since t_outer is a bumpy array. And numpy.float() actually returns a non-numpy float. So I have to specify np.float64() from what I can tell.

openmc/kinetics/solver.py
mkreher13: This one confuses me. These definitions of `OuterState()` and `InnerState()` are identical to how we define `Solver()` in the input file and assign all of its variables. 

We are creating many `states`, each at a different `time_point`. I see that almost all of these variables don't change between time_points (mesh definitions and group boundaries, etc don't change throughout the calculation), so are you suggesting we avoid re-defining all the same variables each time a new `state` is created?

paulromano: I'm suggesting that instead of adding all these attributes to the `State` object, just keep a single attribute that refers to the `Solver` instance, i.e.,
```Python
state.solver = self
```
Then if the state object needs to access, e.g., the amplitude mesh, it can get it via `state.solver.amplitude_mesh`. Does that make sense?

mkreher13: Thanks for helping me understand! I did that, and I think it works really nicely.

mkreher13:
I'm working on the testing now! I didn't want to do it right off the bat, because I wanted to be sure that this could be merged before investing in that effort. Now that you think it's getting close, I'm on it.

mkreher13:
@paulromano Let me know if these tests need any modifications. Looks like coveralls is increased overall.

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,303 @@
...
+        MatChange = {material.name: {},}
+        transient.update(MatChange)
+        for t in t_outer:
+            time_dict = {
+            t: {
+                'density' : {},
+                'temperature' : {},
+                }
+            }
+            transient[material.name].update(time_dict)
```
paulromano, line 268: I think this is equivalent?
```suggestion
            transient[material.name][t] = {
                'density': {},
                'temperature': {}
            }
```

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,303 @@
...
+    for material in model.materials:
+        if material.name == 'Water':
+            transient[material.name][0]['density'] = material.density
+            transient[material.name][0.5]['density'] = material.density*0.9
+            for t in t_outer:
+                transient[material.name][t]['temperature'] = material.temperature
+        else: 
+            for t in t_outer:
+                transient[material.name][t]['density'] = material.density
+                transient[material.name][t]['temperature'] = material.temperature
```
paulromano, line 279: Given this logic, I'm confused as to why you're setting `transient[material.name][t]['density']` (and temperature) to an empty dictionary above only to overwrite it here with a float.

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,303 @@
...
+    solver.tally_groups    = energy_groups
+    solver.geometry        = model.geometry
+    solver.settings        = model.settings
+    solver.materials       = model.materials
+    solver.transient       = transient
+    solver.outer_tolerance = np.inf
+    solver.mgxs_lib        = mg_cross_sections_file
+    solver.multi_group     = True
+    solver.clock           = clock
+    solver.run_kwargs      = {'threads':None, 'mpi_args':None}
```
paulromano, line 299:
```suggestion
    solver.run_kwargs      = {'threads': None, 'mpi_args': None}
```

tests/regression_tests/transient_ce/test.py
```diff
@@ -0,0 +1,241 @@
...
+    for material in model.materials:
+        if material.name == 'Moderator':
+            transient[material.name][0]['density'] = material.density
+            transient[material.name][0.5]['density'] = material.density*0.9
+            for t in t_outer:
+                transient[material.name][t]['temperature'] = material.temperature
+        else: 
+            for t in t_outer:
+                transient[material.name][t]['density'] = material.density
+                transient[material.name][t]['temperature'] = material.temperature
```
paulromano, line 209: Similar changes needed here as what I suggested for the other test

tests/regression_tests/transient_ce/test.py
```diff
@@ -0,0 +1,241 @@
...
+    solver.settings        = model.settings
+    solver.materials       = model.materials
+    solver.transient       = transient
+    solver.outer_tolerance = np.inf
+    solver.initial_power   = 1.0
+    solver.core_volume     = 1.0
+    solver.method          = 'ADIABATIC'
+    solver.multi_group     = False
+    solver.clock           = clock
+    solver.run_kwargs      = {'threads':None, 'mpi_args':None}
```
paulromano, line 230:
```suggestion
    solver.run_kwargs      = {'threads': None, 'mpi_args': None}
```

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,303 @@
...
+        super()._cleanup()
+        output = ['materials.xml', 'geometry.xml', 'settings.xml',
+                  'tallies.xml', 'plots.xml', 'inputs_test.dat']
+        output += glob.glob('*.h5')
+        for f in output:
+            if os.path.exists(f):
+                os.remove(f)
+
+    def _run_transient(self):
+        self.solver.solve()
```
paulromano, line 90: Normally in the `_run_openmc` method we account for the test configurations where MPI is being used. It doesn't look like that is being taken into account here. It should be as simple as setting:
```Python
self.solver.run_kwargs['mpi_args'] = ...
```
for the MPI configuration.

paulromano (changes requested):
Thanks for putting together the tests @mkreher13. We're getting closer!

mkreher13:
I addressed everything, and I'm pretty happy with where the PR is now. But I can't explain the single test failure since it seems unrelated: one system fails on the tally_aggregation test by a one number in the 7th decimal place.

tests/regression_tests/transient_ce/test.py
```diff
@@ -0,0 +1,187 @@
...
+        self.solver = solver
+        self._log_name = log_name
+
+        if model is None:
+            self._model = pwr_core()
+        else:
+            self._model = model
+        self._model.plots = []
+
+        self.inputs_true = "inputs_true.dat" if not inputs_true else inputs_true
```
paulromano, line 30:
```suggestion
        super().__init__(statepoint_name, model, inputs_true)
        self.solver = solver
        self._log_name = log_name
```

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,250 @@
...
+        self.solver = solver
+        self._log_name = log_name
+
+        if model is None:
+            self._model = pwr_core()
+        else:
+            self._model = model
+        self._model.plots = []
+
+        self.inputs_true = "inputs_true.dat" if not inputs_true else inputs_true
```
paulromano, line 32:
```suggestion
        super().__init__(statepoint_name, model, inputs_true)
        self.solver = solver
        self._log_name = log_name
```

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,250 @@
...
+    lower_left = [-0.63, -0.63, -10]
+    upper_right = [0.63, 0.63, 10]
+    uniform_dist = openmc.stats.Box(lower_left, upper_right, only_fissionable=True)
+    model.settings.source = openmc.source.Source(space=uniform_dist)
+
+    sourcepoint = dict()
+    sourcepoint['batches'] = [model.settings.batches]
+    sourcepoint['separate'] = True
+    sourcepoint['write'] = True
+    model.settings.sourcepoint = sourcepoint
```
paulromano, line 195:
```suggestion
    model.settings.sourcepoint = {
        'batches': [model.settings.batches],
        'separate': True,
        'write': True
    }
```

paulromano (changes requested):
A few more small changes

mkreher13:
Question: what do you think of the `transient` dictionary for the user to prescribe the changes during the transient? In Sam Shaner's original code, he added a time variable to geometry, cell, and material. Then he made translation (for cells) and density (for materials) an array of floats instead of a float. That way, a user could define the density, the cell translations, and then the time arrays as the means to prescribe the changes during the transient. I opted for the `transient` dictionary instead to make the transient code more self-contained and avoid too many changes to the original OpenMC code. Do you have a preference? Because I could probably change it back to what Sam had.

examples/c5g7_transient/density_change/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
...
+        universes[univ_name] = openmc.Universe(name=univ_name)
+        for ring in rings:
+            name = '{} {} Bank {}'.format(mat, ring, bank)
+            cells[name] = openmc.Cell(name=name)
+            universes[univ_name].add_cell(cells[name])
+
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].region       = -surfaces['Fuel Outer Clad OR']
+        cells['{} {} Bank {}'.format(mat, 'Moderator', bank)].region  = +surfaces['Fuel Outer Clad OR']
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].fill       = materials[mat]
+        cells['{} {} Bank {}'.format(mat, 'Moderator', bank)].fill  = materials['Moderator Bank {}'.format(bank)]
```
paulromano, line 27:
```suggestion
        cells[f'{mat} Base Bank {bank}'].region       = -surfaces['Fuel Outer Clad OR']
        cells[f'{mat} Moderator Bank {bank}'].region  = +surfaces['Fuel Outer Clad OR']
        cells[f'{mat} Base Bank {bank}'].fill       = materials[mat]
        cells[f'{mat} Moderator Bank {bank}'].fill  = materials[f'Moderator Bank {bank}']
```

examples/c5g7_transient/density_change/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
...
```
paulromano, line 42:
```suggestion
        cells[f'{mat} Base Bank {bank}'].region       = -surfaces['Fuel Outer Clad OR']
        cells[f'{mat} Moderator Bank {bank}'].region  = +surfaces['Fuel Outer Clad OR']
        cells[f'{mat} Base Bank {bank}'].fill       = materials[mat]
        cells[f'{mat} Moderator Bank {bank}'].fill  = materials['Moderator Bank {}'.format(bank)]
```

examples/c5g7_transient/density_change/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
...
+for mat in mats:
+    for bank in range(1,5):
+        for ring in rings:
+            name = '{} {} Bank {}'.format(mat, ring, bank)
+            cells[name] = openmc.Cell(name=name)
+
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].region      = -surfaces['Fuel Outer Clad OR'] & -surfaces['Axial Midplane']
+        cells['{} {} Bank {}'.format(mat, 'Moderator', bank)].region = +surfaces['Fuel Outer Clad OR'] & -surfaces['Axial Midplane']
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].fill      = materials[mat]
+        cells['{} {} Bank {}'.format(mat, 'Moderator', bank)].fill = materials['Moderator Bank {}'.format(bank)]
```
paulromano, line 54:
```suggestion
        cells[f'{mat} Base Bank {bank}'].region      = -surfaces['Fuel Outer Clad OR'] & -surfaces['Axial Midplane']
        cells[f'{mat} Moderator Bank {bank}'].region = +surfaces['Fuel Outer Clad OR'] & -surfaces['Axial Midplane']
        cells[f'{mat} Base Bank {bank}'].fill      = materials[mat]
        cells[f'{mat} Moderator Bank {bank}'].fill = materials['Moderator Bank {}'.format(bank)]
```

examples/c5g7_transient/density_change/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
...
+        universes[univ_name] = openmc.Universe(name=univ_name)
+        for ring in rings:
+            name = '{} {} Bank {}'.format(mat, ring, bank)
+            cells[name] = openmc.Cell(name=name)
+            universes[univ_name].add_cell(cells[name])
+
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].region      = -surfaces['Fuel Outer Clad OR']
+        cells['{} {} Bank {}'.format(mat, 'Moderator', bank)].region = +surfaces['Fuel Outer Clad OR']
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].fill      = materials[mat]
+        cells['{} {} Bank {}'.format(mat, 'Moderator', bank)].fill = materials['Moderator Bank {}'.format(bank)]
```
paulromano, line 69:
```suggestion
        cells[f'{mat} Base Bank {bank}'].region      = -surfaces['Fuel Outer Clad OR']
        cells[f'{mat} Moderator Bank {bank}'].region = +surfaces['Fuel Outer Clad OR']
        cells[f'{mat} Base Bank {bank}'].fill      = materials[mat]
        cells[f'{mat} Moderator Bank {bank}'].fill = materials['Moderator Bank {}'.format(bank)]
```

examples/c5g7_transient/density_change/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
...
+        cells['{} {} Bank {}'.format(mat, 'Core', bank)].region           = -surfaces['Fuel Outer Clad OR'] & +surfaces['Axial Midplane']
+        cells['{} {} Bank {}'.format(mat, 'Core Moderator', bank)].region = +surfaces['Fuel Outer Clad OR'] & +surfaces['Axial Midplane']
+
+        cells['{} {} Bank {}'.format(mat, 'Core', bank)].fill           = materials[mat]
+        cells['{} {} Bank {}'.format(mat, 'Core Moderator', bank)].fill = materials['Moderator Bank {}'.format(bank)]
+
+        cells['{} {} Bank {}'.format(mat, 'Base', bank)].fill = universes['{} {} Bank {}'.format(mat, 'Core', bank)]
+        univ_name = '{} {} Bank {}'.format(mat, 'Base', bank)
+        universes[univ_name] = openmc.Universe(name=univ_name)
+        universes[univ_name].add_cell(cells['{} {} Bank {}'.format(mat, 'Base', bank)])
```
paulromano, line 95:
```suggestion
        cells[f'{mat} Core Bank {bank}'].region           = -surfaces['Fuel Outer Clad OR'] & +surfaces['Axial Midplane']
        cells[f'{mat} Core Moderator Bank {bank}'].region = +surfaces['Fuel Outer Clad OR'] & +surfaces['Axial Midplane']

        cells[f'{mat} Core Bank {bank}'].fill           = materials[mat]
        cells[f'{mat} Core Moderator Bank {bank}'].fill = materials[f'Moderator Bank {bank}']

        cells[f'{mat} Base Bank {bank}'].fill = universes[f'{mat} Core Bank {bank}']
        univ_name = f'{mat} Base Bank {bank}'
        universes[univ_name] = openmc.Universe(name=univ_name)
        universes[univ_name].add_cell(cells[f'{mat} Base Bank {bank}'])
```

examples/c5g7_transient/density_change/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
...
+    universes[univ_name] = openmc.Universe(name=univ_name)
+    for ring in rings:
+        name = '{} {}'.format(mat, ring)
+        cells[name] = openmc.Cell(name=name)
+        universes[univ_name].add_cell(cells[name])
+
+    cells['{} {}'.format(mat, 'Reflector')].region           = -surfaces['Fuel Outer Clad OR']
+    cells['{} {}'.format(mat, 'Reflector Moderator')].region = +surfaces['Fuel Outer Clad OR']
+    cells['{} {}'.format(mat, 'Reflector')].fill           = materials[mat]
+    cells['{} {}'.format(mat, 'Reflector Moderator')].fill = materials['Moderator']
```
paulromano, line 110:
```suggestion
    cells[f'{mat} Reflector'].region           = -surfaces['Fuel Outer Clad OR']
    cells[f'{mat} Reflector Moderator'].region = +surfaces['Fuel Outer Clad OR']
    cells[f'{mat} Reflector'].fill           = materials[mat]
    cells[f'{mat} Reflector Moderator'].fill = materials['Moderator']
```

examples/c5g7_transient/density_change/surfaces.py
```diff
@@ -0,0 +1,36 @@
...
+surfaces = {}
+
+# Instantiate ZCylinder surfaces
+surfaces['Fuel OR']            = openmc.ZCylinder(x0=0, y0=0, r=0.4095, name='Fuel OR')
+surfaces['Fuel Inner Clad IR'] = openmc.ZCylinder(x0=0, y0=0, r=0.4180, name='Fuel Inner Clad IR')
+surfaces['Fuel Inner Clad OR'] = openmc.ZCylinder(x0=0, y0=0, r=0.4750, name='Fuel Inner Clad OR')
+surfaces['Fuel Outer Clad IR'] = openmc.ZCylinder(x0=0, y0=0, r=0.4800, name='Fuel Outer Clad IR')
+surfaces['Fuel Outer Clad OR'] = openmc.ZCylinder(x0=0, y0=0, r=0.5400, name='Fuel Outer Clad OR')
+surfaces['Guide Tube IR']      = openmc.ZCylinder(x0=0, y0=0, r=0.3400, name='Guide Tube IR')
+surfaces['Guide Tube OR']      = openmc.ZCylinder(x0=0, y0=0, r=0.5400, name='Guide Tube IR')
```
paulromano, line 16: For z-cylinders at the origin, you don't need to specify x0 and y0
```suggestion
surfaces['Fuel OR']            = openmc.ZCylinder(r=0.4095, name='Fuel OR')
surfaces['Fuel Inner Clad IR'] = openmc.ZCylinder(r=0.4180, name='Fuel Inner Clad IR')
surfaces['Fuel Inner Clad OR'] = openmc.ZCylinder(r=0.4750, name='Fuel Inner Clad OR')
surfaces['Fuel Outer Clad IR'] = openmc.ZCylinder(r=0.4800, name='Fuel Outer Clad IR')
surfaces['Fuel Outer Clad OR'] = openmc.ZCylinder(r=0.5400, name='Fuel Outer Clad OR')
surfaces['Guide Tube IR']      = openmc.ZCylinder(r=0.3400, name='Guide Tube IR')
surfaces['Guide Tube OR']      = openmc.ZCylinder(r=0.5400, name='Guide Tube IR')
```

examples/c5g7_transient/rod_translation/c5g7-3d-mg-rod.py
```diff
@@ -0,0 +1,122 @@
...
+
+from geometry_mg import materials, surfaces, universes, cells, lattices, geometry, mgxs_lib_file
+from mgxs_lib import mgxs_data
+materials_file = openmc.Materials(geometry.get_all_materials().values())
+os.environ['OPENMC_MG_CROSS_SECTIONS'] = 'mgxs.h5'
+
+###############################################################################
+# Define problem settings
+
+settings_file = openmc.Settings()
```
paulromano, line 15: Again, change `settings_file` to `settings`

examples/c5g7_transient/rod_translation/c5g7-3d-mg-rod.py
```diff
@@ -0,0 +1,122 @@
...
+bounds = [-32.13, -10.71, -64.26, 10.71,  32.13,  64.26]
+entropy_bounds = [-32.13, -10.71, -85.68, 10.71,  32.13,  85.68]
+uniform_dist = openmc.stats.Box(bounds[:3], bounds[3:], only_fissionable=True)
+settings_file.source = openmc.source.Source(space=uniform_dist)
+
+sourcepoint = dict()
+sourcepoint['batches'] = [settings_file.batches]
+sourcepoint['separate'] = True
+sourcepoint['write'] = True
+settings_file.sourcepoint = sourcepoint
```
paulromano, line 33:
```suggestion
settings_file.sourcepoint = {
    'batches': [settings_file.batches],
    'separate': True,
    'write': True
}
```

examples/c5g7_transient/rod_translation/assemblies_mg.py
```diff
@@ -0,0 +1,111 @@
+import openmc
```
paulromano, line 1: This file is almost identical to `density_change/assemblies_mg.py`. I would just remove it and add a symbolic link to the file in the `density_change` directory.

examples/c5g7_transient/rod_translation/materials_mg.py
```diff
@@ -0,0 +1,48 @@
+import openmc
```
paulromano, line 1: This file is almost identical to `density_change/materials_mg.py`. Consider removing it and using a symbolic link instead

examples/c5g7_transient/rod_translation/mgxs_lib.py
```diff
@@ -0,0 +1,282 @@
+#coding=utf-8
```
paulromano, line 1: File is almost identical to `density_change/mgxs_lib.py`. Consider removing and adding a symbolic link instead

examples/c5g7_transient/rod_translation/pin_cells_mg.py
```diff
@@ -0,0 +1,115 @@
+import openmc
```
paulromano, line 1: File is identical to `density_change/pin_cells_mg.py`. Consider removing and adding a symbolic link instead

examples/c5g7_transient/rod_translation/settings.py
```diff
@@ -0,0 +1,31 @@
+import openmc
```
paulromano, line 1: Should this file be removed? It looks like settings are already defined in c5g7-3d-mg-rod.py

examples/c5g7_transient/rod_translation/surfaces.py
```diff
@@ -0,0 +1,36 @@
+import openmc
```
paulromano, line 1: File is identical to `density_change/surfaces.py`. Consider removing and adding a symbolic link instead

examples/c5g7_transient/rod_translation/tallies.py
```diff
@@ -0,0 +1,56 @@
+import openmc
```
paulromano, line 1: This file doesn't appear to do anything. Should it be removed?

examples/pincell_transient/pin-td.py
```diff
@@ -0,0 +1,112 @@
...
+import openmc.mgxs
+import openmc.kinetics as kinetics
+
+from geometry_pin import geometry
+materials_file = openmc.Materials(geometry.get_all_materials().values())
+
+###############################################################################
+# Define problem settings
+
+settings_file = openmc.Settings()
```
paulromano, line 12: Change `settings_file` to `settings` for consistency with other examples

examples/pincell_transient/pin-td.py
```diff
@@ -0,0 +1,112 @@
...
+lower_left = [-0.62992, -0.62992, -182.88]
+upper_right = [0.62992, 0.62992, 182.88]
+uniform_dist = openmc.stats.Box(lower_left, upper_right, only_fissionable=True)
+settings_file.source = openmc.source.Source(space=uniform_dist)
+
+sourcepoint = dict()
+sourcepoint['batches'] = [settings_file.batches]
+sourcepoint['separate'] = True
+sourcepoint['write'] = True
+settings_file.sourcepoint = sourcepoint
```
paulromano, line 29:
```suggestion
settings_file.sourcepoint = {
    'batches': [settings_file.batches],
    'separate': True,
    'write': True
}
```

openmc/kinetics/aux_functions.py
```diff
@@ -0,0 +1,211 @@
+from math import gcd
+import scipy.sparse as sps
+from scipy.sparse.linalg import spsolve
+
+import numpy as np
+from numpy.lib.stride_tricks import as_strided as ast
+
+def lcm(a, b):
```
paulromano, line 8: All the functions in this file need docstrings describing what their purpose is

openmc/kinetics/state.py
```diff
@@ -0,0 +1,1507 @@
+from collections import OrderedDict
+import copy
+import sys
```
paulromano, line 3:
```suggestion
```

tests/regression_tests/transient_ce/test.py
```diff
@@ -0,0 +1,184 @@
+from difflib import unified_diff
+import filecmp
```
paulromano, line 2:
```suggestion
```

tests/regression_tests/transient_ce/test.py
```diff
@@ -0,0 +1,184 @@
...
+import h5py
+import numpy as np
+import os
+
+import openmc
+import openmc.kinetics as kinetics
+import openmc.mgxs
+
+from tests.regression_tests import config
+from tests.testing_harness import PyAPITestHarness, colorize
```
paulromano, line 13:
```suggestion
from tests.testing_harness import PyAPITestHarness
```

tests/regression_tests/transient_mg/test.py
```diff
@@ -0,0 +1,247 @@
+from difflib import unified_diff
+import filecmp
```
paulromano, line 2:
```suggestion
```

paulromano (changes requested):
It looks like the examples could be simplified a bit (see my comments and suggestions below). Regarding the transient input dictionary, I think what you have is fine for now so no need to change that.

paulromano:
@mkreher13 One other comment -- Python source files in OpenMC are supposed to conform to the [PEP8 style guide](https://www.python.org/dev/peps/pep-0008/). Rather than going through all your files and manually making changes, I would recommend applying the [black autoformatter](https://black.readthedocs.io/en/stable/) on new files you've added in this PR.

gridley:
I was just looking at this a little bit more and was curious about something. With quasi-static, aren't we assuming that the precursor conentrations are all at their equilibrium ratios? So, to clarify, shouldn't this fail to accurately solve a prompt jump?

Consider the case where you have two outer time steps: one before the prompt jump and then one right after. In the initial state, the delayed precursors are near their equilibrium ratios. Then the CMFD-based inner method can accurately proceed through the prompt jump part of the transient. After completing that (right after the prompt jump part), you do another MC run, and this MC run (because it has not modified the C++ operation in any way) is assuming that precursors are all present in their equilibrium ratios. However, this is far from reality for a prompt jump. This would introduce some error. Even worse, even if we were using the transforming frequencies which increase the accuracy of this method, this would still not mitigate the problem that precursors are deviating from the equilibrium ratios.

Sjenitzer's method, on the other hand, does account for this effect since the precursor group you sample is always related to the time since it was produced. Moreover, it's not clear to me that the method proposed here is actually more computationally efficient despite its shortcomings in making some approximations which are perhaps not very general at all, despite giving reasonable results for something like C5G7-TD. While it's nice to have larger MC time steps, you can't really overcome the cusping problem with this which is something to really consider in adding transients to MC codes. Sam points out that you need to cut the MC timestep quite a bit to get good results with moving rods, and even the results he presents using the 0.1s MC timestep have some pretty noticeable cusping.

So, if the goal is to have OpenMC solve transients, does it make more sense to implement one very general method with more effort, or alternatively implement one simpler method of questionable generality with the end goal of eventually adding the more general Sjenitzer method, leaving two different routes for users to simulate transients. Of course the manual will always say "use dynamic MC!" because we know it will give the right answer every time, in contrast to something more approximate.

Lastly, even in Sam's thesis, he points out that this method has not been tested on instantaneous reactivity insertions (p 128 last paragraph). There is also not comparison to the results obtained by other codes in his thesis. All of this considered, IMO this should not be merged until this method has been qualified more thoroughly as a generally applicable method for MC transients.

mkreher13:
Of course this method has known shortcomings. Since it assumes that static eigenstates can approximate the transient at the outer time steps, it is intrinsically flawed, as you've pointed out. I think what makes it attractive is that it's easy to merge without any modifications to the source code which has been a major barrier to getting frequencies or the alpha eigenvalue method merged. As it is, the adiabatic method is a decent way to make modifications to material parameters and slow rod movements, and it is a great demonstration of a High-Order/Low-Order method. Ultimately, the OpenMC development team will have to decide what's worth merging and what's not.

mkreher13:
Should I close this PR? As it's been two years, it's a bit out of date from my more recent improvements. I'm not currently in a position to make the changes to bring it up to date, and there doesn't seem to be interest in merging this capability (at least not at this time or in this form). It was a good discussion, but I don't see a need to keep it open.

paulromano:
@mkreher13 Yes, I think at this point it probably makes sense to close this since it hasn't had any traction and is quite out of date. If you do decide to resubmit something along these lines in the future, let's make sure we have a plan that all agree on in advance to avoid another PR suffering the same fate. Sorry this one didn't pan out!

PR #1755, opened 2021-02-04
Allow set_contained to be passed through C-API for cell temp

### LINKED ISSUE #1754: Expose the set_contained option through the C-API
In Cardinal, we're setting the temperatures of pebbles (which are filled with many cells representing TRISOs, etc.) by calling `cell->set_temperature(T, instance, true /* set_contained */)`. Unfortunately, the `runtime_error`s thrown by OpenMC in this function do not get routed through MOOSE, so if OpenMC encounters an error, all that shows up on the MOOSE side is an unhelpful `MPI_Abort`. 

For instance, if we try to set a temperature of `0` to an OpenMC cell, but don't have data at that temperature, we get this error message:

```
Time Step 1, time = 1, dt = 1
Sending temperature to OpenMC...
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=1
```

even though OpenMC internally is throwing an exception with a helpful message:

*in Cell::set_temperature*:
```
      throw std::runtime_error{"Temperature is below minimum temperature at "
        "which data is available."};
```

Because the `openmc_cell_set_temperature` function _does_ catch this exception and saves the error message, I'd like to use `openmc_cell_set_temperature`. *But*, the interface right now does not let you control the `set_contained` parameter.

### Desired Change

Allow `openmc_cell_set_temperature` to control the `set_contained` parameter so that we can catch exceptions for cell temperatures in MOOSE.

aprilnovak:
Allows access to `set_contained` in the C-API, following the same convention that by default it is false.

Closes #1754

### DIFF
--- a/docs/source/capi/index.rst
+++ b/docs/source/capi/index.rst
@@ -105,7 +105,7 @@ Functions
    :return: Return status (negative if an error occurred)
    :rtype: int
 
-.. c:function:: int openmc_cell_set_temperature(index index, double T, const int32_t* instance)
+.. c:function:: int openmc_cell_set_temperature(index index, double T, const int32_t* instance, bool set_contained)
 
    Set the temperature of a cell.
 
@@ -113,6 +113,8 @@ Functions
    :param double T: Temperature in Kelvin
    :param instance: Which instance of the cell. To set the temperature for all
                     instances, pass a null pointer.
+   :param set_contained: If the cell is not filled by a material, whether to set the temperatures
+                         of all filled cells
    :type instance: const int32_t*
    :return: Return status (negative if an error occurred)
    :rtype: int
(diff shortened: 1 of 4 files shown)


pshriwise (approved):
Nice to make this capability consistent between the C++ and C API's. Thanks @aprilnovak! 
 
 This is a pretty straightforward update, so I'll merge it today.

aprilnovak:
Thanks, @pshriwise! After it's merged, if you could rebase your OpenMC branch we're using in Cardinal, that'd be great! If that's at all a hassle though (because it diverged a while ago), then don't worry about it - I hope to get back to your mesh PR soon!

pshriwise:
> Thanks, @pshriwise! After it's merged, if you could rebase your OpenMC branch we're using in Cardinal, that'd be great! If that's at all a hassle though (because it diverged a while ago), then don't worry about it - I hope to get back to your mesh PR soon!

Can do!

PR #1759, opened 2021-02-11
Support rectilinear meshes for CMFD

shikhar413:
This PR adds support for rectilinear meshes for CMFD by exposing the rectilinear mesh implementation through the C API. The actual core of the CMFD implementation did not require much modification since all cross sections and solver quantities should be volume averaged based on the variable `self._hxyz`. These changes are useful for simulating 3-D full-core cases where spacer grids are not placed at equal intervals. A rectilinear mesh therefore allows for the CMFD mesh to be defined exactly at these spacer grid boundaries.

The following plots shows close agreement in the aggregated fission source over 50 active generations for a 200cm slab problem, where the blue line plots the fission source using a 1cm regular mesh and the orange line plots the fission source using the following grid:

```
cmfd_mesh = cmfd.CMFDMesh()
cmfd_mesh.mesh_type = 'rectilinear'
x_grid = [-5., 5.]
y_grid = [-5., 5.]
z_grid = [-100.,  -99.,  -98.,  -97.,  -96.,  -95.,  -94.,  -93.,  -92.,  -91.,
          -90.,  -88.,  -86.,  -84.,  -82.,  -80.,  -78.,  -76.,  -74.,  -72.,
          -70.,  -69.,  -68.,  -67.,  -66.,  -65.,  -64.,  -63.,  -62.,  -61.,
          -60.,  -58.,  -56.,  -54.,  -52.,  -50.,  -48.,  -46.,  -44.,  -42.,
          -40.,  -39.,  -38.,  -37.,  -36.,  -35.,  -34.,  -33.,  -32.,  -31.,
          -30.,  -28.,  -26.,  -24.,  -22.,  -20.,  -18.,  -16.,  -14.,  -12.,
          -10.,   -9.,   -8.,   -7.,   -6.,   -5.,   -4.,   -3.,   -2.,   -1.,
            0.,    2.,    4.,    6.,    8.,   10.,   12.,   14.,   16.,   18.,
           20.,   21.,   22.,   23.,   24.,   25.,   26.,   27.,   28.,   29.,
           30.,   32.,   34.,   36.,   38.,   40.,   42.,   44.,   46.,   48.,
           50.,   51.,   52.,   53.,   54.,   55.,   56.,   57.,   58.,   59.,
           60.,   62.,   64.,   66.,   68.,   70.,   72.,   74.,   76.,   78.,
           80.,   81.,   82.,   83.,   84.,   85.,   86.,   87.,   88.,   89.,
           90.,   92.,   94.,   96.,   98.,  100.]
cmfd_mesh.grid = [x_grid, y_grid, z_grid]
```
image

One C++-specific detail I wasn't sure about is if there was a more elegant way to figure out what sub-class a variable belongs to. For example, for a variable that is defined as a `StructuredMesh`, is there a way to tell if it is of type `RegularMesh` or `RectilinearMesh`? The workaround I used was to define a string variable `type` for the Mesh abstract class which stores the mesh type as a string and called in `openmc_mesh_get_type `. Is there an alternative way of doing this with C++ similar to Python's `isinstance()` function?

Adding @smharper as a reviewer as well since I know he implemented something similar for his Ph.D. work.


smharper:
@shikhar413, nice! It'll be good to have rectilinear CMFD support merged in. I'll take a look at the code soon, but I can respond now to your question about the subclass. There's actually an example of this for meshes now in the `check_mesh` function in `mesh.cpp`. Basically, you start with a pointer of the type `Mesh *`, then you get a `RegularMesh *` pointer via `dynamic_cast`. If it wasn't actually a `RegularMesh` then the returned pointer will be null.

shikhar413:
Thanks @smharper I'll give that a shot

src/mesh.cpp
```diff
@@ -1282,6 +1323,77 @@ openmc_mesh_set_params(int32_t index, int n, const double* ll, const double* ur,
   return 0;
 }
 
+//! Get the rectilinear mesh grid
+extern "C" int
+openmc_rectilinear_mesh_get_grid(int32_t index, double** grid_x, int* nx,
+                       double** grid_y, int * ny, double** grid_z, int* nz)
```
smharper, line 1329:
```suggestion
                       double** grid_y, int* ny, double** grid_z, int* nz)
```

src/mesh.cpp
```diff
@@ -1282,6 +1323,77 @@ openmc_mesh_set_params(int32_t index, int n, const double* ll, const double* ur,
...
+  *nx = m->grid_[0].size();
+  *grid_y = m->grid_[1].data();
+  *ny = m->grid_[1].size();
+  *grid_z = m->grid_[2].data();
+  *nz = m->grid_[2].size();
+
+  return 0;
+}
+
+//! Set the regular mesh parameters
```
smharper, line 1349:
```suggestion
//! Set the rectilinear mesh parameters
```

src/mesh.cpp
```diff
@@ -1282,6 +1323,77 @@ openmc_mesh_set_params(int32_t index, int n, const double* ll, const double* ur,
...
+    }
+    for (int i = 1; i < g.size(); ++i) {
+      if (g[i] <= g[i-1]) {
+        std::cout << g[i] << " " << g[i-1] << "\n";
+        set_errmsg("Values in for x-, y-, and z- grids for "
+          "rectilinear meshes must be sorted and unique.");
+        return OPENMC_E_INVALID_ARGUMENT;
+      }
+    }
+  }
```
smharper, line 1389: Maybe we should put this error checking into something like a `RectilinearMesh::set_grid` function. That way, we can use the same error checking for both the C-API and the normal XML input sequence.

smharper (changes requested):
Apologies for the long delay. And congrats, Dr. @shikhar413!

This code is pretty clean. I have a few minor suggestions here to make the style more consistent with the rest of the code.

I didn't realize before that we were already computing the D_tilde values with an irregular mesh in mind. That's handy! The physics looks good.

shikhar413:
Thanks @smharper. Didnt think about templating the `check_mesh_type` function but that makes sense. I think these changes should address your comments now

smharper (approved):
Looks good!

PR #1761, opened 2021-02-11
Reset n_event in event_death

gridley:
Solves #1760

### DIFF
--- a/src/simulation.cpp
+++ b/src/simulation.cpp
@@ -472,6 +472,9 @@ void initialize_history(Particle& p, int64_t index_source)
   // set progeny count to zero
   p.n_progeny_ = 0;
 
+  // Reset particle event counter
+  p.n_event_ = 0;
+
   // set random number seed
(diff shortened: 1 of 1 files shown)


jtramm:
Great find!

Is there any advantage to moving the `n_event_` reset to the `initialize_history()` function? From my read, it doesn't seem like it would make any difference, but was just curious if you had considered that.

gridley:
Hey @jtramm, thanks. I think you're right it may go better there. Someone may at some point in the future expect to check `n_event_` after a particle death. So, fixed that.

BTW, I seem to have accidentally opened up a new branch within this repository, not my own, as a result of using the online editor since this change was so small. So we need to be sure to delete it after merging  this...

jtramm:
Ah, yes, that's a good point. I could see someone perhaps wanting to make a histogram of # of events per history or something, so retaining that value after death could be useful.

jtramm (approved):
Looks good to me -- thanks @gridley!

PR #1762, opened 2021-02-11
remove unnecessary pass by value for plot class

gridley:
Yet another small PR for today. This doesn't really affect performance, but the plotting routines have some completely unnecessary pass by value of the `Plot` class going on. This PR takes that out.


pshriwise (approved):
Yep. Definitely no need for pass by value here. Thanks @gridley!

PR #1770, opened 2021-02-22
do not set properties of pugixml target if it's not a valid CMake target

### LINKED ISSUE #1769: Need to delete install dir in between builds?
Hi, since commit 489a02bd035140adf0eaf03dd9914a2e905e793f (Merge pull request #1757 from gridley/use_system_pugi_fmt) I find that I need to delete my install directory between builds as cmake finds the installed version of fmt and pugixml, but then it fails to find them later for set_target_properties. I verified I don't get this behaviour with the previous commit. I'm guessing this merge passes because it works from a clean build. This is somewhat inconvenient having to add this step, since I rebuild frequently. As far as I can tell I haven't set an environment variable which is causing the cmake to find these packages, so I'm not aware of anything I can unset on my end which would change this behaviour. (Of course, great if there is - please let me know if I'm incorrect.)  Here's the output from cmake/make.

```
-- Detected MPI wrapper: /usr/bin/mpicxx
-- Configuring DAGMC 3.2.0
-- Found fmt: /home/helen/Projects/openmc/fork/install/lib/cmake/fmt (version 6.1.2)
-- Found pugixml: /home/helen/Projects/openmc/fork/install/lib/cmake/pugixml
-- HDF5: Using hdf5 compiler wrapper to determine C configuration
-- OpenMC C++ flags: -fopenmp;-O2
-- OpenMC Linker flags: -fopenmp
-- Submodule update
-- xtl v0.6.13
-- Building xtensor v0.21.3
-- Found xtl v0.6.13
-- Project 'gsl_lite', package 'gsl-lite' version: '0.36.1'
CMake Error at CMakeLists.txt:416 (set_target_properties):
  set_target_properties Can not find target to add properties to: pugixml


-- Configuring incomplete, errors occurred!
See also "/home/helen/Projects/openmc/fork/openmc/build/CMakeFiles/CMakeOutput.log".
See also "/home/helen/Projects/openmc/fork/openmc/build/CMakeFiles/CMakeError.log".
make: *** [Makefile:3402: cmake_check_build_system] Error 1
```

gridley:
Fixes #1769

OpenMC is now set up to use pugixml if it's already available on the system rather than recompile it, for purposes of distribution [on the AUR](https://aur.archlinux.org/packages/openmc-git/). We set compile properties for pugixml in the CMakeLists.txt file still though, and if we use the system version, it's not available as a CMake target, causing an error as found in #1769.

This small change fixes that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,7 @@ endif()
 
 if (NOT pugixml_FOUND)
   add_subdirectory(vendor/pugixml)
+  set_target_properties(pugixml PROPERTIES CXX_STANDARD 14 CXX_EXTENSIONS OFF)
 endif()
 
 #===============================================================================
@@ -357,8 +358,7 @@ endif()
 
 # Avoid vs error lnk1149 :output filename matches input filename
 if(NOT MSVC)
-set_target_properties(libopenmc PROPERTIES
-  OUTPUT_NAME openmc)
+  set_target_properties(libopenmc PROPERTIES OUTPUT_NAME openmc)
 endif()
 
 target_include_directories(libopenmc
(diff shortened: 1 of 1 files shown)


paulromano (approved):
Simple enough -- thanks @gridley!

PR #1772, opened 2021-02-24
[Fix] 'std::out_of_range' error

AI-Pranto:
Simple enough to fix `terminate called after throwing an instance of ‘std::out_of_range’` error now user can select up to ... well a lot like 1 trillion number of particles to bank per surface :)

### DIFF
--- a/docs/source/io_formats/settings.rst
+++ b/docs/source/io_formats/settings.rst
@@ -751,7 +751,7 @@ This element has the following attributes/sub-elements:
 
     *Default*: None
 
-  :max_particles:
+  :max_surface_particles:
     An integer indicating the maximum number of particles to be banked on
     specified surfaces per processor. The size of source bank in
     ``surface_source.h5`` is limited to this value times the number of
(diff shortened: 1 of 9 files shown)


gridley:
Good catch, looks good to me. Before merging though, can we rename this to "max_surface_particles"? The name used here is really ambiguous.

gridley:
BTW, isn't that ~180 GB of surface source data if we're pushing from 32 bit ints to 64? Must be some crazy calculation you're doing :smile:

AI-Pranto:
I also agree with you @gridley. I don't know why Antonino Meli is trying to do it.
```
Here is the settings.xml file. Btw, have you tried to run with ‘max_particles:’ 1billion or even 1trillion ?
```
> I agree this change is good though, since we've got this variable declared as a 64 bit int we shouldn't be reading it in as a 32.

Yep!

AI-Pranto:
@gridley It's ready to review.

AI-Pranto:
I am going to close this PR and reopen it.

gridley:
Hey Ariful, why close and reopen? Sorry to delay in getting some feedback in on this.

PR #1792, opened 2021-03-10
pre c99 compatible endf.c parsing

gridley:
So, as mentioned in the slack channel, a system I was trying to run on had picked gcc 4.something for compiling endf.c in openmc/data. Rather than figuring out how to make it use -std=c99 (and correspondingly a small change would go into our setup.py file), this small revision should let our endf.c file be compatible with gcc versions of yore.

(ignore the previous PR as i'd accidentally branched off of unmerged commits)


paulromano (approved):
Thanks @gridley for the simple fix!

PR #1797, opened 2021-03-17
Add OpenMC command line arguments to Deplete

dryuri92:
It's a little improvement is going to be useful for small calculation resources to change OpenMC settings.
For example ['-s', '8'] to provide Monte-Carlo simulation with 8 OpenMPI threads.

### DIFF
--- a/openmc/deplete/operator.py
+++ b/openmc/deplete/operator.py
@@ -78,6 +78,9 @@ class Operator(TransportOperator):
         Results from a previous depletion calculation. If this argument is
         specified, the depletion calculation will start from the latest state
         in the previous results.
+    args : list of str, optional
+        Command-line arguments for openmc solver, which might be used to control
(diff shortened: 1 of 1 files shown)


gridley:
If changing the thread count is your goal, you could just set the `OMP_NUM_THREADS` in your environment to the number of threads you'd like to use before running your depletion script. I don't see an application of passing any of the other currently available command line arguments to the solver during depletion, so I'll leave it up to @drewejohnson if he thinks this is worth adding.

Alternatively, we should make it to where we can set the number of threads OpenMC should request through the python API.

dryuri92:
Oh, thank you @gridley! I forget about manipulation number of threads by enviroment variable. Yes this PR is assumed only for such purpose.

drewejohnson:
@dryuri92 using the environment variables, can this PR be closed?

PR #1798, opened 2021-03-18
Added Collision filter

ameliajo:
Added a collision filter to isolate 1st collision events, 2nd, ect. 

Did fixed source calculation for a block of aluminum at 1kev for the first 5 scatter events, and all looks well. Note that the y-axis looks like a weird scale because all the bins sum to ~1 (a bit less to account for absorptions that have happened)
aluminum_scattering


paulromano:
Thanks for the PR @ameliajo! I'll take a look in the next day or two.

src/tallies/filter_collision.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+void CollisionFilter::set_bins(gsl::span<const int> bins)
+{
+  // Clear existing bins
+  bins_.clear();
+  bins_.reserve(bins.size());
+
+  // Copy bins, ensuring they are valid
+  for (gsl::index i = 0; i < bins.size(); ++i) {
+    if (i > 0 && bins[i] <= bins[i - 1]) {
```
gridley, line 32: Two things:

1) This is a possible access out-of-bounds. You check `i>0 && bins[i] <= bins[i-1]`, where potentially `i=0`. You can possibly index into `bins[-1]` in this case, and boom, your computer is up in smoke.

2) I recommend replacing this by two standard library calls. First use [`std::is_sorted`](https://en.cppreference.com/w/cpp/algorithm/is_sorted) to check they're in ascending order (you may need to manually set it to use the <= operator rather than the < operator as is used by default, although I think the < operator is what should be used here anyway). After checking they're in order, push the contents of `bins` into `bins_` by using `std::copy(bins.begin(), bins.end(), std::back_inserter(bins_));`. So this all should be replaced by like three lines of code.

Lastly, I think we are leaning towards calls to `openmc::fatal_error` rather than throwing exceptions now, but @paulromano will have to check me on that one. This is so we can eventually make a our own unified error handling system this way that interfaces with python/GPU perfectly, where exceptions don't work as cleanly.

src/tallies/filter_collision.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+void CollisionFilter::to_statepoint(hid_t filter_group) const
+{
+  Filter::to_statepoint(filter_group);
+  write_dataset(filter_group, "bins", bins_);
+}
+
+std::string CollisionFilter::text_label(int bin) const
+{
+  return fmt::format("Collision Number {}", int(bins_[bin]));
```
gridley, line 69: We use C++ style casts rather than C style in OpenMC to be more explicit about what's going on, so `int(bins_...` should instead be `static_cast<int>(bins_...`.

Although, you shouldn't even need to be casting to an int here, as that's already the data type held by `bins_`.

src/tallies/filter.cpp
```diff
@@ -19,6 +19,7 @@
 #include "openmc/tallies/filter_distribcell.h"
 #include "openmc/tallies/filter_energyfunc.h"
 #include "openmc/tallies/filter_energy.h"
+#include "openmc/tallies/filter_collision.h"
```
gridley, line 22: Put this before distribcell so it's all in alphabetical order please :slightly_smiling_face:

include/openmc/tallies/filter_collision.h
```diff
@@ -0,0 +1,51 @@
+#ifndef OPENMC_TALLIES_FILTER_COLLISIONS_H
```
gridley, line 1: You can replace the ifndefs here at the top of the file (and correspondingly remove the macro closing statement at the bottom of the file) and instead put `#pragma once` at the top of the file to achieve the same effect. Anyone who thinks we shouldn't do this can duke it out with me. There has been pushback against using this in the past out of fear of compiler compatibility, but we're totally inconsistent on that viewpoint because we use gsl, which uses `pragma once`.

src/tallies/filter_collision.cpp
paulromano: Yeah, for now I would recommend just calling `fatal_error`

include/openmc/tallies/filter_collision.h
paulromano: For now please keep the ifdef as is. If we decide to go to `pragma once`, let's do it all at once rather than piecemeal.

openmc/filter.py
```diff
@@ -964,6 +965,127 @@ def get_pandas_dataframe(self, data_size, stride, **kwargs):
...
+        The events whose post-scattering collision number equals one of 
+        the provided values will be counted. 
+    filter_id : int
+        Unique identifier for the filter
+
+    Attributes
+    ----------
+    values : numpy.ndarray
+        An array of integer values representing the number of collisions events
+        by which to filter
```
paulromano, line 984: For this filter, it doesn't really make sense to have a separate `values` attribute. I would just keep bins. For some filters, `values` represents something separate from `bins` (see EnergyFilter for example).

openmc/filter.py
```diff
@@ -964,6 +965,127 @@ def get_pandas_dataframe(self, data_size, stride, **kwargs):
...
+            XML element containing filter data
+
+        """
+        element = super().to_xml_element()
+        element[0].text = ' '.join(str(x) for x in self.values)
+        return element
+
+
+
+
```
paulromano, line 1088:
```suggestion
```

src/tallies/filter_collision.cpp
paulromano: Thinking about this a little bit more, I don't see why we should need to enforce monotonicity at all. We can store an unordered map of (number of collisions) → (index in bins) that will allow us to get the correct index in O(1) time. You can look at CellFilter for an example of how this would work.

src/tallies/filter_collision.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+//==============================================================================
+// C-API functions
+//==============================================================================
+
+extern "C" int openmc_collision_filter_get_bins(
+  int32_t index, const int** energies, size_t* n)
+{
```
paulromano, line 78: Do you anticipate using this filter from the C API? If not, we might want to consider omitting this for now. Not all of our filters are exposed right now, so unless there is a particular use case, my vote would be to leave it out.

openmc/filter.py
```diff
@@ -964,6 +965,127 @@ def get_pandas_dataframe(self, data_size, stride, **kwargs):
...
+    def get_bin_index(self, filter_bin):
+        i = np.where(self.bins[:, 1] == filter_bin[1])[0]
+        if len(i) == 0:
+            msg = 'Unable to get the bin index for Filter since "{0}" ' \
+                  'is not one of the bins'.format(filter_bin)
+            raise ValueError(msg)
+        else:
+            return i[0]
+
+    def get_pandas_dataframe(self, data_size, stride, **kwargs):
```
paulromano, line 1026: I think the default `Filter.get_pandas_dataframe` method should work fine for this filter, in which case you don't need to provide an implementation here

openmc/filter.py
```diff
@@ -964,6 +965,127 @@ def get_pandas_dataframe(self, data_size, stride, **kwargs):
...
+    def bins(self, bins):
+        Filter.bins.__set__(self, np.asarray(bins))
+
+    def check_bins(self, bins):
+        for x in bins:
+            # Values should be integers
+            cv.check_type('filter value', x, Integral)
+            cv.check_greater_than('filter value', x, 0, equality=True)
+
+    def get_bin_index(self, filter_bin):
```
paulromano, line 1017: I think the default implementation of `get_bin_index` should work fine, so you can omit this method

paulromano (changes requested):
This is a great start @ameliajo. Other than the line comments, a few other things are needed:
- Add the filter to `docs/source/pythonapi/base.rst`
- Add a tally with this filter to `tests/regression_tests/tallies/test.py` so that it gets tested. I'd recommend reading through the dev guide section on [tests](https://docs.openmc.org/en/latest/devguide/tests.html#test-suite)
- Add a unit test for this filter in `tests/unit_tests/test_filters.py`

src/tallies/filter_collision.cpp
ameliajo: So you'd recommend removing the ordered check entirely? I'm alright with that. I suppose it while it seems most natural to put them in increasing order there is no real reason to enforce that.

src/tallies/filter.cpp
ameliajo: 1. I can't believe I did not notice that this was in alphabetical order
2. I'm so happy this is something you pick up on.

I'll change it :)

openmc/filter.py
ameliajo: I used EnergyFilter as an example of how filters are typically written in OpenMC, so I kept bins separate from values in case there was some distinction use that I had not picked up on. Glad to know it's nothing like that. I'll make the change, thanks!

src/tallies/filter_collision.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+void CollisionFilter::get_all_bins(
+  const Particle& p, TallyEstimator estimator, FilterMatch& match) const
+{
+  // Get the number of collisions for the particle
+  auto n = p.n_collision_;
+
+  // Bin the collision number. Must fit exactly the desired collision number .
+  if (n >= bins_.front() && n <= bins_.back()) {
```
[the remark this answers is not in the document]
ameliajo: If I go with @paulromano's suggestion to not require sorted bins, I might opt to keep it as it, unless you suggest otherwise @gridley ?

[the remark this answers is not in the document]
gridley: Oh yeah, sounds totally good if you take that route!

src/tallies/filter_collision.cpp
paulromano: Exactly, might as well give the user the flexibility to order them however they want since we can do O(1) lookups with an unordered map anyway

openmc/filter.py
```diff
@@ -964,6 +965,66 @@ def get_pandas_dataframe(self, data_size, stride, **kwargs):
...
+        by which to filter
+    num_bins : int
+        The number of filter bins
+
+    """
+
+    def __init__(self, bins, filter_id=None):
+        self.bins = np.asarray(bins)
+        self.id = filter_id
+
```
paulromano, line 995:
```suggestion
```

openmc/filter.py
```diff
@@ -964,6 +965,66 @@ def get_pandas_dataframe(self, data_size, stride, **kwargs):
...
+        -------
+        element : xml.etree.ElementTree.Element
+            XML element containing filter data
+
+        """
+        element = super().to_xml_element()
+        element[0].text = ' '.join(str(x) for x in self.bins)
+        return element
+
+
```
paulromano, line 1026:
```suggestion
```

src/tallies/filter_collision.cpp
```diff
@@ -0,0 +1,63 @@
...
+  n_bins_ = bins_.size();
+}
+
+void CollisionFilter::get_all_bins(
+  const Particle& p, TallyEstimator estimator, FilterMatch& match) const
+{
+  // Get the number of collisions for the particle
+  auto n = p.n_collision_;
+
+  // Bin the collision number. Must fit exactly the desired collision number .
```
paulromano, line 44:
```suggestion
  // Bin the collision number. Must fit exactly the desired collision number.
```

tests/regression_tests/tallies/test.py
```diff
@@ -169,12 +169,22 @@ def test_tallies():
...
 
+    n_collision = (1,2,5,3,6)
+    collision_filter = CollisionFilter(n_collision)
+    collision_tally = Tally()
+    collision_tally.filters = [collision_filter]
+    collision_tally.scores = ['scatter']
+
+
+
+
```
paulromano, line 180:
```suggestion
```

tests/regression_tests/tallies/test.py
```diff
@@ -169,12 +169,22 @@ def test_tallies():
     fusion_tally.scores = ['H1-production', 'H2-production', 'H3-production',
         'He3-production', 'He4-production', 'heating', 'damage-energy']
 
+    n_collision = (1,2,5,3,6)
```
paulromano, line 172:
```suggestion
    n_collision = (1, 2, 5, 3, 6)
```

tests/regression_tests/tallies/test.py
```diff
@@ -169,12 +169,22 @@ def test_tallies():
...
+
+
     model.tallies += [
         azimuthal_tally1, azimuthal_tally2, azimuthal_tally3,
         cellborn_tally, dg_tally, energy_tally, energyout_tally,
         transfer_tally, material_tally, mu_tally1, mu_tally2,
         polar_tally1, polar_tally2, polar_tally3, legendre_tally,
-        harmonics_tally, harmonics_tally2, harmonics_tally3, universe_tally]
+        harmonics_tally, harmonics_tally2, harmonics_tally3, 
+        universe_tally,collision_tally]
```
paulromano, line 187:
```suggestion
        universe_tally, collision_tally]
```

tests/unit_tests/test_filters.py
```diff
@@ -46,6 +46,22 @@ def test_cell_instance():
     assert instances.apply(lambda x: x in (0, 1, 2)).all()
 
 
+def test_collision():
+    f = openmc.CollisionFilter([1,5,3,2,8])
```
paulromano, line 50:
```suggestion
    f = openmc.CollisionFilter([1, 5, 3, 2, 8])
```

paulromano (changes requested):
@ameliajo The changes are looking great. A few more nitpicky requests on style here. Also, the tallies regression test is currently failing because you need to update the results. This can be done by running the test with the `--update` flag:
```bash
cd tests/regression_tests/tallies
pytest --update .
```

openmc/filter.py
ameliajo: Hey @paulromano, I'll address these, thank you! Sorry for being so late, as I mentioned in another comment I'm having difficulty getting the openmc tests to even run/pass on any of my machines. I'll try the above pytest --update . command again, though

paulromano: Ah, ok, I know a lot of people have had problems with getting tests to match on macs. What I can do for now is update the test and send a pull request to your branch so we can keep this moving.

paulromano (approved):

paulromano:
@gridley do you have further requests before we merge this?

gridley (approved):

gridley:
> @gridley do you have further requests before we merge this?

Ah, thanks for checking and sorry to not mark my previous review as resolved on the requested changes and hold things back. This all looks great to me now.

paulromano:
Thanks again for the contribution @ameliajo!

PR #1801, opened 2021-03-25
moving from dagmc develop to latest tagged release

Shimwell:
Perhaps overdue minor changed to the Dockerfile that tags a specific version of DAGMC.

Occasionally the develop branch [fails](https://app.circleci.com/pipelines/github/svalinn/DAGMC) to build, tagging the latest release is probably advisable

If people like tagged then this PR can be expanded to include tagged versions for other codes such as Embree MOAB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -117,7 +117,7 @@ RUN if [ "$include_dagmc" = "true" ] ; \
 RUN if [ "$include_dagmc" = "true" ] ; \
     then mkdir DAGMC ; \
     cd DAGMC ; \
-    git clone --single-branch --branch develop https://github.com/svalinn/DAGMC.git ; \
+    git clone --single-branch --branch 3.2.0 https://github.com/svalinn/DAGMC.git ; \
     mkdir build ; \
     cd build ; \
     cmake ../DAGMC -DBUILD_TALLY=ON \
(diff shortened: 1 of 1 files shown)


pshriwise:
Thanks for this @Shimwell! It looks like we should go ahead and do the same with CI as well...

pshriwise:
And to answer your question about the other versions. Yeah, I think fixing those as well would be a good idea.

Shimwell:
I can update the CI scripts if that helps
Perhaps this can change to  MOAB_BRANCH='5.2.1' https://github.com/openmc-dev/openmc/blob/aeee1664f363495106c322317dc319bf45412045/tools/ci/gha-install-dagmc.sh#L6

and this could be changed to DAGMC_BRANCH='3.2.0' https://github.com/openmc-dev/openmc/blob/aeee1664f363495106c322317dc319bf45412045/tools/ci/gha-install-dagmc.sh#L11

It looks like Embree is not used in the CI installation scripts

Just to throw this out there as an option, the dockerfile could also be reshaped to allow build arguments and then cover all CI environments but perhaps that is too much change at the moment.

pshriwise:
I like the option of adding optional build args for the Dockerfile in the future. If you'd like to give it a try, go for it!

For CI, we're going to discuss the pros and cons of fixing these optional dependency versions in CI soon, so let's hold off on that for now.

pshriwise (approved):
Thanks @Shimwell! It will be nice to have stable versions in the Dockerfile. They can always be updated to development versions by those using the file if needed.

PR #1807, opened 2021-03-30
Fix calculation of neutron data min and max temperatures

### LINKED ISSUE #1806: Incorrectly reported min/max temperature for neutron data
Hey everyone,

So, the problem here is that given the setup below, I get this out of OpenMC:

```
 Minimum neutron data temperature: 294.0 K
 Maximum neutron data temperature: 250.0 K
```

Well, that's clearly not right. This is on the latest develop branch, btw. To reproduce this, simply go to the pin cell example, turn on temperature interpolation mode in settings.xml, and change the fuel temperature to 600 and the clad temperature to 200, for instance. You should then obtain the above, which is clearly incorrect. It also misses the max fuel temperature of 600.

gridley:
Fix #1806

### DIFF
--- a/src/nuclide.cpp
+++ b/src/nuclide.cpp
@@ -33,7 +33,7 @@ namespace data {
 std::array<double, 2> energy_min {0.0, 0.0};
 std::array<double, 2> energy_max {INFTY, INFTY};
 double temperature_min {0.0};
-double temperature_max {INFTY};
+double temperature_max {0.0};
 std::unordered_map<std::string, int> nuclide_map;
 std::vector<std::unique_ptr<Nuclide>> nuclides;
 } // namespace data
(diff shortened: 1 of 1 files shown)


src/nuclide.cpp
```diff
@@ -169,11 +169,10 @@ Nuclide::Nuclide(hid_t group, const std::vector<double>& temperature)
   // Sort temperatures to read
   std::sort(temps_to_read.begin(), temps_to_read.end());
 
-  double T_min_read = *std::min_element(temps_to_read.cbegin(), temps_to_read.cend());
-  double T_max_read = *std::max_element(temps_to_read.cbegin(), temps_to_read.cend());
-
-  data::temperature_min = std::max(data::temperature_min, T_min_read);
-  data::temperature_max = std::min(data::temperature_max, T_max_read);
+  data::temperature_min =
+    std::min(data::temperature_min, static_cast<double>(temps_to_read.front()));
```
jtramm, line 173: Will this always choose 0 as that's what `temperature_min` is initialized to on line 35?

gridley: Ah, good catch. So we need to initialize temperature_min to INFTY instead.

jtramm: If it's initialized to INFTY, can we have the case where minimum temperature ends up being INFTY while the maximum temperature ends up being zero?

gridley: It should be impossible, so long as any nuclide gets loaded. If a nuclide does not get loaded, they will stay put.

jtramm: In the old method, if no nuclides are loaded then the min is 0 and the max is INFTY, which makes enough sense. In the new method, we end up with min = INFTY and max = 0, which does not make sense. But, perhaps if no nuclides are loaded then we don't care about these fields? I can't think of a case where we would.

gridley: Yeah, definitely agree that a void-only case will have no meaningful interpretation of temperature.

jtramm (approved):
Looks good now -- thanks for tracking this one down and making the fix.

PR #1809, opened 2021-04-01
Reset particle coordinates in `neighbor_list_find_cell`

pshriwise:
I discovered today that particles moving from a rotated to an unrotated cell cause a seg fault because the `LocalCoord::rotated` property isn't updated correctly. I originally updated the `LocalCoords::rotate` method to always update this property in `find_cell_innder`, but I think it's best to ensure that the lower level coordinates always get reset in `neighbor_list_find_cell` in case there are other properties of `LocalCoord` that might cause problems if they aren't updated.

I've added a test that fails for this reason ([failing CI run in my fork](https://github.com/pshriwise/openmc/actions/runs/707299555)) and added a loop that resets the particle's lower level coordinates in `neighbor_list_find_cell` to fix the problem.


gridley:
Oops, sorry to introduce this regression. Good thing the test is in place now!

gridley (approved):

PR #1812, opened 2021-04-07
cover a previously missed quadric edge case

### LINKED ISSUE #1811: segfault with particle first crossing quadric
Hi all,

It seems that if a particle is born, and touches a vacuum BC quadric before anything else happens, we get a segfault. I have prepared a pure void box problem that reproduces the issue below. The quadric should be equivalent to the commented out y plane surface, and you can see that with the y plane, OpenMC runs as expected, but the quadric makes OpenMC choke. I think it has to do with the particle's internal `surf_` state not being set correctly, somehow.

```
<settings>
  <run_mode>fixed source</run_mode>
  <particles>1</particles>
  <batches>1</batches>
  <source strength="1.0">
    <space type="box">
      <parameters>0.5 0.5 0 0 1 1</parameters>
    </space>
    <angle type="monodirectional">
        <reference_uvw>0 1 0</reference_uvw>
    </angle>
  </source>
  <verbosity>10</verbosity>
</settings>

<materials>
</materials>

<geometry>
  <cell material="-1" id="1" name="Cell 1" region="1 -2 3 -4" universe="0" />
  <surface boundary="vacuum" coeffs="-2" id="1" name="left" type="x-plane" />
  <surface boundary="vacuum" coeffs="2" id="2" name="right" type="x-plane" />
  <surface boundary="vacuum" coeffs="-2" id="3" name="bottom" type="y-plane" />
  <surface boundary="vacuum" coeffs="0 0 0 0 0 0 0 1 0 -2" id="4" name="top" type="quadric" />
  <!-- <surface boundary="vacuum" coeffs="2" id="4" name="top" type="y-plane" /> -->
</geometry>
```

gridley:
Closes #1811

There is a measure zero set of quadric/particle orientations that lead to division by zero in the quadric intersection routine. One likely example is if a user inputs a general plane as a quadric, which will cause failure of this routine every time at the moment and segfault OpenMC, as shown in #1811. the code comment explains a little more. I think there are other edge cases this covers, but haven't put time into figuring what exactly they are.

### DIFF
--- a/src/surface.cpp
+++ b/src/surface.cpp
@@ -971,8 +971,11 @@ SurfaceQuadric::distance(Position r, Direction ang, bool coincident) const
   } else if (coincident || std::abs(c) < FP_COINCIDENT) {
     // Particle is on the surface, thus one distance is positive/negative and
     // the other is zero. The sign of k determines which distance is zero and
-    // which is not.
-    if (k >= 0.0) {
+    // which is not. Additionally, if a is zero, it means the particle is on
+    // a plane-like surface.
+    if (a == 0.0) {
+      d = INFTY; // see the below explanation
+    } else if (k >= 0.0) {
       d = (-k - sqrt(quad)) / a;
     } else {
       d = (-k + sqrt(quad)) / a;
--- a/src/surface.cpp
+++ b/src/surface.cpp
@@ -978,6 +978,15 @@ SurfaceQuadric::distance(Position r, Direction ang, bool coincident) const
       d = (-k + sqrt(quad)) / a;
     }
 
+  } else if (a == 0.0) {
+    // Given the orientation of the particle, the quadric looks like a plane in
+    // this case, and thus we have only one solution despite potentially having
+    // quad > 0.0. While the term under the square root may be real, in one
+    // case of the +/- of the quadratic formula, 0/0 results, and in another, a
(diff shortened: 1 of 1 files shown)


paulromano (approved):
Makes sense to me. Thanks @gridley! I'll leave it open for a little bit in case others have thoughts.

src/surface.cpp
```diff
@@ -978,6 +978,15 @@ SurfaceQuadric::distance(Position r, Direction ang, bool coincident) const
       d = (-k + sqrt(quad)) / a;
     }
 
+  } else if (a == 0.0) {
```
pshriwise, line 981: I'm wondering if this case should come before the one above. In the case where a particle is coincident with a planar surface defined as a quadric, it looks like we'll enter the case above and end up dividing by zero still.

pshriwise (commented):
This approach and explanation look solid to me. Thanks for documenting it well @gridley. Just one comment about control flow and I'm happy.

src/surface.cpp
gridley: Ah, true. I never hit this issue since I was only doing quadrics as problem boundaries, but I think non-boundary planar quadrics will give an error here as-is. Good catch!

gridley: OK, the latest commit addresses this. I couldn't immediately come up with a more elegant way to write this.

PR #1817, opened 2021-04-14
added reference to material class and xml output

Shimwell:
Hi all

This is a tiny change to the Material.py file that allows one to add a string reference to a material.

## Motivation

Traceable materials have to be a good thing, ideally one would know where the material information such as density and isotopes came from. Adding a reference attribute to the Material class and printing in the xml file facilitates traceability.

## before

I would save information on the materials with a comment which would be lost in the xml file
```python
import openmc

my_mat=openmc.Material()
my_mat.temperature=500
# Don't forget this material came from PNNL material compendium revion 1. page 174
my_mat.add_element('Kr', 1)
my_mat.set_density('g/cm3', 0.003478)

all_mats=openmc.Materials([my_mat])
all_mats.export_to_xml()
```

## After

With this PR one can save information on the materials with a .reference property
```python
import openmc

my_mat=openmc.Material()
my_mat.temperature=500
my_mat.reference='material from PNNL material compendium revion 1. page 174'
my_mat.add_element('Kr', 1)
my_mat.set_density('g/cm3', 0.003478)

all_mats=openmc.Materials([my_mat])
all_mats.export_to_xml()
```

```xml
and it goes into the xml file

<materials>
  <material id="1" reference="material from PNNL material compendium revion 1. page 174" temperature="500">
    <density units="g/cm3" value="0.003478" />
    <nuclide ao="0.00355" name="Kr78" />
    <nuclide ao="0.02286" name="Kr80" />
    <nuclide ao="0.11593" name="Kr82" />
    <nuclide ao="0.115" name="Kr83" />
    <nuclide ao="0.56987" name="Kr84" />
    <nuclide ao="0.17279" name="Kr86" />
  </material>
</materials>
```

What do people think, is this a worth while idea.

### DIFF
--- a/openmc/material.py
+++ b/openmc/material.py
@@ -91,11 +91,12 @@ class Material(IDManagerMixin):
     next_id = 1
     used_ids = set()
 
-    def __init__(self, material_id=None, name='', temperature=None):
+    def __init__(self, material_id=None, name='', temperature=None, reference=None):
         # Initialize class attributes
         self.id = material_id
         self.name = name
         self.temperature = temperature
+        self.reference = reference
         self._density = None
         self._density_units = 'sum'
         self._depletable = False
@@ -120,6 +121,7 @@ class Material(IDManagerMixin):
         string += '{: <16}=\t{}\n'.format('\tID', self._id)
         string += '{: <16}=\t{}\n'.format('\tName', self._name)
         string += '{: <16}=\t{}\n'.format('\tTemperature', self._temperature)
+        string += '{: <16}=\t{}\n'.format('\tReference', self._reference)
 
         string += '{: <16}=\t{}'.format('\tDensity', self._density)
         string += ' [{}]\n'.format(self._density_units)
@@ -148,6 +150,10 @@ class Material(IDManagerMixin):
     @property
     def temperature(self):
         return self._temperature
+
+    @property
+    def reference(self):
+        return self._reference
 
     @property
     def density(self):
@@ -219,6 +225,12 @@ class Material(IDManagerMixin):
         cv.check_type('Temperature for Material ID="{}"'.format(self._id),
                       temperature, (Real, type(None)))
         self._temperature = temperature
+
+    @reference.setter
+    def reference(self, reference):
+        cv.check_type('Reference for Material ID="{}"'.format(self._id),
+                      reference, (str, type(None)))
+        self._reference = reference
 
     @depletable.setter
     def depletable(self, depletable):
@@ -953,6 +965,10 @@ class Material(IDManagerMixin):
         # Create temperature XML subelement
         if self.temperature is not None:
             element.set("temperature", str(self.temperature))
+
+        # Create reference XML subelement
+        if self.reference is not None:
+            element.set("reference", str(self.reference))
 
         # Create density XML subelement
         if self._density is not None or self._density_units == 'sum':


eepeterson:
I think this is definitely worth it @Shimwell. Thanks for the addition! This would help with QA for sure and would pair well with a future materials database PR.

paulromano:
@Shimwell looking at your example, it seems like `Material.name` could serve this purpose already. How will this `reference` field be any different than `name`?

Shimwell:
I make use of Material().name in DAGMC models as the name corresponds to the h5m tag within the model

However DAGMC tags are limited to 28 characters

So I would typically like something like this that supports longer strings for additional details

```python
my_mat=openmc.Material(name='magnet_shielding')
my_mat.temperature=500
my_mat.reference='material from PNNL material compendium revion 1. page ....'
my_mat.add_element('W', 1)
my_mat.set_density('g/cm3', 0.003478)
```

eepeterson:
@paulromano I can imagine use cases in model generation where somebody may want access to or use the name attribute for materials in model building efforts, but might not want the full reference to be part of the name. Also, if the reference has a url or doi number in it that would make the name attribute pretty cumbersome to use.

gridley:
Why not just add a comment with references to the python code? I don't see any scenario where this needs to be propagated into the XML.

eepeterson:
@gridley If you were to submit a neutronics calculation for QA/QC during design you would likely hand over all the inputs necessary to run openmc, but not necessarily the python file that generated them. The xml files are ultimately the important files to quality check. I think this is a pretty low overhead addition that would actually add value to the code in terms of adopting it in startups that often outsource QA stuff to consultants.

gridley:
Hm, I'm not familiar with the requirements of QA or QC, so have to take your word for that.

Shimwell:
> Hm, I'm not familiar with the requirements of QA or QC, so have to take your word for that.

It is not the most fun topic out there, but I think every bit of traceability helps
https://www.gov.uk/government/collections/quality-assurance-tools-and-guidance-in-decc

I think OpenMC falls in to the "non-Excel" category  :joy:

eepeterson:
We're going through this process right now with the SPARC design and MCNP. This small addition is essentially born out of expediting that process in the hopes that OpenMC can supplant MCNP as the tool of choice in fusion design (sooner rather than later).

gridley:
I see, well, my prior comment was simply due to my lack of knowledge on the tasks you all are facing! As long as we're not adding more fields to the C++ data structures that can impact performance, I see no reason against this aside from the possibility that you two will be the only ones using this feature. If that became the case, this could be considered software bloat.

So, if doing this for materials, why not do similar for geometry or the like? It seems like any object in the python API should have a field for additional documentation, then.

openmc/material.py
```diff
@@ -44,13 +44,19 @@ class Material(IDManagerMixin):
     temperature : float, optional
         Temperature of the material in Kelvin. If not specified, the material
         inherits the default temperature applied to the model.
+    reference : str, optional
+        Reference string that can be added to the material and propagates
+        to the material.xml file. Intended use is to containing additional
```
gridley, line 49: Check the wording here

gridley:
Also, skimming that QA document, I still don't see why you wouldn't just do QA on the python script that creates the input. If the goal of QA is for someone else to catch errors and have everything be thoroughly documented, the python side is the way to go.

eepeterson:
There are just too many processes that happen between reading/writing the text in a python script (which may not contain material definitions that are likely loaded from an external library of previously QA'ed materials either stored as a python file or more likely an xml file) and the materials that show up in the simulation itself. QA is a giant pain and you have to check every "pipe" between your starting point and the end point. If you can cut out some of that process that speeds things up. Plus, when things in the pipes change from time to time they need to be re-qualified. The xml file format for materials seems pretty stable and robust and I don't really see any need for that to change in the C++ layer which makes it a good target for QA'ing once and being done with it more or less. 

As for everything else in the Python API, I don't think this needs to be done for anything like geometry/settings/tallies etc. Those things would get QA'ed in a separate fashion to the material compositions. Ultimately, I think it would be nice to ship a material database along with OpenMC out of the box so to speak since more or less the first thing I did when starting with OpenMC was create a list of materials and store it for future use. Most of these materials would have standard references so I think it would be important to include them. 

I'm also against software bloat (I'm not sure I know anyone who's not), but I think you'd see more people than just Jon and I relying on this feature, especially if we add a common materials library.

gridley:
Thanks for explaining that Ethan. I see where you're coming from now.

On the topic of materials, we definitely need to get something similar to Serpent's `sss2 -comp` functionality. It includes all the materials below by default. It would fit perfectly into `openmc.model`.

eepeterson:
Right. That PNNL document was what I was thinking would make up the initial cut at a material library. I've never used Serpent, so I'm not familiar with the `sss2 -comp` functionality, but I will check it out. The other addition to that library that I would suggest is elemental densities in similar fashion to how we keep the elemental abundances. That way elements with their standard densities could be contained within the built in library (so I don't have to keep going to wikipedia to look up different densities of things...)

nelsonag:
I love this PR and can echo the benefits of making it easier to QA models. @gridley hits on a good question that many have struggled with in the past (revised slightly to make the point): "should we be QA'ing the ASCII input files or the program that made those inputs?" Generally folks prefer to QA the ASCII as part of the engineering QA process, as if you QA the program that made them, then you are in the Software QA space which can be way more involved and time consuming. 

I'd support making it easier for QA however possible, though I do have 2 reservations (which arent necessarily reasons not to do):
1. Documenting in the XML won't help when using the C-API
2. Is the XML here to stay long term or not? If it will be here for at least a few more years, then this makes sense. However, if the intent is to remove for full Python - C API interactions, then this may not be useful for long.

In addition to the above, may I make a proposal though? What if the ``.reference`` attribute was generalized to be ``.comment``?  This also can be added to all  classes (preferably with its own Mixin or in the IDManagerMixin abstract class) with the functionality that an XML comment will be written with the contents of that ``.comment`` attribute.  I think this captures the intent of ``.reference`` while being clearer on how it can be used.

eepeterson:
I like all the points you're making @gridley and @nelsonag. I'd echo the sentiment that QA'ing the ASCII inputs is the preferred method in the engineering QA process (not that I agree with this necessarily, but it seems to be the way things are right now). Whether or not that is a good enough reason to keep XML around - I don't know, but likely not. As for generalizing the `reference` to a `comment` I think that works just fine too, but your point number 2 above is definitely worth discussing more. Is there an existing issue or project created for that topic where I can read what people think?

makeclean:
@nelsonag until you can 100% guarantee that the python script you run, runs the same on my machine as it does yours including your own funky dependencies and environment settings, then I think the xml based input is here to stay. I can be very confident that if I get your nuclear data, your xml file, and the git sha of the version of openmc you used, I can be pretty confident we'll have identical results. That can't be said of the python 'input' which may be sensitive to python versions, dependencies, environment variables etc etc. That being said, the more we can encourage people along the software qa route the better. Further, my experience currently with using OpenMC in a multiphysics environment, an extra bit of python would be really unwelcome and unweildy as part of the running stack.

nelsonag:
@eepeterson, @makeclean, I agree with you 100% that the XML is generally preferable for production-grade computations and I am not going arguing that we should be getting rid of it. There is not an existing discussion/issue/etc that I am aware of; I am simply remembering our decision to emphasize the API in the docs over the XML (years ago). That is it and no more, it was likely very over-zealous of myself to even include that based on that! ;-)

cjwyett:
A (probably poor) thought on the implementation: this would be more flexible if it were extended to a material.metadata like [PyNE ](https://pyne.io/usersguide/material.html#user-defined-metadata)? Some kind of iterable to allow the user to attach whatever information they see fit.

nelsonag:
Also, shall we move discussions (that dont relate to this PR) on things we can do to help strengthen the ability ro do software QA of OpenMC and it's models, and to improve engineering QA of models to a thread on slack/discourse? If we want a dedicated thread, I'd be 100% interested in joining in

eepeterson:
@nelsonag I second the slack/discourse thread. @cjwyett thanks for chiming in! After looking over the link you provided it looks like some of the examples of metadata like units shouldn't really be metadata and are in fact already covered by the existing attributes and methods of the `Material` class in OpenMC. Generally adding flexibility for the sake of flexibility isn't the best plan in my opinion. This PR is proposing a concrete addition to the xml material format to achieve a specific goal (improve material composition traceability for QA purposes) and the idea of arbitrary metadata doesn't really fit within the context of that goal. If you have particular use-cases in mind I'd be happy to hear them, but I would tend to disagree with the implementation of a metadata dict-like attribute.

openmc/material.py
```diff
@@ -44,13 +44,19 @@ class Material(IDManagerMixin):
...
+        details regarding the source of the material for traceability reasons.
 
     Attributes
     ----------
     id : int
         Unique identifier for the material
     temperature : float
         Temperature of the material in Kelvin.
+    reference : str
+        Reference details for the material.
```
Shimwell, line 59:
```suggestion
    comment: str
        Comment attached to the material.
```

openmc/material.py
Shimwell, line 49:
```suggestion
    comment: str, optional
        Comment string that can be added to the material and propagates
        to the materials.xml file. Intended use is to contain additional
```

openmc/material.py
```diff
@@ -91,11 +97,12 @@ class Material(IDManagerMixin):
     next_id = 1
     used_ids = set()
 
-    def __init__(self, material_id=None, name='', temperature=None):
+    def __init__(self, material_id=None, name='', temperature=None, reference=None):
```
Shimwell, line 100:
```suggestion
    def __init__(self, material_id=None, name='', temperature=None, comment=None):
```

openmc/material.py
```diff
@@ -91,11 +97,12 @@ class Material(IDManagerMixin):
...
         # Initialize class attributes
         self.id = material_id
         self.name = name
         self.temperature = temperature
+        self.reference = reference
```
Shimwell, line 105:
```suggestion
        self.comment= comment
```

openmc/material.py
```diff
@@ -120,6 +127,7 @@ def __repr__(self):
         string += '{: <16}=\t{}\n'.format('\tID', self._id)
         string += '{: <16}=\t{}\n'.format('\tName', self._name)
         string += '{: <16}=\t{}\n'.format('\tTemperature', self._temperature)
+        string += '{: <16}=\t{}\n'.format('\tReference', self._reference)
```
Shimwell, line 130:
```suggestion
        string += '{: <16}=\t{}\n'.format('\tComment', self.comment)
```

openmc/material.py
```diff
@@ -149,6 +157,10 @@ def name(self):
     def temperature(self):
         return self._temperature
 
+    @property
+    def reference(self):
+        return self._reference
```
Shimwell, line 162:
```suggestion
    def comment(self):
        return self._comment
```

openmc/material.py
```diff
@@ -220,6 +232,12 @@ def temperature(self, temperature):
                       temperature, (Real, type(None)))
         self._temperature = temperature
 
+    @reference.setter
+    def reference(self, reference):
+        cv.check_type('Reference for Material ID="{}"'.format(self._id),
+                      reference, (str, type(None)))
+        self._reference = reference
+
```
Shimwell, line 240:
```suggestion
    @comment.setter
    def comment(self, comment):
        cv.check_type('Comment for Material ID="{}"'.format(self._id),
                      comment, (str, type(None)))
        self._comment = comment

```

openmc/material.py
```diff
@@ -954,6 +972,10 @@ def to_xml_element(self, cross_sections=None):
         if self.temperature is not None:
             element.set("temperature", str(self.temperature))
 
+        # Create reference XML subelement
+        if self.reference is not None:
+            element.set("reference", str(self.reference))
+
```
Shimwell, line 978:
```suggestion
        # Create reference XML subelement
        if self.comment is not None:
            element.set("comment", str(self.comment))

```

openmc/material.py
Shimwell: Thanks

paulromano:
Good discussion here! A lot of excellent points have been raised, many of which resonate with me. To chime in with thoughts about the future of XML/Python, it's never been a goal of mine to completely eliminate XML and go 100% Python. I personally appreciate (and commonly exercise) the ability to have a set of XML files and a plain `openmc` executable that I can run without having to worry about my Python stack. There's a separate question of whether XML itself is the best format and one could certainly make an argument that perhaps we should be using JSON or YAML or any of the various other formats in existence, but the basic advice we give to users (build your models using the Python API) will be the same.

Not sure how this fits in to the discussion, but there are already things in the code (namely, depletion) that are effectively Python-only and can't be done with XML alone.

With respect to the proposed change here, let me throw out some alternatives and see how quickly they get shot down :smile: One option would be to embed both name/reference information in the `name` attribute, something like:
```Python
m = openmc.Material("awesome metal | This material is from PNNL compendium, page 19728127")
```
As long as model scripts adhere to the convention, it's easy to pull out information on the reference.

Another alternative -- have the model script itself write a separate reference information file:
```Python
references = {}

m1 = openmc.Material(name="petersonium")
references[m1] = "from PNNL compendium, page 02139"

m2 = openmc.Material(name="shimwellium")
references[m2] = "from PNNL compendium, page OX14 3DB"

...
with open("materials_reference.txt", "w") as fh:
    for mat, reference in references.items():
        fh.write("{mat.id}: {reference}\n")
```
Whether you're doing something like this or using the proposed `material.comment` attribute, it's still incumbent on the person building the model to follow the convention required for proper QA. Having the `comment` attribute is, admittedly, just a tad simpler, but either seems like it would be feasible to include in organizational procedures. Also, having a separate references text file could be a uniform convention across multiple codes.

eepeterson:
Thanks for weighing in @paulromano. Of your two suggestions I'd prefer the first over the second since it doesn't involve writing a separate file just for references which could easily get separated from the XML file that ultimately represents the one source of truth as to what materials make it into the simulation. I also don't particularly want to develop a cross code convention with respect to QA for materials that relies on a separate file. I'm probably still partial to a separate `comment` attribute rather than adding to the `name` attribute, but perhaps my view of what constitutes a material name is too narrow (i.e. doesn't include a reference.). Then again if, for example, you have compositions for SS316L from a number of manufacturers that vary slightly I could see people wanting the `name` attribute to be `name='SS316L | manufacturer 1, datasheet url'` rather than just `name='SS316L'`, but I can see it both ways. What I don't want to do is change the meaning of what the `name` attribute means to users by adding additional implied meaning that isn't clear from the name of the attribute itself. This is why I'm slightly in favor of the `.comment` addition, but am also happy to include it in the `name` field if the community prefers to leave it as is.

Shimwell:
Only problem I can see (perhaps just for OpenMC DAGMC users) with including long references in the .name field is that this field is used to link materials with the DAGMC tags so it is limited to 28 chars when using in DAGMC.

PR #1822, opened 2021-04-22
Very partial gpu merge to reduce rebase conflicts

gridley:
Hey everyone,

As per a discussion with @jtramm, we thought it may be helpful to partially merge some of the changes in my ongoing CUDA branch that will substantially reduce the amount of conflicts that occur when I need to rebase my branch on develop, and hopefully also help out John's GPU code.

This PR adds changes only three things, despite being sizable. Firstly, I changed up some of the lattice methods to consistently use `std::array` arguments, passed by const reference rather than value where possible. This made the code there cleaner as there is no longer a need to convert between C-style fixed length arrays, individual integer index variables, and `std::array`. It will additionally potentially reduce stack usage since the compiler will now decide whether it wants to pass by value, rather than us forcing it to.

Secondly, in the GPU branch, I have my own implementation of a few standard library containers for GPU that allow us to keep the same codebase between the CPU side and GPU side to a very large extent, if using CUDA. This entails creating our own `openmc::vector`, `openmc::unique_ptr`, `openmc::make_unique`, `openmc::array`, among a few others. Although what I listed there is by far the majority of places where such a replacement must be made.

So, in order to reduce the previously huge amount of manual merge conflicting I have to do on each rebase, I simply named some header files the same as they're named in my GPU branch, and put `using std::vector` statements and such inside `namespace openmc` so it looks just like my other branch.

Lastly, in order to portably switch between array-of-structures and structure-of-array particle data layout (which is performance critical for GPU code), I have incorporated a subset of my changes here in order to further reduce rebase conflicts. This simply entails separating the particle into a non-virtual base class that's responsible for defining the particle data layout (`ParticleData`), and correspondingly providing accessor methods to the underlying data. For instance, to get the particle energy, we use `p.E()` rather than `p.E_` here. I have measured this to introduce zero performance impact since it compiles to the same thing. From that base class providing the data layout and accessor implementations, the true `Particle` class inherits and provides the methods which do physical operations on the data.

In case anyone is curious, in order to implement structure-of-array mode, you simply have to write a base class for the particle which provides accessor methods to arrays of particle data. For instance, you'd have one big array of particle energies in event mode that this class will provide access to. That way, you can have the exact same physics code for both the structure-of-array and array-of-structure version. From what I understand, this is something that SHIFT was not able to achieve and ended up writing their GPU code to be totally separate from CPU and rely on separate functions for all the physics. I would include an implementation of that if it helped execution in CPU event mode at all, but it doesn't. We'll definitely only use structure-of-array for GPU code in the future.

Yes, this change is big and does not directly benefit the OpenMC community as a whole, and could potentially be confusing to new developers. I believe this will be worthwhile though, as it will allow greater productivity for me to sooner get CUDA support in a public version of OpenMC and benefit @jtramm by making a structure-of-array approach substantially easier to implement for other GPU programming models.


jtramm:
Thanks for putting this together Gavin! I'll be able to take a close look at this one early next week and will plan on submitting a review.

paulromano:
Likewise! I'll try to take a look early next week

gridley:
Seems I screwed something up on the latest rebase. Will fix it soon.

src/material.cpp
```diff
@@ -1144,9 +1144,9 @@ double sternheimer_adjustment(const std::vector<double>& f, const
   return rho;
 }
 
-double density_effect(const std::vector<double>& f, const std::vector<double>&
-  e_b_sq, double e_p_sq, double n_conduction, double rho, double E, double tol,
-  int max_iter)
+double density_effect(const vector<double>& f,
+  const std::vector<double>& e_b_sq, double e_p_sq, double n_conduction,
```
paulromano, line 1148: std::vector in here

src/math_functions.cpp
```diff
@@ -556,7 +556,7 @@ void calc_zn(int n, double rho, double phi, double zn[]) {
   // ===========================================================================
   // Calculate R_pq(rho)
   // Matrix forms of the coefficients which are easier to work with
-  std::vector<std::vector<double>> zn_mat(n + 1, std::vector<double>(n + 1));
+  vector<std::vector<double>> zn_mat(n + 1, std::vector<double>(n + 1));
```
paulromano, line 559: std:: in here

src/track_output.cpp
```diff
@@ -25,35 +25,35 @@ namespace openmc {
...
     hid_t file_id = file_open(filename, 'w');
     write_attribute(file_id, "filetype", "track");
     write_attribute(file_id, "version", VERSION_TRACK);
-    write_attribute(file_id, "n_particles", p.tracks_.size());
+    write_attribute(file_id, "n_particles", p.tracks().size());
     write_attribute(file_id, "n_coords", n_coords);
-    for (auto i = 1; i <= p.tracks_.size(); ++i) {
-      const auto& t {p.tracks_[i-1]};
+    for (auto i = 1; i <= p.tracks().size(); ++i) {
+      const auto& t {p.tracks()[i - 1]};
```
paulromano, line 56: Did you want to add a `tracks(int)` version as you were doing for other members?

include/openmc/particle_data.h
```diff
@@ -0,0 +1,453 @@
...
+enum class ParticleType { neutron, photon, electron, positron };
+
+//! Saved ("banked") state of a particle
+//! NOTE: This structure's MPI type is built in initialize_mpi() of
+//! initialize.cpp. Any changes made to the struct here must also be
+//! made when building the Bank MPI type in initialize_mpi().
+//! NOTE: This structure is also used on the python side, and is defined
+//! in lib/core.py. Changes made to the type here must also be made to the
+//! python defintion.
+struct ParticleBank {
```
paulromano, line 43: I've actually been wanting to separate out this type and rename it as well. I was thinking `SourceSite` might be a better indication of its use in most places. What do you think? It is the type that `Source::sample` returns after all.

include/openmc/particle_data.h
```diff
@@ -0,0 +1,453 @@
...
+  Direction last_dir_;
+#endif
+
+  int64_t n_progeny_ {0}; // Number of progeny produced by this particle
+
+public:
+  //==========================================================================
+  // Methods and accessors
+
+  NuclideMicroXS& neutron_xs(const int& i) { return neutron_xs_[i]; }
```
paulromano, line 286: I don't see any particular reason to pass a const reference instead of just passing by value, which is preferred for simple scalar types like `int` and `double`. Can you change this (and others) to be just `int i`?

include/openmc/particle_data.h
```diff
@@ -0,0 +1,453 @@
...
+//============================================================================
+//! Defines how particle data is laid out in memory
+//============================================================================
+
+class ParticleData {
+
+public:
+  ParticleData();
+
+private:
```
paulromano, line 172: Maybe we should make this `protected:` so that `Particle` still has access to all the members. The thing I don't like about calling, e.g., `alive()`, `material()`, etc. in the methods of `Particle` is that it becomes less clear whether those are accessor methods of the `Particle` class or just external functions. Using the data members directly (`alive_`, `material_`, etc.) immediately signals to the reader that those are data members of the current class. The alternative is to put `this->` in front of everything which gets busy (not in favor of that).

paulromano (changes requested):
Whew, that was a lot of files to look over... overall these are pretty simple changes. Can you elaborate on your decision to separate `ParticleData` from `Particle` itself?

include/openmc/particle_data.h
gridley: Ah, good point. CPP core guidelines agree with you on this. Moreover, they claim there is overhead in passing by const reference. I find that interesting, because I was told over the past summer that the compiler will choose to pass by value if it deems that's more efficient when we're doing pass by reference and don't end up modifying the parameter.

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

include/openmc/particle_data.h
gridley: Totally agree with you that this is less clear. Normally, I would not design a class to be using accessor methods to variables it owns either.

But, this ambiguity is the very reason for putting these as methods. In my structure-of-array code, you call `material()` as in the code in this PR, but it's instead going to access an array rather then return a member variable. So, I template on the base class to decide who provides a particle's energy, whether `E()` is defined to simply return a member variable or return a reference to a position in some array. In my view of it, the `Particle` class should do abstracted physical operations on pieces of data like the energy, position, etc. No need to know how the data is laid out in memory, just that it will do certain things with that data when the particle collides.

If we just accessed the member variables here, it would break compatibility with the SOA code, making us have separate `event_collide` and so on between the CPU and GPU code.

This is the best solution I could come up with. If you have other ideas, my ears are totally open. This is the most simple route that avoids having separate copies of the same code.

include/openmc/particle_data.h
gridley: Yes, I think that's a far more descriptive name. I'll definitely add that in. I like the guidance that if you use a struct outside of a class, it should probably not be embedded in that class.

src/track_output.cpp
gridley: Yeah, suppose so. This variable is unique, because it's the one variable in `Particle` I'm planning to absolutely not support on GPU. The track saving functionality relies on allocating an indefinite amount of memory on the fly. That's why I didn't provide an accessor like that.

Up to you if you'd like me to put that in!

src/math_functions.cpp
gridley: Ah yeah. Forgot the trailing `g` in `sed`. Thanks!

gridley:
> Whew, that was a lot of files to look over... overall these are pretty simple changes. Can you elaborate on your decision to separate `ParticleData` from `Particle` itself?

Hey Paul, yeah. This is because my GPU code currently looks at whether we're using a CUDA compiler, and if so, `Particle` inherits from `SOAParticleData` instead. This replaces the accessor methods with versions that index into arrays on the GPU, so we get the same physics code on CPU and GPU but can work with different data layouts.

Also, I think it's nice to bring the size of the `Particle` file down some. A nuclear engineer can look at the code and easily know what these methods are doing, even if they don't know exactly what `E()` does behind the scenes, it's probably the particle energy.

include/openmc/material.h
```diff
@@ -191,14 +192,14 @@ class Material
...
+double sternheimer_adjustment(const vector<double>& f,
+  const vector<double>& e_b_sq, double e_p_sq, double n_conduction,
+  double log_I, double tol, int max_iter);
 
 //! Calculate density effect correction
-double density_effect(const std::vector<double>& f, const std::vector<double>&
-  e_b_sq, double e_p_sq, double n_conduction, double rho, double E, double tol,
-  int max_iter);
+double density_effect(const vector<double>& f,
+  const std::vector<double>& e_b_sq, double e_p_sq, double n_conduction,
```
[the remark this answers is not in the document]
gridley: Indeed, thanks. Had not had to have it compile for GPU yet.

include/openmc/particle_data.h
```diff
@@ -0,0 +1,453 @@
+#ifndef OPENMC_PARTICLE_REPRESENTATION_H
+#define OPENMC_PARTICLE_REPRESENTATION_H
```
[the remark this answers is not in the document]
gridley: Thanks, good catch.

include/openmc/physics_mg.h
```diff
@@ -5,10 +5,9 @@
 #define OPENMC_PHYSICS_MG_H
 
 #include "openmc/capi.h"
-#include "openmc/particle.h"
 #include "openmc/nuclide.h"
-
-#include <vector>
+#include "openmc/particle.h"
+#include "openmc/vector.h"
```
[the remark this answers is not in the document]
gridley: indeed! Removed that

gridley:
OK, all the remaining std::vector mentions have been removed, and I resolved all the conversations that are addressed in the latest commit.

include/openmc/particle_data.h
gridley: OK, latest commit incorporates your suggestion

gridley:
I'll resolve the merge conflicts once we finish the other discussions.

src/track_output.cpp
paulromano: Nope, I'm fine leaving it out

include/openmc/particle_data.h
paulromano: That all makes complete sense @gridley. Thanks for taking the time to elaborate on the design decisions!

paulromano:
@gridley There are some uses of `Particle::Bank` in `docs/source/usersguide/settings.rst` in the section about custom sources. Can you update those to the new `SourceSite` name?

gridley:
> @gridley There are some uses of `Particle::Bank` in `docs/source/usersguide/settings.rst` in the section about custom sources. Can you update those to the new `SourceSite` name?

Hey Paul, yep, just did that. Thanks for pointing that out. Similarly, there were occurrences in the examples directory, which I've correspondingly replaced, as with `Particle::Type` which I've replaced with `ParticleType`.

And, just fixed up the merge conflicts, so the hopefully ready-to-go PR should be up soon.

paulromano (approved):

paulromano:
Thanks for addressing the outstanding issues @gridley. I'm satisfied on my end to merge at this point. @jtramm It would be good to get your thoughts before we move forward.

jtramm:
Starting my review, was just doing a search to make sure there were no cases of `std::vector`, `std::unique_ptr`, `std::make_unique`, or `std::array` left. Are all the below expected to still be using `std`, or should they be changed as well?

```
jtramm@skylake09:~/gavin_openmc/openmc/src$ grep -rin 'std::vector' ../include/openmc/ .
../include/openmc/mesh.h:377:                       std::vector<int>& bins) const override;
../include/openmc/mesh.h:529:                       std::vector<int>& bins) const override;
../include/openmc/vector.h:20:using std::vector;
./mesh.cpp:1819:                                    std::vector<int>& bins) const
./mesh.cpp:2207:                              std::vector<int>& bins) const
./volume_calc.cpp:245:          std::vector<int> buffer(2*q);
./volume_calc.cpp:264:        std::vector<int> buffer(2*q);

jtramm@skylake09:~/gavin_openmc/openmc/src$ grep -rin 'std::make_unique' ../include/openmc/ .
../include/openmc/memory.h:15:using std::make_unique;
./cell.cpp:1002:    model::cells.push_back(std::make_unique<CSGCell>(cell_node));
./cell.cpp:1021:      model::universes.push_back(std::make_unique<Universe>());
./cell.cpp:1288:    model::cells.push_back(std::make_unique<CSGCell>());
./cross_sections.cpp:239:        data::thermal_scatt.push_back(std::make_unique<ThermalScattering>(
./dagmc.cpp:214:      model::universes.push_back(std::make_unique<Universe>());
./distribution_spatial.cpp:68:    r_ = std::make_unique<Discrete>(x, p, 1);
./distribution_spatial.cpp:79:    phi_ = std::make_unique<Discrete>(x, p, 1);
./distribution_spatial.cpp:90:    z_ = std::make_unique<Discrete>(x, p, 1);
./distribution_spatial.cpp:132:    r_ = std::make_unique<Discrete>(x, p, 1);
./distribution_spatial.cpp:143:    theta_ = std::make_unique<Discrete>(x, p, 1);
./distribution_spatial.cpp:154:    phi_ = std::make_unique<Discrete>(x, p, 1);
./endf.cpp:87:    func = std::make_unique<Tabulated1D>(dset);
./endf.cpp:89:    func = std::make_unique<Polynomial>(dset);
./endf.cpp:91:    func = std::make_unique<CoherentElasticXS>(dset);
./endf.cpp:93:    func = std::make_unique<IncoherentElasticXS>(dset);
./geometry_aux.cpp:160:            univ->partitioner_ = std::make_unique<UniversePartitioner>(*univ);
./initialize.cpp:77:  settings::libmesh_init = std::make_unique<libMesh::LibMeshInit>(argc, argv, comm, n_threads);
./initialize.cpp:80:  settings::libmesh_init = std::make_unique<libMesh::LibMeshInit>(argc, argv, 0, n_threads);
./lattice.cpp:1058:    model::lattices.push_back(std::make_unique<RectLattice>(lat_node));
./lattice.cpp:1061:    model::lattices.push_back(std::make_unique<HexLattice>(lat_node));
./material.cpp:568:  ttb_ = std::make_unique<Bremsstrahlung>();
./material.cpp:1248:    model::materials.push_back(std::make_unique<Material>(material_node));
./material.cpp:1478:    model::materials.push_back(std::make_unique<Material>());
./mesh.cpp:1369:      model::meshes.push_back(std::make_unique<RegularMesh>());
./mesh.cpp:1371:      model::meshes.push_back(std::make_unique<RectilinearMesh>());
./mesh.cpp:1392:    model::meshes.push_back(std::move(std::make_unique<MOABMesh>(mesh_file)));
./mesh.cpp:1399:    model::meshes.push_back(std::move(std::make_unique<LibMesh>(mesh_file)));
./mesh.cpp:1586:  mbi_ = std::make_unique<moab::Core>();
./mesh.cpp:1646:  kdtree_ = std::make_unique<moab::AdaptiveKDTree>(mbi_.get());
./mesh.cpp:2151:  m_ = std::make_unique<libMesh::Mesh>(*settings::libmesh_comm, n_dimension_);
./mesh.cpp:2163:  equation_systems_ = std::make_unique<libMesh::EquationSystems>(*m_);
./mesh.cpp:2381:      model::meshes.push_back(std::make_unique<RegularMesh>(node));
./mesh.cpp:2383:      model::meshes.push_back(std::make_unique<RectilinearMesh>(node));
./mesh.cpp:2386:      model::meshes.push_back(std::make_unique<MOABMesh>(node));
./mesh.cpp:2390:      model::meshes.push_back(std::make_unique<LibMesh>(node));
./nuclide.cpp:203:      reactions_.push_back(std::make_unique<Reaction>(rx_group, temps_to_read));
./nuclide.cpp:1050:    data::nuclides.push_back(std::make_unique<Nuclide>(group, temperature));
./nuclide.cpp:1082:        data::elements.push_back(std::make_unique<PhotonInteraction>(group));
./reaction_product.cpp:72:      distribution_.push_back(std::make_unique<UncorrelatedAngleEnergy>(dgroup));
./reaction_product.cpp:74:      distribution_.push_back(std::make_unique<CorrelatedAngleEnergy>(dgroup));
./reaction_product.cpp:76:      distribution_.push_back(std::make_unique<NBodyPhaseSpace>(dgroup));
./reaction_product.cpp:78:      distribution_.push_back(std::make_unique<KalbachMann>(dgroup));
./settings.cpp:428:      model::external_sources.push_back(std::make_unique<FileSource>(path));
./settings.cpp:438:      model::external_sources.push_back(std::make_unique<CustomSourceWrapper>(path, parameters));
./settings.cpp:440:      model::external_sources.push_back(std::make_unique<IndependentSource>(node));
./settings.cpp:455:    model::external_sources.push_back(std::make_unique<FileSource>(path));
./settings.cpp:460:    model::external_sources.push_back(std::make_unique<IndependentSource>(
./surface.cpp:1057:        model::surfaces.push_back(std::make_unique<SurfaceXPlane>(surf_node));
./surface.cpp:1060:        model::surfaces.push_back(std::make_unique<SurfaceYPlane>(surf_node));
./surface.cpp:1063:        model::surfaces.push_back(std::make_unique<SurfaceZPlane>(surf_node));
./surface.cpp:1066:        model::surfaces.push_back(std::make_unique<SurfacePlane>(surf_node));
./surface.cpp:1069:        model::surfaces.push_back(std::make_unique<SurfaceXCylinder>(surf_node));
./surface.cpp:1072:        model::surfaces.push_back(std::make_unique<SurfaceYCylinder>(surf_node));
./surface.cpp:1075:        model::surfaces.push_back(std::make_unique<SurfaceZCylinder>(surf_node));
./surface.cpp:1078:        model::surfaces.push_back(std::make_unique<SurfaceSphere>(surf_node));
./surface.cpp:1081:        model::surfaces.push_back(std::make_unique<SurfaceXCone>(surf_node));
./surface.cpp:1084:        model::surfaces.push_back(std::make_unique<SurfaceYCone>(surf_node));
./surface.cpp:1087:        model::surfaces.push_back(std::make_unique<SurfaceZCone>(surf_node));
./surface.cpp:1090:        model::surfaces.push_back(std::make_unique<SurfaceQuadric>(surf_node));
./tallies/filter.cpp:78:  auto filter = std::make_unique<T>();
./tallies/tally.cpp:311:  model::tallies.push_back(std::make_unique<Tally>(id));
./tallies/tally.cpp:740:    model::tallies.push_back(std::make_unique<Tally>(node_tal));
./tallies/tally.cpp:919:    model::tallies.push_back(std::make_unique<Tally>(-1));
./thermal.cpp:207:      elastic_.distribution = std::make_unique<CoherentElasticAE>(*xs);
./thermal.cpp:210:        elastic_.distribution = std::make_unique<IncoherentElasticAE>(dgroup);
./thermal.cpp:213:        elastic_.distribution = std::make_unique<IncoherentElasticAEDiscrete>(
./thermal.cpp:235:      inelastic_.distribution = std::make_unique<IncoherentInelasticAE>(dgroup);
./thermal.cpp:238:      inelastic_.distribution = std::make_unique<IncoherentInelasticAEDiscrete>(
./wmp.cpp:255:  nuc->multipole_ = std::make_unique<WindowedMultipole>(group);

jtramm@skylake09:~/gavin_openmc/openmc/src$ grep -rin 'std::array' ../include/openmc/ .
../include/openmc/version.h.in:10:constexpr std::array<int, 3> VERSION {VERSION_MAJOR, VERSION_MINOR, VERSION_RELEASE};
../include/openmc/array.h:15:using std::array;
```

gridley:
Hey John, thanks, indeed, I missed those after the rebase as I have never compiled with libmesh or dagmc. And you're right that I completely forgot to remove the `std::` from make_unique.

gridley:
OK @jtramm, got those final changes in.

jtramm:
One downside to refactoring the lattice code to use std::array is that for device code, we now have to implement a device version of that container. If it was refactored the other way to just be C-style static arrays, e.g., `int arr[3]`, then would the code work on both host and device without further changes? What is the advantage in this use case to using std::array?

gridley:
> One downside to refactoring the lattice code to use std::array is that for device code, we now have to implement a device version of that container. If it was refactored the other way to just be C-style static arrays, e.g., `int arr[3]`, then would the code work on both host and device without further changes? What is the advantage in this use case to using std::array?

Because it's trivial to implement in device code using static arrays, and if we have it, we can use std::arrays along with iterators and such elsewhere in the code. Here's my CUDA implementation.

```
#pragma once
#include <array>

#ifdef __CUDACC__
template<typename T, unsigned long Size>
class array {
public:
  array() = default;
  __host__ __device__ T& operator[](unsigned indx) { return data[indx]; }
  __host__ __device__ T const& operator[](unsigned indx) const
  {
    return data[indx];
  }

  __host__ __device__ array<T, Size>& operator=(array<T, Size> const& other)
  {
    for (unsigned i = 0; i < Size; ++i) {
      data[i] = other.data[i];
    }
    return *this;
  }
  __host__ __device__ array(array<T, Size> const& other)
  {
    for (unsigned i = 0; i < Size; ++i) {
      data[i] = other.data[i];
    }
  }

  __host__ __device__ array(std::initializer_list<T> list)
  {
    unsigned i = 0;
    for (auto const& val : list)
      data[i++] = val;
  }

  __host__ operator std::array<T, Size>() const
  {
    std::array<T, Size> result;
    for (unsigned i = 0; i < Size; ++i) {
      result[i] = data[i];
    }
    return result;
  }

private:
  T data[Size];
};
#else
using std::array;
#endif

```

jtramm:
I guess the way I see it is that C++ already has a non-STL language mechanism for expressing static arrays that works seamlessly on both host and device with **zero** code delta and **regardless of device programming model**. That advantage of simplicity greatly outweighs anything offered by the std::array container, in my opinion. However, I definitely don't feel strongly enough about it to request that you go back through and refactor everything in the other direction, as that really doesn't seem like it would be a good use of time. So, I'll happily accept the device array container strategy.

jtramm:
It looks like there is still a `std::array` being used after the latest commit. Is this intended?

```
jtramm@jlselogin4:~/gavin_openmc/openmc/src$ grep -rin 'std::array' ../include/openmc/. .
../include/openmc/./version.h.in:10:constexpr std::array<int, 3> VERSION {VERSION_MAJOR, VERSION_MINOR, VERSION_RELEASE};
../include/openmc/./array.h:15:using std::array;
```

gridley:
> I guess the way I see it is that C++ already has a non-STL language mechanism for expressing static arrays that works seamlessly on both host and device with **zero** code delta and **regardless of device programming model**. That advantage of simplicity greatly outweighs anything offered by the std::array container, in my opinion. However, I definitely don't feel strongly enough about it to request that you go back through and refactor everything in the other direction, as that really doesn't seem like it would be a good use of time. So, I'll happily accept the device array container strategy.

I see what you're saying. Yeah, a static array would have been fine. It's just that there are other places where we use an `std::array`, so even if it creates a code delta in the lattice code, it removes code deltas in other places.

gridley:
> It looks like there is still a `std::array` being used after the latest commit. Is this intended?
> 
> ```
> jtramm@jlselogin4:~/gavin_openmc/openmc/src$ grep -rin 'std::array' ../include/openmc/. .
> ../include/openmc/./version.h.in:10:constexpr std::array<int, 3> VERSION {VERSION_MAJOR, VERSION_MINOR, VERSION_RELEASE};
> ../include/openmc/./array.h:15:using std::array;
> ```

Yes, constexpr std::array should work on both host and device using `--expt-host-device-constexpr`. Then we don't have to implement a constexpr std::array which is a bit different from a normal static array, I think.

jtramm:
I was able to confirm this works for OpenMP offloading as well, so this sounds good to leave as it is.

jtramm (commented):
Looks pretty good -- **thanks again @gridley for doing all the heavy lifting on this refactor!** This effort will really make life easier going forward for the GPU ports!

I would also note that if we were going to do this PR all over again, it might have been much easier/quicker to review if all the changes had been broken out into independent PRs. This PR combined several massive find & replace all changes with a number of other independent smaller scale logic changes. Plus there were also a number of small undocumented editorial changes sprinkled in throughout. It would have been nice to just review the find/replace PRs with grep without having to look at each line manually and the look through more closely at the smaller manual PRs that had the logic changes. This "omnibus" style PR combined to really create extra work for reviewers as the combined nature of it meant that every line touched needed to be looked at, even though probably 95% were just find/replace all. I found that there are some github tools to help focus on single commits or a range of commits, but there are some limitations with that strategy that meant I still usually ended up looking at the full PR at once.

Anyway, as there will likely to be more GPU refactor PRs down the line, we should try to make them as focused as is reasonably practical, even if it means submitting several PRs in parallel. In particular, mixing of simpler find/replace all refactors and small scale logic/editorial changes should be avoided if at all possible.

Thanks again for getting this one knocked out and reducing the host-device code delta again!

src/particle_restart.cpp
```diff
@@ -93,33 +94,27 @@ void run_particle_restart()
...
   // Force calculation of cross-sections by setting last energy to zero
   if (settings::run_CE) {
-    for (auto& micro : p.neutron_xs_) micro.last_E = 0.0;
+    p.invalidate_neutron_xs();
   }
 
   // Prepare to write out particle track.
-  if (p.write_track_) add_particle_track(p);
-
-  // Every particle starts with no accumulated flux derivative.
```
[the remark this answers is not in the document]
gridley: Ah, I see what you're talking about. I put this stuff in the `ParticleData` constructor. It's because the logic for this looks different between the SOA and non-SOA particle. That way we keep the common stuff outside the particle class, and the uncommon stuff restricted to the `ParticleData` class.

Sorry to not include an explainer on this. You're right that doing multiple smaller PRs would have made documenting changes like this easier. Will definitely do that in the future.

include/openmc/lattice.h
```diff
@@ -234,14 +225,9 @@ class RectLattice : public Lattice
   void to_hdf5_inner(hid_t group_id) const;
 
 private:
-  std::array<int, 3> n_cells_;    //!< Number of cells along each axis
+  array<int, 3> n_cells_;         //!< Number of cells along each axis
   Position lower_left_;           //!< Global lower-left corner of the lattice
   Position pitch_;                //!< Lattice tile width along each axis
-
-  // Convenience aliases
-  int &nx {n_cells_[0]};
```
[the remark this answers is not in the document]
gridley: So, yeah... This screws up device code. It's because we can construct a reference to stuff on the host (probably implemented as a pointer?), but then later run in device code. If you just copy data directly to the device, that can cause a segfault. Removing these convenience aliases doesn't add too much inconvenience since it's something like 50 extra characters total as a result of removing these.

src/geometry.cpp
```diff
@@ -369,48 +366,47 @@ BoundaryInfo distance_to_boundary(Particle& p)
...
+  array<int, 3> level_lat_trans {};
 
   // Loop over each coordinate level.
-  for (int i = 0; i < p.n_coord_; i++) {
-    const auto& coord {p.coord_[i]};
-    Position r {coord.r};
-    Direction u {coord.u};
+  for (int i = 0; i < p.n_coord(); i++) {
+    const auto& coord {p.coord(i)};
+    const Position& r {coord.r};
```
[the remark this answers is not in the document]
gridley: Ah yeah... my thinking here is that we could possibly reduce register pressure by doing const references rather than keeping separate copied values. With a const reference, the compiler should be able to choose between storing a const pointer and keeping a copied value.

src/simulation.cpp
```diff
@@ -464,74 +463,61 @@ void initialize_history(Particle& p, int64_t index_source)
...
   // Force calculation of cross-sections by setting last energy to zero
   if (settings::run_CE) {
-    for (auto& micro : p.neutron_xs_) micro.last_E = 0.0;
+    p.invalidate_neutron_xs();
   }
 
   // Prepare to write out particle track.
-  if (p.write_track_) add_particle_track(p);
-
-  // Every particle starts with no accumulated flux derivative.
```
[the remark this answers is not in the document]
gridley: Yeah, again, they're in the constructor for `ParticleData` now since this code differs between SOA and non-SOA.

gridley:
Hey John,

Thanks for the further review. I definitely see your point of view on having someone throw a wall of code at you and effectively forcing you to take the time to review it. It would be nice to guarantee one PR is only a find-and-replace and then whatever clang-format did.

Anyway, thanks so much for being responsive on this PR. Really hope it'll actually help you out on the OpenMP side.

In the future, I'll definitely follow the advice on submitting divvied up PRs. I think that luckily, this is the biggest delta between CPU and CUDA, though.

src/geometry.cpp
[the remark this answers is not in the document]
jtramm: Makes sense! I'm not sure if this would really make any difference in practice after compiler optimization, but definitely doesn't hurt anything either. For future refactoring PRs, might be best to cull changes like this that are independent from the PR topic and put into a separate optimization PR.

src/simulation.cpp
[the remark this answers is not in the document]
jtramm: I can see that this change works fine for history-based mode, where each particle is constructed and destroyed each power iteration. However, in event-based mode, the particles persist between iterations such that the constructors are only called once per simulation, not once per iteration. Does this cause issues?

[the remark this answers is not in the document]
gridley: Yeah, so, look at the constructor of ParticleData. It's in there. It specifically checks if we're in event mode, then sets things up so that it'll be good for the rest of the simulation. As for the track writing thing directly near your comment here, it's still in initialize_history, just not in inititalize_history_partial, which seemed to only be used in one place and had data-layout-dependent code so I just lumped what isn't now in the ctor of ParticleData into initialize_history.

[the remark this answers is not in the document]
jtramm: The way it currently is in OpenMC, the particle's `flux_derivs_` are set to zero when `initialize_history()` is called (i.e., every batch). By moving that to the particle constructor, the `flux_derivs_` now only get set to zero once at program initialization in event-based mode, not once per iteration. Does this cause issues?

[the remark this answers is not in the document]
gridley: Nope. There is now a method called `zero_flux_derivs`. It gets called in Particle's `from_source`, so the behavior should be the same.

jtramm:
Looks good to me -- thanks again @gridley!

jtramm:
Two last small issues I spotted:

1. The struct name change from `Particle::Bank` to `SourceSite` did not propagate into the MPI section or python. The name for that MPI type was left as `bank` in `message_passing.h`, and as `_Bank` in `lib/core.py`. The code is still correct, but I think it's a little confusing to have the same type called very different things in the different areas of the code. Probably better to make the switch in the MPI areas as well as on the python side.
2. I think it would be helpful to better document (via comments) why `PaticleData` and `Particle` classes are not combined. As there is no reason for them to be that way without the context of different AOS/SOA implementations, someone might come along in the future and try to combine them again. Adding some comments in their headers might help prevent that.

jtramm:
In line with (2) above, it might also be wise to add a comment on the `ParticleData` type for why accessors are defined everywhere. Someone in the future similarly might want to eliminate those, so would be good to have the AOS/SOA device etc motivation for them in comments.

gridley:
Sounds good, will address these later today!

gridley:
OK, latest commit should address those last points

include/openmc/particle_data.h
```diff
@@ -164,6 +164,32 @@ struct BoundaryInfo {
 //! Defines how particle data is laid out in memory
 //============================================================================
 
+/*
+ * This class was added in order to separate the layout and access of particle
+ * data from particle physics operations during a development effort to get
+ * OpenMC running on GPUs. In the event-based Monte Carlo method, one creates
+ * an array of particles on which actions like cross section lookup and surface
+ * crossing are done en masse, which works best on vector computers of yore and
+ * modern GPUs. It has been shown in the below publication that arranging
```
jtramm, line 173: The comment mentions "the below publication," but I'm not seeing where the publication is listed.

gridley: ah, good catch. Putting it in now...

gridley:
OK @jtramm, added that paper reference that I missed!

PR #1824, opened 2021-04-26
fix typo in fmt string for a lattice error

gridley:
Currently, if you make a certain mistake while creating lattices, you get this from OpenMC:

```
 Reading settings XML file...
 Reading cross sections XML file...
 Reading materials XML file...
 Reading geometry XML file...
terminate called after throwing an instance of 'fmt::v6::format_error'
  what():  invalid format string
zsh: abort (core dumped)  openmc
```

This PR fixes that. Now, we get the correct error message:

```
 Reading settings XML file...
 Reading cross sections XML file...
 Reading materials XML file...
 Reading geometry XML file...
 ERROR: Expected 16 universes for a rectangular lattice of size 4x4x1 but 9 were
        specified.

```


paulromano (approved):
Nice find, thanks!

PR #1800, opened 2021-03-22
Mcf example using paramak cad

Shimwell:
This PR is an attempt to add a magnetic confinement fusion example to the collection of Jupyter notebook examples

This attempts to duplicate the style used in @pshriwise DagMC [example notebook] (https://github.com/openmc-dev/openmc/blob/develop/examples/jupyter/cad-based-geometry.ipynb)

The dagmc.h5m model was made using the Paramak and the details of that model creation can be found [here](https://github.com/Shimwell/fusion_example_for_openmc_using_paramak/) The url downloads the dagmc.h5m model from this repo but I notice that the DagMc example used a tiny url so perhaps this can be changed if preferred.

Minimal tallies have been added (TBR cell and tritium production mesh)

PR #1825, opened 2021-04-27
DAGMC CAD-based geometry as universes

pshriwise:
We currently only support DAGMC models where the entire geometry is represented as CAD-based tesselations. This PR adds a DAGMC Universe derived from the standard CSG universe. This allows one to construct geometries from both CAD and CSG. This enables repeated CAD geometries in OpenMC like this lattice of DAGMC pincells from our regression tests

dag_lattice

Or, as a more interesting case, one can rotate a section of a tokamak into a set of CSG cells to model the entire device with a smaller DAGMC memory footprint. (Thanks @Shimwell for providing this quarter-tokamak model as a demonstration.)

Here is the entire tokamak model in Cubit:

<p style="text-align:center;">

</p>

We can add this model as a fill to the appropriate cells and rotate the model 90 degrees to get this X-Y view:

tokamak_ex

Adding two more cells and rotations gives use the entire device:

Screenshot from 2021-04-27 00-08-30

## Input Changes

These updates do require some changes to the current format when the DAGMC model represents the entire geometry. A `geometry.xml` file is now always required. To represent a geometry composed of a single DAGMC model as we do now, a `geometry.xml` file that gives the same behavior should contain

```XML
<geometry>
<dagmc id="1" filename="dagmc.h5m"/>
</geometry>
```

where the `id` is the universe ID and the filename is the name of the DAGMC model file. 

**Note that the `filename` parameter remoes the restriction that DAGMC files have the name `dagmc.h5m`.**

## Notable code changes

  - Addition of the `DAGUniverse` class
  - Addition of `Cell::geom_type` and `Surface::geom_type`
  - Removal of the `settings::dagmc` parameter from the CAPI and Python API
  - Added a `Universe::find_cell` method
  - Fewer `#ifdef DAGMC`'s throughout the code thanks to the addition of the `GeometryType` enum

## ID management

One difficulty of mixing CSG and CAD geometry is the unknown ID spaces of the CAD geometry before simulation initialization. The `DAGUniverse` class on the Python side has an option, `auto_geom_ids`, to automatically set cell and surface IDs outside of the CSG cell & surface space to avoid conflicts. If this option is not enabled, then an error will occur if the ID spaces overlap.

A similar option, `auto_mat_ids`, is also available for materials to address the case where some of the DAGMC universes are using UWUW materials (material definitions embedded in the `.h5m` along with the geometry.

### Coincident DAGMC/CSG surfaces

Thanks to our preference for hitting surface at higher levels in the geometry, this is hasn't been a problem in the models I've used for testing.

### Managing a particle's ray history

The ray history is a record of facets hit by DAGMC rays to avoid duplicate hits when coincident with a surface or reflecting from a surface. This history is reset when

  - a particle changes direction or undergoes a collision,
  - a particle is not on a surface,
  - and a particle is terminated.
 
Now that a particle can move directly from one DAGMC universe to another, another case is added for

  - a particle exiting the current `DAGUniverse`

For more details on this implementation, please see [this google doc](https://docs.google.com/document/d/1Jbg-8VZhxBXTMxbz1r85GgaFgLBYIlkXbvvXLUhV_K0/edit?usp=sharing) w/ the proposal for this work.


makeclean:
Ooh very exciting :)

gridley:
In case anyone else is wondering, this seems to add pretty much zero overhead to non-DAGMC models. Seems like a really useful upgrade.

pshriwise:
Sorry for the initial CI failures here. Needed to make a correction to the properties of the `DAGUniverse` Python class. This should be good to try out now!

Shimwell:
@pshriwise is it worth me adding a rotation to this example https://github.com/openmc-dev/openmc/pull/1800 . I assume that I just remove the reflecting surface (perhaps using the new dagmc_remove_tags package) and rotate 90, 180 and 270 degrees to make the complete geometry

pshriwise:
> @pshriwise is it worth me adding a rotation to this example #1800 . I assume that I just remove the reflecting surface (perhaps using the new [dagmc_remove_tags](https://github.com/svalinn/remove_dagmc_tags) package) and rotate 90, 180 and 270 degrees to make the complete geometry

I'd welcome it! It would make #1800 dependent on this PR though. I'll leave it up to you as to whether or not you're ok with waiting on these changes to get that example in.

src/dagmc.cpp
```diff
@@ -37,189 +36,137 @@ const bool DAGMC_ENABLED = false;
...
-      throw std::runtime_error{"Error reading UWUW materials"};
-    }
+DAGUniverse::DAGUniverse(const std::string& filename,
+                         bool auto_geom_ids,
+                         bool auto_mat_ids)
+: filename_(filename), adjust_geometry_ids_(auto_geom_ids), adjust_material_ids_(auto_mat_ids) {
+  // determine the next universe id
+  int32_t next_univ_id = 0;
+  for (const auto& u : model::universes) {
+    if (u->id_ > next_univ_id) next_univ_id = u->id_;
```
makeclean, line 76: will `u->id_` always be sorted smallest to largest?

src/dagmc.cpp
```diff
@@ -37,189 +36,137 @@ const bool DAGMC_ENABLED = false;
...
 
     // set cell ids using global IDs
-    DAGCell* c = new DAGCell();
-    c->dag_index_ = i+1;
-    c->id_ = model::DAG->id_by_index(3, c->dag_index_);
-    c->dagmc_ptr_ = model::DAG;
-    c->universe_ = dagmc_univ_id; // set to zero for now
+    auto c = std::make_unique<DAGCell>();
+    c->dag_index_ = i + 1;
+    c->id_ = adjust_geometry_ids_ ? next_cell_id++ : dagmc_instance_->id_by_index(3, c->dag_index_);
```
makeclean, line 155: say i tally in cell 3, which is a dagmc cell, do we accumulate tally scores across all the instances of that cell individually, or would I get a score for each 'copy'?

src/dagmc.cpp
```diff
@@ -279,39 +232,32 @@ void load_dagmc_geometry()
   }
 
   if (!graveyard) {
-    warning("No graveyard volume found in the DagMC model."
+    warning("No graveyard volume found in the DagMC model. "
```
makeclean, line 235: Do you need a graveyard even in a universe based model? Do OpenMC cells not bound this?

src/dagmc.cpp
```diff
@@ -279,39 +232,32 @@ void load_dagmc_geometry()
...
 
     // set BCs
     std::string bc_value = DMD.get_surface_property("boundary", surf_handle);
     to_lower(bc_value);
     if (bc_value.empty() || bc_value == "transmit" || bc_value == "transmission") {
-      // Leave the bc_ a nullptr
+      // set to transmission by default (nullptr)
     } else if (bc_value == "vacuum") {
       s->bc_ = std::make_shared<VacuumBC>();
     } else if (bc_value == "reflective" || bc_value == "reflect" || bc_value == "reflecting") {
```
makeclean, line 259: is this sort of thing not handled by DMD? I thought it dealt with synonyms?

src/dagmc.cpp
```diff
@@ -331,24 +277,364 @@ void load_dagmc_geometry()
...
+
+  // generate a string representation of the ID range(s)
+  std::stringstream out;
+
+  int i = 0;
+  int start_id = id_vec[0];
+  int stop_id;
+  while (i < n_cells) {
+    stop_id = id_vec[i];
+
```
makeclean, line 315: quite comment light, i presume you've making a list like 1-100, 101, 105-109?

src/dagmc.cpp
```diff
@@ -279,39 +232,32 @@ void load_dagmc_geometry()
...
-
+  int n_surfaces = dagmc_instance_->num_entities(2);
   for (int i = 0; i < n_surfaces; i++) {
-    moab::EntityHandle surf_handle = model::DAG->entity_by_index(2, i+1);
+    moab::EntityHandle surf_handle = dagmc_instance_->entity_by_index(2, i+1);
 
     // set cell ids using global IDs
-    DAGSurface* s = new DAGSurface();
+    auto s = std::make_unique<DAGSurface>();
     s->dag_index_ = i+1;
```
makeclean, line 248: maybe a note here - reminder dagmc indices are 1 based

tests/regression_tests/rotation/test.py
```diff
@@ -1,6 +1,40 @@
-from tests.testing_harness import TestHarness
+import openmc
```
makeclean, line 1: was this file changed on purpose? seems unrelated to the DAGMC based changes? maybe im wrong?

makeclean (changes requested):
Looks mostly good here, I think a few extra new lines, some extra comments and its good :)

Shimwell:
Looks like code block 17 from the cad-based-geometry.ipynb has the old ```settings.dagmc = True``` method  

https://github.com/pshriwise/openmc/blob/5b20fbbf6a377b16d201a34c5cf65471662ac7cc/examples/jupyter/cad-based-geometry.ipynb#L524

include/openmc/surface.h
```diff
@@ -30,7 +30,7 @@ namespace model {
 } // namespace model
 
 //==============================================================================
-//! Coordinates for an axis-aligned cuboid bounding a geometric object.
+//! Coordinates for an axis-aligned cube that bounds a geometric object.
```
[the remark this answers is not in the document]
pshriwise: Nope. Definitely not. Unsure as to how that came about. Thanks!

src/dagmc.cpp
pshriwise: It isn't a requirement for `model::universes`, no. Are you concerned the next ID isn't correct?

src/dagmc.cpp
pshriwise: When using a `CellFilter` you'll get the accumulated score over all the instances of that cell in the geometry. If you use a `CellInstanceFilter` filter or a `DistribCellFilter` you can get for each instance of the cell in the geometry.

src/dagmc.cpp
pshriwise: The only case where you now need one is if the DAGMC model is the root universe of the Geometry. I'll add some detection for that so this warning doesn't appear when it shouldn't. Thanks!

src/dagmc.cpp
pshriwise: I believe synonyms are for the property type (`mat:`, `surf:`, `rho:`) as opposed to the value. The logic here is matches the surface BC logic [here](https://github.com/openmc-dev/openmc/blob/4399645859ce5459828721f12fbcaf052a63653a/src/surface.cpp#L137). I thought it made sense for these to be consistent.

src/dagmc.cpp
pshriwise: Correct. I'll add some more explanation here.

tests/regression_tests/rotation/test.py
pshriwise: Ah, no you're right. This is the result of a mishandled rebase conflict. I'll sort it out.

src/dagmc.cpp
makeclean: just wondering if it leads to any issues, like does it matter if the order isnt monotonically increasing?

Shimwell:
I've been using the branch on an ARC reactor model and have reproduced the example pictures with a different geometry (still 90 degree rotated 4 times).

I've found the ability to turn of the auto assign id numbers useful
It is nice to have a geometry.xml file when using DAGMC
It is great that this branch allows flexability with the dagmc h5m file (previously only dagmc.h5m was accepted).

Actually this feature really saved me and was exactly what I needed to get the simulation results.

Perhaps this is unrelated to this PR but one thing I noticed was that the terminal prints out a lot of (perhaps too much) information when plotting an DAGMC geometry using this method

```
import os
# makes the 3d "cube" style geometry
vox_plot = openmc.Plot()
vox_plot.type = 'voxel'
vox_plot.width = (1700., 1700., 1800.)
vox_plot.pixels = (300, 300, 300)
vox_plot.filename = 'plot_3d_tokamak'
vox_plot.color_by = 'material'
plots = openmc.Plots([vox_plot])
plots.export_to_xml()
openmc.plot_geometry()
os.system('openmc-voxel-to-vtk plot_3d_tokamak.h5 -o plot_3d_tokamak.vti')
```

pshriwise:
> import os
> # makes the 3d "cube" style geometry
> vox_plot = openmc.Plot()
> vox_plot.type = 'voxel'
> vox_plot.width = (1700., 1700., 1800.)
> vox_plot.pixels = (300, 300, 300)
> vox_plot.filename = 'plot_3d_tokamak'
> vox_plot.color_by = 'material'
> plots = openmc.Plots([vox_plot])
> plots.export_to_xml()
> openmc.plot_geometry()
> os.system('openmc-voxel-to-vtk plot_3d_tokamak.h5 -o plot_3d_tokamak.vti')

Yeah, I'm seeing that too. Thanks for sharing this. Appears to be a warning about a corner case in MOAB's triangle intersection routine. I'll have to look into that separately.

pshriwise:
> Looks like code block 17 from the cad-based-geometry.ipynb has the old `settings.dagmc = True` method

Thanks! Just took care of that one.

src/dagmc.cpp
pshriwise: The IDs can be in any order here. Whenever we find one that's bigger, we update `next_univ_id`. If we happened to find the largest ID first, we'd only update that variable once.

pshriwise:
Got this updated after the merge of #1834.

Shimwell (approved):
I've been using this branch for a while now and not noticed any problems. My C++ skills are not really good enough to understand all the changes. But from a user point of view the docs, examples and usage is all fine.

openmc/universe.py
```diff
@@ -600,3 +627,112 @@ def _determine_paths(self, path='', instances_only=False):
...
+        # Initialize class attributes
+        self.filename = filename
+        self.auto_geom_ids = auto_geom_ids
+        self.auto_mat_ids = auto_mat_ids
+
+    def __repr__(self):
+        fmt_str = '{: <16}=\t{}\n'
+        string = super().__repr__()
+        string += fmt_str.format('\tGeom', 'DAGMC')
+        string += fmt_str.format('\tFile', self.filename)
```
paulromano, line 683: Defining `fmt_str` upfront actually doesn't save any work. In the case where you repeat the literal twice, Python's compiler is smart enough to recognize that it is the same object. By setting it as a various like this, you are actually giving the interpreter more work to do.
```suggestion
        string = super().__repr__()
        string += '{: <16}=\t{}\n'.format('\tGeom', 'DAGMC')
        string += '{: <16}=\t{}\n'.format('\tFile', self.filename)
```

openmc/universe.py
```diff
@@ -600,3 +627,112 @@ def _determine_paths(self, path='', instances_only=False):
...
+    def auto_mat_ids(self):
+        return self._auto_mat_ids
+
+    @auto_mat_ids.setter
+    def auto_mat_ids(self, val):
+        cv.check_type('DAGMC automatic material ids', val, bool)
+        self._auto_mat_ids = val
+
+    def clone(self, clone_materials=True, clone_regions=True, memo=None):
+        pass
```
paulromano, line 714: Should this be a warning or something?

src/particle.cpp
```diff
@@ -377,6 +381,12 @@ Particle::cross_surface()
     int64_t idx = simulation::surf_source_bank.thread_safe_append(site);
   }
 
+  // if we're crossing a CSG surface, make sure the DAG history is reset
+  #ifdef DAGMC
+  if (surf->geom_type_ != GeometryType::DAG) history().reset();
+  #endif
+
+
```
paulromano, line 389:
```suggestion
```

src/surface.cpp
```diff
@@ -198,21 +197,22 @@ Surface::diffuse_reflect(Position r, Direction u, uint64_t* seed) const
...
 {
-  std::string group_name {"surface "};
-  group_name += std::to_string(id_);
+  hid_t surf_group = create_group(group_id, fmt::format("surface {}", id_));
 
-  hid_t surf_group = create_group(group_id, group_name);
+  if (geom_type_ == GeometryType::DAG) {
+    write_string(surf_group, "geom_type", "dagmc", false);
+  }
+  else if (geom_type_ == GeometryType::CSG) {
```
paulromano, line 208:
```suggestion
  } else if (geom_type_ == GeometryType::CSG) {
```

tests/regression_tests/dagmc/legacy/test.py
```diff
@@ -26,6 +26,10 @@ def model():
 
     model.settings.dagmc = True
 
+    # geometry
+    dag_univ = openmc.DAGMCUniverse("dagmc.h5m")
+    model.geometry = openmc.Geometry(root=dag_univ)
```
paulromano, line 31:
```suggestion
    model.geometry = openmc.Geometry(dag_univ)
```

tests/regression_tests/dagmc/refl/test.py
```diff
@@ -27,13 +27,18 @@ def _build_inputs(self):
 
         model.settings.export_to_xml()
 
+        # geometry
+        dag_univ = openmc.DAGMCUniverse("dagmc.h5m", auto_geom_ids=True)
+        model.geometry = openmc.Geometry(root=dag_univ)
```
paulromano, line 32:
```suggestion
        model.geometry = openmc.Geometry(dag_univ)
```

openmc/universe.py
```diff
@@ -600,3 +627,112 @@ def _determine_paths(self, path='', instances_only=False):
...
+
+    def create_xml_subelement(self, xml_element, memo=None):
+        if memo and self in memo:
+            return
+
+        if memo is not None:
+            memo.add(self)
+
+        # Set xml element values
+        dagmc_element = ET.Element('dagmc')
```
paulromano, line 730: Maybe make this 'dagmc_universe'?

tests/regression_tests/rotation/inputs_true.dat
```diff
@@ -0,0 +1,31 @@
+<?xml version='1.0' encoding='utf-8'?>
```
paulromano, line 1: Doesn't look like this file is used? The rotation test is just a normal (non Python API) test.

tests/regression_tests/dagmc/universes/test.py
```diff
@@ -0,0 +1,71 @@
...
+        left = openmc.XPlane(x0=-24.0, name='left', boundary_type='reflective')
+        right = openmc.XPlane(x0=24.0, name='right', boundary_type='reflective')
+        front = openmc.YPlane(y0=-24.0, name='front', boundary_type='reflective')
+        back = openmc.YPlane(y0=24.0, name='back', boundary_type='reflective')
+        bottom = openmc.ZPlane(z0=-20.0, name='bottom', boundary_type='vacuum')
+        top = openmc.ZPlane(z0=20.0, name='top', boundary_type='vacuum')
+
+        box = +left & -right & +front & -back & +bottom & -top
+
+        bounding_cell = openmc.Cell(fill=pincell_univ, region=box)
```
paulromano, line 56: Might be more exciting to put this DAGMC pincell in a lattice so that it appears multiple times?

tests/regression_tests/dagmc/universes/test.py
```diff
@@ -0,0 +1,71 @@
...
+
+        model.geometry = openmc.Geometry(root=[bounding_cell])
```
paulromano, line 58:
```suggestion
        model.geometry = openmc.Geometry([bounding_cell])
```

paulromano (changes requested):
Really awesome feature @pshriwise! Looking forward to having this as part of our next release.

src/dagmc.cpp
pshriwise: Are you happy with this implementation @makeclean?

include/openmc/cell.h
```diff
@@ -118,7 +121,9 @@ class Cell {
 
   //! Write all information needed to reconstruct the cell to an HDF5 group.
   //! \param group_id An HDF5 group id.
-  virtual void to_hdf5(hid_t group_id) const = 0;
+  virtual void to_hdf5(hid_t group_id) const;
```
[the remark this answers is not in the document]
pshriwise: For sure. Thanks!

include/openmc/dagmc.h
```diff
@@ -9,24 +8,114 @@ extern "C" const bool DAGMC_ENABLED;
...
+  virtual bool find_cell(Particle &p) const override;
+
+  // Data Members
+  std::string filename_; //!< Name of the DAGMC file used to create this universe
+  std::shared_ptr<moab::DagMC> dagmc_instance_; //!< DAGMC Instance for this universe
+  std::shared_ptr<UWUW> uwuw_; //!< Pointer to the UWUW instance for this universe
+  int32_t cell_idx_offset_; //!< An offset to the start of the cells in this universe in OpenMC's cell vector
+  int32_t surf_idx_offset_; //!< An offset to the start of the surfaces in this universe in OpenMC's surface vector
+  bool adjust_geometry_ids_; //!< Indicates whether or not to automatically generate new cell and surface IDs for the universe
+  bool adjust_material_ids_; //!< Indicates whether or not to automatically generate new material IDs for the universe
```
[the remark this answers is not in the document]
pshriwise: Made private. Updated constructor to take those as parameters. Added accessors.

openmc/universe.py
pshriwise: I need to think about this a bit more. I don't see any reason why we can't clone a `DAGMCUniverse`. I wouldn't necessarily recommend it, but it should be possible.

tests/regression_tests/rotation/inputs_true.dat
pshriwise: Thanks! looks like this was leftover from my update to the new `develop` history after getting rid of that `plot_settings.pkl` file.

src/dagmc.cpp
pshriwise: Reorganized the code a bit so we can perform this check in 302f12b5ffba88c78ce2cb39722e76ade9672735. Hopefully this is what you had in mind!

tests/regression_tests/dagmc/universes/test.py
pshriwise: Updated
 in 263addf

paulromano:
@pshriwise Thanks for incorporating all the feedback. I have a few more requests:
1. The new `<dagmc_universe>` element needs to be documented in `docs/source/io_formats/geometry.rst`
2. The `geom_type` attribute for surfaces needs to be documented in `docs/source/io_formats/summary.rst`
3. It looks like a geometry that has a DAGMC universe does not "round trip" correctly to and from XML. That is, if you `geom.export_to_xml()` and then `new_geom = openmc.Geometry.from_xml(...); new_geom.export_to_xml('new_geom.xml')`, the `new_geom.xml` file is not the same as the original.

openmc/surface.py
```diff
@@ -453,13 +453,19 @@ def from_hdf5(group):
 
         """
 
+        # If this is a DAGMC surface, do nothing for now
+        geom_type = group['geom_type'][()].decode()
```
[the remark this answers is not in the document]
pshriwise: Addressed this in f2e0af3.

pshriwise:
> @pshriwise Thanks for incorporating all the feedback. I have a few more requests:
> 
> 1. The new `<dagmc_universe>` element needs to be documented in `docs/source/io_formats/geometry.rst`
> 2. The `geom_type` attribute for surfaces needs to be documented in `docs/source/io_formats/summary.rst`
> 3. It looks like a geometry that has a DAGMC universe does not "round trip" correctly to and from XML. That is, if you `geom.export_to_xml()` and then `new_geom = openmc.Geometry.from_xml(...); new_geom.export_to_xml('new_geom.xml')`, the `new_geom.xml` file is not the same as the original.

Thanks for  @paulromano! Docs have all been updated.

I've updated the export/import code for XML so that it handles DAGMC universes properly and can reproduce the same `geometry.xml` reliably. I also decided that we should really be able to reproduce the `geometry.xml` from a `summary.h5` file as well, so I've added a `DAGUniverse::to_hdf5` method that writes out all of the necessary information to reproduce the Python object (and corresponding XML element if desired). I've updated the `summary.h5` documentation to reflect these changes as well.

I noticed along the way that surface names weren't being picked up in `openmc.Surface.from_xml_element`, so I took care of that in 25cf7a3e9712f80b05e918311fbd5e65cfd6d627.

docs/source/io_formats/geometry.rst
```diff
@@ -370,3 +370,44 @@ Here is an example of a properly defined 2d hexagonal lattice:
...
+----------------------------
+``<dagmc_universe>`` Element
+----------------------------
+
+Each ``<dagmc_universe>`` element can have the following attributes or sub-elements:
+
+  :id:
+    A unique integer used to identify the universe.
+
+    *Default*: ""
```
paulromano, line 386:
```suggestion
    *Default*: None
```

docs/source/io_formats/geometry.rst
```diff
@@ -370,3 +370,44 @@ Here is an example of a properly defined 2d hexagonal lattice:
...
+
+  :name:
+    An optional string name to identify the surface in summary output
+    files. This string is limited to 52 characters for formatting purposes.
+
+    *Default*: ""
```
paulromano, line 392:
```suggestion
    *Default*: None
```

docs/source/io_formats/geometry.rst
```diff
@@ -370,3 +370,44 @@ Here is an example of a properly defined 2d hexagonal lattice:
...
+Each ``<dagmc_universe>`` element can have the following attributes or sub-elements:
+
+  :id:
+    A unique integer used to identify the universe.
+
+    *Default*: ""
+
+  :name:
+    An optional string name to identify the surface in summary output
+    files. This string is limited to 52 characters for formatting purposes.
```
paulromano, line 390: As we discussed offline, this limit is probably historical and no longer applies (unless you've found something that indicates we are indeed limited to 52 characters)

docs/source/io_formats/geometry.rst
```diff
@@ -370,3 +370,44 @@ Here is an example of a properly defined 2d hexagonal lattice:
...
+    *Default*: false
+
+  :filename:
+    A required string indicating the file to be loaded representing the DAGMC universe.
+
+    *Default*: None
+
+
+
+
```
paulromano, line 413:
```suggestion
```

openmc/universe.py
```diff
@@ -600,3 +625,163 @@ def _determine_paths(self, path='', instances_only=False):
...
+
+        out = cls(fname, universe_id=id, name=name)
+
+        out.auto_geom_ids = bool(group.attrs['auto_geom_ids'])
+        out.auto_mat_ids = bool(group.attrs['auto_mat_ids'])
+
+        return out
+
+    @classmethod
+    def from_xml(cls, elem):
```
paulromano, line 761: For consistency, let's name this `from_xml_element`

openmc/universe.py
```diff
@@ -600,3 +625,163 @@ def _determine_paths(self, path='', instances_only=False):
...
+        if memo and self in memo:
+            return
+
+        if memo is not None:
+            memo.add(self)
+
+        # Set xml element values
+        dagmc_element = ET.Element('dagmc_universe')
+        dagmc_element.set('id', str(self.id))
+        dagmc_element.set('name', self.name)
```
paulromano, line 726: Do we really want to allow these universes to be named? Right now, although you can give a regular universe a name, nothing is actually done with it since we don't write a `<universe>` element in the XML. At the very least, if the `name` is not assigned, we shouldn't create an empty `name=""` attribute in the XML.

docs/source/io_formats/geometry.rst
paulromano, line 413:
```suggestion
```

paulromano (changes requested):
Thanks for the latest round of updates. A few more requests based on those changes... we're getting close!

openmc/universe.py
pshriwise: Good point. I was realizing this too looking at the import/export XML methods yesterday. I think it's probably best to make the behavior of the name attribute as similar to the regular universe as possible.

docs/source/io_formats/geometry.rst
pshriwise: Nope! I'll go ahead and update other instances of this as well.

openmc/universe.py
pshriwise: One instance where it would be nice to have the universe name is when looking at different levels of the geometry in the plotter so we can display universe names in the status bar. I think it can be dealt with in a separate PR if we feel it's important enough to get those names into the XML files though.

openmc/geometry.py
```diff
@@ -159,6 +159,11 @@ def get_universe(univ_id):
         for s1, s2 in periodic.items():
             surfaces[s1].periodic_surface = surfaces[s2]
 
+        # Add any DAGMC universes
+        for elem in root.findall('dagmc_universe'):
+            dag_univ = openmc.DAGMCUniverse.from_xml(elem)
```
paulromano, line 164:
```suggestion
            dag_univ = openmc.DAGMCUniverse.from_xml_element(elem)
```

paulromano (changes requested):

paulromano (approved):

paulromano:
@makeclean Do you have any further comments on this PR?

makeclean (approved):

makeclean:
LGTM thanks @pshriwise

PR #1835, opened 2021-05-09
Rename "name: dockerhub-publish-develop-dagmc.yml"

isaac-gs:
Cannot clone on windows (easily) due to file name.


Shimwell:
Thanks for correcting the filename. Not quite sure how it ended up with the "name: " at the front but I agree it is better without that part.

PR #1838, opened 2021-05-13
Re-add DAGMC Version Check

pshriwise:
The check for a DAGMC version >= 3.2.0 was accidentally removed in #1748. This PR re-adds that check.


paulromano (approved):
Thanks @pshriwise!

PR #1840, opened 2021-05-25
OpenMC Weight Windows

makeclean:
Picking up where the KIT folks left off. I've refactored the excellent WW work done by them to inherit from the appropriate mesh classes, reworked some of the algorithms and modified so a separate weigh_windows.xml file is needed. In terms of structure, when a WW file exists, it is read into a map of ww parms by particle type and then later accessed by key. I'm at a natural stopping point, where the code functionally works, I think its got some nice code re-use, but needs a little more tidying. 

Im currently running through some validation cases to make sure the different edge cases dont lead to biassing and so on. But there are a  few open questions that I would appreciate some opinions on. 

1. The current weight_window.xml file looks something like this.
```
<weight_windows>
  <mesh id="1" type="rectilinear">
    <x_grid>0 10 20 30 40 50 60 70 80 90 100</x_grid>
    <y_grid>-50 50</y_grid>
    <z_grid>-50 50</z_grid>
  </mesh>
  <particle type="neutron">
    <parameters>
      <upper_bound>5</upper_bound>
      <survival>3</survival>
      <max_split>10</max_split>
      <multiplier>1</multiplier>
    </parameters>
    <energy>
      1e-50 20.0e6
    </energy>
    <lower_ww>
      2.5000E-01
      1.2500E-01
      6.2500E-02
      3.1250E-02
      1.5625E-02
      7.8125E-03
      3.9063E-03
      1.9531E-03
      9.7656E-04
      4.7656E-04
    </lower_ww>
  </particle>
  <particle type="photon">
    <parameters>
      <upper_bound>5</upper_bound>
      <survival>3</survival>
      <max_split>10</max_split>
      <multiplier>1</multiplier>
    </parameters>
    <energy>
    </energy>
    <lower_ww>
      2.5000E-01
      1.2500E-01
      6.2500E-02
      3.1250E-02
      1.5625E-02
      7.8125E-03
      3.9063E-03
      1.9531E-03
      9.7656E-04
      5.7656E-04
    </lower_ww>
  </particle>
</weight_windows>
```
So as you can see, each particle type will have its own set of VR parameters and set of weight window values. I would appreciate, some clarity if we think the format is adequate? Particularly I was thinking the mesh section could instead be cells, and voila you have a bunch of cell based WW's rather than mesh based. There is no reason, why there couldnt be multiple ww sections, a differnent mesh for each particle type, or even cells for some particle types and meshes for others. Once we've reached some agreement, I'll then crack on with the python interface. There is also an optional <upper_ww> section like the lower one, where you could set a completely different upper weight in each mesh region, this is akin to how other MC codes (other than MCNP) do it, so I think thats a useful addition, e.g. the bounds may start quite wide in the first few MFP, then get smaller as we go deeper into the model. 
2. Unit testing - its not particularly clear how I should unit test this? My thoughts are for a simple enough model that runs quickly enough, then maybe use something like the Kolmogorov–Smirnov test on the two flux or particle spectra?
3. I also need to write an mcnp2openmc ww script, but that can come later I think. 
4. Also, should the wwmesh be dumped into the summary.h5 file or the statepoint?


makeclean:
Here are a few examples of runs, a cube of side 1 m, made of Nickel at nominal density, point 14.08 MeV source (1,0,0) cm. You will see results from a no-VR run 1e8 particles/batch 10 batches and a run with WW 1e6 particles/batch 10 batches.

cell_1
Basically near the source
cell_5
Halfway through the block
cell_10
At the rear of the block

src/physics.cpp
```diff
@@ -1160,4 +1166,65 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+  // Check if this particle is in the weight weindow mesh and get the mesh bins in each direction
+  int ijk[3] = {0};     // mesh bin in each direction
+  bool in_mesh;  
+  // this is likely expensive 
+  // would like a more generic wrapper in case of cell based
+  // ww's
+  settings::ww_settings->mesh_->get_indices(pos, ijk, &in_mesh);
+  // no vr if not inside the mesh
+  if (!in_mesh) {
+    std::cout << "particle not in mesh" << std::endl;
```
gridley, line 1190: This should probably have a more descriptive message, and should instead use an `openmc::warning` instead. Would be good to say we're talking about the VR mesh.

include/openmc/settings.h
```diff
@@ -56,7 +58,11 @@ extern bool ufs_on;                   //!< uniform fission site method on?
 extern bool urr_ptables_on;           //!< use unresolved resonance prob. tables?
 extern bool write_all_tracks;         //!< write track files for every particle?
 extern bool write_initial_source;     //!< write out initial source file?
+extern bool weightwindow_on;          //!< use weight window or not
 
+// weight window   
+extern std::shared_ptr<weight_window::WeightWindow>  ww_settings;  //!< weight window settings
```
gridley, line 64: This makes more sense as a `unique_ptr`. No other `shared_ptr` is getting created that refers to this, especially since this is at global scope. Moreover, you can remove the `weightwindow_on` setting. Rather, we can just check if `ww_settings` is a nullptr and assume weight windows are not turned on if so.

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+// given a location tell the particle the right bounds
+ParticleWeightParams WeightWindow::get_params(Particle& p) const
+{	
+  // Particle's position and energy
+  Position pos  = p.r();
+  double E = p.E();
+
+  // get the settings for the weight window
+  ParticleType type = p.type();
+  WWParams ww_settings = weight_params.at(type);
```
gridley, line 256: `WWParams` potentially contains a fairly large amount of data (the std::vectors). You're making a separate copy of it here that will be expensive. I think you should store a const reference instead.

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+    if (!result) {                                                                          
+      fatal_error("Error processing weight_windows.xml file.");                                   
+    }
+    // Get root element                                                                     
+    xml_node root = doc.document_element();     
+    std::cout << " Reading weight window XML file ..." << std::endl;
+    // try and read a weight window
+    settings::ww_settings = std::make_shared<weight_window::WeightWindow>(root);
+  } else {
+    settings::ww_settings = std::make_shared<weight_window::WeightWindow>();
```
gridley, line 33: Just leave `settings::ww_settings` as a `nullptr` if no weight windows are being used.

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,72 @@
...
+  
+  // source weight biasing in energy
+  bool user_defined_biasing;      //!< use user difined weight or not
+  std::vector<double> biasing_energy;   //!< energy group for weight biasing
+  std::vector<double> origin_probability; //!< probability for each group
+  std::vector<double> cumulative_probability; //!< cumulative probability for each group
+  std::vector<double> biasing;   //!< biasing for each energy group
+  std::vector<double> cumulative_biasing;   //!< cumulative probability for biasing for each energy group
+ 
+  // ptr for the structure mesh
```
gridley, line 65: This comment doesn't really add much.

docs/source/io_formats/settings.rst
```diff
@@ -999,3 +999,170 @@ sub-elements/attributes:
      sample points within.
 
      *Default*: None
+     
+-------------------------
+``<weightwindow>`` Element
+-------------------------
+
+The ``<weightwindow>`` element specifies all necessary parameters for weight 
+window mesh function.This element has the following sub-elements/attributes:
```
gridley, line 1008: Should have a space after the period

docs/source/io_formats/settings.rst
```diff
@@ -999,3 +999,170 @@ sub-elements/attributes:
...
+
+      *Default*: 5
+      
+    :multiplier:
+      The multiplier for all lower weight window bounds.
+
+      *Default*: 1
+
+  :user_defined_biasing:
+     The usder defined weight biasing in energy.
```
gridley, line 1159: typo

include/openmc/mesh.h
```diff
@@ -271,7 +272,7 @@ class RectilinearMesh : public StructuredMesh
   // Constructors
   RectilinearMesh() = default;
   RectilinearMesh(pugi::xml_node node);
-
+
```
gridley, line 275: remove added whitespace please. Did you apply clang-format? May want to do that.

include/openmc/mesh.h
```diff
@@ -346,6 +347,11 @@ class UnstructuredMesh : public Mesh {
   void to_hdf5(hid_t group) const override;
 
   // Data members
+  xt::xtensor<int, 1> shape_; //!< Number of mesh elements in each dimension
```
gridley, line 350: If you know this will always be an array of length 3, you should make it an `std::array<int, 3>` to avoid indirection.

include/openmc/physics.h
```diff
@@ -87,6 +87,10 @@ void sample_fission_neutron(int i_nuclide, const Reaction& rx, double E_in,
 void inelastic_scatter(const Nuclide& nuc, const Reaction& rx, Particle& p);
 
 void sample_secondary_photons(Particle& p, int i_nuclide);
+  
+//！Split or Roulette particles based their weight and the lower weight window bound.
```
gridley, line 91: It appears you've added some weird unicode explanation mark. Would be best to replace it with good ol' ascii

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,72 @@
+#ifndef OPENMC_WEIGHT_WINDOWS_H
+#define OPENMC_WEIGHT_WINDOWS_H 1
+
+#include "openmc/mesh.h"
+#include "openmc/particle.h"
+
+namespace openmc {
+
+void read_weight_window_xml(); //!< read the weight window file
```
gridley, line 9: The comment here is not really adding much, since it is basically the same as the name of the function. Maybe you could name the specific variables it fills.

openmc/mesh.py
```diff
@@ -854,4 +854,505 @@ def from_xml_element(cls, elem):
...
         library = get_text(elem, 'library')
 
-        return cls(filename, library, mesh_id)
+        mesh = cls(filename, mesh_id)
+
+        return mesh
+
+  
+class WeightWindowMesh:
+    """A three dimension Weight Window Mesh class
```
gridley, line 863: Might want to say "dimensional" and not uppercase here, unless you're trying to do CamelCase, and if that is so, remove the spaces between the words.

openmc/mesh.py
```diff
@@ -854,4 +854,505 @@ def from_xml_element(cls, elem):
...
+		self._xmesh = None
+		self._xints = None
+		self._ymesh = None
+		self._yints = None
+		self._zmesh = None
+		self._zints = None
+		
+		self._n_energy_group = None
+		self._p_energy_group = None
+        self._lower_ww = None
```
gridley, line 957: Is the indentation here correct?

openmc/settings.py
```diff
@@ -199,6 +199,9 @@ class Settings:
         described in :ref:`verbosity`.
     volume_calculations : VolumeCalculation or iterable of VolumeCalculation
         Stochastic volume calculation specifications
+
```
gridley, line 202: Should the newline be there?

src/mesh.cpp
```diff
@@ -1305,7 +1308,7 @@ void RectilinearMesh::to_hdf5(hid_t group) const
 
   close_group(mesh_group);
 }
-
+
```
gridley, line 1311: is this a clang-format artifact? Should probably remove trailing whitespace

src/mesh.cpp
```diff
@@ -5,6 +5,9 @@
 #include <string>
 #include <gsl/gsl>
 
+#include <sstream>
+#include <fstream>
```
gridley, line 9: Are these new headers being used? I don't see other changes in this file that use them.

src/physics.cpp
```diff
@@ -1160,4 +1166,65 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+  }
+
+  // get the paramters 
+  auto params = settings::ww_settings->get_params(p);
+	
+  // if particles weight is above the weight window 
+  // split until they are within the window
+  if (weight > params.upper_weight) {
+    double number = weight/params.upper_weight;  
+    double num = std::min(std::ceil(number), double(params.max_split));
```
gridley, line 1201: use a static_cast rather than C-style cast

src/physics.cpp
```diff
@@ -1160,4 +1166,65 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+
+    for (int l = 0; l < num - 1; ++l)  { 
+      p.create_secondary(weight/num, p.u(), p.E(), p.type()); 
+    }
+    p.wgt() = weight/num; 
+    p.wgt_last() = p.wgt();
+	// if the particle weight is below the window
```
gridley, line 1208: align the comments here please

src/physics.cpp
```diff
@@ -1160,4 +1166,65 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+    }
+    p.wgt() = weight/num; 
+    p.wgt_last() = p.wgt();
+	// if the particle weight is below the window  
+  // roulette until the weight is high enough
+  } else if (weight <= params.lower_weight) {  
+    double number = weight/params.survival_weight;
+    if (number < 1.0/params.max_split) {
+      number = 1.0/params.max_split;
+      params.survival_weight = weight/number;
```
gridley, line 1214: Why modify the `params`? It seems you use it later, but you could instead just do `p.wgt() = weight/number` later.

src/physics.cpp
```diff
@@ -1160,4 +1166,65 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+
+    for (int l = 0; l < num - 1; ++l)  { 
+      p.create_secondary(weight/num, p.u(), p.E(), p.type()); 
+    }
+    p.wgt() = weight/num; 
+    p.wgt_last() = p.wgt();
+	// if the particle weight is below the window  
+  // roulette until the weight is high enough
+  } else if (weight <= params.lower_weight) {  
+    double number = weight/params.survival_weight;
```
gridley, line 1211: Should change `number` to a more descriptive variable name.

src/settings.cpp
```diff
@@ -416,8 +420,8 @@ void read_settings_xml()
   if (check_for_node(root, "threads")) {
     if (mpi::master) warning("The <threads> element has been deprecated. Use "
       "the OMP_NUM_THREADS environment variable to set the number of threads.");
-  }
-
+  }
```
gridley, line 423: whitespace changes intentional?

src/source.cpp
```diff
@@ -213,20 +213,23 @@ SourceSite IndependentSource::sample(uint64_t* seed) const
                   "one cross section table");
     }
   }
+  
+  // source weight biasing in energy
+  if (settings::weightwindow_on && settings::ww_settings->user_defined_biasing)  { 
+    settings::ww_settings->weight_biasing(site, seed); 
+  } else {
+    // origin code
```
gridley, line 221: the comment here doesn't make sense. Should perhaps expand on what you meant

src/source.cpp
```diff
@@ -213,20 +213,23 @@ SourceSite IndependentSource::sample(uint64_t* seed) const
...
-
+      // Resample if energy falls outside minimum or maximum particle energy
+      if (site.E < data::energy_max[p] && site.E > data::energy_min[p]) break;
+    }
+  } 
+  
   // Set delayed group
   site.delayed_group = 0;
-  // Set surface ID
-  site.surf_id = 0;
```
gridley, line 228: why remove setting the surf_id?

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+{
+  // default parameters
+  n_ww = false;                  // flag for neutron use weight window
+  p_ww = false;                  // flag for photon use weight window
+  user_defined_biasing = false;  // flag for weight biasing in energy
+
+  // upper weight window = upper_ratio * lower weight window
+  weight_params[ParticleType::neutron].upper_ratio = 5.; 
+  // survival weight = survival_ratio * lower weight window  
+  weight_params[ParticleType::neutron].survival_ratio = 3.;
```
gridley, line 53: Surely there is some way to remove the redundant code here. My recommendation is to have the default constructor of whatever's stored in `weight_params` use these reasonable defaults. While it is a `struct` and not a `class`, it would remove the need for all this code that's repeated (especially if we ever start moving other particles).

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+  // multiplier for weight window lower bounds
+  weight_params[ParticleType::photon].multiplier = 1.0; 
+
+}
+
+WeightWindow::WeightWindow(pugi::xml_node node)
+{
+  using namespace pugi;
+
+  WeightWindow(); // call the default constructor
```
gridley, line 74: Don't do this here. It's not actually doing what you think -- this is creating a `WeightWindow` object here that never gets used or assigned to any variable. What you want is a "delegating constructor", please look it up.

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+    xml_node mesh = node.child("mesh");
+    mesh_ = make_unique<RectilinearMesh>(mesh);
+  }
+
+  // make sure we have a particle section
+  if (check_for_node(node, "particle")) {
+    // read the particle section
+    std::string particle_name = "neutron";
+    openmc::ParticleType particle_type;
+    if ( particle_name == "neutron") {
```
gridley, line 87: Please don't use these `if` statements. Do a `switch-case` using `openmc::str_to_particle_type` from `particle.h`.

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+        fatal_error("Survival to lower weight window ratio must bigger than 1"
+                  "and less than the upper to lower weight window ratio.");   
+     
+    // get the max split
+    settings.max_split = std::stod(get_node_value(parameters, "max_split"));
+    if(settings.max_split <= 1) fatal_error("max split must be larger than 1");
+     
+    // multiplier
+    settings.multiplier = std::stod(get_node_value(parameters, "multiplier"));
+    if(settings.multiplier <= 0) fatal_error("multiplier must be larger than 0");
```
gridley, line 139: definitely need to apply `clang-format` here

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+  if (check_for_node(node, "upper_ww")) {
+    settings.upper_ww = get_node_array<double>(node, "upper_ww");  
+    // ensure there are enough entries
+    if (settings.upper_ww.size() != 
+      mesh_->n_bins()*(settings.energy_bounds.size()-1)) {
+         fatal_error("not enough entries in the upper_ww section");
+    }
+  }
+
+  // maybe multply out a vector for the upper bounds for consistency
```
gridley, line 172: spelling. Also, not really sure what you're saying here.

src/weight_windows.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+  return settings;
```
gridley, line 174: I don't think you're going to get copy elision here, although I may be wrong. You should find a way to initialize this `settings` variable in-place rather than copying like this.

gridley (changes requested):
Hey Andrew, nice! It's going to be great to get this into OpenMC. I just left a variety of minor comments on the code like typos to fix, comments that could use expansion, etc.

I think we should at some point look at the tally variance reported by MCNP versus the variance reported by OpenMC if using weight windows, since then we'll know if we're seeing the expected benefits of the new methods.

src/physics.cpp
```diff
@@ -1160,4 +1166,61 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+    }
+   
+    if (prn(p.current_seed()) <= n_split)  {
+      p.wgt() = params.survival_weight;
+      p.wgt_last() = p.wgt();
+    } else {       
+      p.alive() = false;
+      p.wgt() = 0.0;
+      p.wgt_last() = p.wgt();
+    }
```
paulromano, line 1220:
```suggestion
    double n_split = std::max(1/params.max_split, weight/params.survival_weight);
   
    if (prn(p.current_seed()) <= n_split)  {
      p.wgt() /= n_split;
    } else {       
      p.alive() = false;
      p.wgt() = 0.0;
    }
    p.wgt_last() = p.wgt();
```

src/weight_windows.cpp
```diff
@@ -0,0 +1,314 @@
...
+//! source weight biasing in energy
+void WeightWindow::weight_biasing(SourceSite& site, uint64_t* seed) 
+{
+  int i = 0;
+  double r = prn(seed);
+  for (i = 0; i < cumulative_biasing.size() - 1; i++) 
+    if (cumulative_biasing[i] <= r && r < cumulative_biasing[i+1])  break;
+  site.E = biasing_energy[i] + ( biasing_energy[i+1] - biasing_energy[i] ) * prn(seed);
+  site.wgt *= origin_probability[i+1] / biasing[i+1];
+}
```
paulromano, line 255: Source weight biasing is a separate thing from weight windows -- I feel like this shouldn't be part of the `WeightWindow` class and should instead be more intrinsically linked to source definitions.

src/weight_windows.cpp
```diff
@@ -0,0 +1,314 @@
...
+  params.upper_weight = ww_settings.upper_ratio*
+                         params.lower_weight;
+  // set the survival weight
+  params.survival_weight = params.lower_weight*
+                          ww_settings.survival_ratio;
+  // set the max split
+  params.max_split = ww_settings.max_split;
+
+  // set the weight cutoff
+  params.weight_cutoff = ww_settings.weight_cutoff;
```
paulromano, line 309: We could probably make this a bit cleaner by just having a constructor `ParticleWeightParams(ww_settings, indices)`

paulromano (changes requested):
Thanks so much for picking this one up @makeclean! Regarding the format for the new weight_windows.xml file, I think ideally the structure should match how the data is stored within objects (both at the Python and C++ level), i.e. each nested object would correspond to a nesting of XML elements. I would say just give a first shot at what you think makes sense and we can always tweak it during review.

paulromano:
Sorry forgot to respond to your other questions. Regarding a unit test, yeah it's sort of difficult but what you proposed sounds reasonable if you can get something like that to work. The wwmesh data can be dumped into the summary.h5 file rather than the statepoint.

makeclean:
FYI as a validation case.

neutron_mesh

OpenMC with VR - 30 MPI tasks 56 threads - 10 batches 1e8 histories - 1.0264e+04 seconds
OpenMC without VR - 30 MPI tasks 56 threads - 10 batches 1e9 histories - 4.8436e+03 seconds

src/weight_windows.cpp
```diff
@@ -0,0 +1,314 @@
...
+{	
+  // Particle's position and energy
+  Position pos  = p.r();
+  double E = p.E();
+
+  // get the settings for the weight window
+  ParticleType type = p.type();
+
+  // 
+  WWParams ww_settings;
```
gridley, line 269: This is a potentially big class since it stores `std::vector`s. Making a copy of it here is perhaps a source of the slowdown relative to normal OpenMC tracking you're seeing.

gridley:
OK, here's what I think we should do to ensure we're getting good performance before merging this. We should take a normal OpenMC test case, perhaps the `asym_lattice` or something, and apply a weight window mesh such that the particle weights never need to be adjusted and the particles take the same trajectories as in analog Monte Carlo. We should see negligible overhead of the weight windows, in that case. Given the current design of the code, that will possibly not be the case since large data structures are unnecessarily copied in a few places. Replacing with pointers or references should help.

makeclean:
You always see a slow down (in time per history) when applying VR as opposed to not. Any one history takes longer since typically, this is because we split 'important' histories so they end up getting deeper into the model. In this case the attenuation across the model is like 1e6, which means that if you've got a good weight map, ultimately your one source history is split eventually, in this case upto 1e6 times. Now the map used in this case is GVR (e.g. MAGIC, FW-CADIS), global variance reduction, where you try to reduce the variance uniformly across the geometry. In other cases, of regular target based VR (e.g. CADIS), histories are rouletted when they are unimportant which reduces the overall CPU time, where you might (but only by chance) see parity of run times. So longer run times, is just something we have to live with, its also variable upon the quality of the weight window mesh, in this case I've spent some time making sure its 'about right' if anything its slightly under performing and in reality, a slowdown per history of 50x to 100x would be good for a model of this depth. In full fusion geometries, we're talking about attenuation factors of 1e12 or more, where entire histories with the full GVR based weight map and lots of splitting will take wall clock seconds per history. In terms of performance, you may well be right that a shift to pointers will be more efficient, but I think I saw parity in runtimes with and without VR (where the VR was on but no splitting was happening, but I will perform this more scientifically with that timing in mind.

pshriwise:
Some thoughts on the `weight_windows.xml` file:

  - I also mentioned this in the inline comments, but how would we feel about giving the new XML file a more generic name like `variance_reduction.xml`? Any other variance reduction schemes we add could be grouped there in the future. Otherwise, as new methods appear we'll return to this question of adding them to the settings file or to a new XML.
  - Right now it appears that a single mesh can be specified. I think it would be useful to have the `<particle>` node reference a mesh ID, allowing different particle types to have different mesh types/structures.

src/physics.cpp
```diff
@@ -1160,4 +1166,55 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+    p.wgt() = weight/n_split; 
+    p.wgt_last() = p.wgt();
+   
+    // if the particle weight is below the window  
+    // roulette until the weight is high enough
+  } else if (weight <= params.lower_weight) {  
+    double n_split = std::max(1./params.max_split, weight/params.survival_weight);
+
+    if (prn(p.current_seed()) <= n_split)  {
+      p.wgt() /= n_split;
```
pshriwise, line 1209: I'm thinking we can reduce the number of division operations here:

```suggestion
    double n_split = std::min(params.max_split, params.survival_weight/weight);

    if (n_split*prn(p.current_seed()) > 1)  {
      p.wgt() *= n_split;
```

I think it also is a little more clear that we're increasing the weight in this situation. Does this cause any roundoff issues though?

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,81 @@
...
+    // set the survival weight
+    survival_weight = lower_weight*ww_settings.survival_ratio;
+    // set the max split
+    max_split = ww_settings.max_split;
+
+    // set the weight cutoff
+    weight_cutoff = ww_settings.weight_cutoff;  
+  }
+
+  double lower_weight;
```
pshriwise, line 48: We'll want to add some doc strings to these data members.

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,81 @@
...
+  double weight_cutoff;
+  std::vector<double> energy_bounds;
+  std::vector<double> lower_ww;
+  std::vector<double> upper_ww;
+};
+
+// struct to pass specific location data 
+struct ParticleWeightParams {
+  // 0,1,0.5,1,1e-12
+  ParticleWeightParams() : lower_weight(0), upper_weight(1), survival_weight(0.5), max_split(1), weight_cutoff(WEIGHT_CUTOFF) {}
```
pshriwise, line 30: Default values like these can be added as initializers of the data members below. Then we can rely on the default constructor for `ParticleWeightParams`.

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,81 @@
...
+
+namespace openmc {
+
+void read_weight_window_xml(); //!< read the weight window file
+
+namespace weight_window {
+
+// generic struct to contain the ww settings
+struct WWParams {
+  double upper_ratio;
```
pshriwise, line 17: We'll want to include some docstrings for these data members.

pshriwise (changes requested):
Thanks again for continuing this work, @makeclean! I like the updates you've made.

makeclean (commented):
Somehow i started my own review?!

src/physics.cpp
makeclean: Roundoff is potentially an issue here yes, it will usually be the case in fact that there will be some roundoff, making sure the primary particle accrues the difference is likely the best option.

include/openmc/weight_windows.h
```diff
@@ -4,21 +4,14 @@
 #include "openmc/mesh.h"
 #include "openmc/particle.h"
 
+#define WEIGHT_CUTOFF 1.0E-38 // default low weight cutoff
```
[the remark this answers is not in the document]
makeclean: Ok

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,81 @@
...
+  int max_split;
+  double weight_cutoff;
+  std::vector<double> energy_bounds;
+  std::vector<double> lower_ww;
+  std::vector<double> upper_ww;
+};
+
+// struct to pass specific location data 
+struct ParticleWeightParams {
+  // 0,1,0.5,1,1e-12
```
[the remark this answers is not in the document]
makeclean: that just a lingering comment :)

include/openmc/weight_windows.h
```diff
@@ -0,0 +1,81 @@
...
+
+  bool n_ww {false};   //!< flag for neutron use weight window
+  bool p_ww {false};   //!< flag for photon use weight window
+
+private:
+  
+  WWParams read_particle_settings(pugi::xml_node node);
+  
+  // ptr for the structure mesh
+  std::shared_ptr<RectilinearMesh> mesh_; //!< The mesh for the problem
```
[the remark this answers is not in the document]
makeclean: I dont see why not

src/weight_windows.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+WeightWindow::WeightWindow(pugi::xml_node node) : WeightWindow()
+{
+  using namespace pugi;
+  
+  // check for a mesh
+  if (check_for_node(node, "mesh")) { 
+    xml_node mesh = node.child("mesh");
+    mesh_ = make_unique<RectilinearMesh>(mesh);
```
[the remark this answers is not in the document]
makeclean: sure

src/weight_windows.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+  return settings;
+}
+  
+//! Get weight windows parameters given particle - essentially
+// given a location tell the particle the right bounds
+ParticleWeightParams WeightWindow::get_params(Particle& p) const
+{	
+  // Particle's position and energy
+  Position pos  = p.r();
```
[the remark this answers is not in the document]
makeclean: Sure

src/weight_windows.cpp
```diff
@@ -0,0 +1,223 @@
...
+  double E = p.E();
+
+  // get the settings for the weight window
+  ParticleType type = p.type();
```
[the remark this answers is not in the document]
makeclean: Sure thing

tests/regression_tests/weightwindow/settings.xml
```diff
@@ -0,0 +1,2636 @@
...
+  <particles>100000</particles>
+
+  <source>
+    <particle>neutron</particle>
+    <space type="point" parameters="0.001 0.001 0.001"/>
+    <angle type="isotropic" />
+    <energy type="discrete" parameters="14000000 1"/>
+  </source>
+
+  <weightwindow>
```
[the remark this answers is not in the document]
makeclean: Yeah, I've not updated this yet :)

src/weight_windows.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include "openmc/weight_windows.h"
+#include "openmc/xml_interface.h"
+
+namespace openmc {
+
+// read the weight window file
+void read_weight_window_xml() {
+  using namespace weight_window;
+  using namespace settings;
+  using namespace pugi;
```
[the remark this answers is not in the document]
makeclean: That was my original feeling, and I think for a while I changed, and then changed it back. If we're happy with a VR 'module' that contains all VR stuff, then I'd be keen on that.

src/physics.cpp
```diff
@@ -1160,4 +1166,55 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
   }
 }
 
+// split a particle on the basis of the statistical weight
+// of the current location
```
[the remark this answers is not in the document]
makeclean: ok

src/physics.cpp
```diff
@@ -1160,4 +1166,55 @@ void sample_secondary_photons(Particle& p, int i_nuclide)
...
+    double n_split = weight/params.upper_weight;  
+    n_split = std::min(std::ceil(n_split), double(params.max_split));
+        
+    // possibility of some round off 
+    for (int l = 0; l < int(n_split) - 1; l++)  { 
+      p.create_secondary(weight/n_split, p.u(), p.E(), p.type()); 
+    }
+    // todo maybe weight should be weight - sum of child weight
+    p.wgt() = weight/n_split; 
+    p.wgt_last() = p.wgt();
```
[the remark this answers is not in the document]
makeclean: gotcha

src/weight_windows.cpp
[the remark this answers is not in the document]
pshriwise: Yeah, that would be my vote for the namespace at least.

paulromano:
Superseded by #1947

PR #1861, opened 2021-07-29
Test PR for type hinting

Shimwell:
Just making a small PR to see how type hinting is received

This adds:

- type hinting to a coupled function arguments
- return types to some functions
- type hinting support to sphinx build
- docstring simplification (as types are now in args)

I'm wondering if this type hinting something that can be blended in gradually across all the python.

I'm keen to see how the docs look when they build against this PR branch


Shimwell:
Sorry for this, looks like the tests have failed. I think this PR is not close to being ready for review. I shall close it and come back with a better one in the future

PR #1870, opened 2021-08-18
Fix DAGMC and libMesh variable in CMake config

pshriwise:
Currently, the `DAGMC_FOUND` and `LIBMESH_FOUND` variables indicate whether or not these libraries are dependencies of the installed OpenMC library in our CMake config. I've found that this can cause problems if the `*_FOUND` variables are cached in the CMake directory. When the configuration variable `libmesh` is disable after being enabled on a previous build, some of the related variables inside the if-statement for libMesh in our CMake configure file are not populated and cause configurations of external projects to fail. I noticed this thanks to the `cpp_driver` test failing to build as I was testing locally.

This PR updates the DAGMC and libMesh blocks in our CMake config to use the configuration variables `dagmc` and `libmesh` to indicate whether or not those libraries are required. This should make our CMake configuration file more consistent with the configuration of the installed library.


paulromano (approved):
Thanks for the fix @pshriwise!

PR #1873, opened 2021-08-18
Allow constant scaling factor to be applied to libMesh unstructured mesh

### LINKED ISSUE #1872: Allow LibMesh unstructured mesh to apply a scaling factor
For Cardinal, we often couple OpenMC to a MOOSE application in units of meters. It would be convenient if OpenMC's `LibMesh` unstructured mesh constructor could apply a constant scaling factor to a provided mesh file to internally convert it to the units of centimeters required by OpenMC. 

For unstructured mesh tallies in Cardinal, we copy from the bins in the mesh file to the mesh in the `[Mesh]` MOOSE block. For proper transfers to other MOOSE apps in SI units, the `[Mesh]` is typically in units of meters and provided with a file. Having a scaling factor capability in OpenMC would save us the extra step of making essentially the same mesh, once in units of meters (for the MOOSE side of things) and once in centimeters (for the OpenMC side of things).

aprilnovak:
This PR allows a constant scaling factor to be applied to a LibMesh unstructured mesh template. This scaling factor is then used to multiply the entire mesh as it is constructed.

I tested these changes in Cardinal by changing our tests from:

- `[Mesh]` in units of meters, but OpenMC mesh template in units of centimeters

to this newly-enabled option:

- `[Mesh]` in units of meters, OpenMC mesh template in units of meters, with scaling factor applied internally to get to OpenMC-internal units of centimeters

I realize this proposed change is quite specific to Cardinal, so I'm totally open to other design ideas. I can add a test once I get a green light that these changes are feasible.

Closes #1872


gridley (approved):
LGTM! I like how you don't call the libMesh scaling operation if it would scale by one.

src/mesh.cpp
```diff
@@ -2141,15 +2142,32 @@ void MOABMesh::write(const std::string& base_filename) const
...
   filename_ = filename;
+
+  if (length_multiplier != 1.0) {
+    set_length_multiplier(length_multiplier);
+  }
+
   initialize();
 }
 
+void LibMesh::set_length_multiplier(const double length_multiplier)
```
pshriwise, line 2165:
```suggestion
void LibMesh::set_length_multiplier(double length_multiplier)
```

src/mesh.cpp
```diff
@@ -2141,15 +2142,32 @@ void MOABMesh::write(const std::string& base_filename) const
...
+  if (check_for_node(node, "length_multiplier")) {
+    length_multiplier_ = std::stod(get_node_value(node, "length_multiplier"));
+    specified_length_multiplier_ = true;
+  }
+
   initialize();
 }
 
-LibMesh::LibMesh(const std::string& filename)
+LibMesh::LibMesh(const std::string& filename, const double length_multiplier)
```
pshriwise, line 2154:
```suggestion
LibMesh::LibMesh(const std::string& filename, double length_multiplier)
```

include/openmc/mesh.h
```diff
@@ -492,7 +492,7 @@ class LibMesh : public UnstructuredMesh {
 public:
   // Constructors
   LibMesh(pugi::xml_node node);
-  LibMesh(const std::string& filename);
+  LibMesh(const std::string& filename, const double length_multiplier = 1.0);
```
pshriwise, line 495:
```suggestion
  LibMesh(const std::string& filename, double length_multiplier = 1.0);
```

pshriwise (changes requested):
This looks good to me! I've requested a few small changes to the `length_multiplier` parameters here. 

A couple of additional items:
  - We'll want to provide this parameter in the Python API `UnstructuredMesh` class.
  - We'll want to allow this for MOAB meshes as well to match this capability.

I'd be up for taking care of the MOAB mesh scaling piece. Would you be ok with a PR into your branch @aprilnovak? Otherwise, I'm fine with doing that as a follow-on PR to this one.

aprilnovak:
> Would you be ok with a PR into your branch

Sure, go for it! Thanks for handling the MOAB part - I wanted to put this into the UnstructuredMesh class directly, but don't know how to use MOAB.

openmc/mesh.py
```diff
@@ -215,6 +215,7 @@ def from_hdf5(cls, group):
         mesh.lower_left = group['lower_left'][()]
         mesh.upper_right = group['upper_right'][()]
         mesh.width = group['width'][()]
+        mesh.length_multiplier = group['length_multiplier'][()] if 'length_multiplier' in group else 1.0
```
pshriwise, line 218: We'll want to put this in the `UnstructuredMesh.from_hdf5` method.

While I typically like these kind of inline if-else statements, we can take out the `else` since the `length_multiplier is already 1.0 due to the default arg when the new mesh class is created above.

```suggestion
    if 'length_multiplier' in group:
        mesh.length_multiplier = group['length_multiplier'][()] 
```

openmc/mesh.py
```diff
@@ -611,6 +612,8 @@ class UnstructuredMesh(MeshBase):
     ----------
     filename : str
         Location of the unstructured mesh file
+    length_multiplier: float
```
pshriwise, line 615: Let's move this to the end of the list here so the order of the parameters matches the signature of the class constructor.

Also, if you wouldn't mind removing the `size` parameter docstring while we're here it would be much appreciated. I must've left that in by mistake.

pshriwise (changes requested):
Mostly line comments to take care of there and then this looks ready to me.

openmc/mesh.py
pshriwise: Unless I messed up in pulling your most recent changes, it looks like these lines still need to be moved to the `UnstructuredMesh.from_hdf5` method

aprilnovak: I think I added it? I see it on the "Files Changed" tab at least. Your batch suggestion didn't have the right indentation, so I had to change it locally after adding it to the batch. Maybe that did something strange with viewing the changes?

aprilnovak: Oh I'm dumb. No that's my bad - I blame Python's lack of `ClassName::method_name` for why I didn't see that!

pshriwise (approved):
Looks good @aprilnovak! Thanks!

PR #1879, opened 2021-09-09
DAGMC error message and documentation update.

pshriwise:
Currently, if a user tries to run a problem that contains only DAGMC geometry

```xml
<geometry>
  <dagmc_universe filename="dagmc.h5m" id="1" />
</geometry>
```

 without configuring OpenMC to use DAGMC they get the message

```
 ERROR: No cells were found in the geometry.xml file
```

, which isn't very informative.

I've updated the body of `read_dagmc_universes` when `DAGMC` is disabled to produce a more useful error message if DAGMC XML nodes are present.

Additionally, I've added an example of what a `geometry.xml` file would look like if the geometry were made up of one DAGMC universe as the root universe in the documentation for the `geometry.xml` file spec.

### DIFF
--- a/docs/source/io_formats/geometry.rst
+++ b/docs/source/io_formats/geometry.rst
@@ -406,3 +406,14 @@ Each ``<dagmc_universe>`` element can have the following attributes or sub-eleme
     A required string indicating the file to be loaded representing the DAGMC universe.
 
     *Default*: None
+
+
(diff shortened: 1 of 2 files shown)


paulromano (approved):
LGTM. I'll let @makeclean chime in before merging.

PR #1884, opened 2021-09-24
Expanded capabilities of openmc.model.Model container

nelsonag:
This PR introduces some additional features to the openmc.model.Model class. Most importantly, it makes the Model class able to load a model to the C-API and then provide interfaces to perform operations on the python and C representations of the model to ensure consistency.

The end goal of these changes is to have a one-line statement that the user can use to efficiently run a life cycle analysis that includes depletion and critical searches.

This PR is not yet ready to be merged as it requires some items (most importanty, tests); it is provided early to allow for an early check of this potential change in philosophy of the uses of the Model class.

At this point, the Model class can initialize the C-API when asked, use the C-API in model.run and model.deplete, and now it can change the temperature, volume, density, rotation, and transform of materials and cells (as applicable) on *both* the C-API and Python side with the same call.

The remaining items are:

- [x] Creation of tests

- [x] Addition of a volume calculation that can run in the C-API and/or via command line

- [x] Thought given to file management

--------------------------
To get to a point where Model.critical_deplete(...) is possible, #1656 (improved crit search) is necessary and shall be ensured to work on the C-API side. That will be the subject of a future PR.

To document thoughts on the matter, my goal of a Model.critical_deplete(...) method is one where time steps, power histories, and lists of control rod schemes are provided.  These schemes represent the control rod sequences (i.e., withdraw this group of rods, then that group of rods, then rotate these drums, etc) and will be defined by the user with the following for each sequence: translate or rotation, the set of cells to be operated on, and the range of motion for each.  Then the code will step through each range (perhaps with smartness) to find the critical config and phase of the control sequence.

Ideally there would be a way to also run branch calculations in-line or in a separate calculation that uses all output provided by these cases.  Branch calculations would include:
- Xe-Free and peak xenon
- AlphaT calculations
- Differential Rod worth calcs
- Integral rod worth 'walks'


nelsonag:
Quick status update here:

- The desired methods are implemented (as per a commit coming soon), however, there are some TODOs noted in model.py (and in some cases corresponding ones in test_model.py). My current plan is to hold off on these TODOs for my next PR, to keep this one small, and limited to just the python side. These TODOs are (right now):

  - ``openmc.lib.init`` takes and applies command line arguments. That means that if the user wanted to run in geometry debug mode (for example) in run via the C-API, but not in another, then they are required to finalize and re-initialize the model. I would prefer to find alternative solutions. There probably isn't a simple solution to this one for all the command line arguments, so I dont want to promise they all will be fixed in that TODO-fixing PR.
  - Right now when you initialize, there is no way to turn off output to screen. I would like to do that and so I left the output flag in the ``Model.init_C_api`` function, but it is not implemented. When I look in to this more I may decide its not worth the effort, and then I will remove this flag.
  - The ``Settings.event_based`` flag on the C++ side is not exposed to the Py-API. This is one of the params mentioned in the above bullet, but is clearly simple enough that it can be fixed.
  - There is a ``model.plot_geometry`` method that can call ``openmc.lib.plot_.geometry`` if the C-API has been initialized and ``openmc.plot_geometry`` (i.e., the CLI approach) otherwise. However, the ``openmc_init`` function doesn't load ``plots.xml`` unless it has the ``--plots`` argument given. Therefore calling ``openmc.lib.plot_.geometry`` won't do anything. For now we just default to the CLI approach and the C-API block is commented out. This could be fixed on the C++ side by either always reading ``plots.xml`` or by having some C-API command to load the data.
  - Right now there is ``Model.update_material_temperatures(...)` `code that can set material temperatures on both the python and C++ sides simultaneously (e.g., this is intended for alphaT branch calcs and not necessarily thermal feedback). However, there is no C-API ``Material.set_temperature`` method/or a ``Material.temperature`` setter. This needs to be added before this part of the code can be enabled and its test uncommented.

- At the time of writing this comment I have two more tests to write: a test of ``Model.deplete(...)`` and of ``Model.calculate_volume(...)``.
- The new test_model.py is failing on CI but not locally. I haven't yet investigated why.
- For file management, the approach now is to allow the user to run these C-API calls in a directory of their choosing. I envision the future ``Model.critical_deplete(...)`` method would then know how to use this to, for example, create directories of ``step_0``, ``step_1``, ... and so on, to store results. Another option is to condense results to a single HDF5 file, but I worry that creating a new file format that is only needed by a single capability may be overkill at this stage.

openmc/model/model.py
```diff
@@ -254,32 +443,68 @@ def import_properties(self, filename):
                 cell = cells[cell_id]
                 if cell.fill_type in ('material', 'distribmat'):
                     cell.temperature = group['temperature'][()]
+                    if self.C_init:
+                        C_cell = openmc.lib.cells[cell_id]
+                        C_cell.set_temperature(group['temperature'][()])
```
paulromano, line 448:
```suggestion
                        lib_cell = openmc.lib.cells[cell_id]
                        lib_cell.set_temperature(group['temperature'][()])
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+            if self.C_init:
+                # Handle the run options as applicable
+                # First dont allow ones that must be set via init
+                for arg_name, arg, default in zip(
+                    ['threads', 'geometry_debug', 'restart_file', 'tracks'],
+                    [threads, geometry_debug, restart_file, tracks],
+                    [None, False, None, False]):
+                    if arg != default:
+                        msg = "{} must be set via Model.c_init(...)".format(
+                            arg_name)
```
paulromano, line 533:
```suggestion
                        msg = f"{arg_name} must be set via Model.c_init(...)"
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+            else:
+                output_dir = Path.cwd()
+            for sp in output_dir.glob('statepoint.*.h5'):
+                mtime = sp.stat().st_mtime
+                if mtime >= tstart:  # >= allows for poor clock resolution
+                    tstart = mtime
+                    last_statepoint = sp
+        return last_statepoint
+
+    def calculate_volume(self, threads=None, output=True, cwd='.',
```
paulromano, line 580: For consistency
```suggestion
    def calculate_volumes(self, threads=None, output=True, cwd='.',
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+                for vol_calc in vol_calcs:
+                    # First add them to the Python side
+                    self.geometry.add_volume_information(vol_calc)
+
+                    # And now repeat for the C-API
+                    if vol_calc.domain == 'material':
+                        # Then we can do this in the C-API
+                        for domain_id in vol_calc.domains:
+                            self.update_material_volumes(
+                                [domain_id], vol_calc.volumes[domain_id])
```
paulromano, line 662: We should probably encapsulate the ability to update volumes in `openmc.lib.calculate_volumes` itself. What you have here works for now though

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+                self.export_to_xml()
+                openmc.calculate_volumes(threads=threads, output=output,
+                                         openmc_exec=openmc_exec,
+                                         mpi_args=mpi_args)
+
+            # Now we apply the volumes
+            if apply_volumes:
+                # Load the results
+                f_names = \
+                    ["volume_{}.h5".format(i + 1)
```
paulromano, line 648:
```suggestion
                    [f"volume_{i + 1}.h5"
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+            if attrib_name == 'rotation':
+                obj.rotation = value
+            elif attrib_name == 'translation':
+                obj.translation = value
+            elif attrib_name == 'volume':
+                obj.volume = value
+            elif attrib_name == 'temperature':
+                obj.temperature = value
+            elif attrib_name == 'density':
+                obj.set_density(density_units, value)
```
paulromano, line 796:
```suggestion
            if attrib_name == 'density':
                obj.set_density(density_units, value)
            else:
                setattr(obj, attrib_name, value)
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+            raise ValueError('Cannot set a Cell density!')
+        if obj_type == 'material' and attrib_name in ('rotation',
+                                                      'translation'):
+            raise ValueError('Cannot set a material rotation/translation!')
+
+        # Set the
+        if obj_type == 'cell':
+            by_name = self._cells_by_name
+            by_id = self._cells_by_id
+            C_by_id = openmc.lib.cells
```
paulromano, line 759:
```suggestion
            obj_by_id = openmc.lib.cells
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
-            output_dir = Path.cwd()
-        for sp in output_dir.glob('statepoint.*.h5'):
-            mtime = sp.stat().st_mtime
-            if mtime >= tstart:  # >= allows for poor clock resolution
-                tstart = mtime
-                last_statepoint = sp
-        return last_statepoint
+            by_name = self._materials_by_name
+            by_id = self._materials_by_id
+            C_by_id = openmc.lib.materials
```
paulromano, line 763:
```suggestion
            obj_by_id = openmc.lib.materials
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+                if attrib_name == 'rotation':
+                    C_obj.rotation = value
+                elif attrib_name == 'translation':
+                    C_obj.translation = value
+                elif attrib_name == 'volume':
+                    C_obj.volume = value
+                elif attrib_name == 'temperature':
+                    C_obj.set_temperature(value)
+                elif attrib_name == 'density':
+                    C_obj.set_density(value, density_units)
```
paulromano, line 809:
```suggestion
                lib_obj = obj_by_id[id_]
                if attrib_name == 'temperature':
                    lib_obj.set_temperature(value)
                elif attrib_name == 'density':
                    lib_obj.set_density(value, density_units)
                else:
                    setattr(lib_obj, attrib_name, value)
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+
+    def rotate_cells(self, names_or_ids, vector):
+        """Rotate the identified cell(s) by the specified rotation vector.
+        The rotation is only applied to cells filled with a universe.
+
+        Parameters
```
paulromano, line 815:
```suggestion
        The rotation is only applied to cells filled with a universe.

        .. versionadded:: 0.13.0

        Parameters
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+
+        """
+
+        self._change_py_C_attribs(names_or_ids, vector, 'cell', 'rotation')
+
+    def translate_cells(self, names_or_ids, vector):
+        """Translate the identified cell(s) by the specified translation vector.
+        The translation is only applied to cells filled with a universe.
+
+        Parameters
```
paulromano, line 832:
```suggestion
        The translation is only applied to cells filled with a universe.

        .. versionadded:: 0.13.0

        Parameters
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+            the x, y, and z dimensions of the translation.
+
+        """
+
+        self._change_py_C_attribs(names_or_ids, vector, 'cell', 'translation')
+
+    def update_densities(self, names_or_ids, density, density_units='atom/b-cm'):
+        """Update the density of a given set of materials to a new value
+
+        Parameters
```
paulromano, line 848:
```suggestion
        """Update the density of a given set of materials to a new value

        .. versionadded:: 0.13.0

        Parameters
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+
+        """
+
+        self._change_py_C_attribs(names_or_ids, density, 'material', 'density',
+                                  density_units)
+
+    def update_cell_temperatures(self, names_or_ids, temperature):
+        """Update the temperature of a set of cells to the given value
+
+        Parameters
```
paulromano, line 866:
```suggestion
        """Update the temperature of a set of cells to the given value

        .. versionadded:: 0.13.0

        Parameters
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+
+        """
+
+        self._change_py_C_attribs(names_or_ids, temperature, 'cell',
+                                  'temperature')
+
+    def update_material_temperatures(self, names_or_ids, temperature):
+        """Update the temperature of a set of materials to the given value
+
+        Parameters
```
paulromano, line 882:
```suggestion
        """Update the temperature of a set of materials to the given value

        .. vesionadded:: 0.13.0

        Parameters
```

openmc/model/model.py
```diff
@@ -289,23 +514,395 @@ def run(self, **kwargs):
...
+
+        """
+
+        self._change_py_C_attribs(names_or_ids, temperature, 'material',
+                                  'temperature')
+
+    def update_material_volumes(self, names_or_ids, volume):
+        """Update the volume of a set of materials to the given value
+
+        Parameters
```
paulromano, line 898:
```suggestion
        """Update the volume of a set of materials to the given value

        .. versionadded:: 0.13.0

        Parameters
```

paulromano (changes requested):
This is an awesome feature! Thanks a lot for contributing this @nelsonag, and glad to see some inspired ideas from you as a user-developer.

openmc/lib/__init__.py
```diff
@@ -59,3 +59,5 @@ def _libmesh_enabled():
 from .settings import settings
 from .math import *
 from .plot import *
+
+LIB_INIT = False
```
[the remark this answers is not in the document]
nelsonag: Thanks; good point on ``simulation::initialized``. I just added one of the TODOs for that so that is included with C changes

openmc/model/model.py
```diff
@@ -31,6 +40,14 @@ class Model:
...
     plots : openmc.Plots, optional
         Plot information
+    chain_file : str or Path, optional
+        Path to the depletion chain XML file.  Defaults to the chain
+        found under the ``depletion_chain`` in the
+        :envvar:`OPENMC_CROSS_SECTIONS` environment variable if it exists. If a
+        str is provided it will be converted to a Path object.
+    fission_q : dict, optional
+        Dictionary of nuclides and their fission Q values [eV].
+        If not given, values will be pulled from the ``chain_file``.
```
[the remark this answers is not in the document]
nelsonag: My decision to store this as an attribute is because I envision ``model.deplete`` being called multiple times in a given analysis. Since the operator kwargs are unlikely to change for each of these ``model.deplete`` calls, it seems as if that should be saved somewhere to make the user input burden smaller.

Having said that, I do recognize that the ``operator_kwargs`` is a good idea. 
How do you feel about replacing ``Model.chain_file`` and ``Model.fission_q`` with ``Model.operator_kwargs``?

openmc/model/model.py
```diff
@@ -66,6 +91,19 @@ def __init__(self, geometry=None, materials=None, settings=None,
...
 
+        self.chain_file = chain_file
+        self.fission_q = fission_q
+
+        if materials is None:
+            mats = self.geometry.get_all_materials().values()
+        else:
+            mats = self.materials
+        self._materials_by_id = {mat.id: mat for mat in mats}
+        self._materials_by_name = {mat.name: mat for mat in mats}
```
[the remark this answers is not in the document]
nelsonag: Ah, yes, thanks! I meant to leave a note about that in the docs. My original intent was to just leave a documentation ``.. note::`` in the ``rotate_cells``, ``translate_cells``, etc, methods that accessing the correct cell is not guaranteed for non-unique names.

However, right now in thinking about it I think the approach I will take is to store a list of all the cells that share a given name and operate on all of them. This could be super useful as a user by naming all things that need to be moved (e.g.) the same thing.  This will still have a documentation note to explain the behavior.

openmc/model/model.py
```diff
@@ -1,11 +1,20 @@
...
+import subprocess
 
 import h5py
+import numpy as np
 
 import openmc
-from openmc.checkvalue import check_type, check_value
+from openmc.checkvalue import check_type, check_value, check_iterable_type, \
+    check_length
+import openmc.deplete as dep
```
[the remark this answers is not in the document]
nelsonag: The reason I moved it globally is because of the following:

1. I'm probably too used to it now, but I was thinking that having the shared library around is no longer a big hurdle to overcome, especially since so much functionality requires it now.

2. I have an MPI barrier when writing the XML files. I realize now that the better approach could just be to set the communicator at ``Model.__init__``

openmc/model/model.py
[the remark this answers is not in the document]
nelsonag: Aaalright fine. After trying it out, I decided I agree, operator_kwargs could change and thus no reason to set it on the Model class. So I did as you suggested (pass operator_kwargs and integrator_kwargs to deplete and removed the operator parameters from Model)

openmc/model/model.py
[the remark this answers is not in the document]
nelsonag: This is now done

openmc/model/model.py
```diff
@@ -151,8 +218,70 @@ def from_xml(cls, geometry='geometry.xml', materials='materials.xml',
...
+        """
+
+        # TODO: right now the only way to set most of the above parameters via
+        # the C-API are at initialization time despite use-cases existing to
+        # set them for individual runs. For now this functionality is exposed
+        # where it exists (here in init), but in the future the functionality
+        # should be exposed so that it can be accessed via model.run(...)
+
+        # TODO: the output flag is not yet implemented for the openmc.lib.init
+        # command. This will be the subject of future work.
```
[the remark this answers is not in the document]
nelsonag: Got it! Created a contextmanager to handle it and now both ``openmc.lib.init`` and ``openmc.lib.run`` accept an ``output`` parameter.

openmc/lib/__init__.py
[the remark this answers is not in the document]
nelsonag: Regarding using ``simulation::initialized``, this one is set in ``openmc_simulation_init`` which is called by ``openmc_run``. That makes it a bit too late for my purposes here.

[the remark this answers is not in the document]
paulromano: Yeah you're absolutely right. Perhaps we need another variable on the C++ side to track basic initialization. We can leave that as a TODO for another time.

nelsonag:
Ok, all desired functionality is complete and tests are in place. The comments from the review by @paulromano have been addressed, and the tests seem to be on their way to passing!

I previously noted that I was going to hold off on changes to the C++ side until another PR. However, once I/we decided not to address the following two items, there wasnt a strong reason to hold off on the changes. The remaining items are noted with a TODO and are:
- Create and expose some variable on the C++ side which captures the initialization state (i.e., has openmc_init been called) and then expose that to the PyAPI.
- Refactor the openmc_init so that it is not the only way certain items are set (mainly, those from the command line arguments).

I updated my previous comment to reflect this change by striking through items 1 and 2 to reflect that those are 'done' (pending merge)

nelsonag:
@paulromano I think the most interesting changes since your review would be the following:
- Addition of reading plots.xml even if not in the plotting run mode (so that model.plot_geometry can be called at any time; this could be especially useful for verifying rod motion occurs as expected). Doing so required minor changes, but I suppose you may be interested to know that the plots.xml data will be in memory for every future invocation of OpenMC. (Likely very negligible)
- The ``openmc.deplete`` package was effectively where the MPI Communicator was initialized and stored (and if no MPI, a dummy communicator used instead). I moved this upstream to living within ``openmc`` but was sure not to make a circular dependency by not having ``openmc.deplete.__init__`` depend upon it.
- A new ``quiet_dll`` context manager in ``openmc.lib.core``. Pretty excited about this one, that depletion output always annoyed me, and now the output is sleeping with the fishes! This context manager effectively swaps the system's stdout file descriptor for /dev/null and then undoes what it did when exiting the context. I *think* this is only POSIX compliant and will not work on Windows.

openmc/__init__.py
```diff
@@ -30,8 +30,10 @@
 from openmc.search import *
 from openmc.polynomial import *
 from . import examples
+from openmc.mpi import DummyCommunicator, MPI, comm
```
paulromano, line 33: I'm fine with `comm` moving to its own file, but I don't want it to be imported by default since that implies an import of `mpi4py` by default. Let's do the same approach as with lib/deplete -- if `comm` is needed, have a deferred import of `openmc.mpi` in the body of a function/method.

openmc/deplete/abc.py
```diff
@@ -22,7 +22,7 @@
 from openmc.data import DataLibrary
 from openmc.lib import MaterialFilter, Tally
 from openmc.checkvalue import check_type, check_greater_than
-from . import comm
+from openmc import comm
```
paulromano, line 25:
```suggestion
from openmc.mpi import comm
```

openmc/deplete/helpers.py
```diff
@@ -10,7 +10,7 @@
 
 from numpy import dot, zeros, newaxis, asarray
 
-from . import comm
+from openmc import comm
```
paulromano, line 13:
```suggestion
from openmc.mpi import comm
```

openmc/deplete/operator.py
```diff
@@ -20,7 +20,7 @@
 import openmc
 from openmc.checkvalue import check_value
 import openmc.lib
-from . import comm
+from openmc import comm
```
paulromano, line 23:
```suggestion
from openmc.mpi import comm
```

openmc/deplete/results.py
```diff
@@ -9,7 +9,7 @@
 import h5py
 import numpy as np
 
-from . import comm, MPI
+from openmc import comm, MPI
```
paulromano, line 12:
```suggestion
from openmc.mpi import comm, MPI
```

openmc/executor.py
```diff
@@ -5,6 +5,82 @@
 import openmc
 
 
+def process_CLI_arguments(volume=False, geometry_debug=False, particles=None,
+                          plot=False, restart_file=None, threads=None,
+                          tracks=False, event_based=None, openmc_exec='openmc',
+                          mpi_args=None):
+    """Run an OpenMC simulation.
```
paulromano, line 12: Need to change description here

openmc/executor.py
```diff
@@ -5,6 +5,82 @@
 import openmc
 
 
+def process_CLI_arguments(volume=False, geometry_debug=False, particles=None,
```
paulromano, line 8: Probably should make this "private" with a leading underscore -- not something that I'd imagine a user calling

openmc/lib/core.py
```diff
@@ -342,9 +359,19 @@ def next_batch():
     return status.value
 
 
-def plot_geometry():
-    """Plot geometry"""
-    _dll.openmc_plot_geometry()
+def plot_geometry(output=True):
+    """Plot geometry
+
+    .. versionchanged:: 0.13.0
```
paulromano, line 365:
```suggestion
    .. versionchanged:: 0.13.0
       The *output* argument was added.
```

openmc/lib/core.py
```diff
@@ -357,9 +384,19 @@ def reset_timers():
     _dll.openmc_reset_timers()
 
 
-def run():
-    """Run simulation"""
-    _dll.openmc_run()
+def run(output=True):
+    """Run simulation
+
+    .. versionchanged:: 0.13.0
```
paulromano, line 390:
```suggestion
    .. versionchanged:: 0.13.0
       The *output* argument was added.
```

openmc/model/model.py
```diff
@@ -66,6 +91,34 @@ def __init__(self, geometry=None, materials=None, settings=None,
...
+
+        # Get the names maps, but since names are not unique, store a list for
+        # each name key. In this way when the user requests a change by a name,
+        # the change will be applied to all of the same name.
+        self._cells_by_name = {}
+        for cell in cells.values():
+            if cell.name not in self._cells_by_name:
+                self._cells_by_name[cell.name] = [cell]
+            else:
+                self._cells_by_name[cell.name].append(cell)
```
paulromano, line 114:
```suggestion
            if cell.name not in self._cells_by_name:
                self._cells_by_name[cell.name] = set()
            self._cells_by_name[cell.name].add(cell)
```

openmc/model/model.py
```diff
@@ -151,44 +219,152 @@ def from_xml(cls, geometry='geometry.xml', materials='materials.xml',
...
+        """
+
+        openmc.lib.finalize()
+
+    def deplete(self, timesteps, method='cecm', final_step=True,
+                operator_kwargs=None, directory='.', output=True,
+                **integrator_kwargs):
         """Deplete model using specified timesteps/power
 
+        .. versionchanged:: 0.13.0
```
paulromano, line 290:
```suggestion
        .. versionchanged:: 0.13.0
           The *final_step*, *operator_kwargs*, *directory*, and *output*
           arguments were added.
```

tests/unit_tests/test_deplete_chain.py
```diff
@@ -7,7 +7,8 @@
 from pathlib import Path
 
 import numpy as np
-from openmc.deplete import comm, Chain, reaction_rates, nuclide, cram, pool
+from openmc import comm
```
paulromano, line 10:
```suggestion
from openmc.mpi import comm
```

tests/unit_tests/test_deplete_integrator.py
```diff
@@ -14,11 +14,11 @@
 from uncertainties import ufloat
 import pytest
 
+from openmc import comm
```
paulromano, line 17:
```suggestion
from openmc.mpi import comm
```

paulromano (changes requested):

openmc/lib/core.py
```diff
@@ -475,3 +512,43 @@ def __init__(self, uid=None, new=True, index=None):
...
+        # Get a garbage descriptor so we can throw away output
+        devnull = os.open(os.devnull, os.O_WRONLY)
+
+        # Get the current stdout stream and make a duplicate of it
+        new_stdout = os.dup(1)
+        # Copy the garbage output to the stdout stream
+        os.dup2(devnull, 1)
+        os.close(devnull)
+        # Now point stdout to the re-defined stdout
+        sys.stdout = os.fdopen(new_stdout, 'w')
```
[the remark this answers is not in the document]
nelsonag: For the latter, i think the output parameter should suffice for that, especially as its default is True (show output), right?

openmc/model/model.py
nelsonag: And I just learned that ``versionchanged`` requires a description whereas ``versionadded`` does not! (first time using is right in this PR!)

src/material.cpp
```diff
@@ -1480,6 +1480,18 @@ extern "C" int openmc_material_set_name(int32_t index, const char* name)
   return 0;
 }
 
+extern "C" int openmc_material_set_temperature(int32_t index, double temperature)
+{
+  if (index >= 0 && index < model::materials.size()) {
+    auto& m {model::materials[index]};
+    m->temperature_ = temperature;
+    return 0;
```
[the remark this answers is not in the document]
nelsonag: Shoot; you are right. That must be why it wasn't there ;-) I will remove this functionality

openmc/__init__.py
nelsonag: Hmm, the presence of DummyCommunicator provides that protection via Mock and DummyCommunicator. Will take a look and see what the delayed import looks like.

openmc/model/model.py
```diff
@@ -44,6 +65,10 @@ class Model:
         Tallies information
     plots : openmc.Plots
         Plot information
+    intracomm : mpi4py.MPI.Intracomm or openmc.DummyCommunicator
+        MPI intracommunicator; this defaults to the mpi4py world communicator
+        from if present, or a DummyCommunicator otherwise. If an alternative
+        communicator is desired, this parameter should be modified accordingly.
```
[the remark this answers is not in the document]
nelsonag: For those purposes I think we should keep it as-is

openmc/__init__.py
nelsonag: Yeah the ``comm`` object is only present throughout ``openmc.deplete`` and in a few places in ``model.py``. 

For ``openmc.deplete``, its so distributed by necessity that local imports will hinder more than help.

For ``model.py`` the place where its mostly needed is setting the default communicator in ``Model.__init__``. That would be used even if MPI functionality isn't desired. This could be moved elsewhere but then the Model won't know about the MPI state until some later point (probably init?), but I worry there could be some future system state manipulations which should be done in an MPI aware context that won't be. Could I tell you what those are now, not exactly, perhaps export_to_xml writes are one example but I'm sure thats not the best.

So, since the ``openmc.mpi`` package includes import protections already, I think this is probably the right approach.

openmc/lib/core.py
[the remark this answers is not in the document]
paulromano: In the output=True case, the output is shown but not in the right place if you are running from a Jupyter notebook; it only shows up in the terminal. Anyway, definitely not a problem you need to worry about for this PR.

openmc/model/model.py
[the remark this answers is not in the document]
paulromano: The only place it's used is in `lib_init` where it's passed to `openmc.lib.init`. That makes me think we should just allow arbitrary kwargs to be passed through from the `lib_init` method to `openmc.lib.init` method.

openmc/__init__.py
paulromano: Let me restate just to make sure the concern is clear -- importing `mpi4py` actually results in `MPI_Init` getting called under the hood. You really don't want this to happen unless you are definitely going to be calling MPI code. As one example of what can go wrong, if you import `mpi4py` and then call `subprocess.run(['mpiexec', ...])`, it won't work for some MPI implementations. This would be a really annoying side effect of `import openmc` which is why I only want to import `mpi4py` if it is 100% needed.

nelsonag: Ohhhhh thank you. That makes more sense now. I have corrected this behavior.

openmc/lib/core.py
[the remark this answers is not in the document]
nelsonag: Ohhh, I thought you meant you had tried this but it had a deleterious effect on notebooks. Got it now

openmc/model/model.py
```diff
@@ -151,44 +206,156 @@ def from_xml(cls, geometry='geometry.xml', materials='materials.xml',
...
+        if intracomm is not None:
+            self._intracomm = intracomm
+        else:
+            self._intracomm = DummyCommunicator()
+
+        if self._intracomm.rank == 0:
+            self.export_to_xml()
+        self._intracomm.barrier()
+
+        openmc.lib.init(args=args, intracomm=self._intracomm, output=output)
```
paulromano, line 264: Right now this doesn't work if you pass a `DummyCommunicator` but `mpi4py` is installed. I found this out when I tried to run tests on my machine using your branch and were getting failures. An easy fix is:
```suggestion
        openmc.lib.init(args=args, intracomm=intracomm, output=output)
```

tests/unit_tests/test_model.py
```diff
@@ -1,29 +1,279 @@
+from math import pi
+import numpy as np
 import pytest
+from pathlib import Path
+from shutil import which
```
paulromano, line 5:
```suggestion
from math import pi
from pathlib import Path
from shutil import which

import numpy as np
import pytest
```

tests/unit_tests/test_model.py
```diff
@@ -32,3 +282,257 @@ def test_import_properties(run_in_tmpdir, mpi_intracomm):
...
+    sp_path = test_model.run(output=False)
+    with openmc.StatePoint(sp_path) as sp:
+        lib_keff = sp.k_combined
+        lib_flux = sp.get_tally(id=1).get_values(scores=['flux'])[0, 0, 0]
+        lib_fiss = sp.get_tally(id=1).get_values(scores=['fission'])[0, 0, 0]
+
+    # and lets compare results
+    assert abs(lib_keff - cli_keff) < 1e-13
+    assert abs(lib_flux - cli_flux) < 1e-13
+    assert abs(lib_fiss - cli_fiss) < 1e-13
```
paulromano, line 309: In all these cases where you're checking if two floating point values are really close, I'd recommend using:
```python
from pytest import approx

assert x == approx(y)
```
It looks like you have a bunch of cases down below that also use the same pattern and could be updated.

For the comparison of the keff values, you'll need to compare the nominal values (`lib_keff.n == approx(cli_keff.n)`)

openmc/model/model.py
```diff
@@ -1,11 +1,32 @@
...
+from openmc.dummy_comm import DummyCommunicator
+from openmc.executor import _process_CLI_arguments
 from openmc.checkvalue import check_type, check_value
+from openmc.exceptions import InvalidIDError
+
+
+@contextmanager
+def _change_directory(working_dir):
+    """A context manager for executing in a provided working directory"""
+    start_dir = Path().cwd()
```
paulromano, line 23: `cwd` is actually a classmethod:
```suggestion
    start_dir = Path.cwd()
```

openmc/model/model.py
```diff
@@ -151,44 +206,156 @@ def from_xml(cls, geometry='geometry.xml', materials='materials.xml',
...
+        restart_file : str, optional
+            Path to restart file to use
+        tracks : bool, optional
+            Write tracks for all particles. Defaults to False.
+        output : bool
+            Capture OpenMC output from standard out
+        event_based : None or bool, optional
+            Turns on event-based parallelism if True. If None, the value in
+            the Settings will be used.
+        intracomm : DummyCommnicator, mpi4py.MPI.Intracomm or None, optional
```
paulromano, line 234: DummyCommunicator is really an implementation detail that shouldn't be exposed to a user. Passing `None` is the right thing to do if you're not using MPI
```suggestion
        intracomm : mpi4py.MPI.Intracomm or None, optional
```

openmc/model/model.py
```diff
@@ -289,23 +496,365 @@ def run(self, **kwargs):
...
+                openmc.calculate_volumes(threads=threads, output=output,
+                                         openmc_exec=openmc_exec,
+                                         mpi_args=mpi_args)
+
+            # Now we apply the volumes
+            if apply_volumes:
+                # Load the results and add them to the model
+                for i, vol_calc in enumerate(self.settings.volume_calculations):
+                    f_name = f"volume_{i + 1}.h5"
+                    vol_calc.load_results(f_name)
```
paulromano, line 614:
```suggestion
                    vol_calc.load_results(f"volume_{i + 1}.h5")
```

paulromano (changes requested):
Thanks for the updates @nelsonag. A few more small things and then I think this is ready to go

nelsonag:
Thanks for the comments @paulromano, I just implemented the requested changes

paulromano (approved):

paulromano:
Thanks again for this contribution @nelsonag!

PR #1886, opened 2021-09-28
added optional scaling factor to vtk writing

Shimwell:
This PR is an attempt to add a scaling factor to the write_data_to_vtk function so that the tet mesh can be scaled by a user defined number. This could be useful when processing an unstructured mesh from a statepoint file for visualisation. I'm still not totally sure how to do the tet mesh extraction from a vtk file and make use of this write_data_to_vtk function as my moab mesh doesn't appear to have any centroids.

My use case for this is to get energy deposited in joules so I'm keen to scale the tally result by a conversion factor. I guess others might want heating in watts and would need to scale the result in a similar manner

What do you think @pshriwise

### DIFF
--- a/openmc/mesh.py
+++ b/openmc/mesh.py
@@ -719,7 +719,8 @@ class UnstructuredMesh(MeshBase):
         string += '{: <16}=\t{}\n'.format('\tMesh Library', self.mesh_lib)
         return string
 
-    def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
+    def write_data_to_vtk(self, filename, datasets, volume_normalization=True,
+        scaling_factor=None):
         """Map data to the unstructured mesh element centroids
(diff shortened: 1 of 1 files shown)


pshriwise:
Despite the well-implemented change here, I think that I'd recommend users simply modify the data being passed in and mapped to the centroids of the unstructured mesh. The application of the scaling factor can be performed just as easily on the incoming data before passing into the `write_data_to_vtk` function:

```python
# compute scaling factor

for d in datasets.values():
    d *= scaling_factor
 
 umesh.write_data_to_vtk('data.vtk', datasets)
```

pshriwise (commented):
Added a note on an alternative to this modification to the function.

Shimwell:
No need for this PR then, I shall scale the incoming data as suggested

PR #1909, opened 2021-11-16
Removing Legendre filter in diffusion coefficient results

mkreher13:
@nelsonag based on our conversation, here is my solution:

Fixed issue where both P0 and P1 moments were displayed in the diffusion coefficient results. With this fix, the Legendre filter is successfully sliced so all results display only the P1 results. My fix involves modeling the code more closely after the TransportXS class. 

I deleted the method for `get_condensed_xs` in the `DiffuionCoefficient` class since it wasn't working. Somehow, I failed to notice that the results from `get_condensed_xs` weren't condensed! Upon further investigation, I noted that the function just didn't work. 

Since the `MGXS` `get_condensed_xs` method condenses the tallies before calculating the cross section, that is compatible with diffusion coefficient condensation. The most important thing is that the diffusion coefficient not be condensed, and instead have the transport cross section condensed, and then the diffusion coefficient calculated based on that condensed transport cross section. By condensing the tallies involved in the diffusion coefficient calculation, that is exactly what happens. Therefore, using the original `get_condensed_xs` method in the `MGXS` class instead makes sense and is accurate. I tested this on a simple example by making a mgxs_library that had 'transport' and 'diffusion-coefficient' specified. After running, I loaded the data and ensured that the condensed value of the diffusion coefficient was equal to 1/(3* condensed transport). 

I also tested my changes in my personal branch of transient code that leverages the diffusion-coefficient. After my changes, the results were identical and my transient tests passed. 

All mgxs_library test results have been updated.


gridley:
"The most important thing is that the diffusion coefficient not be condensed, and instead have the transport cross section condensed, and then the diffusion coefficient calculated based on that condensed transport cross section."

are you sure about that? I could have sworn that despite it being counterintuitive, condensing 1/(3sigma_tr) preserves leakage better.

bforget:
Gavin is correct on this.  Preserving 1/Sigma_tr preserves the flux weighted mean free path.  Typically, lattice codes will spatially homogenized Sigma_tr and energy condense 1/Sigma_tr.  This is only done to facilitate dealing with voids.  Not super trivial to do in a single tally in MC, so you would need to post-process the energy condensation.  That was the entire motivation behind doing the migration area work.

mkreher13:
You're totally right (The rule is "don't condense transport cross sections" which I always mis-remember as diffusion coefficients!!). Anyway, here, it's all the tallies that are condensed. That includes total and scatter-1 which are then used in the transport cross section. Consequently, the transport correction is re-calculated every time you condense. This is as good as it's going to get without doing migration area, since it's as if you tallied everything with the coarser group structure.

nelsonag:
@mkreher13 I took a quick look now and it looks good. I compared diffusion/nu-diffusion and the transport/nu-transport results from the tests and showed they are related as we expect. In all cases that I could test, including the condensation test, this was true.

What I couldn't look at from here is the mgxs_library_nuclides test because that result is quite large and so our test reference is a hash. Could you (locally) convert that test to ASCII and check a sampling of values to ensure diffusion=1/(3*tr) and report back? 
No need to change the test to an ASCII result in the repository, just looking for confirmation that those values are correct.

mkreher13:
Thanks @nelsonag! I ran the mgxs_library_nuclides test with hashing turned off. I compared the values, and I can confirm that diffusion is 1/(3*tr).

nelsonag:
Great @mkreher13, that sounds good to me. I've reviewed the code and it looks good to me as-is.

One last thing though any idea why test_tally_aggregation is failing?

mkreher13:
It's the 8th decimal place on a single result that's off. This has been happening to me recently, sometimes it fails, sometimes it doesn't. I can't explain it.

nelsonag (approved):
Looks good to me, thanks @mkreher13

nelsonag:
Since this PR should have no impact on that particular test behavior, I will go ahead and make the merge.  Thanks again @mkreher13

PR #1913, opened 2021-11-22
Fixed parsing of xsdir entries with a continuation line

nelsonag:
The previous code would combine lines, and then append the combined lines to the original line, leaving with a duplicate of the 1st line. This super minor change avoids this.


paulromano (approved):
Thanks @nelsonag! I agree with your assessment and fix. We should probably add a test for this but I won't put that burden on you.

PR #1916, opened 2021-11-23
Condensing diffusion instead of transport

mkreher13:
I wanted to build on our discussion from the weekend. What was recently merged was a fix that took a completely broken diffusion coefficient condensing function and made it workable by condensing all the underlying tallies which amounts to condensing the transport cross section and then re-calculating the diffusion coefficient. In this case, condensed_diffusion = 1/(3*condensed_tr). 

However, per our conversation, it is more accurate to condense the diffusion coefficient instead (which amounts to condensing 1/tr). In this PR, I believe I've done that. As you can see in the results of mgxs_library_condense, condensed_diffusion is NOT equal to 1/(3*condensed_tr). 

I think a knowledgeable user will score the diffusion coefficient on a very fine energy grid, and then use the condense function to get the diffusion coefficient on the energy grid they actually want to use for the problem. 

The changes in xs_tally are necessary because when condensing, the diffusion coefficient is calculated, so you can't re-calculate it in xs_tally later (for instance, you can't re-slice the P1 scattering tally since it no longer exists after slicing it once). This change ensures that get_condensed_xs and xs_tally are compatible. 

I'm hoping for feedback. Did I make any mistakes? What does everybody think? @gridley @nelsonag

### DIFF
--- a/openmc/mgxs/mgxs.py
+++ b/openmc/mgxs/mgxs.py
@@ -3154,6 +3154,8 @@ class DiffusionCoefficient(TransportXS):
                 raise ValueError(msg)
 
             # Switch EnergyoutFilter to EnergyFilter
+            # If 'scatter-1' is not in tallies, it is because the transport correction has
+            # already occurred on this MGXS in another function or in a previous call to this function.
             if 'scatter-1' in self.tallies:
                 p1_tally = self.tallies['scatter-1']
                 old_filt = p1_tally.filters[-2]
@@ -3176,7 +3178,7 @@ class DiffusionCoefficient(TransportXS):
 
         return self._xs_tally
 
-    def get_condensed_xs(self, coarse_groups, condense_diff_coef=True):
+    def get_condensed_xs(self, coarse_groups):
         """Construct an energy-condensed version of this cross section.
 
         Parameters
@@ -3202,7 +3204,9 @@ class DiffusionCoefficient(TransportXS):
         # Clone this MGXS to initialize the condensed version
         condensed_xs = copy.deepcopy(self)
 
-        if condense_diff_coef:
+        # If 'scatter-1' is not in tallies, it is because the transport correction has
+        # already occurred on this MGXS in another function or in a previous call to this function.
+        if 'scatter-1' in self.tallies:
             p1_tally = self.tallies['scatter-1']
             old_filt = p1_tally.filters[-2]
             new_filt = openmc.EnergyFilter(old_filt.values)
--- a/openmc/mgxs/mgxs.py
+++ b/openmc/mgxs/mgxs.py
@@ -3154,30 +3154,133 @@ class DiffusionCoefficient(TransportXS):
                 raise ValueError(msg)
 
             # Switch EnergyoutFilter to EnergyFilter
+            if 'scatter-1' in self.tallies:
+                p1_tally = self.tallies['scatter-1']
+                old_filt = p1_tally.filters[-2]
+                new_filt = openmc.EnergyFilter(old_filt.values)
+                p1_tally.filters[-2] = new_filt
+
+                p1_tally = p1_tally.get_slice(filters=[openmc.LegendreFilter], 
+                    filter_bins=[('P1',)],squeeze=True)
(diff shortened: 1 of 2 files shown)


openmc/mgxs/mgxs.py
```diff
@@ -3154,30 +3154,133 @@ def xs_tally(self):
...
+        cv.check_value('upper coarse energy', coarse_groups.group_edges[-1],
+                       [self.energy_groups.group_edges[-1]])
+        cv.check_value('lower coarse energy', coarse_groups.group_edges[0],
+                       [self.energy_groups.group_edges[0]])
+
+        # Clone this MGXS to initialize the condensed version
+        condensed_xs = copy.deepcopy(self)
+
+        if condense_diff_coef:
             p1_tally = self.tallies['scatter-1']
```
nelsonag, line 3206: Would this not fail if the user tried condensing an MGXS object already condensed? (back to the above: `self.tallies['scatter-1']` wouldnt be present.

tests/regression_tests/mgxs_library_condense/results_true.dat
```diff
@@ -214,16 +214,16 @@
 28      2  2  y-min out        1   total  4.548  0.156691
   mesh 1       group in nuclide      mean std. dev.
        x  y  z                                     
-0      1  1  1        1   total  0.757948  0.035459
-2      1  2  1        1   total  0.779112  0.047034
-1      2  1  1        1   total  0.787475  0.050368
-3      2  2  1        1   total  0.769656  0.058065
+0      1  1  1        1   total  0.931945  0.045510
```
nelsonag, line 217: I presume you verified these new values with hand-calculations of the pre-condensed results?

nelsonag (changes requested):
Hi @mkreher13! Sorry for the delay on this one. I have attached some comments which should be addressed. But at a bigger picture: I dont quite like the idea of `self.tallies` being different depending on if it is a previously-condensed MGXS or not (if I'm understanding correctly).

Is there a clean way to create a cloned MGXS object, compute xs_tally on that, and then condense the clone?  That way the original is 'untouched'. Thanks!

mkreher13:
Sorry for waiting 9 months to address your comments, as I have had other priorities. I am circling back to it as I am wrapping up my PhD, and I need good diffusion coefficients for my thesis. This piece of code is actually pretty important, as it is crucial to be able to finely tally the transport cross section and then condense the diffusion coefficients. 

I addressed your concerns by including a comment that described the role of the if-statement relating to `'scatter-1'.` If `'scatter-1'` is not in `self.tallies`, it is not because the MGXS has been condensed, but because the transport correction has already been applied. If that is the case, it is safe to jump ahead to the main body of the calculation. Otherwise, the transport correction needs to be done first. 

You were correct that if `'scatter-1'` is not present, you couldn't re-condense a condensed diffusion coefficient. Essentially, you can always compute multiple condensations directly from the original MGXS by using:
`mgxs_lib.get_mgxs(cell, 'diffusion-coefficient').get_condensed_xs(coarse_groups)` several times in a row with coarser and coarser groups.
But you couldn't do:
`condensed_dc = mgxs_lib.get_mgxs(cell, 'diffusion-coefficient').get_condensed_xs(coarse_groups)`
`condensed_dc .get_condensed_xs(coarser_groups)`

To fix this in `get_condensed_xs()`, I added the if-statement relating to `'scatter-1' `which again ensures that the transport correction is done before jumping into the main calculation. Now there is no issue stringing multiple `get_condensed_xs()` calls one after another. 

I also removed the variable `condense_diff_coef`. That variable existed to give the user the option to condense the underlying tallies rather than the diffusion coefficients themselves. My colleague pointed out that if that option is not "good physics" then it shouldn't be an option. So by removing it, the user has to use the better option of condensing the diffusion coefficients instead. 

I also did a few hand calculations to make sure that the condensation was correct.

nelsonag (approved):

nelsonag:
Thanks @mkreher13 and apologies for the late review! Good luck with the remainder of your dissertation!

PR #1918, opened 2021-12-07
correct spelling of my name

jeffhammond:
i hope this does not need any further explanation


paulromano (approved):
Oops! Thanks for the fix @jeffhammond

PR #1920, opened 2021-12-07
Change errors to warnings in plots. 

### LINKED ISSUE #1919: Change errors to warnings for plotting
When running OpenMC in plot mode, you will currently get a fatal error if there's any materials to which you've assigned colors, but didn't end up in the `materials.xml`. However, if you separate out material definitions into a common module where you also assign colors, you can get errors depending on the end model you eventually build - for instance, if you optionally add a cell to the geometry depending on other settings. It can be tedious to keep track of the color assignments as a function of how the model changes.

We can change the fatal errors in `plot.cpp` to warnings instead to be more forgiving of cases where perfect matches between color assignment and materials is tedious/difficult to achieve (such as when cloning materials). Conceptually, I see assigning colors to non-existent materials to be philosophically similar to `-Wunused`-type flags -- warn about it, but don't error.

aprilnovak:
Changes errors to warnings for flexibility.

Closes #1919

### DIFF
--- a/src/plot.cpp
+++ b/src/plot.cpp
@@ -433,7 +433,7 @@ void Plot::set_user_colors(pugi::xml_node plot_node)
         col_id = model::cell_map[col_id];
         colors_[col_id] = user_rgb;
       } else {
-        fatal_error(fmt::format(
+        warning(fmt::format(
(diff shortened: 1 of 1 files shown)


paulromano (approved):
Makes sense -- thanks @aprilnovak!

PR #1926, opened 2021-12-14
Projection plots

gridley:
So, this lets users define a new type of plot that I am nominally calling a projection plot (open to rename it). It loops over pixels from a camera location to a point specified in the model. A reasonable field of view is set by default. When doing the loop over pixels, it's conceptually easy to draw a wireframe since you just check if the stack of intersected cells differs from one pixel to the next.

TODO

- [x] Hook up python API to generate XML
- [x] Add documentation
- [x] Add orthographic projection (as opposed to perspective)
- [x] Parallelize with OpenMP
- [x] Up direction is currently only +z. Should be arbitrary.
- [x] Should read user-given plot names 
- [x] Improve wireframe generation. Is currently patchy

Example from newly included regression tests:

anim

BEAVRS wireframe drawing. It gets a bit dense. Probably need some way to turn off the generation of lines in certain universes.

projection2


makeclean:
A really nice addon for this, would be to colour by the thickness of (non-vacuum) i.e. the total distance travelled by rays, useful for looking at shielding weak spots

gridley:
@makeclean Through any non-vacuum material? It would be easy to do in principle. In fact, trivial, since this records the stack of material IDs getting passed through as the ray moves, then colors in reverse.

Do you think it would be better to color based on thickness through a material or cell of given ID?

The only tricky thing would be that we'd have to define a color map, which is plenty of extra coding and hardcoding (e.g. adding viridis coloring).

pshriwise:
Really awesome capability @gridley! 💯

gridley:
The projection is looking a lot better now. There are two modes:

Perspective:

persp


Orthographic:

orthographic

gridley:
OK, so, it was a little tough to get this parallelized cleanly. There are two reasons why naive omp parallel for doesn't work here.

1) omp parallel for, if applied to the vertical pixel coordinate loop, breaks the image into n short, long rectangles. On my 12 thread computer, the top rectangle takes about 1/4 as long to draw as the middle rectangles for BEAVRS. Most other problems would look like this: there is more complexity to the geometry in the middle of the image and vacuum surrounding it. So the default parallelization is naturally unbalanced and thus scales poorly.
2) Drawing the wireframe requires inter-thread communication in some way, since we have to compare pixel's intersected stack of cell/surface IDs both horizontally and vertically. The standard omp parallel for would require some wacky caching of these intersection stacks to get the same results as the single thread version.

Say n is the number of threads. To solve both of the above, have thread 0 handles lines 0, n, 2n etc. Thread 1 gets lines 1, n+1, 2n+1, etc. After each thread finished rendering a line, you have it compare the cell/surface intersection stack with its upward neighbor. For thread 0, this comparison comes from the last result that thread n-1 obtained. For all others, thread i compares against thread i-1. To do these comparisons correctly, thread barriers have to be used.

This does give reproducibility against single-thread results, and seems to give 75-80% parallel efficiency on drawing BEAVRS for a 200x200 image. I'm a little novice on the OMP stuff so maybe there are some inefficiencies sticking out like a sore thumb. For instance there is definitely a little false sharing on the variable `this_line_segments`, but I think the cost of that is probably negligible compared to the ray tracing work each thread has to do.

pshriwise:
I'm really itching to try this out and look it over, but I'll wait until you give us the heads up that it's ready. Until then I'll mark it as a draft so you don't get any review comments from others while you're tidying up.

gridley:
OK, so, I've got it set up where you can wireframe only select material IDs now. From the hex lattice example:
hex

gridley:
@pshriwise We are now good to go on a review! Phew... there are so many little details here to fill out. The only thing that's missing, I think, is starting ray tracing in universes below the root universe. I have neither tested nor implemented that, and would like to save that for some other PR when it becomes necessary.

I also think the documentation could be improved a little bit, but am unsure how. I'm very open to rewording suggestions here.

gridley:
As one more note, I think perhaps the camera polar angle values for each of the pixels should maybe not be uniformly  spaced but rather be spaced according to cos(mu) or something. I would need to look at some computer graphics code to figure out how people conventionally set up the rays. Right now, it just creates uniform spacing in mu and phi over the camera's field of view.

gridley:
OK... and there is redundant code in the regression test I added. I wish I could import what's in regression_tests/plot, but pytest doesn't like to import from relative directories, it seems. I guess we'd have to move that code into some common file that they both import, but am unsure where to put that. It's the code that does a hash-based comparison on a list of png files.

gridley:
Alright, I had to resolve a merge conflict related to testing the recent to/from XML functionality that was added to the Plot class by Paul. I'll have to add corresponding to_xml/from_xml methods to the projection plots at this point, if the tests fail.

gridley:
@pshriwise just getting on this again. The latest commit *might* get tests passing again. I'm hoping to test this a bit more extensively before merging. In particular, I'm hoping to get this integrated with openmc-dev/plotter, somehow.

pshriwise:
@gridley looks like it did indeed fix it. Nice. Excited that this could make it into the plotter!

gridley:
Thanks @pshriwise. After looking at what would have to go in to get this working in the plotter, it would be maybe a bit too much work for a side project. My general idea was to display a pixmap rendered by OpenMC after image generation, and for navigating the model, we'd use an OpenGL window with points laid down in 3D that are where the wireframe was last drawn by OpenMC, which would be pretty easy to cache and make accessible in the C API. This way, there would be responsive rotation and panning on a low-quality model, then a user would tap "r" or something like that to redraw the good-looking version of the model.

As a result, I think this is ready for review now. Should I test it with DagMC stuff before we begin, though?

gridley:
@pshriwise is there any desire to get this merged from ANL? If not, I totally get it. Otherwise, if there's anything else I can do to move this along, please let me know.

pshriwise:
> @pshriwise is there any desire to get this merged from ANL? If not, I totally get it. Otherwise, if there's anything else I can do to move this along, please let me know.

There is from me! Sorry I haven't gotten around to it yet. Let's get it moving. I'll give it a try tonight and start my review tomorrow.

> Should I test it with DagMC stuff before we begin, though?

Yeah, I think that would be good. Have you given it a try?

pshriwise:
With a few tweaks, DAGMC geometry seems to work pretty well! Below is the Utah teapot from the CAD-based geometry example. I need to look into some unfortunate output in an upstream function though.

dagmc

docs/source/usersguide/plots.rst
```diff
@@ -121,3 +121,79 @@ will depend on the 3D viewer, but should be straightforward.
...
+----------------
+
+.. image:: ../_images/hexlat_anim.gif
+   :width: 200px
+
+The :class:`openmc.ProjectionPlot` class presents an alternative method
+of producing 3D visualizations of OpenMC geometries. It was developed to
+overcome the primary shortcoming of voxel plots, that an enormous number
+of voxels must be employed to capture detailed geometric features.
+Projection plots do volume rendering on material or
```
pshriwise, line 136:
```suggestion
Projection plots perform volume rendering on material or
```

docs/source/usersguide/plots.rst
```diff
@@ -121,3 +121,79 @@ will depend on the 3D viewer, but should be straightforward.
...
+cell volumes, with colors specified in the same manner as slice plots.
+This is done using the native ray tracing capabilities within OpenMC,
+so any geometry in which particles successfully run without overlaps
+or leaks will work with projection plots.
+
+One drawback of projection plots is that particle tracks cannot be overlaid
+on them at present. Moreover, checking for overlap regions as is possible
+with slice plots is also not available. The image heading this section can
```
pshriwise, line 144:
```suggestion
on them at present. Moreover, checking for overlap regions is not currently possible with projection plots. The image heading this section can
```

docs/source/usersguide/plots.rst
```diff
@@ -121,3 +121,79 @@ will depend on the 3D viewer, but should be straightforward.
...
+resemble those employed by :class:`openmc.Plot`, with the exception of
+the :class:`ProjectionPlot.set_transparent` method and :attr:`ProjectionPlot.xs`
+dictionary. These are used to control volume rendering of material
+volumes. "xs" here stands for cross section, and it defines material
+opacities in units of inverse centimeters. Setting this value to a
+large number would make a material or cell opaque, and setting it to
+zero makes a material transparent. Thus, the :class:`ProjectionPlot.set_transparent`
+can be used to make all materials in the geometry transparent. From there,
+individual material or cell opacities can be tuned to produce the
+desired result.
```
pshriwise, line 184: Does the `orthographic_width` attribute need to be mentioned here as well?

include/openmc/plot.h
```diff
@@ -192,61 +235,127 @@ T PlotBase::get_map() const
...
+    {}
+  };
+
+  std::array<int, 2> pixels_;       // pixel dimension of resulting image
+  double horizontal_field_of_view_ {70.0}; // horiz. f.o.v. in degrees
+  Position camera_position_;        // where camera is
+  Position look_at_;                // point camera is centered looking at
+  Direction up_ {0.0, 0.0, 1.0};    // which way is up
+  std::vector<int>
+    wireframe_ids_; // which color IDs should be wireframed. If empty, all
```
pshriwise, line 312: Looks like the end of this comment got cut off accidentally

pshriwise (changes requested):
Not a complete review by any means, but it’s a start! Mostly some comments on the new doc section and on the inheritance structure of the classes. Sorry it’s taken so long to get a start on this @gridley!

gridley:
Hey @pshriwise, sorry to go off the grid on this one, but it has been on my to-do list.

The doc updates all sound good to me and I'll get to that, and renaming `PlotInterface` to `PlotBase` is certainly appropriate. The inheritance structure you suggest originally made the most sense to me too, but when I got to writing the code like that, I found that I could considerably minimize code delta by going with the design as it stands.

Moving the 3D pixel array into the plot base class doesn't make sense to me, since it doesn't get used at all in the projection plots. I think changing the semantics of what `SlicePlotBase` is would be best. Rather than thinking about it as a plot, I think we should just consider it as a `Slice` that holds a pixel array of some sort of data in the problem. Then the overlap coloring and such can go back into the `Plot` and `VoxelPlot` classes.

Looking at it that way, it makes sense for a `SlicePlot` to inherit from both `PlotBase` and `Slice`. Maybe it could be named `PropertySlice` or something like that. Could be useful elsewhere, too.

pshriwise:
>Moving the 3D pixel array into the plot base class doesn't make sense to me, since it doesn't get used at all in the projection plots. I think changing the semantics of what SlicePlotBase is would be best. Rather than thinking about it as a plot, I think we should just consider it as a Slice that holds a pixel array of some sort of data in the problem. Then the overlap coloring and such can go back into the Plot and VoxelPlot classes.
>
>Looking at it that way, it makes sense for a SlicePlot to inherit from both PlotBase and Slice. Maybe it could be named PropertySlice or something like that. Could be useful elsewhere, too.

Yeah, you're right. I think keeping the data structures with their respective classes is the right call. It would be nice to factor out the voxel plot into its own class, but I agree that limiting the size of the delta is probably more important for now. That can be done later.

src/plot.cpp
```diff
@@ -980,16 +1019,552 @@ RGBColor random_color(void)
...
+      if (tid == n_threads - 1)
+        old_segments = this_line_segments[n_threads - 1];
+
+      if (vert < pixels_[1]) {
+
+        for (int horiz = 0; horiz < pixels_[0]; ++horiz) {
+
+          // Generate the starting position/direction of the ray
+          if (orthographic_width_ == 0.0) { // perspective projection
+            double this_phi = -horiz_fov_radians / 2.0 + dphi * horiz;
```
pshriwise, line 1269: Is this setting up particles to move through the edge or center of a pixel? It looks like edge to me.

src/plot.cpp
```diff
@@ -980,16 +1019,552 @@ RGBColor random_color(void)
...
+            s.r = camera_position_ +
+                  cam_yaxis * x_pix_coord * orthographic_width_ +
+                  cam_zaxis * y_pix_coord * orthographic_width_;
+          }
+
+          p.from_source(&s); // put particle at camera
+          bool hitsomething = false;
+          bool intersection_found = true;
+          int loop_counter = 0;
+          const int max_intersections = 1000000;
```
pshriwise, line 1290: There are a couple of constants in here that are a little hidden. What would you say to making them members of the class?

src/plot.cpp
```diff
@@ -980,16 +1019,552 @@ RGBColor random_color(void)
...
+
+          this_line_segments[tid][horiz].clear();
+
+          int first_surface =
```
pshriwise, line 1294: It seems like factoring out the first intersection would allow you to get rid of some of this state that's only used in the first case in the loop.

pshriwise: Also, is it possible to exit and re-enter the model once the particle is inside?

src/plot.cpp
```diff
@@ -980,16 +1019,552 @@ RGBColor random_color(void)
...
+                dist.distance, std::abs(dist.surface_index));
+
+              // Advance particle
+              for (int lev = 0; lev < p.n_coord(); ++lev) {
+                p.coord(lev).r += dist.distance * p.coord(lev).u;
+              }
+              p.surface() = dist.surface_index;
+              p.n_coord_last() = p.n_coord();
+              p.n_coord() = dist.coord_level;
+              if (dist.lattice_translation[0] != 0 ||
```
pshriwise, line 1327: I think you can also check the lowest coordinate of particle for `lattice >= 0`.

src/plot.cpp
```diff
@@ -980,16 +1019,552 @@ RGBColor random_color(void)
...
+        if (wireframe_thickness_ == 1)
+          data(horiz, vert) = wireframe_color_;
+        for (int i = -wireframe_thickness_ / 2; i < wireframe_thickness_ / 2;
+             ++i)
+          for (int j = -wireframe_thickness_ / 2; j < wireframe_thickness_ / 2;
+               ++j)
+            if (i * i + j * j < wireframe_thickness_ * wireframe_thickness_) {
+              if (horiz + i >= 0 && horiz + i < pixels_[0] && vert + j >= 0 &&
+                  vert + j < pixels_[1])
+                data(horiz + i, vert + j) = wireframe_color_;
```
pshriwise, line 1420: There might be some redundant setting of values entries for wireframes on the boundary, but this is a cleaner to me. Feel free to take or leave it.

```suggestion
              int w_i = std::max(std::min(horiz + i, pixels_[0]), 0);
              int w_j = std::max(std::min(vert + j, pixels_[1]), 0);
              data(i, j) = wireframe_color_;
```

pshriwise (changes requested):
Added some more thoughts/comments for you @gridley. Sorry it's taken me so long to get back to this. It's been fun to return to it! A lot of potential here.

One thing I'll request specifically here is some Python --> XML --> Python --> XML round-trip tests. I was losing the custom output filenames after a Python import/export and I'm wondering if there might be some other info that gets dropped as well.

pshriwise:
For got to add a comment before submitting my review...

What do you think of performing the wireframe update entirely separately from the parallel loop? If I understand correctly, the initial wireframe pass can still be performed in parallel under a separate loop. In addition to adding clarity and simplifying that main loop a bit, separating it might also make it easier to examine in case there is a way to parallelize the second pass too.

gridley:
> For got to add a comment before submitting my review...
> 
> What do you think of performing the wireframe update entirely separately from the parallel loop? If I understand correctly, the initial wireframe pass can still be performed in parallel under a separate loop. In addition to adding clarity and simplifying that main loop a bit, separating it might also make it easier to examine in case there is a way to parallelize the second pass too.

Hey Patrick, thanks for reviewing further! I'm going to get to this stuff after taking a final exam and finishing some class projects, but did want to address this.

The reason it's set up this way is because the memory usage in a two pass approach like that is excessively high. True, you could first loop through all pixels and cache the stack of cell intersections first, then loop over the intersection stacks to compute wireframe locations. That's not what I implemented here though: you only need n+1 lines of pixels where n is the number of threads in use.

This is because I calculated that the two loop approach can take up WAY too much memory for a model of interesting size. Suppose you have on average 50 cell intersections per pixel. Then for a 1920x1080 image, that's about 1.5 GB of data. For large models, it's easy to imagine a scenario where you run out of memory.

Maybe, though, I could add some developer documentation explaining the reasoning for the implemented approach. I could provide a pseudocode implementation of what your'e suggesting, then explain why that's not what's done in practice. Instead, you only need to know the adjacent horizontal line of pixels to correctly find the wireframe locations.

pshriwise:
> The reason it's set up this way is because the memory usage in a two pass approach like that is excessively high. True, you could first loop through all pixels and cache the stack of cell intersections first, then loop over the intersection stacks to compute wireframe locations. That's not what I implemented here though: you only need n+1 lines of pixels where n is the number of threads in use.


Ah, yeah that makes sense. Kinda forgot about processing the image lines in chunks. I'm going to think a bit more about this. Mainly, I think it would be nice to simplify the code in that loop a bit if possible 🤔

pshriwise:
> Maybe, though, I could add some developer documentation explaining the reasoning for the implemented approach. I could provide a pseudocode implementation of what your'e suggesting, then explain why that's not what's done in practice. Instead, you only need to know the adjacent horizontal line of pixels to correctly find the wireframe locations.

Additional comments never hurt 😄, but probably don't need to include pseudo-code or a justification. At least not yet anyway -- I'd like to monkey with the ordering and maybe data structures a bit first if that's okay with you.

gridley:
> > Maybe, though, I could add some developer documentation explaining the reasoning for the implemented approach. I could provide a pseudocode implementation of what your'e suggesting, then explain why that's not what's done in practice. Instead, you only need to know the adjacent horizontal line of pixels to correctly find the wireframe locations.
> 
> Additional comments never hurt smile, but probably don't need to include pseudo-code or a justification. At least not yet anyway -- I'd like to monkey with the ordering and maybe data structures a bit first if that's okay with you.

Oh, if you come up with something simpler, please tear it to shreds! :smile:

gridley:
Was making some input for a graphite pile today. Here's another example plot that I got from it.

image

src/plot.cpp
gridley: Yeah, I forgot about this assumption I made! Are OpenMC models assumed to be convex? If we're simulating radiation transport, it seems like a bad model to ever not be convex as it's clearly physically incorrect: you'd be artificially killing particles that may be streaming into some other place you'd like to tally.

Maybe there is some exception to this where reflective BCs would be employed, but I can't imagine it right now. Maybe it would make sense to note in the documentation that the model must be convex rather than allowing this edge case? I think the logic might be complicated to handle reentrant geometries.

pshriwise: That's a good point about the model itself. We don't disallow non-convex model boundaries, but it certainly could cause some modeling problems you've identified here. Moreso I'm wondering if there are CSG cells inside the model will we get the correct answer? I suppose so because we're using the same logic for particles undergoing transport and assuming we're already inside a cell for the ray tracing logic.

gridley:
@pshriwise have you monkeyed with the ordering and data structures yet? I believe we had talked about it briefly on slack, and my position is that changing up the ordering to make the code more readable can use excessive memory.

docs/source/usersguide/plots.rst
gridley: New commit addresses this.

include/openmc/plot.h
```diff
@@ -66,6 +67,50 @@ struct RGBColor {
 // some default colors
 const RGBColor WHITE {255, 255, 255};
 const RGBColor RED {255, 0, 0};
+const RGBColor BLACK {0, 0, 0};
+
+/*
+ * PlottableInterface classes just have to have a unique ID in the plots.xml
+ * file, and guarantee being able to create output in some way.
+ */
+class PlottableInterface {
```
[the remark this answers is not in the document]
gridley: We have discusses this at this point. The current approach is the minimum code delta. I'm happy to refactor in a followup PR. Do you think it's absolutely necessary to refactor like this?

include/openmc/plot.h
gridley: fixed!

include/openmc/plot.h
```diff
@@ -192,61 +235,127 @@ T PlotBase::get_map() const
...
+
+  RGBColor wireframe_color_ {BLACK}; // wireframe color
+  std::vector<double>
+    xs_; // macro cross section values for cell volume rendering
+
+  /* If starting the particle from outside the geometry, we have to
+   * find a distance to the boundary in a non-standard surface intersection
+   * check. It's an exhaustive search over surfaces in the top-level universe.
+   */
+  static int advance_to_boundary_from_void(Particle& p);
```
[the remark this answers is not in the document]
gridley: done!

include/openmc/plot.h
```diff
@@ -192,61 +235,127 @@ T PlotBase::get_map() const
...
+
+  /* Checks if a vector of two TrackSegments is equivalent. We define this
+   * to mean not having matching intersection lengths, but rather having
+   * a matching sequence of surface/cell/material intersections.
+   */
+  bool trackstack_equivalent(const std::vector<TrackSegment>& track1,
+    const std::vector<TrackSegment>& track2) const;
+
+  // Closed form 3x3 matrix inversion
+  static std::vector<double> invert_matrix(const std::vector<double>& input);
```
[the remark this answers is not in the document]
gridley: Turns out this wasn't necessary anyway :smile:

openmc/plots.py
```diff
@@ -175,15 +175,30 @@ def _get_plot_image(plot, cwd):
 
     return Image(str(png_file))
 
+def get_tuple(elem, name, dtype=int):
```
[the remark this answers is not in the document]
gridley: both addressed now.

openmc/plots.py
```diff
@@ -175,15 +175,30 @@ def _get_plot_image(plot, cwd):
...
 
-class Plot(IDManagerMixin):
-    """Definition of a finite region of space to be plotted.
+    Parameters
+    ----------
+    elem : xml.etree.ElementTree.Element
+        XML element that should contain a tuple
+    name : str
+        Name of the subelement to obtain tuple from
+    dtype : data-type
```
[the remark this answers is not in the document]
gridley: Hm, after I rebased, I don't see "dtype" appearing anywhere in plots.py. So not sure what to do here.

openmc/plots.py
```diff
@@ -800,13 +832,333 @@ def to_ipython_image(self, openmc_exec='openmc', cwd='.'):
...
+        same direction. The width of the starting array must be specified,
+        unlike with the default perspective projection. The height of the
+        array is deduced from the ratio of pixel dimensions for the image.
+
+    wireframe_thickness : int
+        Line thickness employed for drawing wireframes around cells or
+        material regions. Can be set to zero for no wireframes
+    wireframe_color : tuple of ints
+        RGB color of the wireframe lines
+    wireframe_regions : iterable of either Material or Cells
```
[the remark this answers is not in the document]
gridley: OK, got this incorporated!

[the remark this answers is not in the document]
gridley: In the C++ and XML, I have it named `wireframe_ids`. Is this inconsisteny too egregious?

openmc/plots.py
```diff
@@ -800,13 +832,333 @@ def to_ipython_image(self, openmc_exec='openmc', cwd='.'):
...
+        If provided, the wireframe is only drawn around these.
+        If color_by is by material, it must be a list of materials, else cells.
+
```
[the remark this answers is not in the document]
gridley: got this in the new commit.

openmc/plots.py
```diff
@@ -800,13 +832,333 @@ def to_ipython_image(self, openmc_exec='openmc', cwd='.'):
...
+    def __init__(self, plot_id=None, name=''):
+        # Initialize Plot class attributes
+        super().__init__(plot_id, name)
+        self._horizontal_field_of_view = 70.0
+        self._camera_position = (1.0, 0.0, 0.0)
+        self._look_at = (0.0, 0.0, 0.0)
+        self._up = (0.0, 0.0, 1.0)
+        self._orthographic_width = 0.0
+        self._wireframe_thickness = 1
```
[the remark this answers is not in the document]
gridley: done!

openmc/plots.py
```diff
@@ -800,13 +832,333 @@ def to_ipython_image(self, openmc_exec='openmc', cwd='.'):
...
+        if self.color_by == 'material':
+            domains = geometry.get_all_materials().values()
+        else:
+            domains = geometry.get_all_cells().values()
+
+        # Generate random colors for each feature
+        for domain in domains:
+            self.xs[domain] = 0.0
+
+    def to_xml_element(self):
```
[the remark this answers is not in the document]
gridley: Nice comment, should be consolidated in the new commit.

openmc/plots.py
```diff
@@ -800,13 +832,333 @@ def to_ipython_image(self, openmc_exec='openmc', cwd='.'):
...
+                subelement.set("xs", str(self._xs[domain]))
+
+        if self._level is not None:
+            subelement = ET.SubElement(element, "level")
+            subelement.text = str(self._level)
+
+        return element
+
+    @classmethod
+    def from_xml_element(cls, elem):
```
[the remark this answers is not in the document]
gridley: done

src/plot.cpp
```diff
@@ -97,7 +97,7 @@ void PropertyData::set_overlap(size_t y, size_t x)
 namespace model {
 
 std::unordered_map<int, int> plot_map;
-vector<Plot> plots;
+std::vector<std::unique_ptr<PlottableInterface>> plots;
```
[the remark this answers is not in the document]
gridley: done

src/plot.cpp
```diff
@@ -405,28 +454,21 @@ void Plot::set_default_colors(pugi::xml_node plot_node)
...
-void Plot::set_user_colors(pugi::xml_node plot_node)
+void PlottableInterface::set_user_colors(pugi::xml_node plot_node)
 {
-  if (!plot_node.select_nodes("color").empty() && PlotType::voxel == type_) {
-    if (mpi::master) {
-      warning(
-        fmt::format("Color specifications ignored in voxel plot {}", id_));
-    }
-  }
-
```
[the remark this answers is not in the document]
gridley: I deleted these because it's hard to check this in the base class. I'd rather just see the colors not get used. Also, the consequence of not getting warned about this is minimal--was there ever any expectation that colors would carry through to a voxel plot?

src/plot.cpp
gridley: I only see max intersections as a constant. What others are you talking about? The others are completely arbitrary and could be set to any value. I think.

gridley: anyway, i did this for max_intersections!

gridley:
OK, the new commit should have addressed most of those concerns! :smile:

gridley:
Ah, I may have accidentally force pushed and removed those dagmc changes you had incorporated, Patrick. I can try to rebase the latest commits on that stuff. Sorry, totally forgot about that stuff.

gridley:
@pshriwise alright, I cherry-picked your dagmc hacks in, but one of them entails commenting out a section that marks particles as lost. Obviously that's not something we want to merge to develop, so it's up to you how we can handle this.

Do you recall why that was necessary? It seems to me that this is getting triggered only after a ray has left the geometry. In the case of the CSG intersections, we are just returning an infinite distance and invalid surface index. Maybe the dag cell should do the same?

pshriwise:
> @pshriwise alright, I cherry-picked your dagmc hacks in, but one of them entails commenting out a section that marks particles as lost. Obviously that's not something we want to merge to develop, so it's up to you how we can handle this.
> 
> Do you recall why that was necessary? It seems to me that this is getting triggered only after a ray has left the geometry. In the case of the CSG intersections, we are just returning an infinite distance and invalid surface index. Maybe the dag cell should do the same?

Thanks for adding those changes in! And yes, I think the behavior you're suggesting is more consistent with the CSG cells. 👍🏻 

I'm trying out the projection plots with DAGMC geometry now with that update. Looks pretty nice!
image

gridley:
> Thanks for adding those changes in! And yes, I think the behavior you're suggesting is more consistent with the CSG cells. 👍🏻

Beautiful plots! So do you want me to create a commit that changes that behavior in dagmc?

pshriwise:
> > Thanks for adding those changes in! And yes, I think the behavior you're suggesting is more consistent with the CSG cells. 👍🏻
> 
> Beautiful plots! So do you want me to create a commit that changes that behavior in dagmc.cpp?

I've been meaning to add a test using an intentionally broken DAGMC file and this is a good prompt for me to do so. As long as you don't mind one more rebase (shouldn't overlap here too much I don't think). I'll submit a separate PR to this effect.

gridley:
Sounds good! 😃

I'll give dagmc a try one of these days...

pshriwise:
@gridley I'd take a look at #2434 for an explanation as to why we need to keep that behavior in the `DAGCell::distance` method. 

For now, we may just skip marking the particle as lost if OpenMC is running in plotting mode.

gridley:
I think maybe we're good to go here @pshriwise?

src/plot.cpp
pshriwise: Did we figure out which this is doing?

src/plot.cpp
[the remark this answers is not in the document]
pshriwise: No it wasn't an expectation, but it was a nice clarification for users so they don't expect their color specs to show up in paraview/visit or something like that.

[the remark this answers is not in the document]
pshriwise: I think we could wrap the `set_user_color` call in the `SlicePlotBase` class to keep this check though, no?

src/plot.cpp
pshriwise: Are we leaving this then, @gridley?

openmc/plots.py
```diff
@@ -800,13 +832,333 @@ def to_ipython_image(self, openmc_exec='openmc', cwd='.'):
         # Return produced image
         return _get_plot_image(self, cwd)
 
+class ProjectionPlot(PlotBase):
```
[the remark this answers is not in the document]
pshriwise: This class still needs a method from what I can see

pshriwise:
> I think maybe we're good to go here @pshriwise?

Very close! Just a few more comments to resolve potentially

src/plot.cpp
gridley: This is through edges. I see how going through centers might be desirable, do you want that? It should be imperceptible, though.

src/plot.cpp
gridley: eh, sure, I'll commit it! Your code is probably more idiomatic.

src/plot.cpp
gridley: What if the particle is in a lattice, but is not crossing a lattice boundary? Wouldn't the lowest coordinate then have the lattice variable defined, but it wouldn't be appropriate to call cross_lattice?

gridley:
Cool, I think I've replied to all your comments now :smile:

src/plot.cpp
gridley: Eh, again, sure, I'll change it to go through centers!

gridley: OK, new commit and regolded tests go through pixel centers. You can see a slight difference in one of the images, but it's pretty negligible.

openmc/plots.py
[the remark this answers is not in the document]
gridley: should be good now!

src/plot.cpp
pshriwise: Oh yeah totally. My bad!

src/plot.cpp
pshriwise: Thanks for making that adjustment! It's generally preferable (at least in projection plots) when model extents or surface bounding boxes are used to set plot values so that the samples aren't right on the boundary of a surface -- makes it less likely we'll run into numerical artifacts in the plots. It's nice to be consistent in that approach here as well even though, as you stated, it shouldn't be noticeable for these plots.

pshriwise (approved):
Thanks for your patience waiting on reviews and for sticking with this @gridley! I'll merge once CI looks good 👍🏻

gridley:
Yay! :smile:

pshriwise:
Looking forward to using these! Thanks @gridley!

PR #1953, opened 2022-01-19
Remove `MAX_WORD_LEN` and `MAX_LINE_LEN` constants.

pshriwise:
I was going through our constants today and noticed that these two were only used in a few places in the multi-group cross section code and realized we don't need them anymore. Getting rid of `MAX_WORD_LEN` required a small change in the function signature of `get_name` and `MAX_LINE_LEN` isn't needed with the right call to `read_attribute`.


paulromano (approved):
Oh man, those are some really ancient constants. Thanks for getting rid of them!

PR #1954, opened 2022-01-21
Moving some single-use constants into relevant header files.

### LINKED ISSUE #1952: Move single-use constants closer to point of use.
There are a few constants (`MAX_EVENTS`, `FREE_GAS_THRESHOLD`, etc.) in `constants.h` that are only used in one location elsewhere. We'll want to move those constants into relevant headers so they are closer to the code where they are used and minimize recompilation if these values are altered for debugging purposes or otherwise.

pshriwise:
Resolves #1952 

Some of the physical constants that were single-use I left in `constants.h` since I couldn't imagine a case where someone would want to be tweaking those values and recompiling frequently.

The `PARTIAL_FISSION_MAX` constant didn't seem to be used at all, so I removed it as well. If that was a mistake, I'll gladly put it back where it belongs.


paulromano (approved):
Simple enough. Thanks @pshriwise!

PR #1958, opened 2022-02-01
Move satisfy_triggers reset into openmc_reset. 

### LINKED ISSUE #1957: Reset triggers in openmc_reset
If running multiple OpenMC simulations in-memory, the `openmc::simulation::satisfy_triggers` gets set to true at the end of the previous simulation. This currently doesn't get reset to false when starting the next iteration, so any subsequent OpenMC solves kick out at just a single batch because OpenMC thinks the triggers have been satisfied (i.e. based on the previous iteration).

Because a simulation begins by calling `openmc_simulation_init`, which calls `openmc_reset`, we should move the `satisfy_triggers = false` into the `openmc_reset` routine.

aprilnovak:
Move the reset of triggers into `openmc_reset`, which is called through the various C-APIs in Cardinal.

Closes #1957

### DIFF
--- a/src/finalize.cpp
+++ b/src/finalize.cpp
@@ -119,7 +119,6 @@ int openmc_finalize()
   simulation::keff = 1.0;
   simulation::n_lost_particles = 0;
   simulation::need_depletion_rx = false;
-  simulation::satisfy_triggers = false;
   simulation::total_gen = 0;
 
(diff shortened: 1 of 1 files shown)


paulromano (approved):

paulromano (approved):
Thanks @aprilnovak. Will merge once tests are finished running

PR #1983, opened 2022-02-25
fixing logic for spherical mesh string repr

eepeterson:
I ran into a bug when trying to print some info on weight windows using a spherical mesh. The truth value of an array is ambiguous and raises an error. This PR fixes that.


pshriwise (approved):
Thanks for cleaning this up, @eepeterson!

PR #1997, opened 2022-03-10
adding helper method for EnergyFilter creation

eepeterson:
I found myself often grabbing a group structure from `openmc.mgxs.GROUP_STRUCTURES` and using it to create an `EnergyFilter` so I simply added a helper method so this can be done a little easier.


paulromano (commented):
Good idea! Do you want to add a quick unit test for this? Maybe just create an energy group structure using this classmethod and make sure it has the expected number of groups.

eepeterson:
Yep, can do.

paulromano (approved):
Thanks! Will merge when CI gives the green light

PR #2005, opened 2022-03-16
adding centroids and meshgrid methods to mesh classes

eepeterson:
This PR adds helper methods to the `openmc.MeshBase` subclasses that should be helpful in post-processing. Since I find myself reconstructing mesh coordinates in plotting scripts all the time, I thought it would be nice to get this directly from the mesh itself. This is also the first PR in a series of mesh API related changes I would like to implement to make mesh handling more uniform for some of the shutdown dose rate workflows we will be implementing soon. 

To keep this PR from changing too many things at once, the follow on PRs will include things like:
- adding grid arguments to constructors of all mesh classes
- adding translation and rotation methods to mesh classes

### DIFF
--- a/openmc/mesh.py
+++ b/openmc/mesh.py
@@ -126,8 +126,8 @@ class MeshBase(IDManagerMixin, ABC):
             raise ValueError(f'Unrecognized mesh type "{mesh_type}" found.')
 
 
-class StructuredMesh(ABC):
-    """A mixin for structured mesh functionality
+class StructuredMesh(MeshBase):
+    """A base class for structured mesh functionality
 
     Parameters
     ----------
@@ -194,8 +194,7 @@ class StructuredMesh(ABC):
         return (vertices[s0] + vertices[s1]) / 2
 
 
-
-class RegularMesh(StructuredMesh, MeshBase):
+class RegularMesh(StructuredMesh):
     """A regular Cartesian mesh in one, two, or three dimensions
 
     Parameters
@@ -630,7 +629,7 @@ def Mesh(*args, **kwargs):
     return RegularMesh(*args, **kwargs)
 
 
-class RectilinearMesh(MeshBase):
+class RectilinearMesh(StructuredMesh):
     """A 3D rectilinear Cartesian mesh
 
     Parameters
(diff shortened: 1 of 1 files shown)


eepeterson:
I wanted to post this as a draft because I'm not sure if I want to return tuples of arrays like `numpy.meshgrid` does or if we would rather return a single array with the coordinates on the last axis for consistency with the `UnstructuredMesh` class. For a `RectilinearMesh` this would be an array with shape `(nx, ny, nz, 3)`. Do you have a preference @pshriwise?

pshriwise:
> I wanted to post this as a draft because I'm not sure if I want to return tuples of arrays like `numpy.meshgrid` does or if we would rather return a single array with the coordinates on the last axis for consistency with the `UnstructuredMesh` class. For a `RectilinearMesh` this would be an array with shape `(nx, ny, nz, 3)`. Do you have a preference @pshriwise?

I think returning the `ndarray` would be more intuitive to index and slice.

pshriwise:
Looks like we need an implementation for the `_grids` attribute on the `UnstructuredMesh` class. What do you think about a `StructuredMesh` ABC that the regular, rectilinear, cylindrical, and spherical types can inherit from?

eepeterson:
> What do you think about a `StructuredMesh` ABC that the regular, rectilinear, cylindrical, and spherical types can inherit from?

I could implement a `StructuredMesh` mixin to group common functionality for those types of meshes is that what you mean?

pshriwise:
> > What do you think about a `StructuredMesh` ABC that the regular, rectilinear, cylindrical, and spherical types can inherit from?
> 
> 
> 
> I could implement a `StructuredMesh` mixin to group common functionality for those types of meshes is that what you mean?

Yep, exactly.

openmc/mesh.py
```diff
@@ -126,7 +126,76 @@ def from_xml_element(cls, elem):
             raise ValueError(f'Unrecognized mesh type "{mesh_type}" found.')
 
 
-class RegularMesh(MeshBase):
+class StructuredMesh(ABC):
```
pshriwise, line 129: Any reason not to have this inherit from `MeshBase` and then have the structured mesh types inherit directly from it instead of from both `StructuredMesh` and `MeshBase`?

PR #2013, opened 2022-03-25
cmake alias targets

kkiesling:
This creates `OpenMC::` aliases for both targets `libopenmc` and `openmc`.  (I think this is as simple as just adding these two lines?)


paulromano (approved):
Thanks @kkiesling!

PR #2015, opened 2022-03-29
Move lost particle reset from finalize() to reset()

### LINKED ISSUE #2014: Reset number of lost particles
For multiphysics coupling with Picard iteration, many OpenMC runs will be performed back-to-back, with `openmc_reset()` called between iterations to reset before the next iteration. For TRISO simulations in Cardinal, I think the geometric complexity is causing some particles to be lost during tracking. The number of lost particles, `openmc::simulation::n_lost_particles`, is only reset to zero in `openmc_finalize()`, which we don't call until the very end of the Picard iteration.

I think it makes more sense to reset the number of lost particles for each individual OpenMC run so that the proximity to the error tolerance isn't gradually decreasing with iteration number.

aprilnovak:
Moves the reset of accumulated lost particles from finalize() to reset().

Closes #2014

### DIFF
--- a/src/finalize.cpp
+++ b/src/finalize.cpp
@@ -121,7 +121,6 @@ int openmc_finalize()
   settings::write_initial_source = false;
 
   simulation::keff = 1.0;
-  simulation::n_lost_particles = 0;
   simulation::need_depletion_rx = false;
   simulation::total_gen = 0;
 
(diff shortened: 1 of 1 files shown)


paulromano (approved):
Thanks @aprilnovak!

PR #2020, opened 2022-04-01
Create a Plane in XY from x0, y0, and an angle.

tjlaboss:
Frequently, one needs to make a plane for an angle while working in XY-space. Perhaps it would be useful as a class method of `Plane`. This pull request therefore implements `Plane.from_xy_angle()`.

A unit test is included for the following plane, which intersects (+2, -1) and makes a 60 degree angle with the x-axis.

60-degrees

### DIFF
--- a/openmc/surface.py
+++ b/openmc/surface.py
@@ -753,6 +753,38 @@ class Plane(PlaneMixin, Surface):
         d = np.dot(n, p1)
         return cls(a=a, b=b, c=c, d=d, **kwargs)
 
+    @classmethod
+    def from_xy_angle(cls, x0, y0, angle, degrees=False, **kwargs):
+        """Return a plane given a single xy-point and an angle.
+
+        Parameters
+        ----------
+        angle : float, optional
+            Angle CCW from the x-axis.
+        degrees : bool, optional
+            Whether the angle is in degrees. Otherwise, it is in radians.
+            Defaults to False.
+        x0 : float, optional
+            X-coordinate of the "center" of the plane. Defaults to 0.
+        y0 : float, optional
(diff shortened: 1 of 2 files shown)


eepeterson:
Thanks for putting together this PR @tjlaboss! I just wanted to comment that this behavior should be feasible with something like `plane = YPlane().rotate((0, 0, 60).translate((2, -1, 0))` If you were to add a `from_xy_angle` method it would only make sense to add `yz`and `xz` methods as well. If others would really want more ways of creating/manipulating planes I’m happy to hear those arguments, but I think the current way of doing such things is sufficient and eliminates ambiguity of which halfspace is positive that you get with this sort of class method.

paulromano (commented):
Thanks for proposing this classmethod @tjlaboss! I tend to agree with @eepeterson that using the rotate/translate methods could achieve the same purpose without the confusion of which side is positive/negative. The only downside I see to the rotate/translate approach is that it will create three separate surface IDs unless you call:
```Python
plane = YPlane().rotate((0, 0, 60), inplace=True).translate((2, -1, 0), inplace=True)
```
but even that will produce warnings about IDs. If we were to keep your proposed method, I also agree with @eepeterson that it should work for xz and yz as well. One possibility is that we could change your proposed method to just be a thin wrapper around the above, something like:
```Python
@classmethod
def from_rotate_translate(cls, axis='x', rotation, translation=(0., 0., 0.), **kwargs):
    surf = getattr(openmc, f'{axis.upper()}Plane')(**kwargs)
    with warnings.catch_warnings():
        warnings.simplefilter("ignore", openmc.IDWarning)
        return surf.rotate(rotation, inplace=True).translate(translation, inplace=True)
```
Let us know what you think!

PR #2025, opened 2022-04-05
Separate Universe class source files.

pshriwise:
This creates a separate source and header file for the `Universe` and `UniversePartitioner` class, nothing more. Additional capabilities are slated to be added to this class in support of volumetric mesh universes. Separating this class from the cell source files keeps makes sense as these classes become larger and derived classes are added.


paulromano (approved):
Makes sense -- thanks @pshriwise!

PR #2083, opened 2022-06-14
Move TalliedFissionYieldHelper from abc.py to helpers.py

yardasol:
This PR does the following
 - moves the `TalliedFissionYieldHelper` class from `abc.py` to `helpers.py`
 - Removes the `from openmc.lib import Tally, ...` from `abc.py`
 - Removes `TalliedFissionYieldHelper` from the `_all_` list in `abc.py`; adds it to the `_all_` list in `helpers.py`.
 
The motivation for this is to enable an end user to import the `deplete.abc` module without needing to also implicitly import `openmc.lib`. This will -- once all the necessary additions have been made -- allow the end user to use OpenMC depletion without needing to build the C executable.

### DIFF
--- a/openmc/deplete/abc.py
+++ b/openmc/deplete/abc.py
@@ -19,7 +19,6 @@ from warnings import warn
 from numpy import nonzero, empty, asarray
 from uncertainties import ufloat
 
-from openmc.lib import MaterialFilter, Tally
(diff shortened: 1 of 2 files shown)


paulromano (approved):
Makes sense -- thanks @yardasol!

PR #2097, opened 2022-06-28
added multi stages option

shimwell:
Currently we have a nice dockerfile that builds OpenMC which is great

However there are cases where developers and users don't want every component of the docker image

For example:
- if we start to use code spaces then we might not want the source code to be compiled
- if someone is developing in a container they might just want the dependencies installed 
- users might want different nuclear data

This PR adds multi stage build flags to key parts of the Dockerfile

This will allow us to target specific build stages when building the dockerfile

```
# builds the image to completion, makes the same image that we currently have
docker build -t openmc_deps .

# builds an image with just the dependencies
docker build -t openmc_deps --target dependencies .

# builds an image with dependencies and compiled openmc but no nuclear data
docker build -t openmc_deps --target build .

# builds an image with dependencies, compiled openmc and nuclear data
docker build -t openmc_deps --target release .
```

What do you think @RemDelaporteMathurin would this help your use case

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -172,7 +172,7 @@ RUN if [ "$build_libmesh" = "on" ]; then \
         && rm -rf ${LIBMESH_INSTALL_DIR}/build ${LIBMESH_INSTALL_DIR}/libmesh ; \
     fi
 
-FROM dependencies as build
+FROM dependencies AS build
 
 # clone and install openmc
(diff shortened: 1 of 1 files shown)


RemDelaporteMathurin:
A dependencies/development image would be very useful indeed!

paulromano (approved):
Looks good to me! I'll wait the normal 36 hour period before merging in case there are other thoughts.

PR #2098, opened 2022-06-28
Optimize microscopic photoelectric cross section calculation 

myerspat:
The function `PhotonInteraction::calculate_xs` in the current version of OpenMC iteratively adds the subshelll photoelectric XS using a for loop and individual 1D xtensors. In an effort to speedup this computation the subshell photoelectric XS were placed into a 2D xtensor with the first dimension being energy and the second subshell. This ensures the subshell log-log interpolation and addition is contiguous in memory. The subshell XS are placed in the 2D xtensor using their threshold. With this optimization, a 1.06x speedup on a depleted pincell problem was achieved.


paulromano (approved):
Looks good to me! @myerspat since you're probably unaware, we have a [36 hour minimum review period](https://docs.openmc.org/en/latest/devguide/contributing.html#contribution-process) to allow anyone a chance to comment/object to a pull request. I'll leave the PR open until then but if there are no further comments, it should be good to merge.

@amandalund since you are the photon overlord, do you want to take a quick look? :smile:

amandalund:
Oops sorry @paulromano, I totally missed your mention but just happened to open this up now because I was curious.. looks good to me, nice optimization @myerspat!

PR #2100, opened 2022-06-29
Add class to allow running depletion simulation independent of the OpenMC transport solver

yardasol:
This PR adds a new class, `IndependentOperator`, that enables a user to provide a dictionary of nuclides and their concentrations (in at/cm^3) and microscopic cross section data.

This allows the user to run a depletion calculation without needing to use OpenMC's transport solver.

Right now the class supports flux-based and power-based flux normalization.

Specifically, this PR adds the following:
- `deplete/independent_operator.py`
  - `IndependentOperator` class
    - A nested helper class: `_IndependentRateHelper`. This class allows us to use the code from #2115 without duplicating the `number` or `cross_sections` attributes, which saves memory.
    - A nested helper class: `_IndependentNormalizationHelper`. This class allows us to perform constant-power depletion.
  - `deplete/microxs.py`
    - A new class, `MicroXS` that stores microscopic cross section information, along with the following methods to create the class
      - `from_model`: creates one-group library via tallies from an OpenMC simulation
      - `from_csv`: reads in cross section data from a `.csv` file
      - `from_array`: constructs an object based on a numpy array.
- `tests/unit_tests`
  - `test_deplete_independent_operator.py`
  - `test_deplete_microxs.py`
- `tests/integration_tests/`
  - Move current depletion regression test to `/deplete_with_transport`
  - Create new regression test, `/deplete_no_transport` and `/microxs`
- Add some content about `IndependentOperator` and `MicroXS` to the User's guide
- Add `IndependentOperator` and `MicroXS` to the Python API docs.
- Change name of `Operator` to `CoupledOperator`, along with an alias to `Operator` and a warning about the name change on initialization.
- Change name of `TransportOperator` to `DepletionOperator` to reflect the more generic use case for such an operator.
- Updates the code in the Chapter 8 of the user's guide to reflect the new `Operator` API (using `model` instead of `geometry` and `settings`)

Additionally, much of the code between `Operator` and `IndependentOperator` can be shared. This PR refactors this shared code into a a new class, `OpenMCOperator`, as a subclass of `TransportOperator`

Specifically, this PR does the following:
- Adds a new class `OpenMCOperator`. This class has the following abstract functions to be implemented by subclasses:
  - `_differentiate_burnable_mats()`
  - `_load_previous_results()`
  - `_get_nuclides_with_data()`
  - `_get_helper_classes()`
  - `_update_materials()`
  - `__call__()` (inherited from `TransportOperator`)
  - `write_bos_data()` (inherited from `TransportOperator`)
- Moves the following functions from `Operator` to `OpenMCOperator`:
  - `_distribute()`
  - `_get_burnable_mats()`
  - `_extract_number()`
  - `_set_number_from_mat()`
  - `_set_number_from_results()`
  - `Operator._get_tally_nuclides()` --> `OpenMCOperator._get_reaction_nuclides()`
  - `Operator._unpack_tallies_and_normalize()` --> `OpenMCOperator._calculate_reaction_rates()`
  - `get_results_info()`

- Splits the following functions into into reusable and non-reusable parts. The non-resuable parts stay in the original function under `Operator`, and the reusable parts are moved to a function of the same name (unless otherwise noted) under `OpenMCOperator`.
  - `__init__()`. The `OpenMCOperator.__init__()` implementation utilizes the following abstract functions:
    - `_differentiate_burnable_mats()`
    - `_load_previous_results()`
    - `_get_nuclides_with_data()`
    - `_get_helper_classes()`
  - `initial_condition()`
  - `__call__()` (the reusable part moved to `OpenMCOperator` is named `_update_materials_and_nuclides()`)
- Adds new section in the API docs for the `deplete` module on intermediate classes -- classes that implement abstract base classes and hold shared methods for their subclasses
  - `TalliedFissionYieldHelper`
  - `OpenMCOperator`


yardasol:
Added some basic unit tests, nothing major.

Working on getting an integration test up and running. Need to generate the 1-group cross sections first.

A detail I've notices that should be addressed (in the future), is that in practice, a user would need to update the microscopic cross sections each iteration as the flux spectrum as the fuel composition evolves, especially if we add support for multigroup microscopic xs data.

paulromano (changes requested):
This is a great start @yardasol! In addition to the line comments, one thing I notice is that a lot of the code looks very similar to what already exists in operator.py. We should think about whether it's possible to reduce some of the redundant code by moving it into a base class or otherwise combining.

Before getting to the point of merging, we should also consider how this will extend to multiple materials as well as using multigroup cross sections/flux instead of one group.

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +176,39 @@ across all material instances.
...
+OpenMC also supports transport-independent depletion calculations using the
+:class:`FluxDepletionOperator` class. Rather than taking a
+:class:`openmc.model.Model` object, this class accepts a volume,
+a dictionary of nuclide concentrations, a flux spectra, and one-group
+microscopic cross sections as a pandas dataframe. The class includes
+helper functions to constructe the dataframe from a csv file or from
+data arrays::
+
+    ...
+    micro_xs = FluxDepletionOperator.create_micro_xs_from_csv(micro_xs_path)
```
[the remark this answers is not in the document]
yardasol: I'm just trying to demonstrate how one would initiate a `FluxDepletionOperator` object. I have some discussion on the structure of the `micro_xs` parameter in the docstrings for that class, but I can try to add some details about that in the guide if you think it'd be useful.

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,529 @@
...
+                        if nuc in nuc_set]
+        else:
+            nuc_list = None
+
+        # Store list of tally nuclides on each process
+        nuc_list = comm.bcast(nuc_list)
+        return [nuc for nuc in nuc_list if nuc in self.chain]
+
+    def _get_all_nuclides_in_simulation(self):
+        """Determine nuclides that will show up in the simulation.
```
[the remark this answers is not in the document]
yardasol: Correct! I'll make this more explicit.

shimwell:
I'm just taking a look at this branch as I am keen to see if it will reduce the amount of compute I'm currently using for depletion.

Currently I am using the  ```CF4Integrator``` depletion operator with a few time steps and zero source rates in several timesteps.

```python
integrator = openmc.deplete.CF4Integrator(
        operator=operator,
        timesteps=[60, 60, 60, 60],
        source_rates=[1e21, 0, 0, 0],
    )
```

This appears to run **4** openmc transport simulations when I was hoping to get away with just one transport simulation as the source rate is 0 for the last 3 time steps.

I'm wondering if the addition of the ```FluxDepletionOperator``` in this PR will help me perform such a depletion with just 1 transport step?

shimwell:
I'm seeing transport being run on time each step.

I can see the ```if source_rate == 0.0:``` in the operator.py file but I notice the ```openmc.lib.run()``` comes a few lines later and is outside of this if statement.

This is perhas a bit unrelated to your pull request but I'm wondering if we can move the ```openmc.lib.run()``` into an if statement block that checks for zero flux as well

paulromano:
@shimwell The `CFIntegrator` requires four transport solves _per timestep_ (see mathematical definition [here](https://docs.openmc.org/en/latest/pythonapi/generated/openmc.deplete.CF4Integrator.html#openmc.deplete.CF4Integrator)), so the fact that you are seeing only four solves (from the first timestep) indicates that on the subsequent three timesteps there are no transport solves happening.

yardasol:
Some of the test runs keep failing during the `after success` step:
```
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/coveralls/cli.py", line 95, in main
    result = coverallz.wear()
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/coveralls/api.py", line 257, in wear
    return self.submit_report(json_string)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/coveralls/api.py", line 293, in submit_report
    raise CoverallsException(
coveralls.exception.CoverallsException: Could not submit coverage: 502 Server Error: Bad Gateway for url: https://coveralls.io/api/v1/jobs
Error: Process completed with exit code 1.
```

paulromano:
@yardasol Don't worry about coveralls errors -- that seems to be affecting OpenMC as a whole and is not related to anything on your branch

docs/source/usersguide/depletion.rst
```diff
@@ -113,21 +112,24 @@ should be, including indirect components. Some examples are provided below::
     # use a dictionary of fission_q values
     fission_q = {"U235": 202e+6}  # energy in eV
 
+    # create a Model object
+    model  = openmc.model.Model(geometry, settings)
```
paulromano, line 116:
```suggestion
    model  = openmc.Model(geometry, settings)
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +179,45 @@ across all material instances.
...
+.. note::
+   
+   This feature is still under heavy development. API changes are
+   possible and likely in the near future.
+
+OpenMC supports running depletion calculations independent of the OpenMC
+transport solver using the :class:`FluxDepletionOperator` class. This class
+has two ways to initalize it; the default constructor accepts an 
+:class:`openmc.Materials` object, a flux spectra, and one-group microscopic 
+cross sections as a :class:`pandas.Dataframe`, while the `from_nuclides`
```
paulromano, line 194:
```suggestion
cross sections as a :class:`pandas.DataFrame`, while the `from_nuclides` 
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +179,45 @@ across all material instances.
     This will increase the total memory usage and run time due to an increased
     number of tallies and material definitions.
 
+Transport-independent depletion
+-------------------------------
```
paulromano, line 183:
```suggestion
===============================
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +179,45 @@ across all material instances.
...
+
+.. note::
+   
+   This feature is still under heavy development. API changes are
+   possible and likely in the near future.
+
+OpenMC supports running depletion calculations independent of the OpenMC
+transport solver using the :class:`FluxDepletionOperator` class. This class
```
paulromano, line 191:
```suggestion
transport solver using the :class:`~openmc.deplete.FluxDepletionOperator` class. This class
```

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
+"""Pure depletion operator
+
+This module implements a pure depletion operator that uses user provided fluxes
```
paulromano, line 3:
```suggestion
This module implements a pure depletion operator that uses user-provided fluxes
```

openmc/deplete/openmc_operator.py
```diff
@@ -0,0 +1,567 @@
...
+        listed under ``depletion_chain`` in
+        :envvar:`OPENMC_CROSS_SECTIONS` environment variable.
+    prev_results : Results, optional
+        Results from a previous depletion calculation. If this argument is
+        specified, the depletion calculation will start from the latest state
+        in the previous results.
+    diff_burnable_mats : bool, optional
+        Whether to differentiate burnable materials with multiple instances.
+        Volumes are divided equally from the original material volume.
+        Default: False.
```
paulromano, line 69: Defaults are listed in the function signature so not really necessary to put them in the docstring
```suggestion
```

openmc/deplete/openmc_operator.py
```diff
@@ -0,0 +1,567 @@
...
+    fission_q : dict, optional
+        Dictionary of nuclides and their fission Q values [eV].
+    dilute_initial : float, optional
+        Initial atom density [atoms/cm^3] to add for nuclides that are zero
+        in initial condition to ensure they exist in the decay chain.
+        Only done for nuclides with reaction rates.
+        Defaults to 1.0e3.
+    helper_kwargs : dict
+        Arguments for helper classes
```
paulromano, line 78:
```suggestion
        Keyword arguments for helper classes
```

openmc/deplete/openmc_operator.py
```diff
@@ -0,0 +1,567 @@
...
+            Volumes for the above materials in [cm^3]
+        all_nuclides : list of str
+            Nuclides to be used in the simulation.
+        prev_res : Results, optional
+            Results from a previous depletion calculation
+
+        """
+        self.number = AtomNumber(
+            local_mats, all_nuclides, volume, len(
+                self.chain))
```
paulromano, line 276:
```suggestion
        self.number = AtomNumber(local_mats, all_nuclides, volume, len(self.chain))
```

openmc/deplete/openmc_operator.py
```diff
@@ -0,0 +1,567 @@
...
+            Indicates the subclass of :class:`NormalizationHelper` to
+            instantiate.
+        fission_yield_mode : str
+            Indicates the subclass of :class:`FissionYieldHelper` to instatiate.
+        reaction_rate_opts : dict
+            Keyword arguments that are passed to the :class:`ReactionRateHelper`
+            subclass.
+        fission_yield_opts : dict
+            Keyword arguments that are passed to the :class:`FissionYieldHelper`
+            subclass.
```
paulromano, line 365: This doesn't match what arguments are passed (only `helper_kwargs`)

paulromano (changes requested):
Really nice work on this @yardasol! Here's a first round of comments for you:

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
...
+    def _load_previous_results():
+        """Load in results from a previous depletion calculation."""
+        pass
+
+    def _get_nuclides_with_data(self, cross_sections):
+        """Finds nuclides with cross section data
+        """
+        return set(cross_sections.index)
+
+    class FluxTimesXSHelper(ReactionRateHelper):
```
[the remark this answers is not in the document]
yardasol: I renamed it to `_FluxDepletionRateHelper`

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
...
+
+        rates = self.reaction_rates
+        # Get classes to assit working with tallies
+        nuc_ind_map = {ind: nuc for nuc, ind in rates.index_nuc.items()}
+        rxn_ind_map = {ind: rxn for rxn, ind in rates.index_rx.items()}
+
+        self._rate_helper = self.FluxTimesXSHelper(
+            self.reaction_rates.n_nuc, self.reaction_rates.n_react, self)
+        self._rate_helper.nuc_ind_map = nuc_ind_map
+        self._rate_helper.rxn_ind_map = rxn_ind_map
```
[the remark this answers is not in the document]
yardasol: Good catch!

openmc/deplete/operator.py
```diff
@@ -244,65 +224,126 @@ def __init__(self, model, chain_file=None, prev_results=None,
...
+            Indicates the subclass of :class:`NormalizationHelper` to
+            instatiate.
+        fission_yield_mode : str
+            Indicates the subclass of :class:`FissionYieldHelper` to instantiate.
+        reaction_rate_opts : dict
+            Keyword arguments that are passed to the :class:`ReactionRateHelper`
+            subclass.
+        fission_yield_opts : dict
+            Keyword arguments that are passed to the :class:`FissionYieldHelper`
+            subclass.
```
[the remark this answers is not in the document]
yardasol: Done

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
...
+
+        Parameters
+        ----------
+        outer : openmc.deplete.FluxDepletionOperator
+            Reference to the object encapsulate FluxTimesXSHelper.
+            We pass this so we don't have to duplicate the ``number`` object.
+        n_nucs : int
+            Number of burnable nuclides tracked by :class:`openmc.deplete.Operator`
+        n_react : int
+            Number of reactions tracked by :class:`openmc.deplete.Operator`
```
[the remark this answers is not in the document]
yardasol: Done

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
...
+
+    Parameters
+    ----------
+    materials : openmc.Materials
+        Materials to deplete.
+    micro_xs : pandas.DataFrame
+        DataFrame with nuclides names as index and microscopic cross section
+        data in the columns. Cross section units are [cm^-2].
+    flux_spectra : float
+        Flux spectrum [n cm^-2 s^-1]
```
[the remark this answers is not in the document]
yardasol: Done

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
...
+                            # values.
+                            if val < -1.0e-21:
+                                print(
+                                    "WARNING: nuclide ",
+                                    nuc,
+                                    " in material ",
+                                    mat,
+                                    " is negative (density = ",
+                                    val,
+                                    " at/barn-cm)")
```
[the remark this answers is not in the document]
yardasol: Done

openmc/deplete/openmc_operator.py
yardasol: I've removed this. It was cruft which I forgot to move.

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,464 @@
...
+        # Get classes to assit working with tallies
+        nuc_ind_map = {ind: nuc for nuc, ind in rates.index_nuc.items()}
+        rxn_ind_map = {ind: rxn for rxn, ind in rates.index_rx.items()}
+
+        self._rate_helper = self.FluxTimesXSHelper(
+            self.reaction_rates.n_nuc, self.reaction_rates.n_react, self)
+        self._rate_helper.nuc_ind_map = nuc_ind_map
+        self._rate_helper.rxn_ind_map = rxn_ind_map
+
+        self._normalization_helper = SourceRateHelper()
```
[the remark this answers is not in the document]
yardasol: As we discussed, I'm going to add an addtional option to use `ChainFissionHelper`

openmc/deplete/abc.py
```diff
@@ -523,8 +525,10 @@ class Integrator(ABC):
         Power density of the reactor in [W/gHM]. It is multiplied by
         initial heavy metal inventory to get total power if ``power``
         is not speficied.
+    flux : float or iterable of float, optional
+        neutron flux in [neut/s-cm^2] for each interval in :attr:`timesteps`
```
paulromano, line 529: Since this is being treated exactly the same as `source_rates`, I'm not sure we should introduce a new argument just for this. We can expand the description of `source_rates` below to say that it can also represent the neutron flux (which sort of _is_ the source rate in a sense) and then also make it clear in the .rst documentation.

docs/source/usersguide/depletion.rst
```diff
@@ -19,14 +19,13 @@ transmutation equations and the method used for advancing time. At present, the
...
 in principle additional operator classes based on other transport codes could be
 implemented and no changes to the depletion solver itself would be needed. The
-operator class requires a :class:`openmc.Geometry` instance and a
-:class:`openmc.Settings` instance::
+operator class requires a :class:`openmc.model.Model` instance containing 
+material, geometry, and settings information::
 
-    geom = openmc.Geometry()
-    settings = openmc.Settings()
+    model = openmc.model.Model()
```
paulromano, line 25:
```suggestion
    model = openmc.Model()
```

docs/source/usersguide/depletion.rst
```diff
@@ -19,14 +19,13 @@ transmutation equations and the method used for advancing time. At present, the
 :class:`openmc.deplete.Operator` (which uses the OpenMC transport solver), but
 in principle additional operator classes based on other transport codes could be
 implemented and no changes to the depletion solver itself would be needed. The
-operator class requires a :class:`openmc.Geometry` instance and a
-:class:`openmc.Settings` instance::
+operator class requires a :class:`openmc.model.Model` instance containing
```
paulromano, line 22:
```suggestion
operator class requires a :class:`~openmc.Model` instance containing 
```

openmc/deplete/openmc_operator.py
```diff
@@ -0,0 +1,569 @@
...
+
+
+    Attributes
+    ----------
+    materials : openmc.Materials
+        All materials present in the model
+    cross_sections : str or pandas.DataFrame
+            Path to continuous energy cross section library, or object
+            containing one-group cross-sections.
+
```
paulromano, line 93:
```suggestion
```

openmc/deplete/openmc_operator.py
```diff
@@ -0,0 +1,569 @@
...
+            tally_rates = self._rate_helper.get_material_rates(
+                mat_index, nuc_ind, react_ind)
+
+            # Compute fission yields for this material
+            fission_yields.append(self._yield_helper.weighted_yields(i))
+
+            # Accumulate energy from fission
+            if fission_ind is not None:
+                self._normalization_helper.update(tally_rates[:, fission_ind], mat_index=mat_index)
```
paulromano, line 530: Need some line wrapping here

openmc/deplete/operator.py
```diff
@@ -244,173 +222,32 @@ def __init__(self, model, chain_file=None, prev_results=None,
...
-
-        Parameters
-        ----------
-        step : int
-            Current depletion step including restarts
-        """
-        openmc.lib.statepoint_write(
-            "openmc_simulation_n{}.h5".format(step),
-            write_source=False)
+        helper_kwargs = dict()
```
paulromano, line 228:
```suggestion
```

openmc/deplete/operator.py
```diff
@@ -434,137 +271,104 @@ def _differentiate_burnable_mats(self):
...
+            score = "heating" if self.settings.photon_transport else "heating-local"
+            self._normalization_helper = EnergyScoreHelper(score)
+        else:
+            self._normalization_helper = SourceRateHelper()
 
-            self.number.set_atom_density(mat_id, nuclide, atom_per_cc)
+        # Select and create fission yield helper
+        fission_helper = self._fission_helpers[fission_yield_mode]
+        if fission_yield_opts is None:
+            fission_yield_opts = {}
```
paulromano, line 369:
```suggestion
```

openmc/deplete/operator.py
```diff
@@ -720,103 +431,90 @@ def _unpack_tallies_and_normalize(self, source_rate):
...
-        """Loads cross_sections.xml file to find nuclides with neutron data"""
-        nuclides = set()
-        data_lib = DataLibrary.from_xml(cross_sections)
-        for library in data_lib.libraries:
-            if library['type'] != 'neutron':
-                continue
-            for name in library['materials']:
-                if name not in nuclides:
-                    nuclides.add(name)
+                                      ' at/barn-cm)')
```
paulromano, line 492: To be consistent with usage elsewhere in the API
```suggestion
                                      ' atom/b-cm)')
```

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+                            densities.append(val)
+                        else:
+                            # Only output warnings if values are significantly
+                            # negative. CRAM does not guarantee positive
+                            # values.
+                            if val < -1.0e-21:
+                                print(f'WARNING: nuclide {nuc} in material'
+                                      f'{mat} is negative (density = {val}'
+
+                                      ' at/barn-cm)')
```
paulromano, line 447: For consistency
```suggestion
                                      ' atom/b-cm)')
```

tests/regression_tests/deplete_no_transport/test.py
```diff
@@ -0,0 +1,115 @@
...
+from tests.regression_tests import config
+
+
+@pytest.fixture(scope="module")
+def fuel():
+    fuel = openmc.Material(name="uo2")
+    fuel.add_element("U", 1, percent_type="ao", enrichment=4.25)
+    fuel.add_element("O", 2)
+    fuel.set_density("g/cc", 10.4)
+    fuel.depletable=True
```
paulromano, line 25:
```suggestion
    fuel.depletable = True
```

tests/regression_tests/deplete_no_transport/test.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    # Create operator
+    micro_xs_file = Path(__file__).parents[2] / 'micro_xs_simple.csv'
+    micro_xs = FluxDepletionOperator.create_micro_xs_from_csv(micro_xs_file)
+    chain_file = Path(__file__).parents[2] / 'chain_simple.xml'
+
+    if from_nuclides:
+        nuclides = {}
+        for nuc, dens in fuel.get_nuclide_atom_densities().items():
+            nuclides[nuc] = dens * 1e24
```
paulromano, line 56:
```suggestion
        for nuc, atom_per_b_cm in fuel.get_nuclide_atom_densities().items():
            nuclides[nuc] = atom_per_b_cm * 1e24
```

tests/regression_tests/deplete_no_transport/test.py
```diff
@@ -0,0 +1,115 @@
...
+from difflib import unified_diff
+import numpy as np
+import pytest
+import openmc
+from openmc.data import JOULE_PER_EV
+import openmc.deplete
+from openmc.deplete import FluxDepletionOperator
+
+
+from tests.regression_tests import config
```
paulromano, line 16:
```suggestion
from math import floor
import shutil
from pathlib import Path
from difflib import unified_diff

import numpy as np
import pytest
import openmc
from openmc.data import JOULE_PER_EV
import openmc.deplete
from openmc.deplete import FluxDepletionOperator

from tests.regression_tests import config
```

tests/regression_tests/deplete_no_transport/test.py
paulromano: This also makes me wonder if we should just have `from_nuclides` take densities in atom/b-cm rather than atom/cm³

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+            values. Nuclide concentration units are [atom/cm^3].
+        micro_xs : pandas.DataFrame
+            DataFrame with nuclides names as index and microscopic cross section
+            data in the columns. Cross section units are [cm^-2].
+        chain_file : str
+            Path to the depletion chain XML file.
+        keff : 2-tuple of float, optional
+           keff eigenvalue and uncertainty from transport calculation.
+           Default is None.
+        normalization_mode : {"constant-power", "constant-flux"}
```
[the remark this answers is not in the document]
yardasol: So after skimming that paper, I have a few points of understanding I'd like to reach consensus on:
- the `constant-power` normalization mode _is_ constant power but with the assumptions that the flux is the same at the beginning of the depletion step as at the end of the depletion step, which will give a flux that is different from the "true" value, which may not be a very good assumption.
- to get better results for the flux calculated from the constant power, we need to perform substeps to converge to a value of flux that is closer to the "true" value
- Because we can only accept a single `DataFrame` for the one-group cross-sections, even if we wanted to do this kind of substep renormalization, we couldn't because that would require us to calculate new cross sections for each substep.

We keep running into this issue of having to recalculate the cross sections. Implementing this kind of functionality (and additionally all the extra cool things you can do with it, like this substep renormalization) would be an interesting project!

I think the `constant-fission` name is fine, but we probably should try to be consistent with the rest of the `deplete` module and revert it to `source-rate`.

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+            fission_rates : numpy.ndarray
+                fission reaction rate for each isotope in the specified
+                material. Should be ordered corresponding to initial
+                ``rate_index`` used in :meth:`prepare`
+            mat_index : int
+                Material index
+
+        """
+            volume = self._op.number.get_mat_volume(mat_index)
+            densities = np.empty(np.shape(fission_rates))
```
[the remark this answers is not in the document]
yardasol: handy!

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+        # Reload volumes into geometry
+        self.prev_res[-1].transfer_volumes(self.materials)
+
+        # Store previous results in operator
+        # Distribute reaction rates according to those tracked
+        # on this process
+        if comm.size != 1:
+            prev_results = self.prev_res
+            self.prev_res = Results()
+            mat_indexes = _distribute(range(len(self.burnable_mats)))
```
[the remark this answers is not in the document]
yardasol: I added an import statement to fix this

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+        nuc_ind_map : dict of int to str
+            Dictionary mapping the nuclide index to nuclide name
+        rxn_ind_map : dict of int to str
+            Dictionary mapping reaction index to reaction name
+
+        """
+
+        def __init__(self, n_nuc, n_react, op):
+            super().__init__(n_nuc, n_react)
+            rates = op.reaction_rates
```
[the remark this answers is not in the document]
yardasol: Hadn't considered that. Good suggestion!

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+            self.reaction_rates.n_nuc, self.reaction_rates.n_react, self)
+        if normalization_mode == "constant-power":
+            self._normalization_helper = self._FluxDepletionNormalizationHelper(self)
+        else:
+            self._normalization_helper = SourceRateHelper()
+
+        # Select and create fission yield helper
+        fission_helper = ConstantFissionYieldHelper
+        if fission_yield_opts is None:
+            fission_yield_opts = {}
```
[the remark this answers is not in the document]
yardasol: This suggestion won't work with the current structure of the code, which is why I didn't incorporate it from your previous round of review, but perhaps I should have said something about this.

In `__init__`, we have
```python
def __init__(self,
                  materials,
                  micro_xs,
                  chain_file,
                  keff=None,
                  normalization_mode='constant-flux',
                  fission_q=None,
                  prev_results=None,
                  reduce_chain=False,
                  reduce_chain_level=None,
                  fission_yield_opts=None):
         .
         .
         .
         helper_kwargs = {'normalization_mode': normalization_mode,
                          'fission_yield_opts': fission_yield_opts}
         .
         .
         .
``` 

The default value of `fission_yield_opts` is `None`, and since it is always passed to `helper_kwargs`, the `helper_kwargs.get('fission_yield_opts', {})` fuction inside of `_get_helper_classes` will always return the value of `fission_yield_opts` stored in `helper_kwargs`, and never return the alternate value given (in this case, an empty `dict`). So we need the `if` statement on 373 to avoid an error from trying to pass a `None` value to `fission_helper`.

If you can think of a good way to restructure this so we don't need the `if` statement, please make a suggestion!

openmc/deplete/flux_operator.py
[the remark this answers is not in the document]
paulromano: > * with the assumptions that the flux is the same at the beginning of the depletion step as at the end of the depletion step

That _is_ the constant flux assumption :smile: The only difference really is how you got that constant flux in the first place — specifying it explicitly or deducing it from the power.

> * we need to perform substeps to converge to a value of flux

Yes, that is the only way I could see to adjust the flux so that the power is not decreasing over the full step.

> * Because we can only accept a single `DataFrame` for the one-group cross-sections, even if we wanted to do this kind of substep renormalization, we couldn't

Cross sections are only one piece of the puzzle. We _would_ have updated number densities, so even if the cross sections are constant, you need a higher flux to account for the decrease in the density of fissionable nuclides to maintain constant power. I think that's actually the primary effect -- I wouldn't expect the micros to change much over a single step since the spectrum is probably about the same.

> Implementing this kind of functionality ... would be an interesting project!

Agreed!

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+
+    def __call__(self, vec, source_rate):
+        """Obtain the reaction rates
+
+        Parameters
+        ----------
+        vec : list of numpy.ndarray
+            Total atoms to be used in function.
+        source_rate : float
+            Power in [W] or source rate in [neutron/sec]
```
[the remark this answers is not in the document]
yardasol: I've replaced `source rate` with flux and changed the units, because passing a source rate to this class isn't quite correct.

Should I add a warning statement to the `Integrator` class to alert a user that they may get unexpected results if passing a `source_rate` parameter instead of `flux` or `power` if the `operator` is a `FluxDepletionOperator` object?

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+        ----------
+        nuclides : list of str
+            List of nuclide symbols for that have data for at least one
+            reaction.
+        reactions : list of str
+            List of reactions. All reactions must match those in ``chain.REACTONS``
+        data : ndarray of floats
+            Array containing one-group microscopic cross section information for each
+            nuclide and reaction.
+        units : {'barn', 'cm^2'}, optional
```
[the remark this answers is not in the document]
yardasol: Hmm, probably not. that would actually be quite ugly now that I think about it. I'll remove this.

openmc/deplete/flux_operator.py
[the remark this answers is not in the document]
paulromano: Ah, right! In that case, let's just move the `if` block further up, i.e.
```Python
if fission_yield_opts is None:
    fission_yield_opts = {}
helper_kwargs = ...
```

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+    def _validate_micro_xs_inputs(nuclides, reactions, data):
+        check_iterable_type('nuclides', nuclides, str)
+        check_iterable_type('reactions', reactions, str)
+        check_type('data', data, np.ndarray, expected_iter_type=float)
+        for reaction in reactions:
+            check_value('reactions', reaction, _valid_rxns)
+
+
+    @staticmethod
+    def generate_1g_cross_sections(model, reaction_domain, reactions=['(n,gamma)', '(n,2n)', '(n,p)', '(n,a)', '(n,3n)', '(n,4n)', 'fission'],  energy_bounds=(0, 20e6), write_to_csv=True, filename='micro_xs.csv'):
```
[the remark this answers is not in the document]
yardasol: I'm actually glad you think so because I was not sure where to put it when I wrote it.

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+        """Helper function to generate a one-group cross-section dataframe
+        using OpenMC. Note that the ``openmc`` C executable must be compiled.
```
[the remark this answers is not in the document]
yardasol: I keep thinking that OpenMC is written in C and not C++...

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+        tallies = openmc.Tallies()
+        xs = dict()
+        for rxn in reactions:
+            if rxn == 'fission':
+                xs[rxn] = FissionXS(domain=reaction_domain, groups=groups, by_nuclide=True)
+            else:
+                xs[rxn] = ArbitraryXS(rxn, domain=reaction_domain, groups=groups, by_nuclide=True)
+            tallies += xs[rxn].tallies.values()
+
+        model.tallies = tallies
```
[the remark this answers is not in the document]
yardasol: Is this so, if a user is running in a notebook, they can rerun without needing to reinitialize their `model` object?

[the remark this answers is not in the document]
yardasol: I like option 2

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+            df.index = df['nuclide']
+            df.drop(['nuclide', xs[rxn].domain_type, 'group in', 'std. dev.'], axis=1, inplace=True)
+            df.rename({'mean':rxn}, axis=1, inplace=True)
+            micro_xs = pd.concat([micro_xs, df], axis=1)
+
+        if write_to_csv:
+            micro_xs.to_csv(filename)
+            return None
+        else:
+            return micro_xs
```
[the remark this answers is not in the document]
yardasol: I also thought this should be the case when I looked at it again today.

openmc/deplete/abc.py
yardasol: I agree I think we can just add more spec to `source_rates`. Where in the .rst documentation are you referring to?

openmc/deplete/operator.py
yardasol: The same comment applies as it does to this `if` statement in `FluxDepletionOperator._get_helper_classes()`

tests/regression_tests/deplete_no_transport/test.py
yardasol: I think it's normal for people to have nuclide concentrations in units other than atom/b-cm. Perhaps we should add a unit specifier to `from_nuclides`?

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,594 @@
...
+        chain_file : str
+            Path to the depletion chain XML file.
+        keff : 2-tuple of float, optional
+           keff eigenvalue and uncertainty from transport calculation.
+           Default is None.
+        normalization_mode : {"constant-power", "constant-flux"}
+            Indicate how reaction rates should be calculated.
+            ``"constant-power"`` uses the fission Q values from the depletion
+            chain to compute the flux based on the power. ``"constant-flux"``
+            uses the value stored in `_normalization_helper` as the flux.
```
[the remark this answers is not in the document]
yardasol: Fixed it

tests/regression_tests/deplete_no_transport/test.py
yardasol: I've added such a specifier

yardasol:
@paulromano I think I addressed your second round of comments. I've left suggestions that I'd like feedback on my solution to unresolved.

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,601 @@
...
+        df.drop(['nuclide', xs[rxn].domain_type, 'group in', 'std. dev.'], axis=1, inplace=True)
+        df.rename({'mean':rxn}, axis=1, inplace=True)
+        micro_xs = pd.concat([micro_xs, df], axis=1)
+
+    if write_to_csv:
+        micro_xs.to_csv(filename)
+
+    return micro_xs
+
+class FluxDepletionOperator(OpenMCOperator):
```
paulromano, line 102: I'm reconsidering whether this name is apt. What do you think of `IndependentOperator`? I think the "Depletion" part is probably not necessary since it's already namespaced in `openmc.deplete`.

paulromano: Another alternative: `ExternalOperator` (i.e., transport is run externally)

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,601 @@
...
+        compute the flux based on the power. ``"source-rate"`` uses a the
+        source rate (assumed to be neutron flux) to calculate the
+        reaction rates.
+    fission_q : dict, optional
+        Dictionary of nuclides and their fission Q values [eV]. If not given,
+        values will be pulled from the ``chain_file``. Only applicable
+        if ``"normalization_mode" == "fission-q"``.
+    reduce_chain : bool, optional
+        If True, use :meth:`openmc.deplete.Chain.reduce` to reduce the
+        depletion chain up to ``reduce_chain_level``. Default is False.
```
paulromano, line 137:
```suggestion
        depletion chain up to ``reduce_chain_level``.
```

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,601 @@
...
+                self.prev_res.append(new_res)
+
+
+    def _get_nuclides_with_data(self, cross_sections):
+        """Finds nuclides with cross section data"""
+        return set(cross_sections.index)
+
+    class _FluxDepletionNormalizationHelper(ChainFissionHelper):
+        """Class for calculating one-group flux
+        based on a power.
```
paulromano, line 330:
```suggestion
        """Class for calculating one-group flux based on a power.
```

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,601 @@
...
+                            # values.
+                            if val < -1.0e-21:
+                                print(f'WARNING: nuclide {nuc} in material'
+                                      f'{mat} is negative (density = {val}'
+
+                                      ' atom/b-cm)')
+                            number_i[mat, nuc] = 0.0
+
+    @staticmethod
+    def create_micro_xs_from_data_array(nuclides, reactions, data):
```
paulromano, line 528: So now we have:
- `generate_1g_cross_sections`
- `FluxDepletionOperator.create_micro_xs_from_data_array`
- `FluxDepletionOperator.create_micro_xs_from_csv`

I'm wondering if maybe introducing a class to store 1g cross sections (that subclasses pandas.DataFrame) might be worthwhile. In that case, we could have:
```Python
class MicroXS(pandas.DataFrame):
    @classmethod
    def from_model(cls, ...):
        ...

    @classmethod
    def from_array(cls, ...):
        ...

    @classmethod
    def from_csv(cls, ...):
        ...
```
What do you think? I could also envision other uses of this class in the future, e.g., interpolating between multiple sets of micros.

paulromano (changes requested):
A few more suggestions and thoughts:

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,601 @@
...
+    xs = {}
+    for rxn in reactions:
+        if rxn == 'fission':
+            xs[rxn] = FissionXS(domain=reaction_domain, groups=groups, by_nuclide=True)
+        else:
+            xs[rxn] = ArbitraryXS(rxn, domain=reaction_domain, groups=groups, by_nuclide=True)
+        tallies += xs[rxn].tallies.values()
+
+    model.tallies = tallies
+    statepoint_path = model.run()
```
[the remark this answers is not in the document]
yardasol: Great idea.

openmc/deplete/flux_operator.py
```diff
@@ -0,0 +1,601 @@
...
+    micro_xs = pd.DataFrame()
+    for rxn in xs:
+        df = xs[rxn].get_pandas_dataframe(xs_type='micro')
+        df.index = df['nuclide']
+        df.drop(['nuclide', xs[rxn].domain_type, 'group in', 'std. dev.'], axis=1, inplace=True)
+        df.rename({'mean':rxn}, axis=1, inplace=True)
+        micro_xs = pd.concat([micro_xs, df], axis=1)
+
+    if write_to_csv:
+        micro_xs.to_csv(filename)
```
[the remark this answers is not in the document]
yardasol: I think that makes sense.

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +179,57 @@ across all material instances.
...
+This class supports both constant-flux (``source-rate`` normalization) and 
+constant-power depletion (``fission-q`` normalization).
+
+.. important::
+
+   Make sure you set the correct parameter in the :class:`openmc.abc.Integrator` class. Use the ``source_rates`` parameter when ``normalization_mode == source-rate``, and use ``power`` or ``power_density`` when ``normalization_mode == fission-q``.
+
+.. warning::
+
+   The accuracy of results when using ``fission-q`` is entirely dependent on your depletion chain. Make sure it has sufficient data to resolve the dynamics of your particular scenario.
```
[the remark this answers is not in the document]
yardasol: Sure thing

openmc/deplete/flux_operator.py
yardasol: I agree. It's fine as a preliminary name but we should pick something more polished before this gets merged.

yardasol: I think `ExternalOperator` makes the most sense

openmc/deplete/flux_operator.py
yardasol: I like this idea. I'll add it to the PR. Any idea on where it should go? Probably not in `deplete`...

paulromano: I'd say for now, you can leave it in `deplete` since that's the only immediate application. We can always reconsider where the best place is for it to live later

yardasol:
@paulromano thoughts on changing `Operator` to `CoupledOperator`?

yardasol:
Also, let's wait for #2139 before merging so we can add that functionality to the `from_nuclides` method.

yardasol:
>@paulromano thoughts on changing Operator to CoupledOperator and TransportOperator to Operator? This might clear up the purpose of each with the introduction of IndependentOperator into the mix. We should also update some of the terminology in the user's guide and API docpages to reflect the expanded scope of Operator (previously called TransportOperator)

I discussed this point in private messages with @paulromano. It was brought up that changing `TransportOperator` to `Operator` would completely break backwards compatibility. I agree that we do not want to do this. I still want to change the name of `TransportOperator`, perhaps `GenericOperator` or `ReactionRateOperator`?

I still also think we should change `Operator` to `CoupledOperator`, but we should probably include an alias so that a user can still use `Operator` to initalize a class, and perhaps a deprecation warning as well?

openmc/deplete/coupled_operator.py
```diff
@@ -0,0 +1,529 @@
...
+                 fission_yield_mode="constant", fission_yield_opts=None,
+                 reaction_rate_mode="direct", reaction_rate_opts=None,
+                 reduce_chain=False, reduce_chain_level=None):
+        # warn of name change
+        warn(
+            "The Operator(...) class has been renamed and will "
+            "be removed in a future version of OpenMC. Use "
+            "CoupledOperator(...) instead.",
+            FutureWarning
+        )
```
paulromano, line 201: This warning should only be displayed if the user actually calls `Operator`. An easy way to do this is:
```Python
def Operator(*args, **kwargs):
    warn(...)
    return CoupledOperator(*args, **kwargs)
```

paulromano (changes requested):
Thanks for the updates! Here are a few more comments

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,178 @@
...
+            List of nuclide symbols for that have data for at least one
+            reaction.
+        reactions : list of str
+            List of reactions. All reactions must match those in
+            :data:`openmc.deplete.chain.REACTIONS`
+        data : ndarray of floats
+            Array containing one-group microscopic cross section values for
+            each nuclide and reaction
+        units : {'b', 'cm^2'}
+            Units of the cross section values in ``data``
```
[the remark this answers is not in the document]
yardasol: I added it back in because the `from_arrays` and `from_csv` methods have no way to actually check the units of the data. But I can remove it if you think it isn't useful.

yardasol:
@paulromano thanks for the comments! Any feedback on my updates to the docpages?

openmc/deplete/abc.py
```diff
@@ -76,14 +75,15 @@ def change_directory(output_dir):
         os.chdir(orig_dir)
 
 
-class TransportOperator(ABC):
-    """Abstract class defining a transport operator
+class DepletionOperator(ABC):
```
paulromano, line 78: I think `TransportOperator` more accurately describes what this series of classes is all about: given a composition, return k<sub>eff</sub> and reaction rates. Obviously the new `IndependentOperator` doesn't do any transport, but it also doesn't do any depletion. `ReactionRateOperator` could be more apt, although it doesn't reflect the k<sub>eff</sub> part...

docs/source/pythonapi/deplete.rst
```diff
@@ -12,19 +12,20 @@ Primary API
...
-and retaining important information required for depletion. The most common examples
-are reaction rates and power normalization data. The latter is responsible for
-projecting reaction rates and compositions forward in calendar time across
-some step size :math:`\Delta t`, and obtaining new compositions given a power
-or power density. The :class:`Operator` is provided to handle communicating with
-OpenMC. Several classes are provided that implement different time-integration
-algorithms for depletion calculations, which are described in detail in Colin
-Josey's thesis, `Development and analysis of high order neutron
-transport-depletion coupling algorithms <http://hdl.handle.net/1721.1/113721>`_.
+The former is responsible for calcuating retaining important information required for depletion. The most common examples are reaction rates and power
```
paulromano, line 18:
```suggestion
The former is responsible for calculating and retaining important information required for depletion. The most common examples are reaction rates and power
```

docs/source/pythonapi/deplete.rst
```diff
@@ -189,24 +194,57 @@ total system energy.
    helpers.FissionYieldCutoffHelper
    helpers.FluxCollapseHelper
 
+The :class:`openmc.deplete.IndependentOperator` uses inner class subclassed from
+those listed to perform similar calculations.
```
paulromano, line 198:
```suggestion
The :class:`openmc.deplete.IndependentOperator` uses inner classes subclassed from
those listed above to perform similar calculations.
```

docs/source/usersguide/depletion.rst
```diff
@@ -4,50 +4,30 @@
...
-
-.. important:: The volume must be specified for each material that is depleted by
-               setting the :attr:`Material.volume` attribute. This is necessary
-               in order to calculate the proper normalization of tally results
-               based on the source rate.
+OpenMC supports transport-coupled and transport-independent depletion, or
+burnup, calculations through the :mod:`openmc.deplete` Python module. OpenMC
+uses transmutation reaction rates to solve a set of transmutation equations
+that determine the evolution of nuclide densities within a material. The
+nuclide densities predicted as some future time are then used to determine
```
paulromano, line 11:
```suggestion
nuclide densities predicted at some future time are then used to determine
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,163 @@ across all material instances.
...
     number of tallies and material definitions.
 
+Transport-independent depletion
+===============================
+
+.. note::
+   
+   This feature is still under heavy development and has yet to be verifed
+   code-to-code . API changes and feature additions are possible and likely in
+   the near future.
```
paulromano, line 202:
```suggestion
   This feature is still under heavy development and has yet to be rigorously
   verified. API changes and feature additions are possible and likely in
   the near future.
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,163 @@ across all material instances.
...
+    micro_xs = MicroXS.from_csv(micro_xs_path)
+
+    nuclides = ['U234', 'U235', 'U238']
+    reactions = ['fission', '(n,gamma)']
+    data = np.array([[0.1, 0.2],
+                     [0.3, 0.4],
+                     [0.01, 0.5]])
+    micro_xs = MicroXS.from_array(nuclides, reactions, data)
+
+.. important ::
```
paulromano, line 285:
```suggestion
.. important::
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,163 @@ across all material instances.
...
+normalizing reaction rates:
+
+.. important::
+
+   Make sure you set the correct parameter in the :class:`openmc.abc.Integrator`
+   class. Use the ``source_rates`` parameter when
+   ``normalization_mode == source-rate``, and use ``power`` or ``power_density``
+   when ``normalization_mode == fission-q``.
+
+1. ``soure-rate`` normalization, which assumes the ``source-rate`` provided by
```
paulromano, line 309:
```suggestion
1. ``source-rate`` normalization, which assumes the ``source-rate`` provided by
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,163 @@ across all material instances.
...
+   the time integrator is a flux, and obtains the reaction rates by multiplying
+   the cross-sections by the ``source-rate``.
+2. ``fission-q`` normalization, which assumes the ``source-rate`` provided by
```
paulromano, line 312: In this case, doesn't the user pass `power` or `power_density`, not `source-rate`?

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,163 @@ across all material instances.
...
+
+   where :math:`\sum_i` is the sum over all nuclides :math:`i`. This equation
+   makes the same assumptions and issues as discussed in
+   :ref:`energy-deposition`. Unfortunately, the proposed solution in that
+   section does not apply here since we are decoupled from transport code.
+   However, there is a method to converge to a more accurate value for flux by
+   using substeps during time integration.
+   `This paper <https://doi.org/10.1016/j.anucene.2016.05.031>`_ provides a
+   good discussion of this method. Hopefully such a method will be implemented
+   in OpenMC in the near future.
```
paulromano, line 328: I try to avoid making promises in the documentation
```suggestion
   good discussion of this method.
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,163 @@ across all material instances.
...
+Multiple Materials
+~~~~~~~~~~~~~~~~~~
+
+Running a depletion simulation with multiple materials using the
+``source-rate`` normalization method treats each material as completely
+separate with respect to reaction rates. This can be useful for running many
+different cases of a particular scenario. However, running a depletion
+simulation with multiple materials using the ``fission-q`` normalization method
+treats each material as part of the same "reactor" due to how ``fission-q``
+normalization conglomerates energy values from each material to a single value.
```
paulromano, line 345:
```suggestion
normalization accumulates energy values from each material to a single value.
```

paulromano (changes requested):

docs/source/usersguide/depletion.rst
yardasol: Correct. I think I was referring to how internally in `Integrator`, the variable passed to the operator classes no matter what the user provides is `source_rate`, but that doesn't really make a difference to the user. I'll adjust the wording here.

docs/source/usersguide/depletion.rst
yardasol: haha that's a good recommendation

openmc/deplete/abc.py
yardasol: Okay, I'll switch it back to TransportOperator then.

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,169 @@
...
+        micro_xs = cls()
+        for rxn in xs:
+            df = xs[rxn].get_pandas_dataframe(xs_type='micro')
+            df.index = df['nuclide']
+            df.drop(['nuclide', xs[rxn].domain_type, 'group in', 'std. dev.'], axis=1, inplace=True)
+            df.rename({'mean': rxn}, axis=1, inplace=True)
+            micro_xs = concat([micro_xs, df], axis=1)
+
+        micro_xs._units = 'b'
+
```
paulromano, line 93:
```suggestion
```

docs/source/pythonapi/deplete.rst
```diff
@@ -12,19 +12,20 @@ Primary API
...
-and retaining important information required for depletion. The most common examples
-are reaction rates and power normalization data. The latter is responsible for
-projecting reaction rates and compositions forward in calendar time across
-some step size :math:`\Delta t`, and obtaining new compositions given a power
-or power density. The :class:`Operator` is provided to handle communicating with
-OpenMC. Several classes are provided that implement different time-integration
-algorithms for depletion calculations, which are described in detail in Colin
-Josey's thesis, `Development and analysis of high order neutron
-transport-depletion coupling algorithms <http://hdl.handle.net/1721.1/113721>`_.
+The former is responsible for calcuating and retaining important information required for depletion. The most common examples are reaction rates and power
```
paulromano, line 18:
```suggestion
The former is responsible for calculating and retaining important information
required for depletion. The most common examples are reaction rates and power
```

docs/source/methods/depletion.rst
```diff
@@ -103,16 +103,16 @@ integrate over the entire timestep.
...
-its error converges as :math:`\mathcal{O}(h)`. The CE/CM method requires two
-evaluations and is thus twice as expensive as the predictor method, but achieves
-an error of :math:`\mathcal{O}(h^2)`. An exhaustive description of time
-integration methods and their merits can be found in the `thesis of Colin Josey
+accuracy of the method and its computational expense. In the case of
+transport-coupled depletion, the expense is driven almost entirely by the time
+to compute a transport solution, i.e., to evaluate :math:`\mathbf{A}` for a
+given :math:`\mathbf{n}`. Thus, the cost of a method scales with the number of
+:math:`\mathbf{A}` evaluations that are performed per timestep. On the other
+hand, methods that require more evaluations generally achieve higher accuracy. The predictor method only requires one evaluation and its error converges as
```
paulromano, line 111:
```suggestion
hand, methods that require more evaluations generally achieve higher accuracy.
The predictor method only requires one evaluation and its error converges as
```

docs/source/usersguide/depletion.rst
```diff
@@ -56,11 +36,40 @@ time::
     results = openmc.deplete.Results("depletion_results.h5")
     time, keff = results.get_keff()
 
-Note that the coupling between the transport solver and the transmutation solver
-happens in-memory rather than by reading/writing files on disk.
+Note that the coupling between the reaction rate solver and the transmutation
+solver happens in-memory rather than by reading/writing files on disk. OpenMC has two categories of transport operators for obtaining transmutation reaction
```
paulromano, line 40:
```suggestion
solver happens in-memory rather than by reading/writing files on disk. OpenMC
has two categories of transport operators for obtaining transmutation reaction
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,162 @@ across all material instances.
...
+    import openmc
+
+    model = openmc.Model.from_xml()
+
+    micro_xs = openmc.deplete.MicroXS.from_model(model, model.materials[0])
+
+The :meth:`~openmc.deplete.MicroXS.from_model()` method will produce a
+:class:`~openmc.deplete.MicroXS` object with microscopic cross section data in
+units of ``b``, which is what :class:`~openmc.deplete.IndependentOperator`
+expects the units to be. The :class:`~openmc.deplete.MicroXS` class also includes functions to read in cross section data directly from a ``.csv`` file or from data arrays::
```
paulromano, line 274:
```suggestion
expects the units to be. The :class:`~openmc.deplete.MicroXS` class also
includes functions to read in cross section data directly from a ``.csv`` file
or from data arrays::
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,162 @@ across all material instances.
     This will increase the total memory usage and run time due to an increased
     number of tallies and material definitions.
 
+Transport-independent depletion
+===============================
+
+.. note::
```
paulromano, line 198: Maybe change this to warning?
```suggestion
.. warning::
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,162 @@ across all material instances.
...
+and a path to a depletion chain file::
+
+    # load in the microscopic cross sections
+    materials = openmc.Materials()
+    ...
+
+    micro_xs = openmc.deplete.MicroXS()
+    ...
+
+    op = IndependentOperator(materials, micro_xs, chain_file)
```
paulromano, line 221:
```suggestion
    op = openmc.deplete.IndependentOperator(materials, micro_xs, chain_file)
```

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +192,162 @@ across all material instances.
...
+
+    import openmc
+
+    model = openmc.Model.from_xml()
+
+    micro_xs = openmc.deplete.MicroXS.from_model(model, model.materials[0])
+
+The :meth:`~openmc.deplete.MicroXS.from_model()` method will produce a
+:class:`~openmc.deplete.MicroXS` object with microscopic cross section data in
+units of ``b``, which is what :class:`~openmc.deplete.IndependentOperator`
```
paulromano, line 273:
```suggestion
units of barns, which is what :class:`~openmc.deplete.IndependentOperator`
```

docs/source/pythonapi/deplete.rst
```diff
@@ -12,19 +12,20 @@ Primary API
 The two primary requirements to perform depletion with :mod:`openmc.deplete`
 are:
 
-    1) A transport operator
+    1) A transpor operator
```
paulromano, line 15:
```suggestion
    1) A transport operator
```

openmc/deplete/abc.py
```diff
@@ -79,11 +78,12 @@ def change_directory(output_dir):
 class TransportOperator(ABC):
     """Abstract class defining a transport operator
 
-    Each depletion integrator is written to work with a generic transport
+    Each depletion integrator is written to work with a generic depletion
```
paulromano, line 81:
```suggestion
    Each depletion integrator is written to work with a generic transport
```

openmc/deplete/abc.py
```diff
@@ -598,7 +604,7 @@ def __init__(self, operator, timesteps, power=None, power_density=None,
             else:
                 source_rates = [p*operator.heavy_metal for p in power_density]
         elif source_rates is None:
-            raise ValueError("Either power, power_density, or source_rates must be set")
+            raise ValueError("Either power, power_density, source_rates must be set")
```
paulromano, line 607:
```suggestion
            raise ValueError("Either power, power_density, or source_rates must be set")
```

openmc/deplete/helpers.py
```diff
@@ -610,7 +615,7 @@ class FissionYieldCutoffHelper(TalliedFissionYieldHelper):
         Default: 0.0253 [eV]
     fast_energy : float, optional
         Energy of yield data corresponding to fast yields.
-        Default: 500 [kev]
+        Default: 500 [KeV]
```
paulromano, line 618: Default is in the signature so let's just remove it
```suggestion
```

paulromano (changes requested):
I think we're almost there. By the way -- I'm not sure if you're aware but with suggested changes, you can go to the "Files changed" tab, add them to a batch and commit all in one go without having to make the changes yourself manually. I would also consider it a best practice since the person who makes the suggested changes gets attribution on the commit as a coauthor.

yardasol:
I didn't know that, thanks for the tip!
> I would also consider it a best practice since the person who makes the suggested changes gets attribution on the commit as a coauthor.

Whoops! Definitely good to know this for the future. Don't want to be stealing anyone's work.

paulromano (approved):

yardasol:
@paulromano Please hold off on merging. I have a small change I'd like to propose that will make using the `MicroXS.from_model()` method  bit more useful and practical. Right now it will only return cross sections from nuclides that are present in the `material` contained in the model, but a user would realistically have a chain file that contains hundreds and thousands of nuclides, not all of which are going to be present in their initial material.

paulromano:
@yardasol If it's not too difficult, let's go ahead and include that update with this PR. I agree that, in light of those arguments, the `MicroXS.from_model` class wouldn't be terribly useful.

yardasol:
Okay, I've implemented this feature using `dilute_initial` in an identical way to how it's used in `CoupledOperator`. I've also added `chain_file` as a positional argument to `from_model`

docs/source/usersguide/depletion.rst
```diff
@@ -177,3 +193,166 @@ across all material instances.
...
+coupled to any transport solver. The
+:class:`~openmc.deplete.IndependentOperator` class requires a
+:class:`openmc.Materials` object, a :class:`~openmc.deplete.MicroXS` object,
+and a path to a depletion chain file::
+
+    # load in the microscopic cross sections
+    materials = openmc.Materials()
+    ...
+
+    micro_xs = openmc.deplete.MicroXS
```
paulromano, line 219: Probably should add `from_...(...)`?

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,229 @@
...
+        nuclides = set()
+        data_lib = DataLibrary.from_xml(cross_sections)
+        for library in data_lib.libraries:
+            if library['type'] != 'neutron':
+                continue
+            for name in library['materials']:
+                if name not in nuclides:
+                    nuclides.add(name)
+
+        return nuclides
```
paulromano, line 157: This is exactly the same as `CoupledOperator._get_nuclides_with_data`. Let's just make this a normal function and then call it from both the `CoupledOperator` and `MicroXS` classes.

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,229 @@
...
+            df.rename({'mean': rxn}, axis=1, inplace=True)
+            micro_xs = concat([micro_xs, df], axis=1)
+
+        # Revert to the original tallies
+        model.tallies = original_tallies
+
+        return micro_xs
+
+    @classmethod
+    def _add_dilute_nuclides(cls, chain_file, model, dilute_initial):
```
paulromano, line 110: Add a docstring on this -- obviously not important for users since this is internal but docstrings are also useful for us developers :smile:

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,229 @@
...
+        groups = EnergyGroups(energy_bounds)
+
+        # Set up the reaction tallies
+        original_tallies = model.tallies
+        tallies = Tallies()
+        xs = {}
+        reactions, diluted_materials = cls._add_dilute_nuclides(chain_file,
+                                                                model,
+                                                                dilute_initial)
+        diluted_materials.export_to_xml('diluted_materials.xml')
```
paulromano, line 75: Doesn't look like this is used
```suggestion
```

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,229 @@
...
+        reactions = chain.reactions
+        cross_sections = _find_cross_sections(model)
+        nuclides_with_data = cls._get_nuclides_with_data(cross_sections)
+        burnable_nucs = [nuc.name for nuc in chain.nuclides
+                               if nuc.name in nuclides_with_data]
+        diluted_materials = Materials()
+        for material in model.materials:
+            if material.depletable:
+                nuc_densities = material.get_nuclide_atom_densities()
+                dilute_density = 1.E-24 * dilute_initial
```
paulromano, line 121: To match prevailing style throughout OpenMC codebase
```suggestion
                dilute_density = 1.e-24 * dilute_initial
```

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,229 @@
...
+                material.set_density('sum')
+                for nuc, density in nuc_densities.items():
+                    material.remove_nuclide(nuc)
+                    material.add_nuclide(nuc, density)
+                for burn_nuc in burnable_nucs:
+                    if burn_nuc not in nuc_densities:
+                        material.add_nuclide(burn_nuc,
+                                             dilute_density)
+            diluted_materials.append(material)
```
paulromano, line 130: This seems like it could be a potentially annoying side effect of calling `MicroXS.from_model`. Namely, the `Model` object that's passed in will be changed as a result of this block of code. Probably what we should do is 1) create new `Material` objects that match the original ones but have the dilute nuclides added, 2) create a new copy of `Model` that uses the same geometry but the new materials.

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,229 @@
...
+        return micro_xs
+
+    @classmethod
+    def _add_dilute_nuclides(cls, chain_file, model, dilute_initial):
+        chain = Chain.from_xml(chain_file)
+        reactions = chain.reactions
+        cross_sections = _find_cross_sections(model)
+        nuclides_with_data = cls._get_nuclides_with_data(cross_sections)
+        burnable_nucs = [nuc.name for nuc in chain.nuclides
+                               if nuc.name in nuclides_with_data]
```
paulromano, line 116:
```suggestion
        burnable_nucs = [nuc.name for nuc in chain.nuclides
                         if nuc.name in nuclides_with_data]
```

paulromano (changes requested):

openmc/deplete/microxs.py
yardasol: Fair point

openmc/deplete/microxs.py
yardasol: This was for testing purposed and forgot to remove it when I committed.

openmc/deplete/microxs.py
yardasol: In that case there are a few other places this needs to be updated.

openmc/deplete/microxs.py
yardasol: done

openmc/deplete/microxs.py
yardasol: I attempted implementing this but am now getting a weird error:

```python
    def test_from_model(model):
        ref_xs = MicroXS.from_csv('test_reference.csv')
>       test_xs = MicroXS.from_model(model, model.materials[0], CHAIN_FILE)

/home/ooblack/projects/openmc/tests/regression_tests/microxs/test.py:50: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/ooblack/projects/openmc/openmc/deplete/microxs.py:98: in from_model
    df = xs[rxn].get_pandas_dataframe(xs_type='micro')
/home/ooblack/projects/openmc/openmc/mgxs/mgxs.py:2112: in get_pandas_dataframe
    df['mean'] /= np.tile(densities, tile_factor)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/generic.py:11387: in __itruediv__
    return self._inplace_method(
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/generic.py:11352: in _inplace_method
    result = op(self, other)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/ops/common.py:70: in new_method
    return method(self, other)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/arraylike.py:124: in __truediv__
    return self._arith_method(other, operator.truediv)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/series.py:5636: in _arith_method
    return base.IndexOpsMixin._arith_method(self, other, op)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/base.py:1295: in _arith_method
    result = ops.arithmetic_op(lvalues, rvalues, op)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/ops/array_ops.py:222: in arithmetic_op
    res_values = _na_arithmetic_op(left, right, op)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/ops/array_ops.py:163: in _na_arithmetic_op
    result = func(left, right)
/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/computation/expressions.py:239: in evaluate
    return _evaluate(op, op_str, a, b)  # type: ignore[misc]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

op = <built-in function truediv>, op_str = '/'
a = array([1.98362959e-04, 1.04605679e-02, 1.92481982e-02, 3.95592500e-05,
       3.47830212e-06, 7.17484119e-09])
b = array([], dtype=float64)

    def _evaluate_standard(op, op_str, a, b):
        """
        Standard evaluation.
        """
        if _TEST_MODE:
            _store_test_result(False)
>       return op(a, b)
E       ValueError: operands could not be broadcast together with shapes (6,) (0,)

/home/ooblack/miniconda3/envs/openmc-env/lib/python3.10/site-packages/pandas/core/computation/expressions.py:69: ValueError

```

It's unclear how my change is effecting the `mgxs` side of things. I'll poke around tomorrow to see if I can fix it, but putting it here in the meantime in case anyone want's to take a crack at it.

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,235 @@
...
+        nuclides_with_data = _get_nuclides_with_data(cross_sections)
+        burnable_nucs = [nuc.name for nuc in chain.nuclides
+                         if nuc.name in nuclides_with_data]
+        diluted_materials = Materials()
+        reset_auto_ids()
+        for material in model.materials:
+            if material.depletable:
+                nuc_densities = material.get_nuclide_atom_densities()
+                dilute_density = 1.e-24 * dilute_initial
+                diluted_material = Material()
```
paulromano, line 151: Rather than use `reset_auto_ids` (which itself is another undesired side effect), let's just silence the warning from creating another material with the same ID:
```Python
import warnings

with warnings.catch_warnings():
    warnings.simplefilter("ignore", openmc.IDWarning)
    diluted_material = Material(material_id=material.id)
...
```

paulromano (changes requested):

openmc/deplete/microxs.py
paulromano: I think the problem may be something like the following:
- `reaction_domain` is set to `model.materials[0]`
- Under the hood, we change model.materials to the new materials with dilute nuclides
- `reaction_domain` is passed to `FissionXS`, which then tries to determine the nuclides based on that material (i.e., `model.materials[0]`, not the new material with dilute nuclides)
- Thus, the tallies that are created don't include all the nuclides we need

yardasol: subtle point! I'll keep this in mind.

yardasol: So my workaround for this is just to create a `deepcopy` of the original materials object, do all the dilution to the `Materials` object like I did before, and then once the openmc simulation is done, set `model.materials` to the deepcopy.

openmc/deplete/microxs.py
```diff
@@ -0,0 +1,233 @@
...
+            df.index = df['nuclide']
+            df.drop(['nuclide', xs[rxn].domain_type, 'group in', 'std. dev.'], axis=1, inplace=True)
+            df.rename({'mean': rxn}, axis=1, inplace=True)
+            micro_xs = concat([micro_xs, df], axis=1)
+
+        # Revert to the original tallies and materials
+        model.tallies = original_tallies
+        model.materials = original_materials
+
+        return cls(micro_xs)
```
paulromano, line 109: `micro_xs` was created using `cls` above -- why is it necessary to pass it to `cls` again?

yardasol: We are using `pandas.concat`, which returns a `pandas.DataFrame` instead of a `MicroXS`, so we need initialize the `MicroXS` object again.

paulromano, line 109: Here's a nice simplification to make -- this pulls out the 'mean' column as a `Series`, puts it in a dictionary, and then constructs the dataframe from the dict of series. Also, our standard abbreviation in OpenMC for "reaction" is `rx`
```suggestion
        series = {}
        for rx in xs:
            df = xs[rx].get_pandas_dataframe(xs_type='micro')
            series[rx] = df.set_index('nuclide')['mean']

        # Revert to the original tallies and materials
        model.tallies = original_tallies
        model.materials = original_materials

        return cls(series)
```

yardasol:
@paulromano I think this should be okay to merge now, unless anyone has any further comments.

paulromano (approved):
Thanks for getting this to the finish line @yardasol!

PR #2106, opened 2022-07-06
histogram plot with auto bins for source energy

shimwell:
I find myself often plotting histograms of the source energy quite often and I thought this tiny snippet of code added to the source object might be handy for others. I quite like the auto binning that numpy does behind the scenes. I've followed the same plotting function layout that we see elsewhere in the code ([e.g. tracks.py](https://github.com/openmc-dev/openmc/blob/6549067c43bb00750fe4f3d772795217575380ea/openmc/tracks.py#L148-L179))

Plotting the source energy can then be performed with ```Source.plot_energy()```

This makes use of the ```sample``` method that @pshriwise recently added to the ```Univariate``` in PR #2080 and uses the same arg names ```n_samples``` and ```seed```.

Here is an example plot of the Muir DT fusion energy spectra

openmc-source-plotter-fusion-energy

```
import openmc

my_source = openmc.Source()

my_source.energy = openmc.stats.Muir(e0=14080000.0, m_rat=5.0, kt=20000.0)

my_source.plot_energy()

plt.show()
```

### DIFF
--- a/openmc/source.py
+++ b/openmc/source.py
@@ -198,7 +198,7 @@ class Source:
             element.append(self.time.to_xml_element('time'))
         return element
 
-    def plot_energy(self, n_samples=10000, seed=None, axes=None):
+def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
         """Produce a histogram plot of source particle energy.
 
         Parameters
@@ -230,7 +230,7 @@ class Source:
         else:
             ax = axes
 
-        ax.hist(energy, bins='auto')
+        ax.hist(energy, bins='auto', **kwargs)
 
         return ax
 
--- a/openmc/source.py
+++ b/openmc/source.py
@@ -198,7 +198,7 @@ class Source:
             element.append(self.time.to_xml_element('time'))
         return element
 
-def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
+    def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
         """Produce a histogram plot of source particle energy.
 
         Parameters
@@ -223,8 +223,7 @@ def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
 
         # Setup axes is one wasn't passed
         if axes is None:
-            fig = plt.figure()
-            ax = plt.axes()
+            ax = plt.gca()
             ax.set_xlabel('Energy [eV]')
             ax.set_ylabel('Frequency')
         else:
--- a/openmc/source.py
+++ b/openmc/source.py
@@ -198,6 +198,43 @@ class Source:
             element.append(self.time.to_xml_element('time'))
         return element
 
+    def plot_energy(self, n_samples=10000, seed=None, axes=None):
+        """Produce a histogram plot of source particle energy.
+
+        Parameters
+        ----------
+        n_samples : int
+            Number of sampled values to generate
+        seed : int or None
+            Initial random number seed.
+        axes : matplotlib.axes.Axes, optional
+            Axes for plot
+
+        Returns
+        -------
+        axes : matplotlib.axes.Axes
+            Axes for plot
+
+        """
+
+        energy = self.energy.sample(n_samples=n_samples, seed=seed)
+
+        import matplotlib.pyplot as plt
+
+        # Setup axes is one wasn't passed
+        if axes is None:
+            fig = plt.figure()
+            ax = plt.axes()
+            ax.set_xlabel('Energy [eV]')
+            ax.set_ylabel('Frequency')
+        else:
+            ax = axes
+
+        ax.hist(energy, bins='auto')
+
+        return ax
+
+
     @classmethod
     def from_xml_element(cls, elem):
         """Generate source from an XML element


openmc/source.py
```diff
@@ -198,6 +198,43 @@ def to_xml_element(self):
...
+            fig = plt.figure()
+            ax = plt.axes()
+            ax.set_xlabel('Energy [eV]')
+            ax.set_ylabel('Frequency')
+        else:
+            ax = axes
+
+        ax.hist(energy, bins='auto')
+
+        return ax
```
RemDelaporteMathurin, line 235: I would return the output of `hist()` here. Something like:

```suggestion
        hist = ax.hist(energy, bins='auto')

        return hist
```

openmc/source.py
```diff
@@ -198,6 +198,43 @@ def to_xml_element(self):
...
+        """
+
+        energy = self.energy.sample(n_samples=n_samples, seed=seed)
+
+        import matplotlib.pyplot as plt
+
+        # Setup axes is one wasn't passed
+        if axes is None:
+            fig = plt.figure()
+            ax = plt.axes()
```
RemDelaporteMathurin, line 227: This will create a new figure by default. How about using `plt.gca()` (for Get Current Axis)?

This way, users will be able to effortlessly include this function into their plotting workflow
```suggestion
            ax = plt.gca()
```

openmc/source.py
```diff
@@ -198,6 +198,43 @@ def to_xml_element(self):
             element.append(self.time.to_xml_element('time'))
         return element
 
+    def plot_energy(self, n_samples=10000, seed=None, axes=None):
```
RemDelaporteMathurin, line 201: I think the convention for this case is to use `ax` or `axis` (not `axes` plural)
```suggestion
    def plot_energy(self, n_samples=10000, seed=None, ax=None):
```

RemDelaporteMathurin, line 201: I would expose the `hist` arguments for more flexibility
```suggestion
    def plot_energy(self, n_samples=10000, seed=None, axes=None, **kwargs):
```

openmc/source.py
```diff
@@ -198,6 +198,43 @@ def to_xml_element(self):
...
+        # Setup axes is one wasn't passed
+        if axes is None:
+            fig = plt.figure()
+            ax = plt.axes()
+            ax.set_xlabel('Energy [eV]')
+            ax.set_ylabel('Frequency')
+        else:
+            ax = axes
+
+        ax.hist(energy, bins='auto')
```
RemDelaporteMathurin, line 233: see comment above
```suggestion
        ax.hist(energy, bins='auto', **kwargs)
```

openmc/source.py
RemDelaporteMathurin: The current way of combining this function with another plot would be to do:

```python
plt.plot([1, 2, 3])

my_source.plot_energy(ax=plt.gca())

plt.ylabel("My custom y label")

plt.show()
```

Which is imho less convenient than what's proposed in this comment:

```python
plt.plot([1, 2, 3])

my_source.plot_energy()

plt.ylabel("My custom y label")

plt.show()
```

Furthermore calling it without anything done before will automatically create a new figure just as the current implementation.

openmc/source.py
RemDelaporteMathurin: This is just so that users have control over the objects of the histogram. It adds flexibility:
```python

n, bins, patches = my_source.plot_energy()

for i, p in enumerate(patches):
    # change transparency, cause why not?
    p.set_alpha((i/len(patches)))
```
image

shimwell: Nice example of why hist should be returned, perhaps we should change tracks.py sometime as well
https://github.com/openmc-dev/openmc/blob/6549067c43bb00750fe4f3d772795217575380ea/openmc/tracks.py#L179

openmc/source.py
```diff
@@ -198,6 +198,43 @@ def to_xml_element(self):
...
+        import matplotlib.pyplot as plt
+
+        # Setup axes is one wasn't passed
+        if axes is None:
+            fig = plt.figure()
+            ax = plt.axes()
+            ax.set_xlabel('Energy [eV]')
+            ax.set_ylabel('Frequency')
+        else:
+            ax = axes
```
RemDelaporteMathurin, line 231: I don't think this is needed.

openmc/source.py
shimwell:
```suggestion
def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
```
How about ```axes``` as it is a matplotlib.axes.Axes object and tracks.py uses this arg name

openmc/source.py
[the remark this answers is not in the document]
RemDelaporteMathurin: Fine with me

openmc/source.py
shimwell: Are you sure that we can remove the ``` fig = plt.figure() ``` line as it is also here https://github.com/openmc-dev/openmc/blob/6549067c43bb00750fe4f3d772795217575380ea/openmc/tracks.py#L166

RemDelaporteMathurin: a new figure is created if there's no active axis

RemDelaporteMathurin: On this snipset, adding `plt.figure()` or not doesn't change anything
```python
import numpy as np

plt.figure()

n, bins, patches = plt.hist(np.random.random(100))

for i, p in enumerate(patches):
    p.set_alpha((i/len(patches)))


plt.show()
```

openmc/source.py
```diff
@@ -198,6 +198,43 @@ def to_xml_element(self):
             element.append(self.time.to_xml_element('time'))
         return element
 
+def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
```
RemDelaporteMathurin, line 201: This is under-indented and that's why the tests are failing
```suggestion
    def plot_energy(self, n_samples=10000, seed=None, axes=None,  **kwargs):
```

shimwell: Thanks Remi

shimwell:
No worries, I shall put source plotting functions over in this package.

https://github.com/fusion-energy/openmc_source_plotter

If I come up with anything more interesting I shall be back with source plotting ideas.

PR #2110, opened 2022-07-11
Fixed write_data_to_vtk dimension mismatch

### LINKED ISSUE #2109: write_data_to_vtk producing scambled vtk files
Hi all

While using the nice new ```write_data_to_vtk``` with regular mesh tallies. I've been getting a vtk file which appears to be a bit messed up and I just wanted to check if I am doing things correctly.

I've defined a mesh like this
```
mesh = openmc.RegularMesh()

mesh.dimension = [100, 50, 100]  # number of mesh cells in each dimension (resolution)

mesh.lower_left = [-750, 0, -750]  # x,y,z coordinates
mesh.upper_right = [750, 750, 750]  # x,y,z coordinates
```

I've extracted a mesh tally like this and written it to vtk
```
statepoint = openmc.StatePoint('statepoint.2.h5')

my_tbr_tally = statepoint.get_tally(name='tbr_on_mesh')

mesh.write_data_to_vtk(
    filename="tbr_tally_on_mesh5.vtk",
    datasets={"mean": my_tbr_tally.mean}
)
```

However I get a vtk geometry like this

Screenshot from 2022-07-11 12-17-42

RemDelaporteMathurin:
This fixes #2109 

Somehow during the refactor of #2096 this bug occured.

I'm working on a test to catch this bug!

### DIFF
--- a/openmc/mesh.py
+++ b/openmc/mesh.py
@@ -240,7 +240,7 @@ class StructuredMesh(MeshBase):
 
         vtk_grid = vtk.vtkStructuredGrid()
 
-        vtk_grid.SetDimensions(*self.dimension)
+        vtk_grid.SetDimensions(*[dim + 1 for dim in self.dimension])
 
         vtkPts = vtk.vtkPoints()
         vtkPts.SetData(nps.numpy_to_vtk(points, deep=True))


pshriwise:
Thanks for the quick response @RemDelaporteMathurin! Let me know when that test is in and this is ready for review!

RemDelaporteMathurin:
> Thanks for the quick response @RemDelaporteMathurin! Let me know when that test is in and this is ready for review!

I'm still thinking how we could test it. Any idea?

makeclean:
Isn't the issue here that the code was not correct, not that it was used incorrectly?

RemDelaporteMathurin:
> 
Not sure I get what you mean but yeah it was an error in the code

makeclean:
Well my point is the thing you're trying to test under failure conditions is incorrect, you are trying to create a test that asserts correct behaviour under the new conditions, but a user cannot create the conditions of the original problem. I'm not sure you can add a test here that verifies correctness, other than creating say a simple enough structured mesh, say 8 elements, you can export it to vtk, re-ingest it into vtk, then perform a range of queries to assert that the mesh is indeed as you would expect, including connectivity, the values in the cells etc

pshriwise:
> ...other than creating say a simple enough structured mesh, say 8 elements, you can export it to vtk, re-ingest it into vtk, then perform a range of queries to assert that the mesh is indeed as you would expect, including connectivity, the values in the cells etc

Sounds like a nice test to have to me!

RemDelaporteMathurin:
Oh sorry for the misunderstanding.

I wanted a test to add to the CI so that this behaviour is caught in the future.
I didn't mean a test on the user side

shimwell:
Thanks @RemDelaporteMathurin you are a star. The regular mesh tally is making nice vtk files. Thanks for maintaining the nice feature you put in.

paulromano:
Pretty clear to me that this fix is needed, so I'll go ahead and merge. @RemDelaporteMathurin if you are able to figure out an appropriate test, we would of course welcome an additional PR!

PR #2151, opened 2022-08-05
Spatial sampling on an unstructured mesh

NybergWISC:
Includes sampling over tetrahedral meshes (libmesh and moab) implemented as a separate independent source class MeshIndependent


src/mesh.cpp
```diff
@@ -2004,6 +2039,34 @@ std::string MOABMesh::library() const
...
+
+  moab::EntityHandle tet_ent = get_ent_handle_from_bin(tet_bin);
+
+  // Get vertex coordinates for MOAB tet
+  vector<moab::EntityHandle> conn1;
+  moab::ErrorCode rval = mbi_->get_connectivity(&tet_ent, 1, conn1);
+  if (rval != moab::MB_SUCCESS) {
+    fatal_error("Failed to get tet connectivity");
+  }
+  moab::CartVect p[4];
```
lewisgross1296, line 2053: Nit pick, but just having 4 is a magic number. It would be better to make a descriptive int variable so that anyone who looks at it later knows why this number. Then you should replace the 4 that appear after with the variable name where it appears. Also, if for some reason this 4 changes in the future, then you only need to replace it in one spot. Otherwise you may miss one and it could cause a hard-to-find error.

openmc/stats/multivariate.py
```diff
@@ -3,12 +3,14 @@
 from math import pi, cos
 from numbers import Real
 from xml.etree import ElementTree as ET
+from xmlrpc.client import boolean
```
NybergWISC, line 6: Boolean is currently not used, will remove import

src/distribution_spatial.cpp
```diff
@@ -180,6 +185,78 @@ Position SphericalIndependent::sample(uint64_t* seed) const
...
+  mesh_weights_.resize(tot_bins_);
+
+  // Create cdfs for sampling for an element over a mesh
+  // Volume scheme is weighted by the volume of each tet
+  // File scheme is weighted by an array given in the xml file
+
+  if (check_for_node(node, "weights_from_file")) {
+    mesh_weights_ = get_node_array<double>(node, "weights_from_file");
+    if (mesh_weights_.size() != tot_bins_){
+      write_message("WARNING: The size of the weights array from the xml file does not equal the number of elements in the mesh.");
```
NybergWISC, line 222: I wondered whether this should be a warning or an error, in either case this line should use a real warning rather than just printing it out

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+
+    materials.export_to_xml()
+
+    ### Geometry ###
+    dimen = 10
+    size_hex = 20.0/dimen
+
+    ### Geometry ###
+    cell = np.empty((dimen, dimen, dimen), dtype=object)
+    surfaces = np.empty((dimen+1, 3), dtype=object)
```
pshriwise, line 116:
```suggestion
    surfaces = np.empty((dimen + 1, 3), dtype=object)

```

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+
+    geometry = openmc.Geometry()
+    universe = openmc.Universe(universe_id=1, name="Contains all hexes")
+
+    for i in range(0,dimen+1):
+        surfaces[i][0] = openmc.XPlane(-10.0+i*size_hex, name="X plane at "+str(-10.0+i*size_hex))
+        surfaces[i][1] = openmc.YPlane(-10.0+i*size_hex, name="Y plane at "+str(-10.0+i*size_hex))
+        surfaces[i][2] = openmc.ZPlane(-10.0+i*size_hex, name="Z plane at "+str(-10.0+i*size_hex))
```
pshriwise, line 124:
```suggestion
        coord = -10.0+i*size_hex
        surfaces[i][0] = openmc.XPlane(coord, name=f"X plane at {coord}")
        surfaces[i][1] = openmc.YPlane(coord, name=f"Y plane at {coord}")
        surfaces[i][2] = openmc.ZPlane(coord, name=f"Z plane at {coord}")

```

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+
+        surfaces[i][0].boundary_type = 'vacuum'
+        surfaces[i][1].boundary_type = 'vacuum'
+        surfaces[i][2].boundary_type = 'vacuum'
+
+    for k in range(0,dimen):
+        for j in range(0,dimen):
+            for i in range(0,dimen):
+                cell[i][j][k] = openmc.Cell(name=("x " + str(i) +" y " + str(j) +" z " + str(k)))
```
pshriwise, line 133:
```suggestion
                cell[i][j][k] = openmc.Cell(name=("x = {i}, y = {j}, z = {k}"))
```

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+                cell[i][j][k].region = +surfaces[i][0] & -surfaces[i+1][0] & \
+                                    +surfaces[j][1] & -surfaces[j+1][1] & \
+                                    +surfaces[k][2] & -surfaces[k+1][2]
+                cell[i][j][k].fill = water_mat
```
pshriwise, line 137: Looking at this model again, I think we can make every cell void since all we care about is where the particles are sourced.

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+    tallies.export_to_xml()
+
+    ### Settings ###
+    settings = openmc.Settings()
+    settings.run_mode = 'fixed source'
+    settings.particles = 5000
+    settings.batches = 2
+
+    # settings.create_fission_neutrons = False
+    settings.tracks = [(1, 1, 1)]
```
pshriwise, line 164: After reading the documentation on the new tracks capability, I don't _think_ we need this line.

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+    energy = openmc.stats.Discrete(x=[15.e+06], p=[1.0])
+    source = openmc.Source(space=space, energy=energy)
+    settings.source = source
+
+    model = openmc.model.Model(geometry=geometry,
+                               materials=materials,
+                               tallies=tallies,
+                               settings=settings)
+
+    harness = UnstructuredMeshSourceTest('statepoint.2.h5',
```
pshriwise, line 186: Indentation fixes needed here

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+import pytest
+from tests.testing_harness import PyAPITestHarness
+from tests.regression_tests import config
+from subprocess import call
+
+TETS_PER_VOXEL = 12
+
+class UnstructuredMeshSourceTest(PyAPITestHarness):
+    def __init__(self, statepoint_name, model, inputs_true, schemes):
+
```
pshriwise, line 18:
```suggestion

```

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+import os
+
+import openmc
+import openmc.lib
+import numpy as np
+
+import pytest
+from tests.testing_harness import PyAPITestHarness
+from tests.regression_tests import config
+from subprocess import call
```
pshriwise, line 12:
```suggestion
import glob
from itertools import product
import os

import pytest
import numpy as np

import openmc
import openmc.lib

from tests.testing_harness import PyAPITestHarness
from tests.regression_tests import config
from subprocess import call
```

src/distribution_spatial.cpp
NybergWISC: I used this mostly for testing, it was easier to use a small array and have everything else be zero rather than create a 12000 size array. Current tests should not be affected

openmc/stats/multivariate.py
```diff
@@ -613,6 +617,120 @@ def from_xml_element(cls, elem):
...
+        A list of values which represent the weights of each element
+
+
+    Attributes
+    ----------
+    elem_weight_scheme : str, optional
+        Weighting scheme for sampling element from mesh, defaults to volume
+    mesh : openmc.MeshBase
+        The mesh instance used for sampling, mesh is written into settings.xml, mesh.id is written into the source distribution
+    weights_from_file : Iterable of Real, optional
```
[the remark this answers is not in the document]
NybergWISC: Sounds good to me

tests/regression_tests/unstructured_mesh/source_sampling/test.py
NybergWISC: Should they be lined up with the first variable set? So for instance the m from model should be in the same column as the ' from 'statepoint.2.h5'?

pshriwise: Yep yep

include/openmc/mesh.h
```diff
@@ -453,7 +461,12 @@ class UnstructuredMesh : public Mesh {
   static const std::string mesh_type;
   virtual std::string get_mesh_type() const override;
 
+  //! Sample a tetrahedron for an unstructured mesh
```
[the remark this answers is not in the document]
NybergWISC: I believe since I'm calling the method from a derived class it must be protected rather than private. That is what will be implemented in the next commit

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,190 @@
...
+            kwargs['mpi_args'] = [config['mpiexec'], '-n', config['mpi_np']]
+
+        openmc.run(**kwargs)
+
+    def _compare_results(self):
+
+        average_in_hex = 10.0
+
+
+        call(['../../../../scripts/openmc-track-combine', '-o', 'tracks.h5'] +
```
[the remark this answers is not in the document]
NybergWISC: I can't quite get openmc.Tracks.combine to work. I have left it as is in the current commit (commit as of ~11:30 8/8) but think this is a good suggestion

src/mesh.cpp
NybergWISC: I agree in principle. I think I will leave this code as is because it matches a similar implementation in a different file and the '4' in this case will never change and should be recognized as the number of vertices on a tet

lewisgross1296: Gotcha, yeah I'd say that is up to you and @pshriwise. It makes sense that four won't change since a tet always has four vertices. If it matches another file, then maybe it does make more sense to leave it, so the code is parallel to itself in style (or change everywhere it appears 😉 )

openmc/stats/multivariate.py
```diff
@@ -613,6 +617,120 @@ def from_xml_element(cls, elem):
         origin = [float(x) for x in elem.get('origin').split()]
         return cls(r, phi, z, origin=origin)
 
+class MeshIndependent(Spatial):
```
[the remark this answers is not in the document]
NybergWISC: I think MeshSpatial would be the best change, it seems like there aren't any naming conventions if it is not part of the *Independent group.

include/openmc/distribution_spatial.h
```diff
@@ -94,6 +95,29 @@ class SphericalIndependent : public SpatialDistribution {
...
+  Position sample(uint64_t* seed) const;
+
+private:
+  UnstructuredMesh* umesh_ptr_;
+  int32_t mesh_map_idx_;
+  std::string sample_scheme_;
+  double total_strength_;
+  std::vector<double> mesh_CDF_;
+  std::vector<double> mesh_strengths_;
+  int64_t tot_bins_;
```
pshriwise, line 118: Since the parameters for mesh min are limited to `int23_t`, `tot_bins` can't exceed the max value of that type -- we can use `int32_t` for `tot_bins` as well.

openmc/settings.py
```diff
@@ -1275,6 +1281,12 @@ def _keff_trigger_from_xml_element(self, root):
     def _source_from_xml_element(self, root):
         for elem in root.findall('source'):
             self.source.append(Source.from_xml_element(elem))
+            if isinstance(Source.from_xml_element(elem), MeshIndependent):
+                mesh_id = int(get_text(elem, 'mesh'))
+                path = f"./mesh[@id='{mesh_id}']"
+                mesh_elem = root.find(path)
+                if mesh_elem is not None:
+                    self.source.space.mesh = MeshBase.from_xml_element(mesh_elem)
```
pshriwise, line 1289: Let's read in the source object and add it to the `Source.source` attribute at the very end to simplify the code a bit.
```suggestion
        for elem in root.findall('source'):
            src = Source.from_xml_element(elem)
            if isinstance(src.space, MeshIndependent):
                mesh_id = int(get_text(elem, 'mesh'))
                path = f"./mesh[@id='{mesh_id}']"
                mesh_elem = root.find(path)
                if mesh_elem is not None:
                    src.space.mesh = MeshBase.from_xml_element(mesh_elem)
            self.source.append(src)
```
This also avoids an additional read of the source element in the if-statement.

pshriwise: Let's raise a `RuntimeError` if the mesh isn't found in the `settings.xml` file as well. We'll have no way to set that mesh correctly after this point so we should make sure the source object isn't hanging around without it.

include/openmc/mesh.h
```diff
@@ -454,6 +462,8 @@ class UnstructuredMesh : public Mesh {
   virtual std::string get_mesh_type() const override;
 
   // Overridden Methods
+  // TODO Position sample(uint64_t* seed) const=0;
```
pshriwise, line 465: can this go or is there more work to do here?

openmc/stats/multivariate.py
```diff
@@ -613,6 +616,117 @@ def from_xml_element(cls, elem):
...
+    """Spatial distribution for a mesh.
+
+    This distribution allows one to specify a mesh to sample over and the scheme to sample over the entire mesh.
```
pshriwise, line 622: We should update this a bit now that we're just relying on `strengths` and `volume_normalized`.

openmc/stats/multivariate.py
```diff
@@ -613,6 +616,117 @@ def from_xml_element(cls, elem):
...
+        The mesh instance used for sampling, mesh is written into settings.xml, mesh.id is written into the source distribution
+    strengths : Iterable of Real, optional
+        A list of values which represent the weights of each element
+
+    """
+
+    def __init__(self, mesh, strengths=None, volume_normalized=False):
+        self.mesh = mesh        
+        self.strengths = strengths
+        self.volume_normalized = volume_normalized
```
pshriwise, line 646: Need a docstring for `volume_normalized`.

openmc/stats/multivariate.py
```diff
@@ -613,6 +616,117 @@ def from_xml_element(cls, elem):
...
+    Attributes
+    ----------
+    mesh : openmc.MeshBase
+        The mesh instance used for sampling, mesh is written into settings.xml, mesh.id is written into the source distribution
+    strengths : Iterable of Real, optional
+        A list of values which represent the weights of each element
+
+    """
+
+    def __init__(self, mesh, strengths=None, volume_normalized=False):
```
pshriwise, line 643: I'd say the default here would be `True` in light of #2142. What do you think @eepeterson?

openmc/stats/multivariate.py
pshriwise: Also, this is a misnomer now that I think about it. `volume_weighted` is probably more accurate. What do you think @NybergWISC?

src/distribution_spatial.cpp
```diff
@@ -180,6 +182,78 @@ Position SphericalIndependent::sample(uint64_t* seed) const
...
+  mesh_CDF_.resize(tot_bins_+1);
+  mesh_CDF_[0] = {0.0};
+  total_strength_ = 0.0;
+  mesh_strengths_.resize(tot_bins_);
+
+  // Create cdfs for sampling for an element over a mesh
+  // Volume scheme is weighted by the volume of each tet
+  // File scheme is weighted by an array given in the xml file
+
+  if (check_for_node(node, "strengths")) {
```
pshriwise, line 216: I think this section of code can be refactored and simplified a bit:

```cpp
mesh_strengths_ = std::vector<double>(umesh_ptr->n_bins(), 1.0);

if (check_for_node(node, "strengths") {
    strengths = get_node_array<double>(node, "strengths");
    // size check for 'strengths' here
    mesh_strengths_ = strengths;
}

if (get_node_value_bool(node, "volume_normalized") {
    for (int i = 0; i < umesh_ptr_->n_bins(); i++) {
      mesh_strengths_[i] *= umesh_ptr->volume(i);
  }
}

// compute CDF here
```

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,182 @@
...
+    ### Geometry ###
+    cell = np.empty((dimen, dimen, dimen), dtype=object)
+    surfaces = np.empty((dimen + 1, 3), dtype=object)
+
+    geometry = openmc.Geometry()
+    universe = openmc.Universe(universe_id=1, name="Contains all hexes")
+
+    for i in range(0,dimen+1):
+        coord = -10.0 + i * size_hex
+        surfaces[i][0] = openmc.XPlane(coord, name="X plane at "+str(coord))
```
pshriwise, line 116:
```suggestion
        surfaces[i][0] = openmc.XPlane(coord, name=f"X plane at {coord}")

```

Same for others like this

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,182 @@
...
+        surfaces[i][1].boundary_type = 'vacuum'
+        surfaces[i][2].boundary_type = 'vacuum'
+
+    for k in range(0,dimen):
+        for j in range(0,dimen):
+            for i in range(0,dimen):
+                cell[i][j][k] = openmc.Cell(name=("x = {}, y = {}, z = {}".format(i,j,k)))
+                cell[i][j][k].region = +surfaces[i][0] & -surfaces[i+1][0] & \
+                                    +surfaces[j][1] & -surfaces[j+1][1] & \
+                                    +surfaces[k][2] & -surfaces[k+1][2]
```
pshriwise, line 130:
```suggestion
                cell[i][j][k].region = +surfaces[i][0] & -surfaces[i+1][0] & \
                                       +surfaces[j][1] & -surfaces[j+1][1] & \
                                       +surfaces[k][2] & -surfaces[k+1][2]

```

pshriwise (changes requested):
Here's another pass from me @nybergwisc. It's looking quite good!

include/openmc/mesh.h
```diff
@@ -68,6 +68,12 @@ class Mesh {
 
   // Methods
 
+  //! Sample a mesh volume using a certain seed
+  //
+  //! \param[in] seed Seed to use for random sampling
+  //! \param[out] r Position within tet
+  virtual Position sample(uint64_t* seed, int32_t tet_bin) const=0;
```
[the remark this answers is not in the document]
NybergWISC: Yes, will do. Must have forgotten to change it and resolved the comment once I addressed the second part of it

include/openmc/distribution_spatial.h
```diff
@@ -94,6 +95,29 @@ class SphericalIndependent : public SpatialDistribution {
...
+  //! \return Sampled position
+  Position sample(uint64_t* seed) const;
+
+private:
+  UnstructuredMesh* umesh_ptr_;
+  int32_t mesh_map_idx_;
+  std::string sample_scheme_;
+  double total_strength_;
+  std::vector<double> mesh_CDF_;
+  std::vector<double> mesh_strengths_;
```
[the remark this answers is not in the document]
NybergWISC: Yes, this is the variable that used to be mesh_weights_. Is used to store the strengths coming out of the xml node

include/openmc/mesh.h
NybergWISC: Good catch, will delete and check for other TODOs

openmc/stats/multivariate.py
NybergWISC: I think we were initially trying to avoid using "weights" because that is normally related to the particle instead of the source distributions. I think normalized works in this case, shows whether the volume of the tet is taken into account

include/openmc/distribution_spatial.h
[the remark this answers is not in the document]
pshriwise: I'm not seeing it used outside of the `MeshSpatial` constructor, so I don't think we need to store it on the class. We just need the CDF when sampling test. Unless I'm missing something...

src/mesh.cpp
pshriwise: It's true, the number of tet vertices is hard-coded throughout `mesh.cpp`. It would be good to add a constant in `mesh.h` to improve clarity. I can handle that in a follow-on PR if you like @NybergWISC.

tests/regression_tests/unstructured_mesh/source_sampling/test.py
```diff
@@ -0,0 +1,181 @@
...
+
+        openmc.run(**kwargs)
+
+    def _compare_results(self):
+        # There are 10000 particles and 1000 hexes, this leads to the average
+        # shown below
+        average_in_hex = 10.0
+
+        # Load in tracks
+        openmc.Tracks.combine(glob.glob('tracks_p*.h5'))
```
pshriwise, line 38:
```suggestion
        if config['mpi']:
            openmc.Tracks.combine(glob.glob('tracks_p*.h5'))

```

openmc/stats/multivariate.py
NybergWISC: I ended up changing the default value to True in the latest commit. I think it may be better to have the default be False so that first time users are not confused about why the relative strengths are different than the values passed in if they do not use/know about the volume normalized option

openmc/stats/multivariate.py
```diff
@@ -613,6 +616,117 @@ def from_xml_element(cls, elem):
...
+    def mesh(self):
+        return self._mesh
+
+    @mesh.setter
+    def mesh(self, mesh):
+        if mesh != None:
+            cv.check_type('Unstructured Mesh instance', mesh, MeshBase)
+        self._mesh = mesh
+
+    @property
```
eepeterson, line 658: part of me wonders if it is more straightforward to drop the `volume_normalized` stuff in favor of something like `mesh_spatial_obj.strengths = strengths*mesh.volumes` or something. This forces the user to know what they are doing when setting the `strengths` property.

openmc/stats/multivariate.py
```diff
@@ -613,6 +616,117 @@ def from_xml_element(cls, elem):
...
+    Attributes
+    ----------
+    mesh : openmc.MeshBase
+        The mesh instance used for sampling, mesh is written into settings.xml, mesh.id is written into the source distribution
+    strengths : Iterable of Real, optional
+        A list of values which represent the weights of each element
+
+    """
+
+    def __init__(self, mesh, strengths=None, volume_normalized=True):
```
eepeterson, line 643: If we dropped `volume_normalized` you could set the strengths to be the volumes by default if `strengths is None` which makes sense to uniformly sample by default.

openmc/stats/multivariate.py
pshriwise: That would be nice, but for unstructured meshes we don't know the element volumes until runtime. :-/

We could extract this info with the CAPI, but I think that would unecessarily complicate any analysis workflows using this capability.

src/distribution_spatial.cpp
```diff
@@ -180,6 +182,70 @@ Position SphericalIndependent::sample(uint64_t* seed) const
...
+  }
+}
+
+Position MeshSpatial::sample(uint64_t* seed) const
+{ 
+  // Create random variable for sampling element from mesh
+  float eta = prn(seed);
+  // Sample over the CDF defined in initialization above
+  int32_t tet_bin = lower_bound_index(mesh_CDF_.begin(), mesh_CDF_.end(), eta);
+  return mesh_ptr_->sample(seed, tet_bin);
```
eepeterson, line 245:
```suggestion
  int32_t elem_idx = lower_bound_index(mesh_CDF_.begin(), mesh_CDF_.end(), eta);
  return mesh_ptr_->sample(seed, elem_idx);
```
Just making names more general for other non-tet `MeshSpatial` types

openmc/stats/multivariate.py
eepeterson: mmm pesky. Is it worth adding to `UnstructuredMesh` in the Python API? It may help for metadata manipulation later on to have more functionality on the python side generally speaking.

openmc/stats/multivariate.py
NybergWISC: Patrick and I had a long discussion on this either on this PR or one on my fork. I think this way allows the most functionality since if you ignore volume_normalized you get the expected behavior for both passed in strengths and uniform sampling, but it allows for 4 modes of operation instead of two. Someone technically could want un-normalized passed in variables, or normalized, or uniform sampling.

eepeterson: Ok I buy it, thanks @NybergWISC. I'm still a little hung up on the word `normalized` in the `volume_normalized` kwarg but as long as the docstring clearly explains its use (that it multiplies the strengths by element volumes) I'm fine with it.

openmc/stats/multivariate.py
eepeterson: I'd agree with @pshriwise that the default should be `True`. I think it makes sense for the default UM sampling to be uniform by volume not uniform by index.

openmc/stats/multivariate.py
NybergWISC: I also agree there. I originally used "volume_weighting", but we wanted to avoid using weights as it could be confused for VR rather than sampling

eepeterson: One alternate possibility is `scale_by_volumes=True` but this is relatively small potatoes.

openmc/stats/multivariate.py
pshriwise: We _can_ query element volumes from the `UnstructuredMesh` class, but it has to be loaded from a statepoint file. This is similar to tallies in that they don't have any data unless they're loaded from a statepoint. 

Being able to query element connectivity, vertex coordinates, etc. before running a simulation would be challenging and require us to support some interface to NetCDF files -- a nontrivial lift I'm afraid 😞

paulromano:
Closing this in favor of #2334, which is close to being merged

PR #2159, opened 2022-08-10
minor typo fix

shimwell:
Just adding a space between words in an error message that I happened to notice


pshriwise (approved):
Thanks @shimwell!

PR #2161, opened 2022-08-10
Add extern function to delete a tally by its index in the tallies vector

### LINKED ISSUE #2155: No extern function to delete a tally from a model in tally.cpp
During a simulation with adaptive mesh refinement, as well as in some other cases (e.g. FETs), it would be useful to delete a tally that will no longer be needed. For the adaptivity application, if a mesh changes, the tally need to be reinitialized with new bins that account for additions to and deletions from the mesh.  To accomplish this, an `extern` function will be added to `openmc/src/tallies/tally.cpp`

I will look into how to do this and create a new PR. @aprilnovak @pshriwise let me know if I missed anything important to the description of this issue.

lewisgross1296:
Closes #2155. This PR adds a new function to `openmc/src/tallies/tally.cpp` with the signature `extern "C" void openmc_remove_tally_from_tallies(int32_t index)` that can delete a tally from the tally vector via its index. This will be useful in adaptive mesh refinement problems where a tally will need to be deleted and re-added to the problem every time the mesh, and thus the bins, change. Additionally, this function is added to the C++ API so it can be called, and it is tested in via a unit test in `openmc/tests/unit_tests/test_tallies.py`.

EDIT: After discussion in slack, it is likely possible to do AMR without this feature. Still, this is something worth adding to the C++ API. See #2162 for a future, low priority PR that cleans up the objects after a delete operation.

### DIFF
--- a/include/openmc/capi.h
+++ b/include/openmc/capi.h
@@ -118,6 +118,7 @@ int openmc_regular_mesh_get_params(
 int openmc_regular_mesh_set_dimension(int32_t index, int n, const int* dims);
 int openmc_regular_mesh_set_params(int32_t index, int n, const double* ll,
   const double* ur, const double* width);
+int openmc_remove_tally(int32_t id);
 int openmc_reset();
 int openmc_reset_timers();
 int openmc_run();
--- a/openmc/lib/tally.py
+++ b/openmc/lib/tally.py
@@ -87,6 +87,9 @@ _dll.openmc_tally_set_type.errcheck = _error_handler
 _dll.openmc_tally_set_writable.argtypes = [c_int32, c_bool]
 _dll.openmc_tally_set_writable.restype = c_int
 _dll.openmc_tally_set_writable.errcheck = _error_handler
+_dll.openmc_remove_tally_from_tallies.argtypes = [c_int32]
+_dll.openmc_remove_tally_from_tallies.restype = None
+_dll.openmc_remove_tally_from_tallies.errcheck = _error_handler
 _dll.tallies_size.restype = c_size_t
 
 
--- a/openmc/lib/tally.py
+++ b/openmc/lib/tally.py
@@ -390,13 +390,7 @@ class Tally(_FortranObjectWithID):
 
 class _TallyMapping(Mapping):
     def __getitem__(self, key):
-        index = c_int32()
-        try:
-            _dll.openmc_get_tally_index(key, index)
-        except (AllocationError, InvalidIDError) as e:
-            # __contains__ expects a KeyError to work correctly
-            raise KeyError(str(e))
-        return Tally(index=index.value)
+        return Tally(index=self._get_tally_index(key))
 
     def __iter__(self):
         for i in range(len(self)):
@@ -407,5 +401,19 @@ class _TallyMapping(Mapping):
 
     def __repr__(self):
         return repr(dict(self))
+
+    def __delitem__(self,key):
+        """Delete a tally from tally vector and remove the ID,index pair from tally"""
+        _dll.openmc_remove_tally(self._get_tally_index(key))
+
+    def _get_tally_index(self,key):
+        """Given the ID, return the index"""
+        index = c_int32()
+        try:
+            _dll.openmc_get_tally_index(key, index)
+        except (AllocationError, InvalidIDError) as e:
+            # __contains__ expects a KeyError to work correctly
+            raise KeyError(str(e))
+        return index.value
 
 tallies = _TallyMapping()
--- a/openmc/lib/tally.py
+++ b/openmc/lib/tally.py
@@ -87,9 +87,9 @@ _dll.openmc_tally_set_type.errcheck = _error_handler
 _dll.openmc_tally_set_writable.argtypes = [c_int32, c_bool]
 _dll.openmc_tally_set_writable.restype = c_int
 _dll.openmc_tally_set_writable.errcheck = _error_handler
-_dll.openmc_remove_tally_from_tallies.argtypes = [c_int32]
-_dll.openmc_remove_tally_from_tallies.restype = c_int
-_dll.openmc_remove_tally_from_tallies.errcheck = _error_handler
+_dll.openmc_remove_tally.argtypes = [c_int32]
+_dll.openmc_remove_tally.restype = c_int
+_dll.openmc_remove_tally.errcheck = _error_handler
 _dll.tallies_size.restype = c_size_t
 
 
--- a/openmc/lib/tally.py
+++ b/openmc/lib/tally.py
@@ -406,7 +406,7 @@ class _TallyMapping(Mapping):
         """Delete a tally from tally vector and remove the ID,index pair from tally"""
         _dll.openmc_remove_tally(self._get_tally_index(key))
 
-    def _get_tally_index(self,key):
+    def _get_tally_index(self, key):
         """Given the ID, return the index"""
         index = c_int32()
         try:
--- a/openmc/lib/tally.py
+++ b/openmc/lib/tally.py
@@ -88,7 +88,7 @@ _dll.openmc_tally_set_writable.argtypes = [c_int32, c_bool]
 _dll.openmc_tally_set_writable.restype = c_int
 _dll.openmc_tally_set_writable.errcheck = _error_handler
 _dll.openmc_remove_tally_from_tallies.argtypes = [c_int32]
-_dll.openmc_remove_tally_from_tallies.restype = None
+_dll.openmc_remove_tally_from_tallies.restype = c_int
 _dll.openmc_remove_tally_from_tallies.errcheck = _error_handler
 _dll.tallies_size.restype = c_size_t
 
--- a/openmc/lib/tally.py
+++ b/openmc/lib/tally.py
@@ -390,7 +390,7 @@ class Tally(_FortranObjectWithID):
 
 class _TallyMapping(Mapping):
     def __getitem__(self, key):
-        return Tally(index=self._get_tally_index(key))
+        return Tally(index=self[key])
 
     def __iter__(self):
         for i in range(len(self)):
@@ -404,16 +404,6 @@ class _TallyMapping(Mapping):
 
     def __delitem__(self, key):
         """Delete a tally from tally vector and remove the ID,index pair from tally"""
-        _dll.openmc_remove_tally(self._get_tally_index(key))
-
-    def _get_tally_index(self, key):
-        """Given the ID, return the index"""
-        index = c_int32()
-        try:
-            _dll.openmc_get_tally_index(key, index)
-        except (AllocationError, InvalidIDError) as e:
-            # __contains__ expects a KeyError to work correctly
-            raise KeyError(str(e))
-        return index.value
+        _dll.openmc_remove_tally(self[key]._index)
 
 tallies = _TallyMapping()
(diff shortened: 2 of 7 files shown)


tests/unit_tests/test_tallies.py
```diff
@@ -39,3 +39,6 @@ def test_xml_roundtrip(run_in_tmpdir):
     assert new_tally.triggers[0].trigger_type == tally.triggers[0].trigger_type
     assert new_tally.triggers[0].threshold == tally.triggers[0].threshold
     assert new_tally.triggers[0].scores == tally.triggers[0].scores
+    # delete tally and check that new_tallies now has length zero
+    new_tally.openmc_remove_tally_from_tallies(tally.id)
```
pshriwise, line 43: This test performs checks on the pure Python API as opposed to the `openmc.lib` module, which is the Python interface to the CAPI. This check looks good to me, but we'll want to add something similar in `tests/unit_tests/test_lib.py`.

pshriwise (changes requested):
Thanks for this @lewisgross1296! Should be useful for those AMR simulations. A few notes from me.

tests/unit_tests/test_tallies.py
lewisgross1296: Would it make sense to just cut these lines from `test_tallies.py` and do it instead in `test_lib.py` with a similar setup?

src/tallies/tally.cpp
```diff
@@ -1271,4 +1273,13 @@ extern "C" size_t tallies_size()
   return model::tallies.size();
 }
 
+// given a tally ID, remove it from the tallies vector
+extern "C" void openmc_remove_tally_from_tallies(int32_t id)
+{
+  // query map for index corersponding to the given id
+  int32_t index = model::tally_map[id];
```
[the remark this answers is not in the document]
lewisgross1296: Went with your suggested changes and added testing that the ID exists, as well as deleting the tally ID by key after successfully removing it from the tally vector.

tests/unit_tests/test_lib.py
```diff
@@ -339,6 +339,20 @@ def test_new_tally(lib_init):
...
+    new_tally = openmc.lib.Tally()
+    new_tally.scores = ['flux']
+    new_tally_with_id = openmc.lib.Tally(10)
+    new_tally_with_id.scores = ['flux']
+    assert len(openmc.lib.tallies) == 5
+    # delete tally and check length is one less than before
+    openmc_remove_tally(10)
+    assert len(openmc.lib.tallies) == 4
+
+# def test_delete_wrong_id():
```
lewisgross1296, line 354: @pshriwise I think i should add a test to delete a wrong ID here, but I'm not sure exactly what needs to go here. Not sure if I should just use the `lib_init` and then try and delete something like `delete(9876243184073)` or if I should do something from scratch so I'm exactly sure which tally IDs exist in my setup. Do you have thoughts?

pshriwise: I'd recommend gathering all of the tally ID's from the `lib_init` model and then 
try to delete the `max(IDs) + 1`.

lewisgross1296: That is a good idea! I try that out

src/tallies/tally.cpp
```diff
@@ -1276,7 +1276,7 @@ extern "C" size_t tallies_size()
 extern "C" int openmc_remove_tally(int32_t id)
 {
   // check that id is in the map
-  if (!model::tally_map.contains(id)) {
+  if (model::tally_map.count(id)!=1) {
```
lewisgross1296, line 1279: During the test, I see that 

> -- CXX_STANDARD: 11

which caused this error, as I believe C++20 introduced contains 

> /home/runner/work/openmc/openmc/src/tallies/tally.cpp:1279:25: error: 'class std::unordered_map<int, int>' has no member named 'contains'
>  1279 |   if (!model::tally_map.contains(id)) {

stack overflow recommended using count, which for a map, returns 1 if the element exists and 0 otherwise, as maps must have unique keys. Is this okay syntax to check whether it exists? It compiled for me with contains, but I think I need to use count for the test to pass

pshriwise: Ah, yeah someday we'll get methods that are more descriptive for these containers...

You could also use `(model::tally_map.find(id) == model::tally_map.end())`. But I think using `count` is fine. One small tweak would be to use `== 0` for comparison.

include/openmc/capi.h
```diff
@@ -118,6 +118,7 @@ int openmc_regular_mesh_get_params(
 int openmc_regular_mesh_set_dimension(int32_t index, int n, const int* dims);
 int openmc_regular_mesh_set_params(int32_t index, int n, const double* ll,
   const double* ur, const double* width);
+int openmc_remove_tally(int32_t id);
```
paulromano, line 121: All of our C API functions related to tallies pretty much uniformly take the _index_ as an argument rather than the id, so I would propose we change this to take the index as well. If you only know the id, you can use `openmc_tally_get_id` to determine the index.

openmc/lib/tally.py
```diff
@@ -87,6 +87,9 @@
 _dll.openmc_tally_set_writable.argtypes = [c_int32, c_bool]
 _dll.openmc_tally_set_writable.restype = c_int
 _dll.openmc_tally_set_writable.errcheck = _error_handler
+_dll.openmc_remove_tally_from_tallies.argtypes = [c_int32]
+_dll.openmc_remove_tally_from_tallies.restype = c_int
+_dll.openmc_remove_tally_from_tallies.errcheck = _error_handler
```
paulromano, line 92:
```suggestion
_dll.openmc_remove_tally.argtypes = [c_int32]
_dll.openmc_remove_tally.restype = c_int
_dll.openmc_remove_tally.errcheck = _error_handler
```

paulromano (changes requested):

include/openmc/capi.h
lewisgross1296: Sounds good! I will fix this

src/tallies/tally.cpp
lewisgross1296: I think I'd prefer the `== 0` so I will switch it to that

include/openmc/capi.h
lewisgross1296: For this though, I currently have 
```
  // check that id is in the map
  if (model::tally_map.count(id)!=1) {
    return OPENMC_E_INVALID_ID;
  }
```
If I am given an index instead, does this mean I'll just compare the `index` to the size of the tallies vector? If the `index` is greater than the size, the tally does not exist. Should I still return `OPENMC_E_INVALID_ID` (since technically it is an invalid index) or is there a more appropriate return type?

paulromano: Take a look at any of the other `openmc_tally_*` functions in tally.cpp as they all have consistent logic for treating this

pshriwise: > If I am given an index instead, does this mean I'll just compare the index to the size of the tallies vector?

Yeah I would do a similar check as other functions that take an index into global vectors. Something like:

`if (index >= 0 && index < model::tallies.size()) { ... }`

>  Should I still return OPENMC_E_INVALID_ID (since technically it is an invalid index) or is there a more appropriate return type?

I'd say that returning `OPENMC_E_OUT_OF_BOUNDS` is the way to go here.

lewisgross1296:
All tests are passing. Let me know if there are any other desired changes or tests to add. @pshriwise @paulromano

src/tallies/tally.cpp
```diff
@@ -1271,4 +1272,20 @@ extern "C" size_t tallies_size()
...
+// given a tally ID, remove it from the tallies vector
+extern "C" int openmc_remove_tally(int32_t index)
+{
+  // check that id is in the map
+  if (index < 0 || index > model::tallies.size()) {
+    return OPENMC_E_OUT_OF_BOUNDS;
+  }
+  // grab tally so it's ID can be obtained to remove the (ID,index) pair from tally_map
+  auto& tally = model::tallies[index];
+  model::tally_map.erase(tally->id_);
```
paulromano, line 1284: I believe this should already happen in the `Tally` destructor during the call to `tallies.erase` below

include/openmc/capi.h
paulromano, line 121:
```suggestion
int openmc_remove_tally(int32_t index);
```

openmc/lib/tally.py
```diff
@@ -405,4 +402,18 @@ def __len__(self):
     def __repr__(self):
         return repr(dict(self))
 
+    def __delitem__(self,key):
```
paulromano, line 405:
```suggestion
    def __delitem__(self, key):
```

openmc/lib/tally.py
```diff
@@ -405,4 +402,18 @@ def __len__(self):
...
+
+    def _get_tally_index(self,key):
+        """Given the ID, return the index"""
+        index = c_int32()
+        try:
+            _dll.openmc_get_tally_index(key, index)
+        except (AllocationError, InvalidIDError) as e:
+            # __contains__ expects a KeyError to work correctly
+            raise KeyError(str(e))
+        return index.value
```
paulromano, line 417: I think this should work:
```suggestion
        _dll.openmc_remove_tally(self[key]._index)
```

paulromano (changes requested):

src/tallies/tally.cpp
lewisgross1296: Nice catch, thanks!

openmc/lib/tally.py
pshriwise: Ah, good call. Nice and elegant. I suggested the refactor here, but this is better.

lewisgross1296: My Class knowledge in python is still in development. I can guess that `self[key]._index` retrieves the proper index for this function, but how does this work exactly? Also, I think because I pushed another change in this file, this is now outdated, and I can't "Commit suggestoin," for the co-author, but I will credit you in the commit.

lewisgross1296: Just want to confirm, but if I delete `_get_tally_index` then I also need to modify `__getitem__` to be
```
    def __getitem__(self, key):
        return Tally(index=self[key])
```

lewisgross1296: Only thing I'm concerned about is that we've now lost the `KeyError` checking

paulromano: `self[key]` is the equivalent of `self.__getitem__(key)`. The `__getitem__` method already has the logic for checking the key and will raise a `KeyError` is you pass an invalid key

lewisgross1296: Good to know then! I will add this in and then I think it will be ready for merge, assuming everything passes and there's no new corrections.

openmc/lib/tally.py
```diff
@@ -387,13 +390,7 @@ def ci_width(self, alpha=0.05):
...
 class _TallyMapping(Mapping):
     def __getitem__(self, key):
-        index = c_int32()
-        try:
-            _dll.openmc_get_tally_index(key, index)
-        except (AllocationError, InvalidIDError) as e:
-            # __contains__ expects a KeyError to work correctly
-            raise KeyError(str(e))
-        return Tally(index=index.value)
+        return Tally(index=self[key])
```
pshriwise, line 393:
```suggestion
        index = c_int32()
        try:
            _dll.openmc_get_tally_index(key, index)
        except (AllocationError, InvalidIDError) as e:
            # __contains__ expects a KeyError to work correctly
            raise KeyError(str(e))
```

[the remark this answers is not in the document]
lewisgross1296: Did not even realize this! Thanks for figuring it out

pshriwise (approved):
Thanks for this addition @lewisgross1296!

paulromano (approved):

PR #2187, opened 2022-08-23
Adding pulse schedule plot

shimwell:
I've just been using the nice deplete features and had the need to plot the source strength and timesteps used in the simulation for different pulse scenarios (e.g. maintenance, waste, high duty).

I wrote a little method that takes the timesteps linearizes them and plots them with the source strength in a subplot so one can see check the pulse schedule and timesteps are as expected.

I thought it might be handy for others (perhaps @eepeterson) so made this PR adds a ```.plot()``` method to the integrator class.

Here is an example plot of a 100% pulse for 20 seconds with a timestep in the middle followed by a cooling time and then a smaller ramped pulse with a short cooling time that has timestep readings with it.

shimwell_integrator_plot

Code to reporduce

```python
import openmc
import openmc.deplete

material = openmc.Material()
material.add_nuclide('Fe59', 1)
material.volume = 100
material.depletable = True
materials = openmc.Materials([material])

model = openmc.Model()
model.materials = materials

operator = openmc.deplete.Operator(
        model=model,
        chain_file="chain-nndc-b7.1.xml",
        normalization_mode="source-rate"
    )

integrator = openmc.deplete.PredictorIntegrator(
    operator=operator,
    timesteps=[10, 10, 40, 10, 10, 4,6],
    source_rates=[100, 100, 0, 50, 80, 0,0],
    timestep_units='d',
)

integrator.plot()
plt.show()
```

### DIFF
--- a/openmc/deplete/abc.py
+++ b/openmc/deplete/abc.py
@@ -12,9 +12,7 @@ from numbers import Real, Integral
 from contextlib import contextmanager
 import os
 from pathlib import Path
-import sys
 import time
-from turtle import color
 from warnings import warn
 
 from numpy import nonzero, empty, asarray
--- a/openmc/deplete/abc.py
+++ b/openmc/deplete/abc.py
@@ -35,6 +35,7 @@ __all__ = [
 _SECONDS_PER_MINUTE = 60
 _SECONDS_PER_HOUR = 60*60
 _SECONDS_PER_DAY = 24*60*60
+_SECONDS_PER_JULIAN_YEAR = 365.25*24*60*60
 
 OperatorResult = namedtuple('OperatorResult', ['k', 'rates'])
 OperatorResult.__doc__ = """\
@@ -530,11 +531,11 @@ class Integrator(ABC):
         each interval in :attr:`timesteps`
 
         .. versionadded:: 0.12.1
-    timestep_units : {'s', 'min', 'h', 'd', 'MWd/kg'}
+    timestep_units : {'s', 'min', 'h', 'd', 'a', 'MWd/kg'}
         Units for values specified in the `timesteps` argument. 's' means
-        seconds, 'min' means minutes, 'h' means hours, and 'MWd/kg' indicates
-        that the values are given in burnup (MW-d of energy deposited per
-        kilogram of initial heavy metal).
+        seconds, 'min' means minutes, 'h' means hours, 'a' means Julian years
+        and 'MWd/kg' indicates that the values are given in burnup (MW-d of
+        energy deposited per kilogram of initial heavy metal).
     solver : str or callable, optional
         If a string, must be the name of the solver responsible for
         solving the Bateman equations.  Current options are:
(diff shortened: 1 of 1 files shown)


openmc/deplete/abc.py
```diff
@@ -833,6 +833,46 @@ def integrate(self, final_step=True, output=True):
...
+        axes[0].stairs(self.source_rates, linear_time_steps, linewidth=2)
+
+        for timestep in linear_time_steps:
+            x_vals = [timestep, timestep]
+            y_vals = [0, 1]  # arbitrary heights selected as axis has no y scale
+            axes[1].plot(x_vals, y_vals, '-', color='red')
+
+        axes[1].set_xlabel(f'Timesteps [{self.timestep_units}]')
+        axes[1].set_ylim(0, 1)
+        axes[1].set_xticks(linear_time_steps)
```
shimwell, line 872:
```suggestion
```

Sometimes the numbers can overlap so perhaps best to remove this custom x label

This plot for example would have overlapping x values, if this line is left in
 waste-compressed

shimwell: done in commit #e0dcf17

openmc/deplete/abc.py
```diff
@@ -592,6 +592,7 @@ def __init__(self, operator, timesteps, power=None, power_density=None,
                         self._num_stages))
         self.operator = operator
         self.chain = operator.chain
+        self.timestep_units = timestep_units
```
paulromano, line 595: This approach won't always work because in some cases the `timestep_units` argument is ignored, e.g.:
```Python
Integrator(timesteps=[(3., 'd'), (1000, 's'), (15, 'h')], ...)
```
What I would suggest is to just have `timestep_units` be an argument on the `plot` method.

paulromano (commented):
After thinking about this one a bit, I'm leaning toward not accepting. Most of the body of the new method is just customizing a plot, and that is usually something a user wants control over.

openmc/deplete/abc.py
shimwell: I have now add ```timestep_units``` as an argument on the plot method.

shimwell:
> After thinking about this one a bit, I'm leaning toward not accepting. Most of the body of the new method is just customizing a plot, and that is usually something a user wants control over.

No worries, lets leave it open for a few days just in case anyone else is interested then if no one is desperate for such a feature I shall close the PR.

Thanks for reviewing

shimwell:
Some of the code (Julian years support in Integrator) in this PR has been separated out and put into PR #2195

shimwell:
Just in case anyone is reading this and would like to make such plots in the future then I've added these pulse schedule plots to the ```openmc_depletion_plotter``` package. One can make the plots by:


```
pip install openmc_depletion_plotter
```

```python
from openmc_depletion_plotter import IntegratorWithPlotting  # class extends openmc depletion classes

# arbitrary lists provided here, but this would be defined by the the pulse schedule
source_rates = [1e21] + [0] * 24  # single pulse followed by 0 neutrons
time_steps = [1] + [3600] * 24   # one second followed by 1 hour periods for 24 hours

integrator = IntegratorWithPlotting(    # just inherits and extends the openmc.deple.Integrator class
    operator=operator,
    timesteps=time_steps,
    source_rates=source_rates,
    timestep_units='d',
)
integrator.plot()
plt.savefig('pulse_single_shot.png')
```

I'm now closing this PR as part of the code has moved to #2195 and the rest is in ```openmc_depletion_plotter```. Naturally I'm happy to reopen and merge in this plotting capability in the future if it becomes more appropriate to add such niche plotting capabilities.

PR #2144, opened 2022-08-02
Update python setup

shimwell:
Work in progress, not ready for review

I just wanted to show that I've been working on migrating from the now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) setup.py to a shiny new setup.cfg. :snake: 

The new setup appears to work but will need a lot more testing before I'm confident to move this from a draft pr to a full pr. Still I think this is decent progress and worth showing to anyone interested

I was not able to move over [these parts](https://github.com/openmc-dev/openmc/blob/8a6dc7e9f8075b92038a20b1e77a2569ac71a2cc/setup.py#L82-L86) the setup.cfg but it is close to complete

PR #2191, opened 2022-08-24
added main method to scripts

shimwell:
When experimenting with moving to setup.cfg on PR #2144 I noticed that a few scripts were missing  ```if __name__ == "__main__":``` functions. These main methods are super handy to allow function importing, packaging using setup.cfg and other benefits.

This PR adds a ```if __name__ == "__main__":``` to three scripts

- openmc-ace-to-hdf5
- openmc-voxel-to-vtk
- openmc-validate-xml

I also made a few other improvements while in the area

adds ``````if __name__ == "__main__":``` functions :heavy_check_mark: 
removed unused imports :heavy_check_mark: 
added some type hints to args :heavy_check_mark:
black formatting applied :heavy_check_mark:
added script description to ArgumentParser :heavy_check_mark:
changed format to f strings :heavy_check_mark:

Overall I think this is a reasonable improvement, the scripts are still not perfect but perhaps better than they were

### DIFF
--- a/scripts/openmc-ace-to-hdf5
+++ b/scripts/openmc-ace-to-hdf5
@@ -28,102 +28,126 @@ import openmc.data
 from openmc.data.ace import TableType
 
 
-class CustomFormatter(argparse.ArgumentDefaultsHelpFormatter,
-                      argparse.RawDescriptionHelpFormatter):
-    pass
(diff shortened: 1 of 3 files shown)


paulromano (approved):
Thanks for improving these scripts @shimwell. All good to go once tests pass

PR #2193, opened 2022-08-24
Switch math.log10 to numpy.log10 so test_lethargy_bin_width() passes

lewisgross1296:
During local testing, `test_filters::test_lethargy_bin_width()` failed due to a disagreement about the last decimal place in an assert statement. It was suggested by @shimwell to switch the test's use of `math.log10` to `numpy.log10`. Locally, for me, this resolved the test to now pass, so I thought it may be a good idea to change this for everyone. See #2190 for full discussion.

This pass or fail probably depends on `numpy`/`python` version. I am using `python 3.10.5` and `numpy 1.23.0`. If anyone has hesitation about this switch, please feel free voice your concern. There may be good reasons to leave this as `math.log10 ` that I am unaware of.

### DIFF
--- a/tests/unit_tests/test_filters.py
+++ b/tests/unit_tests/test_filters.py
@@ -1,4 +1,3 @@
-import math
 import numpy as np
 import openmc
 from pytest import fixture, approx
--- a/tests/unit_tests/test_filters.py
+++ b/tests/unit_tests/test_filters.py
@@ -252,6 +252,6 @@ def test_energy():
 def test_lethargy_bin_width():
     f = openmc.EnergyFilter.from_group_structure('VITAMIN-J-175')
     assert len(f.lethargy_bin_width) == 175
-    energy_bins = openmc.mgxs.GROUP_STRUCTURES['VITAMIN-J-175'] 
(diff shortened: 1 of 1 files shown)


shimwell (approved):
Thanks for fixing this, looks easy enough to review.

The changes look fine to me. We should wait to see if the are any objections otherwise we can merge after the standard review time

shimwell:
@lewisgross1296 I just noticed that we have an ```import math``` at the top of the file that we probably don't need anymore, would it be possible to remove this line as well

lewisgross1296:
@shimwell of course! I'll remove it

lewisgross1296:
Just making a note about advice from #2190. It is generally better to use `math.log10(arg)` when the argument is a scalar. Though the right hand side is a scalar in this test, `openmc/openmc/filter.py` which the LHS is returned from uses `numpy.log10` on an array and then accesses the first and last element of that array for the test. Thus, it would be better to use `numpy.log10` in this case, even though the RHS is a scalar. There will be some extra checking in numpy, but both sides will agree to all decimal places.

shimwell:
Thanks for these  improvements

PR #2199, opened 2022-08-26
added is openmc.Material.is_natural_abundance() method and tests

shimwell:
This PR is an attempt to add ```openmc.Material.is_natural_abundance()``` which is returns a boolean to indicate if the material is made from natural abundance of isotopes or not. The new method also has an argument for filtering by elements if the user is interested in whether a specific element in the material is natural or not.

This method will be useful when squashing materials down to their elemental definition as it will allow us to check if the material is natural or not.

Tests have also been added

Requires PR ~#2197~ #2200 to be merged first

### DIFF
--- a/openmc/material.py
+++ b/openmc/material.py
@@ -5,7 +5,7 @@ from numbers import Real
 from pathlib import Path
 import os
 import re
-import typing  # imported separately as py3.8 requires typing.Iterable
+import typing  # imported separately as py3.8 requires typing.Iterableo
 import warnings
 from typing import Optional, Union
 from xml.etree import ElementTree as ET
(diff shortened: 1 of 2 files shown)


shimwell:
As #2200 got merged I've updated this PR branch to contain the latest development additions and move this out of "on hold"

paulromano (commented):
Thanks for proposing this @shimwell. Two comments on this:
1. The approach here looks like it will be very sensitive to floating point issues and might return false negatives.
2. If we refactor the `Material` class to store element densities directly (deferring element expansion to the C++ layer), it seems like this method wouldn't be necessary.

shimwell:
Yes adding the elements to the materials class (that is quite a big refactor) would remove the need for this and sounds like a much nicer way to do the expansion. Thanks for taking a look and pointing out the correct path

Closing this now, partly to clear out the excessive nuber of PRs I've got open.

PR #2224, opened 2022-09-21
helpful error message for excessive energy spectrum rejection rate

gridley:
This has OpenMC include rejection of sampling from the energy spectrum as counting towards the rejection rate counter. It's possible that a user can define a source energy spectrum that falls WAY outside the allowable energy ranges for a problem, and when this happens, OpenMC can become incredible slow for no apparent reason.

Created in [response to this](https://openmc.discourse.group/t/is-multigroup-mode-prohibitively-slow/2297).


paulromano (approved):
Good idea @gridley; thanks!

PR #2226, opened 2022-09-22
Fix elastic distribution

### LINKED ISSUE #2225: Elastic distributions are not used in thermal.py
I think I came across a bug. [Here](https://github.com/openmc-dev/openmc/blob/cbd8376238a71c36a57385df4c18471d5a785d64/openmc/data/thermal.py#L733) and [here](https://github.com/openmc-dev/openmc/blob/cbd8376238a71c36a57385df4c18471d5a785d64/openmc/data/thermal.py#L736) the elastic distributions are assigned to "dist", but [here](https://github.com/openmc-dev/openmc/blob/cbd8376238a71c36a57385df4c18471d5a785d64/openmc/data/thermal.py#L745) the elastic reaction is created with "distribution".

It is an easy fix, just change "dist" for "distribution" in the above two lines.

marquezj:
Elastic distributions were created but not used.

Fixes #2225

### DIFF
--- a/openmc/data/thermal.py
+++ b/openmc/data/thermal.py
@@ -730,10 +730,10 @@ class ThermalScattering(EqualityMixin):
 
             if ace.nxs[5] == 3:
                 xs = incoherent_xs
-                dist = incoherent_dist
+                distribution = incoherent_dist
             elif ace.nxs[5] == 4:
                 xs = coherent_xs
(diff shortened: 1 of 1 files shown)


paulromano (approved):
Mea culpa.. thanks for fixing @marquezj!

PR #2243, opened 2022-10-03
Adding a python build option for openmc executable

shimwell:
Work in progress

This PR is an attempt to update the python setup to include compiling the c++ side of the code.

This makes use of [scikit-build](https://scikit-build.readthedocs.io/en/latest/index.html) to build openmc.

To install openmc python package we can do the same command as normal. Just with this PR we also get the openmc executable installed instead of just the python library
```
pip install .
```

With this PR we can also build the openmc executable using python. Build tools are automatically installed during the process. This doesn't remove the option of using CMake directly but just adds a python option
```
python -m build
```

This last command also [supports](https://scikit-build.readthedocs.io/en/latest/usage.html#command-line-options) parsing in CMake args to the command line

Once built the openmc wheel appears in the ```dist``` folder. The whl file can then be installed with ```pip install dist/*.whl``` which installs the python library and the openmc executable.

Changing the python setup could help with PyPi distribution, related issues #916  #1242)

A disadvantage of this approach is that Python 3.6 becomes more difficult to support. However py3.6 is [no longer supported](https://devguide.python.org/versions/) by Python so perhpas that is not a concern 

Anyone wanting to try this pip install can pip install from this branch with 
```
pip install git+https://github.com/shimwell/openmc.git@update_python_build
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,9 +34,6 @@ jobs:
         vectfit: [n]
 
         include:
-          - python-version: 3.6
-            omp: n
-            mpi: n
           - python-version: 3.7
             omp: n
             mpi: n
--- a/docs/source/devguide/styleguide.rst
+++ b/docs/source/devguide/styleguide.rst
@@ -142,7 +142,7 @@ Style for Python code should follow PEP8_.
 
 Docstrings for functions and methods should follow numpydoc_ style.
 
-Python code should work with Python 3.6+.
+Python code should work with Python 3.7+.
 
 Use of third-party Python packages should be limited to numpy_, scipy_,
 matplotlib_, pandas_, and h5py_. Use of other third-party packages must be
--- a/docs/source/usersguide/install.rst
+++ b/docs/source/usersguide/install.rst
@@ -511,7 +511,7 @@ to install the Python package in :ref:`"editable" mode <devguide_editable>`.
 Prerequisites
 -------------
 
-The Python API works with Python 3.6+. In addition to Python itself, the API
+The Python API works with Python 3.7+. In addition to Python itself, the API
 relies on a number of third-party packages. All prerequisites can be installed
 using Conda_ (recommended), pip_, or through the package manager in most Linux
 distributions.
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -60,6 +57,9 @@ jobs:
             python-version: 3.8
             omp: n
             mpi: y
+          - python-version: 3.9
+            omp: n
+            mpi: n
```
shimwell, line 62:
```suggestion
```

shimwell: should be a separate PR

shimwell:
I just realized that if anyone wants to try this then you can install this openmc package from the fork / branch using pip

```bash
pip install git+https://github.com/shimwel/openmc.git@update_python_build
```

setup.cfg
```diff
@@ -0,0 +1,48 @@
...
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers,
+    Intended Audience :: End Users/Desktop,
+    Intended Audience :: Science/Research,
+    License :: OSI Approved :: MIT License
+    Natural Language :: English
+    Topic :: Scientific/Engineering
+    Programming Language :: C++
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
```
shimwell, line 20:
```suggestion
```

shimwell:
I am getting an error with the latest attempt

```
 >       return float(ENDF_FLOAT_RE.sub(r'\1e\2\3', s))
E       ValueError: could not convert string to float: '+ 2 . 3+ 1'
```

I think this is because I no longer have the cythonise part in the setup

https://github.com/openmc-dev/openmc/blob/6c698e3b88999d7f9d5bbc22f3ca863fe9398112/setup.py#L81-L88

I shall look into bringing it back

shimwell:
Just to mention I've not given up on this quite yet, I've just been tinkering on another branch to keep the number of commits down. Still stuck on the same error but I've tried a few things to no avail over here https://github.com/shimwell/openmc/pull/9

PR #2254, opened 2022-10-06
assume cython is installed

shimwell:
Cython is in the pyproject.toml build requirements section

https://github.com/openmc-dev/openmc/blob/9a86c9c065e545a7ece086c34dc7dca38cfa9892/pyproject.toml#L2

Therefore I think it is safe to assume that cython is installed when building openmc with ```pip install .```

This allows us to remove some cython checking from the setup.py and simplify the setup.py a little bit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,11 +5,7 @@ import sys
 import numpy as np
 
 from setuptools import setup, find_packages
-try:
-    from Cython.Build import cythonize
-    have_cython = True
-except ImportError:
-    have_cython = False
+from Cython.Build import cythonize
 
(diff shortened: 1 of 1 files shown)


paulromano (approved):

shimwell:
@paulromano I wasn't sure how far to go with this PR

I was wondering if other parts of the code could perhaps be removed

https://github.com/openmc-dev/openmc/blob/47ff27e77ff4f9999dfbd7af5b4cc3afdb03a239/openmc/data/endf.py#L17-L21

PR #2256, opened 2022-10-07
Centre for cylinder & spherical meshes

### LINKED ISSUE #2248: Cylinder meshes to allow non central location
As mentioned by @paulromano in PR #2233 the ```CylindricalMesh``` is central around a location.

We could allow for an x,y,z offset that defaults to (0,0,0).

Also tagging @RemDelaporteMathurin

RemDelaporteMathurin:
This fixes #2248 by adding a centre attribute to `CylindricalMesh` and `SphericalMesh`

**Usage**
```python
mesh = openmc.CylindricalMesh()
mesh.phi_grid = np.linspace(0, 2*np.pi, 21)
mesh.z_grid = np.linspace(-1, 1, 11)
mesh.r_grid = np.linspace(0, 1, 11)

mesh.origin = [0.1, 0.2, 0.3]
```


RemDelaporteMathurin:
@pshriwise I need to declare `centre` in mesh.cpp. However as I don't speak C++ (yet?) I don't really know how to do this. Would you mind giving this a look?

src/mesh.cpp
```diff
@@ -1223,6 +1223,9 @@ StructuredMesh::MeshIndex SphericalMesh::get_indices(
       mapped_r[2] += 2 * M_PI;
   }
 
+  // test
+  coucou = grid_[0]
```
shimwell: I knew a "coucou" or two would turn up one day with you Remi :rofl:

RemDelaporteMathurin: image

RemDelaporteMathurin:
> @pshriwise I need to declare `centre` in mesh.cpp. However as I don't speak C++ (yet?) I don't really know how to do this. Would you mind giving this a look?

I think I managed to add everything we needed for this in the end 😄

RemDelaporteMathurin:
The tests are failing.

One regression test is failing --> I think we just need to update the gold standard input file.
Some unit tests are failing in tests/unit_tests/test_deplete_activation.py although I'm not certain why: @shimwell do you confirm the default centre for cylindrical and spherical meshes is (0, 0, 0) ?

pshriwise:
> > @pshriwise I need to declare `centre` in mesh.cpp. However as I don't speak C++ (yet?) I don't really know how to do this. Would you mind giving this a look?
> 
> 
> 
> I think I managed to add everything we needed for this in the end 😄 


That's great! I'll take a look first thing on Monday -- I'm off today for a long weekend.

pshriwise (changes requested):
Hi @RemDelaporteMathurin! Thanks for taking this on. 🤩  It's been needed for a while and will be a much-used, much-appreciated upgrade.

A couple of minor comments from me here. Additionally, I'd prefer to use the term `origin` over `centre` here to be consistent with our description of a sphere surface. Would you mind updating that attribute name?

include/openmc/mesh.h
```diff
@@ -359,6 +359,8 @@ class CylindricalMesh : public StructuredMesh {
   void to_hdf5(hid_t group) const override;
 
   array<vector<double>, 3> grid_;
+  vector<double> centre_;
```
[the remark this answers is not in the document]
RemDelaporteMathurin: I made this change but it seems to break things.

openmc/mesh.py
```diff
@@ -165,7 +165,7 @@ def _grids(self):
         pass
 
     @property
-    def vertices(self):
+    def vertices(self):  # TODO needs changing
```
RemDelaporteMathurin, line 168: @pshriwise do we need to adapt this or can it remain in cylindrical/spherical coordinates?

openmc/mesh.py
```diff
@@ -178,7 +178,7 @@ def vertices(self):
         return np.stack(np.meshgrid(*self._grids, indexing='ij'), axis=-1)
 
     @property
-    def centroids(self):
+    def centroids(self):  # TODO needs changing
```
RemDelaporteMathurin, line 181: @pshriwise do we need to adapt the centroids?

openmc/mesh.py
pshriwise: We'll want to make sure we translate all of the vertices at the end using the origin if it is set. Same for `centroids.

RemDelaporteMathurin: Hm. I don't see where the conversion from cylindrical to cartesian is done for the vertices.
It doesn't seem to be exported to xml so it's unused in the C++ part am I right?

Currently, it doesn't seem that vertices (or centroids) of a CylindricalMesh are converted to cartesian (at least not in mesh.py).

pshriwise (changes requested):
A couple more comments..

openmc/mesh.py
```diff
@@ -1052,6 +1052,7 @@ def __init__(self, mesh_id=None, name=''):
         self._r_grid = None
         self._phi_grid = [0.0, 2*pi]
         self._z_grid = None
+        self._origin = [0., 0., 0.]
```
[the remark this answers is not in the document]
RemDelaporteMathurin: That's fine with me.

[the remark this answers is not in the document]
pshriwise: Yeah I suppose it's more explicit to return `(0.0, 0.0, 0.0)` than `None`. This is true.

I guess I'd still like to make sure we're skipping that translation step when it isn't needed. NumPy might be smart enough to recognize that as a no-op, but I'm not sure so some clarity on that would be good. If it isn't going to skip it then I'll suggest that we keep a check for `self.origin != (0.0, 0.0, 0.0)` in that translation block.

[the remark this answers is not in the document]
RemDelaporteMathurin: Alright I added a check before offsetting. I didn't understand it was for performance purposes.

RemDelaporteMathurin:
@pshriwise I talked with @eepeterson about the `vertices` method and it seems we are keen on leaving it as is (ie in cylindrical or spherical coordinates).

Anything else we need to do for this PR?

openmc/mesh.py
RemDelaporteMathurin: According to @eepeterson no

openmc/mesh.py
RemDelaporteMathurin: According to @eepeterson we don't need to adapt this

eepeterson:
> @pshriwise I talked with @eepeterson about the `vertices` method and it seems we are keen on leaving it as is (ie in cylindrical or spherical coordinates).
> 
I think changing the properties to getter methods if you want the ability to get back cartesian coordinates rather than the natural coordinates would be an option. I think for ease of use we might also want to change the shape of `centroids` and `vertices` so that you can unpack it in a single line more naturally like so: `xx, yy, zz = mesh.get_centroids()` with kwargs like `basis="xyz"` if you want a `CylindricalMesh` transformed to x, y, z coordinates rather than r, phi, z. You could also add a `squeeze=True` kwarg to call `np.squeeze` before returning so that the user could pass these arrays directly into `matplotlib` calls with no additional manipulation.

RemDelaporteMathurin:
> > @pshriwise I talked with @eepeterson about the `vertices` method and it seems we are keen on leaving it as is (ie in cylindrical or spherical coordinates).
> 
> I think changing the properties to getter methods if you want the ability to get back cartesian coordinates rather than the natural coordinates would be an option. I think for ease of use we might also want to change the shape of `centroids` and `vertices` so that you can unpack it in a single line more naturally like so: `xx, yy, zz = mesh.get_centroids()` with kwargs like `basis="xyz"` if you want a `CylindricalMesh` transformed to x, y, z coordinates rather than r, phi, z. You could also just squeeze the array before unpacking so you can pass the arrays into `matplotlib` calls with no additional manipulation for the common use case of a mesh with only 1 bin in one of the three dimensions.

Right, I feel like this could be another PR. I shall open an issue to track this

pshriwise:
@RemDelaporteMathurin this kind of fell off my radar.  Apologies.

Regarding the coordinate systems:

I think that returning these coordinates in the mesh/local coordinate system leads to some inconsistencies in the meaning of the `vertices` and `centroids` properties that are common to the `MeshBase` class. The values returned from `RegularMesh` and `RectilinearMesh` are absolute, not relative to any single location in the mesh whereas the values returned from the `CylindricalMesh` and `SphericalMesh` are relative to the mesh origin.

I agree that I don't think we should hold up this PR further and that we can discuss it in an issue.

I also agree on making the values easier to unpack. 💯

openmc/mesh.py
```diff
@@ -1540,6 +1576,12 @@ def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
         pts_cartesian[:, 1] = r * np.sin(phi) * np.sin(theta)
         pts_cartesian[:, 2] = r * np.cos(phi)
 
+        # offset with origin
+        if self.origin != [0.0, 0.0, 0.0]:
```
pshriwise, line 1580:
```suggestion
        if any(self.origin != (0.0, 0.0, 0.0)):
```

src/mesh.cpp
```diff
@@ -1206,6 +1214,11 @@ StructuredMesh::MeshIndex SphericalMesh::get_indices(
   Position mapped_r;
 
   mapped_r[0] = r.norm();
+
+  mapped_r[0] += origin_[0];
+  mapped_r[1] += origin_[1];
+  mapped_r[2] += origin_[2];
```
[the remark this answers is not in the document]
pshriwise: There are other methods in the C++ code that are going to need this translation applied like `distance_to_grid_boundary`, etc. It's worth going through the call stack under both `StructuredMesh::bins_crossed` to make sure this adjustment is being applied in the right places.

pshriwise (changes requested):
A few more things to address here @RemDelaporteMathurin, but it's looking good! I think we'll want to add a regression test to make sure that the mesh is being translated correctly in the C++ code as well since applying the origin is going to require offsetting the incoming position in a couple of places.

src/mesh.cpp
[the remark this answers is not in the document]
RemDelaporteMathurin: > It's worth going through the call stack under both `StructuredMesh::bins_crossed` to make sure this adjustment is being applied in the right places.

Don't really know what I'm looking at in `StructuredMesh::bins_crossed` nor where the coordinates must be offset

openmc/mesh.py
```diff
@@ -1253,6 +1266,8 @@ def from_xml_element(cls, elem):
         mesh.r_grid = [float(x) for x in get_text(elem, "r_grid").split()]
         mesh.phi_grid = [float(x) for x in get_text(elem, "phi_grid").split()]
         mesh.z_grid = [float(x) for x in get_text(elem, "z_grid").split()]
+        mesh.origin = [float(x) for x in get_text(elem, "origin").split()]
```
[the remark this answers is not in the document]
pshriwise: Statepoint file major/minor versions can be found in this file:

https://github.com/openmc-dev/openmc/blob/a17fe6cc828dd6a54166aa4065eafc0deb145dad/include/openmc/constants.h#L28

openmc/mesh.py
```diff
@@ -1306,6 +1321,12 @@ def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
         pts_cartesian[:, 0] = r * np.cos(phi)
         pts_cartesian[:, 1] = r * np.sin(phi)
 
+        # offset with origin
+        if self.origin != [0.0, 0.0, 0.0]:
```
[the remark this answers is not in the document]
RemDelaporteMathurin: I think this broke a few things....

openmc/mesh.py
```diff
@@ -1052,6 +1052,7 @@ def __init__(self, mesh_id=None, name=''):
         self._r_grid = None
         self._phi_grid = [0.0, 2*pi]
         self._z_grid = None
+        self._origin = (0., 0., 0.)
```
pshriwise, line 1055: Since we're passing a valid value for the origin, we can use the setter here to ensure that they type is converted to a NumPy array. This is what was causing the tests to fail.
```suggestion
        self.origin = (0., 0., 0.)
```

openmc/mesh.py
```diff
@@ -1354,6 +1378,7 @@ def __init__(self, mesh_id=None, name=''):
         self._r_grid = None
         self._theta_grid = [0, pi]
         self._phi_grid = [0, 2*pi]
+        self._origin = (0., 0., 0.)
```
pshriwise, line 1381: Similar fix to the one above using the property setter.
```suggestion
        self.origin = (0., 0., 0.)
```

RemDelaporteMathurin:
@pshriwise added some changes and better tests to ensure the feature was working.

tests/unit_tests/test_cylindrical_mesh.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # we've translated half of the cylinder mesh above the model,
+        # so ensure that half of the bins are populated
+        assert np.count_nonzero(tally.mean) == tally.mean.size / 2
+
+        # check that the lower half of the mesh contains a tally result
+        # and that the upper half is zero
+        mean = tally.mean.reshape(mesh.dimension[::-1]).T
+        # assert np.count_nonzero(mean[:, :, :5]) == mean.size / 2
+        # assert np.count_nonzero(mean[:, :, 5:]) == 0
```
RemDelaporteMathurin, line 98: @pshriwise any particular reason these tests are commented out? (similar comment for test_spherical_mesh.py)

pshriwise:
> EDIT: the CI doesn't pass so need to investigate a bit

Yeah as @shimwell noted in #2366 we have an issue there to sort out.

tests/unit_tests/test_cylindrical_mesh.py
pshriwise: Oh yeah I had an indexing issue to sort out there after extending the test. I'll update that shortly.

RemDelaporteMathurin:
@pshriwise the spherical mesh test seems to fail. Don't really understand why. I'll try to have a look at it later on

RemDelaporteMathurin:
@pshriwise @eepeterson @paulromano I think this PR is now ready.

I was looking for more documentation to add but couldn't find a dedicated page on meshes.

pshriwise:
@eepeterson @paulromano if one of you wouldn't mind taking a look at this I'd appreciate it. I wrote a decent number of lines in the tests and it would be good to get one more set of eyes on it.

pshriwise:
> I was looking for more documentation to add but couldn't find a dedicated page on meshes.

Might be a good thing for us to add in the future. Thanks for including the documentation here, @RemDelaporteMathurin!

openmc/mesh.py
```diff
@@ -1049,6 +1049,8 @@ class CylindricalMesh(StructuredMesh):
         The default value is [0, 2π], i.e. the full phi range.
     z_grid : numpy.ndarray
         1-D array of mesh boundary points along the z-axis.
+    origin : tuple
```
pshriwise, line 1052: We'll want to update the XML spec as well https://github.com/openmc-dev/openmc/blob/develop/docs/source/io_formats/tallies.rst#mesh-element

RemDelaporteMathurin: Done

openmc/mesh.py
```diff
@@ -1351,6 +1378,8 @@ class SphericalMesh(StructuredMesh):
     phi_grid : numpy.ndarray
         1-D array of mesh boundary points along the phi-axis in radians.
         The default value is [0, 2π], i.e. the full phi range.
+    origin : tuple
```
eepeterson, line 1381: Same comment as above re `numpy.ndarray`

openmc/mesh.py
```diff
@@ -1315,6 +1336,12 @@ def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
         pts_cartesian[:, 0] = r * np.cos(phi)
         pts_cartesian[:, 1] = r * np.sin(phi)
 
+        # offset with origin
+        if any([coord != 0 for coord in self.origin]):
+            pts_cartesian[:, 0] = pts_cartesian[:, 0] + self.origin[0]
+            pts_cartesian[:, 1] = pts_cartesian[:, 1] + self.origin[1]
```
eepeterson, line 1342: I think just add the origin coordinates in lines 1336 and 1337 instead of bothering to check

openmc/mesh.py
```diff
@@ -1549,6 +1597,12 @@ def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
         pts_cartesian[:, 1] = r * np.sin(phi) * np.sin(theta)
         pts_cartesian[:, 2] = r * np.cos(phi)
 
+        # offset with origin
```
eepeterson, line 1600: same comment as above

eepeterson (changes requested):
@RemDelaporteMathurin nice work. First pass on the python side I just have a few comments. I'll take a look at the C++ stuff as soon as I can.

openmc/mesh.py
RemDelaporteMathurin: I tend to agree, but @pshriwise wanted to keep this check. Patrick was it for performance issues?

> Yeah I suppose it's more explicit to return `(0.0, 0.0, 0.0)` than `None`. This is true.
> 
> I guess I'd still like to make sure we're skipping that translation step when it isn't needed. NumPy might be smart enough to recognize that as a no-op, but I'm not sure so some clarity on that would be good. If it isn't going to skip it then I'll suggest that we keep a check for `self.origin != (0.0, 0.0, 0.0)` in that translation block.

eepeterson: I don't think this is a big deal from a performance perspective, but I'll try to track down proof. I also don't like using `!=` with floats. So, to avoid these comparisons or using `np.isclose` and specifying a tolerance I would say the clearest/cleanest approach is to just to always add them.

pshriwise: Yeah, true. Good to avoid floating point comparisons if possible. Overall, I'd probably prefer this property be set to `None` by default instead of `(0.0, 0.0, 0.0)`.

pshriwise (commented):
Just adding comments b/c GH categorized them as review comments.

openmc/mesh.py
eepeterson: > Yeah I guess I don't know enough to know if NumPy will skip this or if we'll be adding a bunch of zeros to our points. @eepeterson do you know?
> 
> I'll admit it's cleaner without this check..

I don't think NumPy will skip this. I think it will do the addition, but I don't think there will be any significant overhead if it is added to the coordinate transformation line. I think the `r*np.cos(phi)` portion is much more expensive for large arrays, but we have to do it. FWIW `pts_cartesian[:, 2] += self.origin[2]` actually seems to be ~2x faster for large arrays (testing with 1e8 x 3 array).

eepeterson: Sorry no in comparison to `pts_cartesian[:, 2] = pts_cartesian[:, 2] + self.origin[2]`

pshriwise: Oh! Yeah the in-place operator should be faster here. Nice catch

eepeterson: Also on line 1334 if you are going to end up filling the whole `pts_cartesian` array anyway you don't need to copy the cylindrical points values just use `np.empty_like(pts_cylindrical)` instead, it's much faster. Same goes for line 1594.

src/mesh.cpp
```diff
@@ -1435,6 +1447,7 @@ void SphericalMesh::to_hdf5(hid_t group) const
   write_dataset(mesh_group, "r_grid", grid_[0]);
   write_dataset(mesh_group, "theta_grid", grid_[1]);
   write_dataset(mesh_group, "phi_grid", grid_[2]);
+  write_dataset(mesh_group, "origin", origin_);
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Hm I don't see where we document this in the statepoint specs (like we don't document phi_grid or r_grid, etc).
However, I did add the change to the MeshFilter specs
https://github.com/openmc-dev/openmc/blob/f961fc90c2bc5c0565be5179a6a4c789cc553729/docs/source/io_formats/tallies.rst?plain=1#L354-L355

openmc/mesh.py
```diff
@@ -1543,12 +1592,17 @@ def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
...
+        pts_cartesian = np.empty_like(pts_spherical)
         r, theta, phi = pts_spherical[:, 0], pts_spherical[:, 1], pts_spherical[:, 2]
         pts_cartesian[:, 0] = r * np.sin(phi) * np.cos(theta)
         pts_cartesian[:, 1] = r * np.sin(phi) * np.sin(theta)
         pts_cartesian[:, 2] = r * np.cos(phi)
 
+        # offset with origin
+        pts_cartesian[:, 0] += self.origin[0]
+        pts_cartesian[:, 1] += self.origin[1]
+        pts_cartesian[:, 2] += self.origin[2]
```
eepeterson, line 1604:
```suggestion
        pts_cartesian[:, 0] = r * np.sin(phi) * np.cos(theta) + self.origin[0]
        pts_cartesian[:, 1] = r * np.sin(phi) * np.sin(theta) + self.origin[1]
        pts_cartesian[:, 2] = r * np.cos(phi) + self.origin[2]
```

openmc/mesh.py
```diff
@@ -1310,11 +1332,16 @@ def write_data_to_vtk(self, filename, datasets, volume_normalization=True):
...
+        pts_cartesian = np.empty_like(pts_cylindrical)
         r, phi = pts_cylindrical[:, 0], pts_cylindrical[:, 1]
         pts_cartesian[:, 0] = r * np.cos(phi)
         pts_cartesian[:, 1] = r * np.sin(phi)
 
+        # offset with origin
+        pts_cartesian[:, 0] += pts_cartesian[:, 0]
+        pts_cartesian[:, 1] += pts_cartesian[:, 1]
+        pts_cartesian[:, 2] += pts_cartesian[:, 2]
+
```
eepeterson, line 1344:
```suggestion
        pts_cartesian[:, 0] = r * np.cos(phi) + self.origin[0]
        pts_cartesian[:, 1] = r * np.sin(phi) + self.origin[1]
        pts_cartesian[:, 2] += self.origin[2]
        
```

src/mesh.cpp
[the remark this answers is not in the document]
pshriwise: Thanks! 

and you're right, the cylindrical/spherical grids aren't listed there, but they should be. Would you be up for adding them?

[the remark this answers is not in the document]
RemDelaporteMathurin: Sure, only the grids or also library, filename?

include/openmc/mesh.h
```diff
@@ -241,6 +241,7 @@ class StructuredMesh : public Mesh {
   xt::xtensor<double, 1> lower_left_;  //!< Lower-left coordinates of mesh
   xt::xtensor<double, 1> upper_right_; //!< Upper-right coordinates of mesh
   std::array<int, 3> shape_; //!< Number of mesh elements in each dimension
+  Position origin_ {0.0, 0.0, 0.0};
```
eepeterson, line 244: Do we want this here? in the docs and python API it seems like we are only allowing `origin` for `CylindricalMesh` and `SphericalMesh`, but now all the `StructuredMesh` classes will have this member. As we are trying to make our Mesh functionality more consistent across the code base it might be a good time to clean this up too.

tests/unit_tests/test_cylindrical_mesh.py
```diff
@@ -0,0 +1,105 @@
...
+    with openmc.StatePoint(sp_filename) as sp:
+        tally = sp.tallies[1]
+
+        # we've translated half of the cylinder mesh above the model,
+        # so ensure that half of the bins are populated
+        assert np.count_nonzero(tally.mean) == tally.mean.size / 2
+
+        # check that the half of the mesh that is outside of the geometry
+        # contains the zero values
+        mean = tally.mean.reshape(mesh.dimension[::-1]).T
```
eepeterson, line 97: perhaps use `get_reshaped_data` with `expand_dims` here to make it clearer what we are indexing at least in the test?

tests/unit_tests/test_spherical_mesh.py
```diff
@@ -0,0 +1,109 @@
...
+    with openmc.StatePoint(sp_filename) as sp:
+        tally = sp.tallies[1]
+
+        # we've translated half of the spherical mesh above the model,
+        # so ensure that half of the bins are populated
+        assert np.count_nonzero(tally.mean) == tally.mean.size / 2
+
+        # check that the half of the mesh that is outside of the geometry
+        # contains the zero values
+        mean = tally.mean.reshape(mesh.dimension[::-1]).T
```
eepeterson, line 101: same comment about reshaping as above.

eepeterson (changes requested):
@RemDelaporteMathurin and @pshriwise just a few more questions and comments

include/openmc/mesh.h
RemDelaporteMathurin: Should this only be in `CylindricalMesh` and `SphericalMesh` then?

tests/unit_tests/test_cylindrical_mesh.py
RemDelaporteMathurin: Sorry, is `get_reshaped_data` a method of `mesh`?

pshriwise: It's a method on the `openmc.Tally` class. @eepeterson just made some nice updates to it to expand the filter dimensions correctly for a `MeshFilter`.

https://docs.openmc.org/en/latest/pythonapi/generated/openmc.Tally.html#openmc.Tally.get_reshaped_data

include/openmc/mesh.h
pshriwise: I debated internally about this too. The problem I ran into with placing this in the `CylindricalMesh`/`SphericalMesh` classes is that the origin needs to be accounted for in the `StructuredMesh::raytrace_mesh` method. I think I have a better way to handle this now though. I'll push a commit later today.

pshriwise:
I've just attempted to handle the conflicts in `mesh.py`. I'll keep an eye on CI to make sure that was successful. 

Aside from that, I think there's one outstanding comment from @eepeterson remaining. @RemDelaporteMathurin are you up for addressing that one?

RemDelaporteMathurin:
> @RemDelaporteMathurin are you up for addressing that one?

Thanks for this Patrick, I don't have the bandwith right now but I can have a look whenever I have time

paulromano:
@RemDelaporteMathurin I just sent a PR to your fork to address the outstanding request from @eepeterson:
https://github.com/RemDelaporteMathurin/openmc/pull/7

docs/source/io_formats/statepoint.rst
```diff
@@ -78,6 +78,12 @@ The current version of the statepoint file format is 17.0.
              of mesh.
            - **width** (*double[]*) -- Width of each mesh cell in each
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Thanks @paulromano for taking care of this!

RemDelaporteMathurin:
@paulromano @eepeterson @pshriwise I think all the comments have been addressed and the tests pass! 🎸

eepeterson (approved):
LGTM, thanks for the work team!

src/mesh.cpp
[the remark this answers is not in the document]
eepeterson: @pshriwise can you confirm this is no longer an issue?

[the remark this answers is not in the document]
pshriwise: Yeah it was taken care of in one of the PRs to the PR 👍🏻

pshriwise (approved):

PR #2271, opened 2022-10-19
Include volume in xml

eepeterson:
This PR just aims to make cell volumes able to be written/read to/from `geometry.xml` in the Python API.


paulromano (approved):
Simple enough; thanks @eepeterson!

PR #2275, opened 2022-10-24
ww_bounds bug fix in setter and getter

eepeterson:
Simple bug fix for a place we were intending to modify an array, but were not. Namely `bounds.reshape` doesn't modify the array in place.


shimwell:
Thanks for this fix @eepeterson 
I was thinking is it worth adding a unit test to protect against this bug returning in the future.
If you are keen on a unit test then I've made a PR against your branch https://github.com/eepeterson/openmc/pull/1

openmc/weight_windows.py
```diff
@@ -341,6 +341,11 @@ def from_xml_element(cls, elem, root):
         particle_type = get_text(elem, 'particle_type')
         survival_ratio = float(get_text(elem, 'survival_ratio'))
 
+        ww_shape = (len(e_bounds) - 1,) + mesh.dimension[::-1]
+        print(ww_shape)
```
eepeterson, line 345:
```suggestion
```

shimwell:
The code looks fine to me, just to check my understanding, is this reshape needed when using WW with multiple energy groups

eepeterson:
@shimwell this is essentially just fixing the `from_xml` method of the `WeightWindows` class. If you read in weight windows from a `settings.xml` file into a python script the indexing will be wrong (transposed to what it should be). If you did happen to write them out again they will not be ordered in the way the openmc executable assumes.

shimwell (approved):

shimwell:
> 

this might explain why the[ minimal magic](https://gist.github.com/shimwell/b14a674dff0cf94be79d69610f50be69) script @pshriwise and myself have been looking into has been struggling with improving the ww on each itteration

pshriwise (approved):
Thanks for this fix, @eepeterson!

pshriwise:
> > 
> 
> Thanks Ethan this might explain why the[ minimal magic](https://gist.github.com/shimwell/b14a674dff0cf94be79d69610f50be69) script @pshriwise and myself have been looking into has been struggling with improving the ww on each iteration as we read in the settings.xml after writing it each loop

I think this is fixing a read into Python from XML as opposed to fixing a read into the C++ API. I've had success generating WWs with multiple energy groups on several models with that script as-is. I think our ww gen problem may have more to do with an instantiation of a Numpy array as integers rather than floats -- a change that I have in my branch on this but not that gist I sent you. My bad!

PR #2276, opened 2022-10-24
small grammar fix in `Settings.temperature` docstring

yardasol:
This PR adds a missing word in the docstring for `Settings.temperature`.

### DIFF
--- a/openmc/settings.py
+++ b/openmc/settings.py
@@ -188,7 +188,7 @@ class Settings:
         Kelvin. The value for 'method' should be 'nearest' or 'interpolation'.
         If the method is 'nearest', 'tolerance' indicates a range of temperature
         within which cross sections may be used. The value for 'range' should be
-        a pair a minimum and maximum temperatures which are used to indicate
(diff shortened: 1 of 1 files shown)


openmc/settings.py
```diff
@@ -188,7 +188,7 @@ class Settings:
         Kelvin. The value for 'method' should be 'nearest' or 'interpolation'.
         If the method is 'nearest', 'tolerance' indicates a range of temperature
         within which cross sections may be used. The value for 'range' should be
-        a pair a minimum and maximum temperatures which are used to indicate
+        a pair of a minimum and maximum temperature which are used to indicate
         that cross sections be loaded at all temperatures within the
```
lewisgross1296, line 192: small nitpick I think this would read best
```The value for 'range' should be
        a pair of minimum and maximum temperatures which are used to indicate ...
```

yardasol: I like this better

paulromano (approved):
Thanks @yardasol!

PR #2284, opened 2022-10-27
Adding __eq__ method and roundtrip test for weight windows

pshriwise:
I ran into an issue reading a `settings.xml` file containing weight windows with multiple energy groups that had to do with determining the correct shape of arrays for the weight window bounds.

This fixes that problem and adds a test for successfully roundtripping weight window objects to and from XML. This includes adding a `WeightWindows.__eq__` method to the class. We can't check for mesh equality at the moment, but all other relevant settings are checked here.

### DIFF
--- a/openmc/weight_windows.py
+++ b/openmc/weight_windows.py
@@ -166,7 +166,7 @@ class WeightWindows(IDManagerMixin):
 
         # TODO: add ability to check mesh equality
 
-        # check a few more attributes directly
+        # check several attributes directly
         attrs = ('particle_type',
                  'survival_ratio',
                  'max_lower_bound_ratio',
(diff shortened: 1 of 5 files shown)


openmc/weight_windows.py
```diff
@@ -341,7 +370,7 @@ def from_xml_element(cls, elem, root):
         particle_type = get_text(elem, 'particle_type')
         survival_ratio = float(get_text(elem, 'survival_ratio'))
 
-        ww_shape = (len(e_bounds) - 1,) + mesh.dimension[::-1]
+        ww_shape = (len(e_bounds) - 1,) + tuple(mesh.dimension[::-1])
```
eepeterson, line 373: should we instead insure `mesh.dimension` always returns a tuple since it is so similar to `shape` for `numpy.ndarray`?

eepeterson (commented):
So far, so good @pshriwise! Just one small question.

openmc/weight_windows.py
pshriwise: Good point. The rest of the `dimension` properties return a tuple. `RegularMesh` should too.

pshriwise:
Looks like I'll need to update some other expected test results here...

eepeterson (approved):
Thanks for the fix @pshriwise!

PR #2288, opened 2022-10-31
Change to get_reshaped_data for consistency with mesh dimensions

eepeterson:
A number of users have been experiencing issues with mesh tallies because the dimension of a mesh can not be used directly to reshape the mesh tally results correctly. This is because when the mesh tally results are written to the statepoint file they are written flat with with the first index varying the fastest and the last the slowest (Fortran style indexing). When reading these results into the Python API for post processing, naively reshaping the results using `mesh.dimension` will give you the incorrect result. Short of refactoring either the Python API or the C++ side to be consistent with the other, one option would be to manipulate the shapes to be more sensible in the `get_reshaped_data` method as done in this PR.

### DIFF
--- a/openmc/arithmetic.py
+++ b/openmc/arithmetic.py
@@ -604,6 +604,10 @@ class AggregateFilter:
     def num_bins(self):
         return len(self.bins) if self.aggregate_filter else 0
 
+    @property
+    def shape(self):
+        return (self.num_bins,)
+
     @type.setter
     def type(self, filter_type):
         if filter_type not in _FILTER_TYPES:
--- a/openmc/filter.py
+++ b/openmc/filter.py
@@ -102,6 +102,8 @@ class Filter(IDManagerMixin, metaclass=FilterMeta):
         Unique identifier for the filter
     num_bins : Integral
         The number of filter bins
+    shape : tuple
+        The shape of the filter
 
     """
 
@@ -204,6 +206,10 @@ class Filter(IDManagerMixin, metaclass=FilterMeta):
     @property
     def num_bins(self):
         return len(self.bins)
+
+    @property
+    def shape(self):
+        return (self.num_bins,)
 
     def check_bins(self, bins):
         """Make sure given bins are valid for this filter.
@@ -838,6 +844,10 @@ class MeshFilter(Filter):
                 self.bins = list(range(len(mesh.volumes)))
         else:
             self.bins = list(mesh.indices)
+
+    @property
+    def shape(self):
+        return self.mesh.dimension
 
     @property
     def translation(self):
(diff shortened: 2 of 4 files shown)


pshriwise:
This looks really nice @eepeterson! As I was messing around with some tally data yesterday, I realized that expanding these dimensions does mess with expectations about the dimensionality of the returned array and makes it harder to determine the dimension for the next filter as it is now dependent on the dimensionality of the mesh from the mesh filter. 

What would you think about adding a default argument here (something like `expand_dims=False` that would indicate whether or not the dimensions will be expanded?

eepeterson:
Sure I can add that. Although I would like the behavior to be the same for expanding and not expanding. I.e change the ordering of the data for mesh filters even if you aren't expanding dimensions so that the user could reshape properly once they get the data back - does that seem reasonable? Although it will probably break post-processing for people.

eepeterson:
The `MeshSurfaceFilter` class is causing the tests to fail. There may be some slight refactoring necessary for the `Filter` API to solve this properly, rather than a series of bandaids...

openmc/arithmetic.py
```diff
@@ -604,6 +604,10 @@ def bins(self):
     def num_bins(self):
         return len(self.bins) if self.aggregate_filter else 0
 
+    @property
+    def shape(self):
+        return (self.num_bins,)
```
shimwell, line 609:
```suggestion
        return (self.num_bins)
```

openmc/filter.py
```diff
@@ -205,6 +207,10 @@ def bins(self, bins):
     def num_bins(self):
         return len(self.bins)
 
+    @property
+    def shape(self):
+        return (self.num_bins,)
```
shimwell, line 212:
```suggestion
        return (self.num_bins)
```
Just wondering if the comma is needed

openmc/arithmetic.py
eepeterson: yes it is because we want `shape` to be a tuple. I think what you have here will just return the integer number of bins

pshriwise:
> Sure I can add that. Although I would like the behavior to be the same for expanding and not expanding. I.e change the ordering of the data for mesh filters even if you aren't expanding dimensions so that the user could reshape properly once they get the data back - does that seem reasonable? Although it will probably break post-processing for people.

The only reason I hesitate to do that is that it would give you the data in a different order than you would get using other methods on the `Tally` class. I think if we were going to enforce special behavior for the ordering of `MeshFilter` data we'd want to do that uniformly for the class.

pshriwise (approved):
This looks good to me @eepeterson. I'll get the tests running again and get this merged tomorrow if there are no other comments.

pshriwise:
Might need you to update this against `develop` actually. Looks like there are some stale jobs in CI and it won't let me merge this without confirmation that those jobs have run.

PR #2298, opened 2022-11-13
added acceptable keys to error message

shimwell:
Just a tiny change to the ```openmc.config``` keyError message. I think that returning the acceptable keys in the message could help users quickly correct their code. For example I keep forgetting if the key is cross_section or cross_sections and now it tells me what it should be


paulromano (approved):
Simple enough -- thanks @shimwell!

PR #2299, opened 2022-11-14
Correction to error message regarding WW shape

pshriwise:
This is a mini-PR with a correction to the message displayed when the wrong number of values is provided to the `WeightWindows` class.

### DIFF
--- a/src/weight_windows.cpp
+++ b/src/weight_windows.cpp
@@ -184,7 +184,7 @@ WeightWindows::WeightWindows(pugi::xml_node node)
       fmt::format("In weight window domain {} the number of spatial "
                   "energy/spatial bins ({}) does not match the number "
                   "of weight bins ({})",
-        id_, num_energy_bins, num_weight_bins);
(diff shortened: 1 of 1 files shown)


src/weight_windows.cpp
```diff
@@ -184,7 +184,7 @@ WeightWindows::WeightWindows(pugi::xml_node node)
       fmt::format("In weight window domain {} the number of spatial "
```
eepeterson, line 184: Should we remove the additional word "spatial" so it just reads "the number of energy/spatial bins" ?

pshriwise: Ohh yeah for sure. I think it could also be more clear which number is which. (The expected number vs. the size of the array passed in).

I'll make those updates shortly. Thanks!

eepeterson (approved):
simple enough, much clearer now (and correct) thanks @pshriwise!

paulromano (approved):
LGTM

PR #2314, opened 2022-12-01
added method to get filled cells

shimwell:
Inspired by @eepeterson R2SModel branch I wondered if it might be useful to provide a method for getting all the filled cells.

Perhaps adds a tiny amount of convenience if one is only interested in getting the cells in a universe or geometry that have a material

### DIFF
--- a/openmc/geometry.py
+++ b/openmc/geometry.py
@@ -322,6 +322,19 @@ class Geometry:
         else:
             return OrderedDict()
 
+    def get_all_filled_cells(self):
+        """Return all cells that are contained within the universe that are
+        filled with a material
+
+        cells : collections.OrderedDict
+            Dictionary whose keys are cell IDs and values are :class:`Cell`
+            instances
+
+        """
+
+        return {id:cell for id, cell in self.root_universe.get_all_cells().items() if cell.fill is not None}
+
+
     def get_all_universes(self):
         """Return all universes in the geometry.
 
--- a/openmc/universe.py
+++ b/openmc/universe.py
@@ -512,6 +512,19 @@ class Universe(UniverseBase):
 
         return cells
 
+    def get_all_filled_cells(self):
+        """Return all cells that are contained within the universe that are
+        filled with a material
+
+        cells : collections.OrderedDict
+            Dictionary whose keys are cell IDs and values are :class:`Cell`
+            instances
+
+        """
+
+        return {id:cell for id, cell in self.get_all_cells().items() if cell.fill is not None}
+
+
     def get_all_materials(self, memo=None):
         """Return all materials that are contained within the universe
 
--- a/tests/unit_tests/test_geometry.py
+++ b/tests/unit_tests/test_geometry.py
@@ -88,6 +88,17 @@ def test_get_all_cells():
     all_cells = set(geom.get_all_cells().values())
     assert not all_cells ^ set(cells + cells2)
 
+def test_get_all_filled_cells():
+    mat1= openmc.Material()
+    cell1 = openmc.Cell(name='1')
+    cell2 = openmc.Cell(name='2')
+    cell1.fill = mat1
+    geom = openmc.Geometry([cell1, cell2])
+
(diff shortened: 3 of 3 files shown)


openmc/geometry.py
```diff
@@ -322,6 +322,19 @@ def get_all_cells(self):
...
+        """Return all cells that are contained within the universe that are
+        filled with a material
+
+        cells : collections.OrderedDict
+            Dictionary whose keys are cell IDs and values are :class:`Cell`
+            instances
+
+        """
+
+        return {id:cell for id, cell in self.root_universe.get_all_cells().items() if cell.fill is not None}
```
nelsonag, line 335: Any reason not to use the new Universe.get_all_filled_cells() method here?

openmc/universe.py
```diff
@@ -512,6 +512,19 @@ def get_all_cells(self, memo=None):
...
+        """Return all cells that are contained within the universe that are
+        filled with a material
+
+        cells : collections.OrderedDict
+            Dictionary whose keys are cell IDs and values are :class:`Cell`
+            instances
+
+        """
+
+        return {id:cell for id, cell in self.get_all_cells().items() if cell.fill is not None}
```
nelsonag, line 525: This doesn't return an OrderedDict, is that an intentional move away from OrderedDicts since we now support only later versions of python with ordered dictionaries?

tests/unit_tests/test_geometry.py
```diff
@@ -88,6 +88,17 @@ def test_get_all_cells():
...
 
+def test_get_all_filled_cells():
+    mat1= openmc.Material()
+    cell1 = openmc.Cell(name='1')
+    cell2 = openmc.Cell(name='2')
+    cell1.fill = mat1
+    geom = openmc.Geometry([cell1, cell2])
+
+    all_cells = geom.get_all_filled_cells().values()
+    assert len(all_cells)==1
```
nelsonag, line 99: Add space around ==

nelsonag (changes requested):
Thanks @shimwell, just a few super minor comments! I am pretty sure I have needed something like this in the past!

openmc/geometry.py
shimwell: Thanks for spotting, I shall change to use the new method, it shortens the line nicely

openmc/universe.py
shimwell: Ah sorry, I think as we are not using specific order dict methods like ```move_to_end()``` or ```.popitem()``` and python 3.7 is now the[ minimum version](https://github.com/openmc-dev/openmc/blob/d7fc8e59a27fd9581450b7ea25d89a7637c6cb03/setup.py#L64) we can use the normal dictionary? I based this on a [blog post on the topic](https://realpython.com/python-ordereddict/) but am happy with anything. For now I shall update the doc string to dict

tests/unit_tests/test_geometry.py
shimwell: Thanks

shimwell:
Many thanks for reviewing, I think the review comments have now been addressed in the last commit.

openmc/universe.py
nelsonag: Yeah, no doubt we should migrate to dict as we don't use the additional OrderedDict methods anywhere (I think?), but we probably should just keep this return type to be consistent with the other similarly named methods and return an OrderedDict.

openmc/geometry.py
```diff
@@ -322,6 +322,19 @@ def get_all_cells(self):
         else:
             return OrderedDict()
 
+    def get_all_filled_cells(self):
```
pshriwise, line 325: How does this differ from `Geometry.get_all_material_cells`?

pshriwise, line 325: I'm a little confused. According to the docstrings, this would do the same thing as `Geometry.get_all_material_cells`. How does this method differ?

shimwell:
Sorry, closing this PR as I didn't realise we already have this functionality

PR #2316, opened 2022-12-02
added energy cut for decay photons

shimwell:
This topic has come up on discourse before on [discourse](https://openmc.discourse.group/t/shutdown-dose-rate-calculations/2289/6) and also on @eepeterson R2Smodel branch.

Perhaps it is worth adding some low energy cut to ```material.decay_photon_energy``` to allow users to conveniently remove low energy photons from the simulation. These low energy photons cause issues during transport so we could even have a default set to remove them automatically but perhaps that is a step to far.

I'm sure there are more efficient ways of doing this, perhaps in the C++ layer we could automatically not transport these photons. However this approach allow users to be aware that they are getting removed from the source.

So I've thrown together this PR for your consideration it is based on @paulromano comment on discourse and @eepeterson R2sModel branch. If people think this is worth considering then I would be keen to add some tests

### DIFF
--- a/openmc/material.py
+++ b/openmc/material.py
@@ -291,12 +291,13 @@ class Material(IDManagerMixin):
                 dists.append(source_per_atom)
                 probs.append(num_atoms)
 
-        if min_energy_cutoff and dists:
-            not_cut_energies = dists.x > min_energy_cutoff
-            dists.x = dists.x[not_cut_energies]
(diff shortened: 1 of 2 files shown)


shimwell:
Thanks for looking into this Paul, much appreciated 👍.

Closing this PR in favour of doing it probably in the c++ layer in PR ##2319

PR #2339, opened 2023-01-04
Correct typo in how score is indexed for trigger console. 

### LINKED ISSUE #2338: Possible typo with printing "score" when using triggers with > 1 score/tally
I'm making some changes in Cardinal that allow users to share the same trigger among multiple scores on the same tally. I _think_ there's a minor typo in how the most-limiting score is being printed to the console. My example case has two scores, `kappa_fission` and `damage_energy`. I run the same input file 2 times, in each case only creating a tally with a single score and seeing how many batches _that score_ needs. This shows me:

- `kappa_fission` needs 141 batches
- `damage_energy` needs 45 batches

But when I add both scores simultaneously, the console shows

```
 Triggers unsatisfied, max unc./thresh. is 1.000098 for damage-energy in tally 1
      141/1    1.05364    1.02976 +/- 0.01272
 Triggers satisfied for batch 141
```

It should instead be printing `kappa-fission` instead of `damage-energy` (since I know that my most limiting score is the kappa-fission).

aprilnovak:
As described in #2338, I'm fairly certain there is minor typo in how the limiting score is determined. This fix proposed here does give me the expected behavior, but could use a double-check.

Closes #2338

### DIFF
--- a/src/tallies/trigger.cpp
+++ b/src/tallies/trigger.cpp
@@ -101,7 +101,7 @@ void check_tally_triggers(double& ratio, int& tally_id, int& score)
           // If this is the most uncertain value, set the output variables.
           if (this_ratio > ratio) {
             ratio = this_ratio;
-            score = t.scores_[trigger.score_index];
+            score = t.scores_[score_index];
             tally_id = t.id_;
           }
         }


aprilnovak:
Closing because this was addressed in #2342

PR #2344, opened 2023-01-05
Fix triggers

### LINKED ISSUE #2342: Tally triggers evaluated with incorrect score
@aprilnovak's correction in #2339 revealed a bug in our tally trigger system. 

In `check_tally_triggers`, the uncertainty/threshold ration is updated using the uncertainty of every score on the tally is checked as opposed to just checking the score on the trigger of interest in the loop. The fix in #2339 corrects the code so that the correct score is reported to the console, but based on the loop over the tally scores it isn't necessarily going to be the score of interest.

https://github.com/openmc-dev/openmc/blob/e01f0ea94473ce0bef7f055c32974debe0f1e4f8/src/tallies/trigger.cpp#L53-L111

To convince myself of this, I created a tally on relatively simple fixed source problem I've been using to evaluate weight window performance. It is a set of nested spheres containing a few materials with monoenergetic 10 MeV neutron source at the origin:

image

I then applied a tally with material filter on the water shell with `absorption` and `scattering` scores and set a trigger on the tally with a relative error threshold of 0.01. I was able to get changes in the number of batches required to satisfy the tally trigger depending on the presence of scores that were not specified on a trigger.

|  Tally Scores | Trigger Score   | Batches Required  | 
|---|---|---|
|  `['scatter', 'absorption']` |  `'absorption'` |  47 | 
|  `['scatter', 'absorption']` |  `'scatter'` |   47 |
|  `['absorption']` |  `'absorption'` | 42 |
|  `['scatter']` | `'scatter'`  | 47 |

This is an indicator that if `'scatter'` is present on the tally, then the trigger isn't being represented correctly. This was reflected in the output fix from @aprilnovak. 

**This bug results in conservative behavior, meaning that the simulation runs until the score with the highest uncertainty meets the trigger threshold.**

pshriwise:
This PR resolves #2342 and #2343. 

For #2342, The loop over tally scores in `check_tally_trigger` has been removed -- only the score index on the trigger is checked. A test was added ensuring that the addition of unrelated scores on a tally doesn't influence the number of batches needed to satisfy a tally trigger.

For #2343, the `get_tally_uncertainty` function returns an invalid result for a tally value if the mean is zero. This invalid result prompts the `check_tally_trigger` function to set the outgoing uncertainty-to-trigger threshold ratio value to infinity and exit early (no other tally result will produce a larger ratio, so we can do this safely). A ratio of infinity is used to trigger a different output to the console to indicate that the reason the trigger is unsatisfied is there was no contribution for the specified score.

Update: Adding a link to a CI run on my fork that contains only the first commit from this PR that adds the tests https://github.com/pshriwise/openmc/actions/runs/3850717720/jobs/6561182511

### DIFF
--- a/src/tallies/trigger.cpp
+++ b/src/tallies/trigger.cpp
@@ -37,6 +37,10 @@ std::pair<double, double> get_tally_uncertainty(
 
   int n = tally->n_realizations_;
   auto mean = sum / n;
+
+  // if the result has no contributions, return an invalid pair
+  if (mean == 0) return {-1 , -1};
+
   double std_dev = std::sqrt((sum_sq / n - mean * mean) / (n - 1));
   double rel_err = (mean != 0.) ? std_dev / std::abs(mean) : 0.;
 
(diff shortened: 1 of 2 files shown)


aprilnovak:
This fixes the issue I was seeing - thanks for this really thorough fix!

aprilnovak (approved):

PR #2356, opened 2023-01-18
adding optional arg to matrix_func in Pool.py (for depletion)

bam241:
This should allow to pass additional args to pool.deplete when using user specified `matrix_func`

I figured this could be useful, if one need to use a user provided func to build the depletion matrix to allow additional argument to be provided. (And I kinda need it, it would allow me not to rewrite the `depletion/pool.py` file...)

I am un-aware of the exact development guide lines, feel free to point me in the right direction/documentation :)

### DIFF
--- a/openmc/deplete/pool.py
+++ b/openmc/deplete/pool.py
@@ -38,7 +38,7 @@ def deplete(func, chain, x, rates, dt, matrix_func=None, **func_args):
         Expected to return the depletion matrix required by
         ``func``
     func_args : dict
-        Remaining keyword arguments passed to the matrix_func
+        Remaining keyword arguments passed to matrix_func when used
 
     Returns
     -------
--- a/openmc/deplete/pool.py
+++ b/openmc/deplete/pool.py
@@ -15,7 +15,7 @@ USE_MULTIPROCESSING = True
 NUM_PROCESSES = None
 
 
-def deplete(func, chain, x, rates, dt, matrix_func=None):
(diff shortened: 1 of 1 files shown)


drewejohnson:
Welcome @bam241! I got pinged because I'm one a [codeowner for the depletion module](https://github.com/openmc-dev/openmc/blob/6218becb1710780cc1e5ab82b896a0609375a2b6/CODEOWNERS#L7-L10)

I don't see any technical problems with these changes. Can you explain the workflow that necessitates this change?

bam241:
Thank you @drewejohnson for the answer.

I am contracted by [NAAREA](https://www.naarea.fr) to do this work, and am constrain on what I can disclose about the content of my work. (specially in the open... )

Basically we are trying to setup some online extraction at the depletion step. To be able to do transfert nuclide from a material to another during depletion one need to be able to provide additional information to the `Chain`, that will build the depletion matrix.

Basically the problem we retrying to solve is the following:

- dN1/dt = AN1 - BN1 (where B is the extraction matrix, with efficiency etc.)
- dN2/dt = CN2 + BN1

in this case, one need to be able to compute the term BN1 during the dt steps of the depletion, therefore be able to tweak the way the Chain would build the Bateman/Reaction matrix for specific material.

Being able to have custom parameters when using a custom `matrix_function` would simplify the amount of overloading and parameters passing required.


I hope my explanation make sense, if not, please let me know. I also am on the OpenMC slack, and available for a direct/audio chat (on the French time frame).

Also, if you have a better/simpler idea on how to implement something like this, please let me know.

PR #2360, opened 2023-01-22
multipole documentation typo fix

gridley:
`multipole.py` documentation incorrectly claims that the total micro XS is returned as the first tuple element on evaluation, but it's actually the scattering cross section.


paulromano (approved):
Thanks @gridley!

Issue #2357, opened 2023-01-19
R-Z Polygon incorrectly rendered


hassec:
Reproducer: 

```py
import numpy as np
import openmc as mc
import matplotlib.pyplot as plt
from matplotlib.patches import Polygon


rz_points = np.array(
    [
        [6.88, 3.02],
        [6.88, 2.32],
        [7.63, 0.0],
        [5.75, 0.0],
        [5.75, 1.22],
        [6.30, 1.22],
        [6.30, 3.02],
        [6.88, 3.02],
    ]
)

# plot matplotlib polygon
poly = Polygon(rz_points)
ax = plt.subplot()
ax.add_patch(poly)
ax.autoscale()
plt.savefig("mpl.png")


p0 = mc.model.Polygon(rz_points, basis="rz")
p0.boundary_type = "vacuum"

cell = mc.Cell(region=(-p0), name="my_cell")
geom = mc.Geometry([cell])

model = mc.Model()
model.geometry = geom
plot = mc.Plot()
plot.origin = (0, 0, 0)
plot.width = (20, 20)
plot.basis = "xz"
model.plots = [plot]
model.plot_geometry(False)
```

The matplotlib plot shows the shape of the polygon: 
![image](https://user-images.githubusercontent.com/11351671/213424594-20b29ae7-35fa-47e7-ad4d-5f93384d16cd.png)

But the geometry of the cell that is created by OpenMC doesn't correctly reproduce the top left cutout but has a diagonal instead.

![image](https://user-images.githubusercontent.com/11351671/213424787-1823db85-8664-44c6-bda0-2883c058f588.png)

fyi @eepeterson

hassec:
Adding a plot of the delaunay triangulation: 
![image](https://user-images.githubusercontent.com/11351671/214304179-960c5c5c-12ae-487c-b5d5-7e186396e34c.png)

If I understand the current algorithm correctly it assumes that it can build the original polygon back from sets of convex subpolygons which are created by grouping the simplices created by the Delaunay triangulation. 

But, this doesn't seem to be possible for this specific case. 

Maybe a different approach to splitting the polygon into convex subsets, could work?

eepeterson:
[...]

PR #2362, opened 2023-01-25
try to constrain triangulation

eepeterson:
This PR attempts to fix an issue reported in #2357 when a triangulation in the `Polygon` `CompositeSurface` class doesn't generate a facet on one or more of the polygon edges. The `scipy.spatial.Delaunay` object does not support constrained triangulation so without pulling in additional mesh libraries as dependencies, one way to work around this is to find the offending edges if a triangulation "fails" and to break those edges into multiple edges. 

The figure below shows the resulting triangulation from #2357 with this fix implemented. I can't guarantee this will always work, but it certainly should be more robust. Curious to see if this solves your use case @hassec 

Figure_1

This results in the domain decomposition into 3 regions shown below:
Figure_2

### DIFF
--- a/openmc/model/surface_composite.py
+++ b/openmc/model/surface_composite.py
@@ -670,10 +670,11 @@ class Polygon(CompositeSurface):
         check_length('points', points, 3)
 
         # Order the points counter-clockwise (necessary for offset method)
-        self._points = self._make_ccw(points)
+        points = self._make_ccw(points)
 
-        # Create a triangulation of the points.
-        self._tri = Delaunay(self._points, qhull_options='QJ')
+        # Create a constrained triangulation of the points.
+        # The constrained triangulation is set to _tri attribute
+        self._constrain_triangulation(points)
 
         # Decompose the polygon into groups of simplices forming convex subsets
         # and get the sets of (surface, operator) pairs defining the polygon
@@ -723,7 +724,7 @@ class Polygon(CompositeSurface):
 
     @property
     def points(self):
-        return self._points
+        return self._tri.points
 
     @property
     def basis(self):
@@ -738,7 +739,7 @@ class Polygon(CompositeSurface):
         # Get the unit vectors that point from one point in the polygon to the
         # next given that they are ordered counterclockwise and that the final
         # point is connected to the first point
-        tangents = np.diff(self._points, axis=0, append=[self._points[0, :]])
+        tangents = np.diff(self.points, axis=0, append=[self.points[0, :]])
         tangents /= np.linalg.norm(tangents, axis=-1, keepdims=True)
         # Rotate the tangent vectors clockwise by 90 degrees, which for a
         # counter-clockwise ordered polygon will produce the outward normal
@@ -783,6 +784,41 @@ class Polygon(CompositeSurface):
 
         return points[::-1, :]
 
+    def _constrain_triangulation(self, points):
+        """Generate a constrained triangulation.
+
+        Parameters
+        ----------
+        points : np.ndarray (Nx2)
+            An Nx2 array of coordinate pairs describing the vertices. These
+            points represent a planar straight line graph.
+
+        Returns
+        -------
+        None
+        """
+
+        tri = Delaunay(points, qhull_options='QJ')
+        # Loop through the boundary edges of the polygon. If an edge is not
+        # included in the triangulation, break it into two line segments.
+        n = len(points)
+        new_pts = []
+        for i, j in zip(range(n), range(1, n +1)):
+            # If both vertices of any edge are not found in any simplex, insert
+            # a new point between them
+            if not any([i in s and j % n in s for s in tri.simplices]):
+                newpt = (points[i, :] + points[j % n, :]) / 2
+                new_pts.append((j, newpt))
+
+        # If all the edges are included in the triangulation set it, otherwise
+        # try again with additional points inserted on offending edges
+        if not new_pts:
+            self._tri = tri
+        else:
+            for i, pt in new_pts[::-1]:
+                points = np.insert(points, i, pt, axis=0)
+            self._constrain_triangulation(points)
+
     def _group_simplices(self, neighbor_map, group=None):
         """Generate a convex grouping of simplices.
 
@@ -819,7 +855,7 @@ class Polygon(CompositeSurface):
                     continue
                 test_group = group + [n]
                 test_point_idx = np.unique(self._tri.simplices[test_group, :])
-                test_points = self._tri.points[test_point_idx]
+                test_points = self.points[test_point_idx]
                 # If test_points are convex keep adding to this group
                 if len(test_points) == len(ConvexHull(test_points).vertices):
                     group = self._group_simplices(neighbor_map, group=test_group)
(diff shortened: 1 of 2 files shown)


hassec:
@eepeterson thanks a lot for the lighting fast fix 🎉 
Can confirm that this fixes my original problem as well 👍  

While there are many solutions to this partitioning problem, (e.g. in C++ in https://github.com/CGAL/cgal) I didn't find a good python lib to use. So I think your idea of splitting the edge and retrying Delaunay is great!

>  I can't guarantee this will always work, but it certainly should be more robust.

I think from a users perspective the most important thing for me is, that it doesn't silently fail. 
Correct me if I'm wrong, but if I understand your algo correctly, it should always work as long as the precondition is met that you have simplexes that cover the outer edges of the polygon, right? 

So that only leaves `constrain_triangulation` as possible point of failure. 
Right now it just keeps on trying to add points, but I'm not sure if that is actually guaranteed to converge to a solution. 
So it might be worth to limit that recursion with a user configurable maximum? 
That would avoid that OpenMC gets stuck and then eventually blows it's stack from endless recursion 💥

openmc/model/surface_composite.py
```diff
@@ -783,6 +784,41 @@ def _make_ccw(self, points):
...
+                new_pts.append((j, newpt))
+
+        # If all the edges are included in the triangulation set it, otherwise
+        # try again with additional points inserted on offending edges
+        if not new_pts:
+            self._tri = tri
+        else:
+            for i, pt in new_pts[::-1]:
+                points = np.insert(points, i, pt, axis=0)
+            self._constrain_triangulation(points)
```
hassec, line 820: As mentioned I'd suggest constraining the recursion level here, and if over the max, throwing an error.

openmc/model/surface_composite.py
```diff
@@ -783,6 +784,41 @@ def _make_ccw(self, points):
 
         return points[::-1, :]
 
+    def _constrain_triangulation(self, points):
+        """Generate a constrained triangulation.
```
hassec, line 788: I'd add a sentence as to what "constrained means". 
E.g. "Ensure that all Polygon edges are representable by edges of simplices"

openmc/model/surface_composite.py
```diff
@@ -783,6 +784,41 @@ def _make_ccw(self, points):
 
         return points[::-1, :]
 
+    def _constrain_triangulation(self, points):
```
hassec, line 787: What do you think about adding a test for this function given the points from the reproducer?

openmc/model/surface_composite.py
eepeterson: done

openmc/model/surface_composite.py
```diff
@@ -773,16 +763,77 @@ def _make_ccw(self, points):
...
+        # Check if polygon is self-intersecting by comparing edges pairwise
+        n = len(points)
+        is_self_intersecting = False
+        for i in range(n):
+            p1x, p1y = points[i, :]
+            p2x, p2y = points[(i + 1) % n, :]
+            for j in range(i, n):
+                p3x, p3y = points[j, :]
+                p4x, p4y = points[(j + 1) % n, :]
+
```
hassec, line 785: @eepeterson is it possible that you forgot to commit something here?  
Seems like there should be a check that could potentially set `is_self_intersecting` to `False`, no?

eepeterson: yeah I haven't found an easy way of checking if a polygon doesn't self intersect without writing a lot of code or pulling in another dependency so I stalled on this just a bit. To catch edge cases and such it can be a bit involved.

openmc/model/surface_composite.py
```diff
@@ -773,15 +763,140 @@ def _make_ccw(self, points):
...
         # Calculates twice the signed area of the polygon using the "Shoelace
         # Formula" https://en.wikipedia.org/wiki/Shoelace_formula
+        # If signed area is positive the curve is oriented counter-clockwise.
+        # If the signed area is negative the curve is oriented clockwise.
         xpts, ypts = points.T
+        if np.sum(ypts*(np.roll(xpts, 1) - np.roll(xpts, -1))) < 0:
+            points = points[::-1, :]
+
+
+        # Check if polygon is self-intersecting by comparing edges pairwise
```
hassec, line 789: If there is a source for the below method? If so, it might be worth adding a link such that potential debugging in the future is easier.

openmc/model/surface_composite.py
```diff
@@ -773,15 +763,140 @@ def _make_ccw(self, points):
...
+                    xmin2, xmax2 = min(p2[0], p3[0]), max(p2[0], p3[0])
+                    ymin2, ymax2 = min(p2[1], p3[1]), max(p2[1], p3[1])
+                    xlap = xmin1 < xmax2 and xmin2 < xmax1
+                    ylap = ymin1 < ymax2 and ymin2 < ymax1
+                    if xlap or ylap:
+                        raise ValueError('Polygon cannot be self-intersecting')
+                    continue
+
+                else:
+                    warnings.warn('Unclear if Polygon is self-intersecting')
```
hassec, line 857: What do you think about replacing the general `else` here by an `elif num_zeros == 3` above, to make it obvious that all cases are handled?

hassec (approved):
2 small comments, but also happy if you want to move forward without them. 

Looks great, and now works for all my cases I tested locally 🚀 👍

openmc/model/surface_composite.py
eepeterson: It's loosely based on the link I've added as a comment

openmc/model/surface_composite.py
eepeterson: done

openmc/model/surface_composite.py
```diff
@@ -773,15 +763,147 @@ def _make_ccw(self, points):
...
+            An Nx2 array of coordinate pairs describing the vertices. These
+            points represent a planar straight line graph.
+
+        Returns
+        -------
+        None
+        """
+        # Only attempt the triangulation up to 3 times.
+        if depth > 2:
+            raise RuntimeError('Could not create a valid triangulation after 3'
```
shimwell, line 884: do we want a test case that raises this ```RuntimeError```

openmc/model/surface_composite.py
```diff
@@ -773,15 +763,147 @@ def _make_ccw(self, points):
...
+
+        # Order the points counter-clockwise (necessary for offset method)
         # Calculates twice the signed area of the polygon using the "Shoelace
         # Formula" https://en.wikipedia.org/wiki/Shoelace_formula
+        # If signed area is positive the curve is oriented counter-clockwise.
+        # If the signed area is negative the curve is oriented clockwise.
         xpts, ypts = points.T
+        if np.sum(ypts*(np.roll(xpts, 1) - np.roll(xpts, -1))) < 0:
+            points = points[::-1, :]
+
```
shimwell, line 787:
```suggestion
```

openmc/model/surface_composite.py
```diff
@@ -773,15 +763,146 @@ def _make_ccw(self, points):
...
+                elif num_zeros == 3:
+                    warnings.warn('Unclear if Polygon is self-intersecting')
+                    continue
+
+                else:
+                    # All 4 cross products are zero
+                    # Determine number of unique points, x span and y span for
+                    # both line segments
+                    #unique_pts = np.unique(np.vstack((p0, p1, p2, p3)), axis=0)
```
pshriwise, line 854:
```suggestion
```

paulromano (approved):

PR #2368, opened 2023-01-30
Add volume calc standard deviation as a retrievable property

### LINKED ISSUE #2221: make volume calc standard deviation retrievable
Right now, when volume calculation results are linked to cells/materials/etc through the summary or statepoint file, information about the standard deviation on the volume calc is lost/not reported. The `volume.h5` files do contain this information (see [here](https://docs.openmc.org/en/stable/io_formats/volume.html)), but after linking, only the first float is retained. I _think_ it would be relatively easy to update `add_volume_information()` for cells, materials, and universes to report both values.

kkiesling:
resolves #2221

I added a property to `Cell`, `Material`, and `Universe` that sets the volume standard deviation property (`volume_std`) if results are loaded from a stochastic volume calc. I intentionally did not add it as something that can be set manually because there wouldn't be a standard deviation if it wasn't from loading the results of a volume calc, but I can easily change that if people would prefer it to be able to be set manually as well. I also added the standard deviation to the string representation for Cell if the information is present (Material and Universe already don't print that info).

This update to just add another property is the easy fix as well. A more involved update would be to change the `.volume` property to be an `uncertainties` variable instead of type `float` and report it in the same property. But that could break backward compatibility. Still, I am open to trying to do it that way if it's preferred. Also, I didn't set anything downstream to propagate this in calculations (eg calculation of nuclide densities).

### DIFF
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -85,6 +85,9 @@ class Cell(IDManagerMixin):
         calculated in a stochastic volume calculation and added via the
         :meth:`Cell.add_volume_information` method. For 'distribmat' cells
         it is the total volume of all instances.
+    volume_std : float
+        Standard deviation in cm^3 of the stochastic volume calculation, added
+        via :meth:`Cell.add_volume_information` method.
     atoms : collections.OrderedDict
         Mapping of nuclides to the total number of atoms for each nuclide
         present in the cell, or in all of its instances for a 'distribmat'
@@ -110,6 +113,7 @@ class Cell(IDManagerMixin):
         self._paths = None
         self._num_instances = None
         self._volume = None
+        self._volume_std = None
         self._atoms = None
 
     def __contains__(self, point):
@@ -187,6 +191,10 @@ class Cell(IDManagerMixin):
     @property
     def volume(self):
         return self._volume
+
+    @property
+    def volume_std(self):
+        return self._volume_std
 
     @property
     def atoms(self):
@@ -364,6 +372,7 @@ class Cell(IDManagerMixin):
         if volume_calc.domain_type == 'cell':
             if self.id in volume_calc.volumes:
                 self._volume = volume_calc.volumes[self.id].n
+                self._volume_std = volume_calc.volumes[self.id].s
                 self._atoms = volume_calc.atoms[self.id]
             else:
                 raise ValueError('No volume information found for this cell.')
@@ -520,6 +529,7 @@ class Cell(IDManagerMixin):
 
             clone = openmc.Cell(name=self.name)
             clone.volume = self.volume
+            clone.volume_std = self.volume_std
             if self.temperature is not None:
                 clone.temperature = self.temperature
             if self.translation is not None:
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -529,7 +529,7 @@ class Cell(IDManagerMixin):
 
             clone = openmc.Cell(name=self.name)
             clone.volume = self.volume
-            clone.volume_std = self.volume_std
+            clone._volume_std = self.volume_std
             if self.temperature is not None:
                 clone.temperature = self.temperature
             if self.translation is not None:
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -144,9 +144,9 @@ class Cell(IDManagerMixin):
                                                   self.temperature)
         string += '{: <16}=\t{}\n'.format('\tTranslation', self.translation)
         if self._volume_std is not None:
-            string += '{: <16}=\t{} +/- {}\n'.format('\tVolume', self.volume, self.volume_std)
+            string += '{: <16}=\t{} +/- {} [cm^3]\n'.format('\tVolume', self.volume, self.volume_std)
         else:
-            string += '{: <16}=\t{}\n'.format('\tVolume', self.volume)
+            string += '{: <16}=\t{} [cm^3]\n'.format('\tVolume', self.volume)
 
         return string
 
--- a/openmc/cell.py
+++ b/openmc/cell.py
@@ -5,7 +5,7 @@ from numbers import Real
 from xml.etree import ElementTree as ET
 
 import numpy as np
-from uncertainties import UFloat
+from uncertainties import UFloat, ufloat
 
 import openmc
 import openmc.checkvalue as cv
@@ -80,14 +80,11 @@ class Cell(IDManagerMixin):
         :meth:`Geometry.determine_paths` method.
     num_instances : int
         The number of instances of this cell throughout the geometry.
-    volume : float
+    volume : float or uncertainties.ufloat
         Volume of the cell in cm^3. This can either be set manually or
         calculated in a stochastic volume calculation and added via the
         :meth:`Cell.add_volume_information` method. For 'distribmat' cells
         it is the total volume of all instances.
-    volume_std : float
-        Standard deviation in cm^3 of the stochastic volume calculation, added
-        via :meth:`Cell.add_volume_information` method.
     atoms : collections.OrderedDict
         Mapping of nuclides to the total number of atoms for each nuclide
         present in the cell, or in all of its instances for a 'distribmat'
@@ -113,7 +110,6 @@ class Cell(IDManagerMixin):
         self._paths = None
         self._num_instances = None
         self._volume = None
-        self._volume_std = None
         self._atoms = None
 
     def __contains__(self, point):
@@ -143,10 +139,7 @@ class Cell(IDManagerMixin):
             string += '\t{0: <15}=\t{1}\n'.format('Temperature',
                                                   self.temperature)
         string += '{: <16}=\t{}\n'.format('\tTranslation', self.translation)
-        if self._volume_std is not None:
-            string += '{: <16}=\t{} +/- {} [cm^3]\n'.format('\tVolume', self.volume, self.volume_std)
-        else:
-            string += '{: <16}=\t{} [cm^3]\n'.format('\tVolume', self.volume)
+        string += '{: <16}=\t{} [cm^3]\n'.format('\tVolume', self.volume)
 
         return string
 
@@ -194,10 +187,6 @@ class Cell(IDManagerMixin):
     @property
     def volume(self):
         return self._volume
-
-    @property
-    def volume_std(self):
-        return self._volume_std
 
     @property
     def atoms(self):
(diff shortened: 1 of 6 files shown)


openmc/cell.py
```diff
@@ -139,7 +143,10 @@ def __repr__(self):
             string += '\t{0: <15}=\t{1}\n'.format('Temperature',
                                                   self.temperature)
         string += '{: <16}=\t{}\n'.format('\tTranslation', self.translation)
-        string += '{: <16}=\t{}\n'.format('\tVolume', self.volume)
+        if self._volume_std is not None:
+            string += '{: <16}=\t{} +/- {}\n'.format('\tVolume', self.volume, self.volume_std)
+        else:
+            string += '{: <16}=\t{}\n'.format('\tVolume', self.volume)
```
shimwell, line 149:
```suggestion
            string += '{: <16}=\t{} +/- {} [cm]\n'.format('\tVolume', self.volume, self.volume_std)
        else:
            string += '{: <16}=\t{} [cm]\n'.format('\tVolume', self.volume)
```

Would it be ok to add in units here, I notice we have units for other repr entries such as density

tests/unit_tests/test_geometry.py
```diff
@@ -43,6 +43,9 @@ def test_volume(run_in_tmpdir, uo2):
         nucs = set(domain.get_nuclide_densities())
         assert not nucs ^ {'U235', 'O16'}
 
+        # make sure standard deviation is also set from the calculation
+        assert domain.volume_std is not None
```
shimwell, line 47:
```suggestion
        assert isinstance(domain.volume_std, float)
```
Wondering if it is worth checking we have a float here instead which I guess also checks if it is not None

shimwell:
ah sorry I mean to propose that last commit instead of just throwing it in. But now that it there what do you think of adding the ```volume``` and ```volume_std``` to the ```material.__repr__``` @kkiesling :smile:

shimwell:
I quite like this proposal and would be keen to see it merged in. I would be making use of it as a user straight away

kkiesling:
> ah sorry I mean to propose that last commit instead of just throwing it in. But now that it there what do you think of adding the `volume` and `volume_std` to the `material.__repr__` @kkiesling 😄

Ha that's okay! Yes, I think that is an okay thing to add in as long as no one else has any objections.

openmc/cell.py
kkiesling: yes, good idea, though it should be `[cm^3]`

openmc/material.py
```diff
@@ -140,6 +144,11 @@ def __repr__(self):
         string += '{: <16}=\t{}\n'.format('\tName', self._name)
         string += '{: <16}=\t{}\n'.format('\tTemperature', self._temperature)
 
+        if self._volume_std is not None:
+            string += '{: <16}=\t{} +/- {} [cm]\n'.format('\tVolume', self.volume, self.volume_std)
+        else:
+            string += '{: <16}=\t{} [cm]\n'.format('\tVolume', self.volume)
+
```
kkiesling, line 151:
```suggestion
        if self._volume_std is not None:
            string += '{: <16}=\t{} +/- {} [cm^3]\n'.format('\tVolume', self.volume, self.volume_std)
        else:
            string += '{: <16}=\t{} [cm^3]\n'.format('\tVolume', self.volume)
            
```

kkiesling: a suggested change for you to review @shimwell - to match my other comment

shimwell (approved):

openmc/cell.py
shimwell, line 149: I got mixed up with the other similar doc string for a second there. Now in [cm^3]

```suggestion
            string += '{: <16}=\t{} +/- {} [cm^3]\n'.format('\tVolume', self.volume, self.volume_std)
        else:
            string += '{: <16}=\t{} [cm^3]\n'.format('\tVolume', self.volume)
```

shimwell: Yes good spot, I have updated the suggestion above to cm^3

shimwell (approved):

kkiesling:
Ooof- may want to either remove the `[cm^3]` from the strings, or set it to not print the unit if there is no volume. Either that or we need to update a bunch of test output:
Screen Shot 2023-02-07 at 12 54 17 PM

kkiesling:
> We have other prints outs that are None so I guess to be consistent with the others we should still print it even when volume is None.

I will update the tests then. It actually is only 2, so not as difficult as I thought.

shimwell:
tests are passing now, nice work Kalin :tada:

I'm happy with volume_std being set only via a volume calc as I can't currently think of cases where we would want to manually set this. Perhaps if we can think of a use case this can be changed in the future.

openmc/cell.py
```diff
@@ -85,6 +85,9 @@ class Cell(IDManagerMixin):
         calculated in a stochastic volume calculation and added via the
         :meth:`Cell.add_volume_information` method. For 'distribmat' cells
         it is the total volume of all instances.
+    volume_std : float
+        Standard deviation in cm^3 of the stochastic volume calculation, added
+        via :meth:`Cell.add_volume_information` method.
```
paulromano, line 90: I'm not a huge fan of adding a separate property that basically conveys information related to the volume (which we already have a property for). Given that `ufloat` is already being used elsewhere to indicate uncertainties, the separate property would create an inconsistency across the API in how uncertainties are handled. So, allowing `volume` to be either a plain float or a `uncertainties.ufloat` is certainly cleaner in that regard, but I do see that it would create problems elsewhere. Can you take a shot at trying to use `ufloat` with the `volume` property and we'll see what it looks like?

paulromano (changes requested):

openmc/cell.py
kkiesling: Sure thing- I will look into if/how that can be done.

kkiesling:
As far as I can tell, setting the `volume` property as a `uncertainties.ufloat` variable isn't actually that much change (see minimal diffs now). If volume is set as a `ufloat` from a vol calc and then subsequently used for other calculations all other values are propagated as a `ufloat` automatically as well but the code seems to be able to handle if it is a float or ufloat automatically so I didn't encounter any errors with this change I don't think.

These are the calculated values that seem to be affected by this type change (ie, anything returned by these methods use volume to calculate the value and therefore it will be returned as a `ufloat` now instead of `float` if volume is `ufloat`):

`cell.py` and `universe.py`:
- `self.atoms` property
- `get_nuclide_densities()`

`material.py`:
- `get_nuclide_densities()`
- `get_activity()`
- `get_decay_heat()`
- `get_nuclide_atoms()`
- `get_mass()`

The original string representation works for volume regardless of type so the if-statement has been removed.

If this change is okay, then I will search through documentation and examples where this may need to be updated.

kkiesling:
I am very unsure what is causing this error. There is no error message at all, and no tests fail before it. I relaunched CI once thinking it might be a fluke but it reappeared at the same point. I tried to replicate it locally but haven't been able to figure out the cause. I can run each individual test separately with no issue and everything passes. I can run a small subset of tests with no issue. If I run `pytest` on the entire `unit_tests` directory at once, it fails at the same spot, but locally there is no error message, it just goes back to the command prompt:

CI failure, runs through 5 of 6 tests in `test_source.py` before exiting:
Screen Shot 2023-02-15 at 12 48 18 PM

Local success/fails:

Running just `test_source.py` = success (runs all 6 tests):
Screen Shot 2023-02-15 at 12 52 59 PM

Running a subset of tests together = success:
Screen Shot 2023-02-15 at 12 52 43 PM

Running all tests at once = exits before finishing the 6th test of `test_source.py` with no message (ignore failed data tests here, I don't have my data setup properly locally for those):
Screen Shot 2023-02-15 at 12 48 18 PM

kkiesling:
I have done some more digging and have more info on this failure though I am at a _complete loss_ as to why it is occurring and how it is at all related to the changes I have made here.

Running pytest with standard out printed- when running all the tests together (`pytest -s`), `test_source.py:test_rejection` fails at the line [`openmc.lib.init()`](https://github.com/kkiesling/openmc/blob/volcalc_enhancements/tests/unit_tests/test_source.py#L127) with the error [`Need to specify number of particles.`](https://github.com/kkiesling/openmc/blob/volcalc_enhancements/src/settings.cpp#L138) which occurs during the init process when the XML settings files are being run. This error does not occur when running the test file individually (`pytest -s test_source.py`).

I played around with printing the XML file contents and the node and found that in both cases (meaning running the full test suite vs running the single test), the contents of the `settings.xml` file are complete and contain the `particles` node. I confirmed this by doing `subprocess.call(["cat", "settings.xml"])` right before the init is called in the test. Then in `get_run_parameters()` where this error appears, I also printed the contents of the node that is being processed (this can be done in cpp with `node_base.print(std::cout)`. In the case of running the individual test, the node contents is the entire settings.xml file and matches exactly the contents from doing `cat settings.xml`. In the case of running full test suite together, the contents of this node is empty, which is why it fails thinking particles have not been set.

I have NOT been able to figure out why this node being passed to `get_run_parameters()` is complete in one case but empty in the other case. Nor, as I said, do I understand at all how this is related to the changes I have made here.

paulromano (commented):
So after thinking about this, I'm not sure I'd be keen to accept this PR. I can see it creating problems downstream if we're not very careful. Material volumes are used in depletion, so the uncertainties would propagate into the depletion calculation, which will certainly cause problems (likely the source of at least some of your failing tests). On the other hand, if we separate out the uncertainty into a separate attribute, the information is disjoint which is also not desirable. Can you and/or @shimwell elaborate on what the use case would be for having the uncertainty? Can the need not be met easily enough by user-managed code (e.g., pulling the uncertainties from the volume calculation h5 file when needed).

shimwell:
I was mainly keen to the improved ```__repr__``` for materials to include volume so I made make #2377

Then I noticed this PR and thought it would be best to include the ```repr``` changes in here so I closed my PR and I pitched in to this PR

But if this isn't going ahead I've reopened #2377  :smile:

kkiesling:
> I can see it creating problems downstream if we're not very careful.

I completely agree. It is not as simple as just converting the `float` to `uncertainties` like the current state of this PR, as is evident by the oddly failing CI. So to answer your first question in your original review- it would take a lot of overhaul to make it one variable.

> Can the need not be met easily enough by user-managed code (e.g., pulling the uncertainties from the volume calculation h5 file when needed).

How can this be done exactly? The need I see for it is that we have OpenMC integrated into the PyARC workflow now and it does the volume calculations. I'd like to be able to report the uncertainty of those volume calcs since they are stochastic and do not exactly match the deterministic calculations done by other codes. PyARC uses the OpenMC's Python API and I did not see a way to retrieve this information through python.

paulromano:
> How can this be done exactly? The need I see for it is that we have OpenMC integrated into the PyARC workflow now and it does the volume calculations. I'd like to be able to report the uncertainty of those volume calcs since they are stochastic and do not exactly match the deterministic calculations done by other codes. PyARC uses the OpenMC's Python API and I did not see a way to retrieve this information through python.

If you keep the `volume_#.h5` around, you can load it with `VolumeCalculation.from_hdf5(...)` and the uncertainties are captured in that object. So, it would be stored separately from the cell/material objects, but it's easy to luck them up on the volume calculation object using the cell/material IDs.

kkiesling:
Thanks- I will give that a try to retrieve the information.

Since this is not as easy/straight forward of an enhancement as I expected, I am fine with closing this PR and forgoing this update.

PR #2395, opened 2023-02-18
Use more generalized error message.

### LINKED ISSUE #2394: Generalize error message about DAG vs. CSG cells
If you try to add a DAGCell with an ID already found in the geometry, OpenMC throws an error like:

>  ERROR: Cell ID 1 exists in both DAGMC Universe 2 and the CSG geometry. Setting
        auto_geom_ids to True when initiating the DAGMC Universe may resolve this issue

This error implies that the DAGMC cell is conflicting with a `CSGCell` of the same ID, but this isn't necessarily true. This error can also be caused if the geometry has already loaded a `DAGCell` with the indicated ID. This was confusing, because it made me think my geometry somewhere had a CSG cell in it, when I was only working with DAG geometries.

aprilnovak:
Closes #2394

### DIFF
--- a/src/dagmc.cpp
+++ b/src/dagmc.cpp
@@ -182,8 +182,8 @@ void DAGUniverse::init_geometry()
       model::cell_map[c->id_] = model::cells.size();
     } else {
       warning(fmt::format("DAGMC Cell IDs: {}", dagmc_ids_for_dim(3)));
-      fatal_error(fmt::format("DAGMC Universe {} contains cell ID {}, which "
-                              "already exists in the geometry. Setting auto_geom_ids "
+      fatal_error(fmt::format("DAGMC Universe {} contains a cell with ID {}, which "
+                              "already exists elsewhere in the geometry. Setting auto_geom_ids "
                               "to True when initiating the DAGMC Universe may "
                               "resolve this issue",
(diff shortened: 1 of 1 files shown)


src/dagmc.cpp
```diff
@@ -182,11 +182,11 @@ void DAGUniverse::init_geometry()
       model::cell_map[c->id_] = model::cells.size();
     } else {
       warning(fmt::format("DAGMC Cell IDs: {}", dagmc_ids_for_dim(3)));
-      fatal_error(fmt::format("Cell ID {} exists in both DAGMC Universe {} "
-                              "and the CSG geometry. Setting auto_geom_ids "
+      fatal_error(fmt::format("DAGMC Universe {} contains cell ID {}, which "
+                              "already exists in the geometry. Setting auto_geom_ids "
```
pshriwise, line 186:
```suggestion
      fatal_error(fmt::format("DAGMC Universe {} contains a cell with ID {}, which "
                              "already exists elsewhere in the geometry. Setting auto_geom_ids "
```

pshriwise (approved):

pshriwise:
Since this is limited to an error message update, I'll merge it once CI has passed.

PR #2401, opened 2023-02-26
Implementing a point detector for variance reduction

Itay-max:
This is our try of creating a point detector (F5 like) for OpenMC.
We focused on the special case of elastic collision, however, the implemntaion for other reactions is very similar.
The main challenges were:
- Creating a ghost particle and calclute it's mean free path to the detector.
- Calculating the probabilty density function (PDF) for the ghost particle in the LAB frame (espically in the perticluar case of thermal scattering).


itayhorin:
As a proof of concept,
we tested this version on the follwing configuration of experiment:
image

We got an excellent match between MCNP and OpenMC at high energies, but at low energies we started seeing some unexpected behaviour.

image

PR #2405, opened 2023-03-02
Extra tests for cross section plotter

shimwell:
I made a few tests while tinkering with the ```calculate_cexs``` function and think they are probably worth including as it looks like we don't test this part of the code at the moment.


shimwell:
It looks like the CI has passed but reported as timed out, not sure what is going on there. But these new tests pass on the CI and locally

paulromano (approved):
Thanks so much for improving our coverage @shimwell! :pray: I'll try restarting the tests and hopefully they pass the second time around.

shimwell:
Thanks Paul. I hope to tinker with the plotter.py in the future so great to get some tests in beforehand.

